PR #2, opened 2017-02-19
[WIP] SHM helpers

vberger:
For now, the API is designed.

The actual memory-map logic remains to be implemented in shm/pool.rs


vberger:
Globally works, only the SIGBUS handler part was not tested, I don't manage to actually trigger a SIGBUS...

PR #9, opened 2017-03-07
First draft of a glutin backend

Drakulix:
So here is a good amount of code I have been experimenting and playing with and that needs a lot of discussion, which is why I am uploading it (please don't merge yet, it doesn't even compile).

- had to enable the "dlopen" feature of wayland-server because otherwise things become broken when compiling with glutin (probably because glutin is still using wayland-client 0.7).
- I am thinking about dropping the whole `backend/graphics` module right now. The OpenglRenderer is currently only coping traits and types from `glutin` and `glium`. I also made sure there is a way to layer `gfx-rs` on top of that, but currently there is a lot of code duplication happening right now. I feel there is probably need for a general opengl context library, that just describes how a `GlContext` should look like and that every library uses. But that will not happen any time soon and I don't find it troublesome, if every backend just provides it's own rendering api. We can perhaps layer a very simple optional one for drawing windows on top of all of them. But it is fact, that every rendering backend will work totally different and it does not seem reasonable to merge all OpenGL backends together.
- On the other hand the input trait seems to work out, but it requires a lot of additional discussion about the api:
    - What kind of devices can be up to the compositor to handle? I have no idea, how for example gamepad input works on linux. Is that even abstracted or do all games just talk to the raw device?
    - How does on describe touch events?
    - Do we save Modifiers? (I think we should)
    - I feel like everybody is using it's own types again and that a generic input library just describing how input should look, is needed here. The format in existing libraries is so different, that handling each backend individually makes much less sense here (compared to graphics).
    - I am not sure, if I should continue with the ID's or should just make `InputDevice` generic. But I think this approach avoids much code duplication when dealing with multiple backends at once. If you don't care about the internal identifier of the different backends you should be able to handle all of them with a single handler.
- how difficult would it be to run closures on the event loop for every tick? Getting multi-threaded with opengl contexts is not much fun.


vberger:
Woaw, that's quite some work you've done here!

I haven't yet looked in detail at your code, but first of all I have some reactions to your PR message:

> had to enable the "dlopen" feature of wayland-server

Yeah, that's because wayland-client and wayland-server depend of wayland-sys, and glutin depends on wayland-client[dlopen]. But I would expect this to just work™ nonetheless. The version mismatch is indeed probably the cause of our trouble.

> What kind of devices can be up to the compositor to handle?

The core wayland protocol only handles mouses, keyboards and touchscreens. Other input devices (like gamepads) are afaik often handled directly by the clients which need them, using specialized libs.

Any way, for now lets just stick to these 3.

> How does on describe touch events? [and regarding different input libraries]

My opinion on handling inputs in general is that, as the core goal is to run a compositor on the bare OS, libinput should be the final target, and we could design the input backend api using it as an inspiration. The goal is not to perfectly match it, but it to be the lowest-overhead backend.

> Do we save Modifiers? (I think we should)

`libxkbcommon` does it for us. So I'd say it'll not be the role of the input backend, but of the `libxkbcommon` "middlewayre".

I think I cannot give answers to your other points without reading your code, so I'll do that right now.

vberger:
Regarding the glutin backend and the event loop, actually things are likely to get much simpler with the upcoming new api of winit/glutin: there will be 2 objects, the `Window` which is used to draw, and the `EventsLoop` used to fetch the events. This will likely help us separate the "input" part from the "graphics" part.

src/backend/graphics/opengl.rs
```diff
@@ -0,0 +1,17 @@
...
+
+pub trait OpenGLRenderer
+{
+    fn swap_buffers(&self) -> Result<(), SwapBuffersError>;
+    unsafe fn get_proc_address(&self, symbol: &str) -> *const c_void;
+    fn get_framebuffer_dimensions(&self) -> (u32, u32);
+    fn is_current(&self) -> bool;
+    unsafe fn make_current(&self);
+    fn get_api(&self) -> Api;
+    fn get_pixel_format(&self) -> PixelFormat;
```
vberger: What is you general feeling about this trait?

I'm not extremely used to OpenGL, but fom what I've seen in glutin/glium, it looks like it pretty much covers the fundamental abstraction of "an OpenGL canvas to draw on".

On a nitpick, we should not directly expose glutin's types, but rather expose our own and do the translation in the impl of the backend (possibly via a `From<_>` impl).

src/backend/input.rs
```diff
@@ -0,0 +1,31 @@
+use std::hash::Hash;
+
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub enum InputDevice<T: PartialEq + Eq + Hash + Clone + Copy>
```
vberger: I'm not a big fan of this "ID" thing regarding the input devices...

I'd rather follow the "seat" route of libinput (which matches pretty closely the wayland protocol, probably by design): there can be one or more logical `Seat`s, but each of them has at most 1 of each input device type.

In this case, we could rather have a `SeatHandler` trait, with a type of which an instance is created for each seat, so the trait would probably have to include a constructor method.

In the glutin case, there would be only one seat.

examples/simple.rs
```diff
@@ -0,0 +1,36 @@
...
+        let state = event_loop.state();
+        state.get_handler::<ShmGlobal>(handler_id).get_token()
+    };
+
+    // Initialize a simple backend for testing
+    let (renderer, input) = glutin::init_windowed().unwrap();
+
+    //TODO render stuff
+
+    //TODO put input handling on the event loop
```
vberger: This will probably be blocked on some work in `wayland-server`, I need to implement the support of secondary event FDs (needed for libinput support for example) as well as a mechanism to send messages to an event loop.

src/backend/graphics/opengl.rs
Drakulix: It's okay-ish. I am not happy with copying most of the stuff from glutin and glium. It works, but I think we should drop it completely. A backend should just expose it's own rendering backend.

src/backend/input.rs
Drakulix: So we could not distinguish the devices on one seat? How would that work out with different keyboards and layouts. I still try to understand how xkbcommon would handle that stuff.

Drakulix:
> Regarding the glutin backend and the event loop, actually things are likely to get much simpler with the upcoming new api of winit/glutin: there will be 2 objects, the Window which is used to draw, and the EventsLoop used to fetch the events. This will likely help us separate the "input" part from the "graphics" part.

I am hoping the EventsLoop will be `Send`, but I expect some platforms (especially OSX) to break that.

src/backend/graphics/opengl.rs
vberger: Isn't this part of glutin/glium a direct expression of how OpenGL works?

I mean, for example I expect all graphics backends (glutin/x11/wayland/native) to be directly usable with OpenGL wrappers such as glium and gfx...

So I guess my question here is: what in this trait is glutin/glium-specific?

src/backend/input.rs
vberger: I think there would simply be one xkbcommon context per seat. And the idea is that `seat` is a logical representation, and libinput has a system-wide configuration setup apparently.

So basically, it would be up to the computer user to decide whether their two keyboard should be treated as a single one, or appear as two different logical seats.

vberger:
As per winit's API, the `EventsLoop` must be `Send` : https://github.com/tomaka/winit/blob/master/tests/events_loop.rs#L3

src/backend/input.rs
vberger: You can find some explanation about libinput's seats here: https://wayland.freedesktop.org/libinput/doc/latest/seats.html

vberger: And the udev config for libinput: https://wayland.freedesktop.org/libinput/doc/latest/udev_config.html

src/backend/graphics/opengl.rs
Drakulix: `gluim` does not need `get_api` or `get_pixel_format`, actually no real library needs it, `glutin` just provides both of them. `glx` does not even need any of these, except `get_proc_address`. It just depends on the gl context being the *current* one and also expects that not to change over time and figures out the rest itself via low-level c calls internally. However anybody using raw GL might want functionally such as `get_api` to ease development.
I am just not happy with how directly linked this backend is to `glium`s requirements. It does not feel very generic or useful, if you just copy types and traits from `glutin`.

However you are right, it should be quite easy to put this over any backend type, we are currently aiming at. And I feel like it is already very complete.

src/backend/input.rs
Drakulix: > So basically, it would be up to the computer user to decide whether their two keyboard should be treated as a single one, or appear as two different logical seats.

Logically seats are free to create and destroy as the user wishes right? So this is more of an issue on how to make this easily configurable, I see.

Drakulix: The question remains on how we want to represent seats?
Also does one logical libinput seat match a WlSeat? And can we have multiple WlSeats?

vberger: The wayland protocol explicitly support multiple WlSeat, and my feeling is that libinput and the wayland protocol have been specifically developped together such that 1 libinput seat = 1 wl_seat.

Seats can then be created and destroyed as devices are plugged/unplugged, according to the udev configuration of the user. I believe the default one is probably "all is merged into a single seat", so from smithay's point of view, plugging a 2nd keyboard would be invisible (as it would be merged with the already existing one).

However, with some specific udev config, plugging a keyboard could trigger the creation of a 2nd seat, yeah.

src/backend/graphics/opengl.rs
vberger: I see you concerns. I didn't realise half of these methods where only required by glium.

I think the scope we should target for this backend is the union of the requirements of the various opengl libraries that may be used with smithay, *as long as it does not imply any significant overhead* (be it a runtime cost or a development cost).

But overall, I'd rather be in favour of providing too many methods than too few.

Drakulix: I am also in favour of implementing traits of said opengl libraries. This will ease integration but most importantly always make sure at least the advertised libraries are working on the trait. If the situation gets to complicated at any time, we can still discuss dropping it.

src/backend/input.rs
Drakulix: What implications do multiple seats have for the user? I have seen that multiple seats *can* mean multiple cursors and window focuses (https://www.youtube.com/watch?v=WO2L_ihO_rI), but is this required or can a compositor just *merge* all of them into one pointer and input for a given focused window, if it wants to do that.

My primary use case for this is having multiple keyboards attached (laptop and external) with different layouts. And switching layouts is really not much fun with `wlc` currently and I want to have the possibility to configure this transparently for the user, without him dealing with a multi-seat environment.

Also do maybe just want to pass a `WlSeat` object instead of the Id in that case?

src/backend/graphics/opengl.rs
vberger: Yes, we can completely add a `glium` and a `gfx` cargo features (or others), each of them activating the boilerplate to crate a glium / gfx context from the `OpenGLRenderer` trait. 👍

src/backend/input.rs
vberger: From what I know of the wayland protocol, each seat has its own pointer/keyboard focus.

Also, as the compositor forwards raw key events to the clients and they are supposed to handle the keymap themselves, via libxkbcommon. From a protocol point of view, switching keyboard layout implies sending a new `wl_keyboard::keymap` event to all clients, and these clients are probably supposed to load a new keymap themselves. So, merging two keyboards with different layouts sounds complicated...

However, I think that properly handling runtime switch of keyboard layout is a feature we want to have, and the helpers smithay will provide to handle the `wl_seat` should provide it.

Drakulix: > From a protocol point of view, switching keyboard layout implies sending a new wl_keyboard::keymap event to all clients, and these clients are probably supposed to load a new keymap themselves.

Thats some very interesting info. Can we as a libinput user actually detect, which keyboard is sending a key and just switch the layout if the other is used?

vberger: If the two keyboards has been configured libinput-side to be on different seats, yes.

My take on this would be:

- Have the `InputBackend`s expose seats like libinput (glutin backend would have a single seat, wayland backend would just forward the seats it receives, x11 backedn I don't know)
- Have the `WlSeat` helper expect input organized as seats
- Allow the user of smithay to map them to each other as they please

This way, a compositor built on smithay could merge two seats if it wants, and juggle with the keymaps if needed.

Then, in a second time, we could provide in smithay a simple helper to match the two sides of this pipeline 1-1, for compositors that just want a simple case. But it's important to still expose the very low APIs. I don't want smithay to be very opinionated as how compositors should behave.

Drakulix: > If the two keyboards has been configured libinput-side to be on different seats, yes.

Again the same problem. I see this is not fixable easily. I don't really get why the protocol supports different keymaps per window, but not between different keyboards (on the same seat).

> My take on this would be

Seems reasonable, lets start with that.

Drakulix:
Any idea how quickly this will be available in glutin?

Anyway I know what I need to do for now and will ping you once I got some new stuff to show of.

src/backend/input.rs
vberger: It's not really "different keymap per window", but rather "the compositor advertizes the current keymap to the clients, but the clients are free to use their own". The `wl_keyboard::keymap` event is only informative. Client are free to use their own keymap config.

vberger:
> Any idea how quickly this will be available in glutin?

I don't know, all I can say is that we can follow https://github.com/tomaka/glutin/pull/864

Drakulix:
Okay I have pushed a second version. It is still missing some stuff I want to implement, but at first the changes:

- `OpenglRenderer` should now be completely fleshed out
- `InputHandler` now follows libinput's design very closely, I will most likely start some libinput bindings soon, now I got a grasp of the api. It is a little simplified, but I think for the better. Could you look over it and the relevant libinput documentation and tell me your thoughts? The docs are:
    - [KeyboardEvents](https://wayland.freedesktop.org/libinput/doc/latest/group__event__keyboard.html)
    - [PointerEvents](https://wayland.freedesktop.org/libinput/doc/latest/group__event__pointer.html)
    - [TouchEvents](https://wayland.freedesktop.org/libinput/doc/latest/group__event__touch.html)
- Key differences to how libinput does things:
    - Mouse coordinates are absolute and no raw device data is given, just the translated screen coordinates
    - Time only guarantees order of the events, it can not be actually used for measuring time between any events. This is potentially limiting, but this is what wlc does and I never had a problem with this. This makes other backends like glutin far easier to manage. If needed this can be changed.
     - Slot is just an ID right now. We might want to change that to a `WlSeat`?

One interesting note here is, that - if I am not mistaken - libinput has no real multi-touch support outside of gestures. The api does not allow to differentiate between different fingers like glutin's api does.

[The remaining libinput events](https://wayland.freedesktop.org/libinput/doc/latest/modules.html) are not a very high priority, but I also plan to add them later down the line.

Documentation is obviously missing and will be added.

The glutin translation should mostly be ready, but is of course (like most of the code) untested. glium's Backend trait is also not yet provided, also something on my todo list.

vberger:
Regarding the API we are converging to for GraphicsBackend and InputBackend traits, I really like the way it's going! 👍 

Regarding touch events, and especially multitouch: my understanding is that `libinput_event_touch_get_slot()` returns a different ID for each finger touching the touchscreen. This would coincide with what the wayland protocol does: https://github.com/wayland-project/wayland/blob/master/protocol/wayland.xml#L2219

> Mouse coordinates are absolute and no raw device data is given, just the translated screen coordinates

I don't know what's best between this and relative events. We want to support "teleporting" the pointer if the compositors wants to (and the backend has this capability, which I believe native and x11 have, and possibly glutin on x11 too). But let's keep that this way for now, we can change it later.

> Time only guarantees order of the events, it can not be actually used for measuring time between any events. [...] If needed this can be changed.

Okay. As for the previous point, this'll do for now, and we can increase the guarantees on that afterwards if needed.

> Slot is just an ID right now. We might want to change that to a `WlSeat`?

Hmm, not sure what you would mean by that. `WlSeat` is an object instantiated in the wayland protocol, and thus linked to client...

src/backend/input.rs
```diff
@@ -1,31 +1,59 @@
...
-    fn on_touch(&mut self, device: Touchscreen<T>, modifiers: ModifierMask, at: (u32, u32)); //TODO this is most likely to raw. We should aggregate touch events
+pub trait InputBackendHandler {
+    fn on_seat_created(&mut self, seat: &Seat);
+    fn on_seat_destroyed(&mut self, seat: &Seat);
+    fn on_keyboard_key(&mut self, seat: &Seat, time: u32, key_code: u32, state: KeyState, count: u32);
+    fn on_pointer_move(&mut self, seat: &Seat, time: u32, to: (u32, u32));
+    fn on_pointer_button(&mut self, seat: &Seat, time: u32, button: MouseButton, state: MouseButtonState);
+    fn on_pointer_scroll(&mut self, seat: &Seat, time: u32, axis: Axis, source: AxisSource, amount: f64);
+    fn on_touch(&mut self, seat: &Seat, time: u32, event: TouchEvent);
 }
```
vberger: We need a callback to advertise the capabilities of a seat (keyboard/mouse/touch), the wayland protocol requires this.

vberger:
I'm also in favour a creating an `InputBackend` trait, even if its definition is as simple as

```rust
pub trait InputBackend<H: InputHandler> {
    pub fn set_handler(&mut self, handler: H);
    pub fn get_handler(&mut self) -> &mut H;
    pub fn clear_handler(&mut self);
}
```

In order to allow writing code generic over input backends, as the `GraphicBackend` trait allows.

vberger:
Hmm, github is handling poorly my comments on your second commit, why does it show them as on an "outdated diff"?

src/backend/input.rs
Drakulix: Do we need a callback? I would suggest promoting `Seat` to a real type and add this via a method. So you can do something like: `seat.capabilities()` or even `seat.capabilities.contain(Capability::Mouse)` or something along those lines.

src/backend/glutin.rs
```diff
@@ -143,44 +147,126 @@ pub enum GlutinBackendError
...
                 match event {
-                    Event::MouseEntered => {
-                        handler.on_input_attached(InputDevice::Mouse(()));
-                    },
-                    Event::MouseLeft => {
-                        handler.on_input_detached(InputDevice::Mouse(()));
-                    },
                     Event::Focused(focused) => if focused {
-                        handler.on_input_attached(InputDevice::Keyboard(()));
+                        handler.on_seat_created(&0u32);
```
Drakulix: This was done mostly because I was too lazy to do this outside of `process_events` and because it helps testing, if you don't need to restart to trigger this event. In theory a compositor should be able to handle this, but I see your point. I will change it.

Drakulix:
> Hmm, github is handling poorly my comments on your second commit, why does it show them as on an "outdated diff"?

Really no idea. Seems very weird.

> Regarding touch events, and especially multitouch: my understanding is that libinput_event_touch_get_slot() returns a different ID for each finger touching the touchscreen. This would coincide with what the wayland protocol does: https://github.com/wayland-project/wayland/blob/master/protocol/wayland.xml#L2219

I totally looked over that! I have probably mistaken it for `libinput_event_touch_get_seat_slot()`. I will fix this.

> I don't know what's best between this and relative events. We want to support "teleporting" the pointer if the compositors wants to (and the backend has this capability, which I believe native and x11 have, and possibly glutin on x11 too). But let's keep that this way for now, we can change it later.

I would suggest adding platform specific behavior to the different `InputBackends` instead. Absolute coordinates are easier to handle in most cases, you should just be able to do `glutin_input.set_pointer(x, y)` if supported.

> Okay. As for the previous point, this'll do for now, and we can increase the guarantees on that afterwards if needed.

I will document this for later. Supporting this is no issue for libinput at all, just a little more difficult on glutin. Do we want microsecond precision? `Duration` would allow that easily.

> Hmm, not sure what you would mean by that. WlSeat is an object instantiated in the wayland protocol, and thus linked to client...

Didn't realize this was client only. I have no much knowledge of the wayland protocol, I just thought we are maybe linking libinput's with wayland's seats somehow. I guess an id or a full blown type implementing `Eq` and `Hash` is enough in that case?

src/backend/input.rs
vberger: That could be a way too, but we'll need to properly handle the case where a seat gains or loses a capability during its lifetime. In this context, the callback is the best guarantee I have in mind that the handler will be notified about this change.

vberger:
> I would suggest adding platform specific behavior to the different InputBackends instead. Absolute coordinates are easier to handle in most cases, you should just be able to do `glutin_input.set_pointer(x, y)` if supported.

Yes, in most cases absolute coordinates are the simpler. But things like mouse-grabbing protocol extensions (for games mostly) can require to process the relative motion without moving the pointer at all on the screen... Anyway, this is not a priority, but something we'll be able to add/change afterwards, depending on how it appears effectively needed.

> I guess an id or a full blown type implementing Eq and Hash is enough in that case?

For the seats, probably, yes. I'm in favour of a type with `Eq` and `Hash`, and doing the same for the touch ids as well.

src/backend/input.rs
Drakulix: https://wayland.freedesktop.org/libinput/doc/latest/group__event.html#ga8a49a089181960f7536344f7637ea777

^ It is actually possible to differentiate between devices. And it is necessary for device configuration. I guess we need a way to access devices or more callbacks now.

src/backend/input.rs
Drakulix: What about doing `seat.capabilities` and `on_seat_changed` as a new callback. I feel like capabilities is not something all backends will have and emulation is pretty annoying. This way we could also add other properties to seat's of the other backends.

Drakulix:
Third draft:

- Rename `InputBackendHandler` to `InputHandler`
- Add `InputBackend` - allows setting an `InputHandler`. Encourages to use a `Box` internally to not be bound to a single handler type. I can imagine use cases, where you want to exchange handlers and making the trait generic over a certain handler makes this more difficult. 
- `set_cursor_position` api. Relative to the client input area. Hopefully will allow us to work on wayland in the future, when the pointer grabbing/wrapping api has been worked out. Returns a result to handle unsupported platforms (equivalent to what glutin does).
- Promote `Seat` to a type. Visibility issues did drive me to create a `NewIdType` trait. Name is ugly, location is ugly, do you have a better idea? I really want crate visibility in rust, but that is only on nightly right now, right?
- Add `TouchSlot`, already handled by glutin
- `glutin` does now send seat events on setting and clearing the `InputHandler`. Note that this api means, that we maybe maybe need to cache seats on other apis and send them to new `InputHandler`s. Alternative would be to provide one handler on construction of the `InputBackend`s only.

Whats missing:
- [ ] Handle `time` better most likely with `Duration`. How does libinput's time variable work? Is it absolute since the beginning of the context? Is it relative to the last event? The docs don't tell and I will not start emulation of that on glutin, until I know.
- [ ] Provide device callbacks for configuration. (see https://wayland.freedesktop.org/libinput/doc/latest/group__device.html) Most likely backend specific device types, so we need some more generics.
- [ ] Seat capabilities and callbacks.
- [ ] Impl `glium::Backend` on to of `OpenglRenderer` and hide it behind a `glium` feature.

Drakulix:
Alright the fourth draft fixes all remaining issue, that I wanted to solve before merging.
The rest (as seen above) makes sense to postpone until we have actual `libinput` bindings to experiment with (which is my next goal, if you don't have any objections).

Please give me some final feedback and once everything is good, I will rewrite and cleanup the (ugly) commit history of this branch and open it for merging.

Drakulix:
Could you restart the nightly travis job? @vberger 
Should work now, after the broken nightly yesterday

vberger:
Sure!

Cargo.toml
```diff
@@ -4,10 +4,15 @@ version = "0.1.0"
 authors = ["Victor Berger <victor.berger@thalesgroup.com>"]
 
 [dependencies]
-wayland-server = "0.8.4"
+wayland-server = { version = "0.8.4", features = ["dlopen"] }
```
vberger: Update to 0.8.6 for fixing the missing symbol error.

Also, we should make the "dlopen" feature conditioned on the glutin feature of smithay

src/backend/glutin.rs
```diff
@@ -0,0 +1,359 @@
...
+    /// `Window` active. Otherwise the window may no respond to user interaction and no
+    /// input events will be received by a set `InputHandler`.
+    ///
+    /// Returns an error if the `Window` the window has been closed. Calling
+    /// `process_new_events` again after the `Window` has been closed is considered an
+    /// application error and unspecified baviour may occur.
+    ///
+    /// The linked `GlutinWindowedRenderer` will error with a lost Context and should
+    /// not be used anymore as well.
+    pub fn process_new_events(&mut self) -> Result<(), GlutinInputError>
```
vberger: I've started a trend of calling these kings of methods using the word `dispatch`, as they are called in the wayland libs. To fit this pattern, I'd name this method `dispatch_pending_events`.

src/backend/glutin.rs
```diff
@@ -0,0 +1,359 @@
...
+    {
+        for event in self.window.poll_events()
+        {
+            if let Some(ref mut handler) = self.handler {
+                match event {
+                    Event::KeyboardInput(state, key_code/*TODO: Is this really the keycode? glutins docs don't tell*/, _) => handler.on_keyboard_key(&self.seat, self.time_counter, key_code as u32, state.into(), 1),
```
vberger: For the wayland backend, of glutin, this is the raw keycode (truncated from `u16` to `u8`). It's probably the same for others, afaik.

Also, this `1` means we do not track simultaneous key presses? See my comment on the definition of the trait.

src/backend/graphics/opengl.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Operations will have no effect. Functions that read textures, buffers, etc.
+    /// from OpenGL will return uninitialized data instead.
+    ///
+    /// A context loss usually happens on mobile devices when the user puts the
+    /// application on sleep and wakes it up later. However any OpenGL implementation
+    /// can theoretically lose the context at any time.
+    ContextLost,
+    /// The buffers have already been swapped.
+    ///
+    /// This error can be returned when `swap_buffers` has been called multiple times
```
vberger: "... multiple times without any modification in between.", is that what you mean here?

src/backend/graphics/opengl.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// OpenGL for the web. Very similar to OpenGL ES.
+    WebGl,
+}
+
+/// Describes the pixel format of the main framebuffer
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PixelFormat {
+    /// Is the format hardware accelerated
+    pub hardware_accelerated: bool,
+    /// bits used for colors
```
vberger: "number of bits...", same for the other fields of this struct

src/backend/graphics/opengl.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// bits used for depth channel
+    pub depth_bits: u8,
+    /// bits used for stencil buffer
+    pub stencil_bits: u8,
+    /// is stereoscopy enabled
+    pub stereoscopy: bool,
+    /// is double buffering enabled
+    pub double_buffer: bool,
+    /// multisampling format used if enabled
+    pub multisampling: Option<u16>,
```
vberger: what does this u16 represent here is it an enum? (maybe it's obvious, I don't know a lot about OpenGL)

src/backend/graphics/opengl.rs
```diff
@@ -0,0 +1,87 @@
...
+{
+    /// Swaps buffers at the end of a frame.
+    fn swap_buffers(&self) -> Result<(), SwapBuffersError>;
+
+    /// Returns the address of an OpenGL function.
+    ///
+    /// Supposes that the context has been made current before this function is called.
+    unsafe fn get_proc_address(&self, symbol: &str) -> *const c_void;
+
+    /// Returns the dimensions of the window, or screen, etc.
```
vberger: In what units?

src/backend/graphics/opengl.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn get_framebuffer_dimensions(&self) -> (u32, u32);
+
+    /// Returns true if the OpenGL context is the current one in the thread.
+    fn is_current(&self) -> bool;
+
+    /// Makes the OpenGL context the current context in the current thread.
```
vberger: We need to tell why this is an unsafe method (and I don't know why it is)

src/backend/input.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// Not guaranteed to be supported on every backend. The result usually
+    /// depends on the capability of the backend, but may also fail for certain
+    /// specific actions. See the backends documentation.
+    fn set_cursor_position(&mut self, x: u32, y: u32) -> Result<(), ()>;
+}
+
+/// Implement to receive input events from any `InputBackend`.
+pub trait InputHandler {
+    /// Called when a new `Seat` has been created
+    fn on_seat_created(&mut self, seat: &Seat);
```
vberger: Given `Seat` is just a newid, do we really need to pass it by reference? (This applies to all methods)

src/backend/input.rs
```diff
@@ -0,0 +1,277 @@
...
+    fn on_seat_destroyed(&mut self, seat: &Seat);
+    /// Called when a new keyboard event was received.
+    ///
+    /// # Arguments
+    ///
+    /// - `seat` - The `Seat` the event belongs to
+    /// - `time` - A upward counting variable useful for event ordering. Makes no gurantees about actual time passed between events.
+    /// - `key_code` - Code of the pressed key. See linux/input-event-codes.h
+    /// - `state` - `KeyState` of the event
+    /// - `count` - Amount of key presses registered
```
vberger: "registered"? Is that the number of simultaneous keys currently pressed?

src/backend/input.rs
```diff
@@ -0,0 +1,277 @@
...
+    ///
+    /// # TODO:
+    /// - check if events can arrive out of order.
+    /// - Make stronger time guarantees
+    fn on_keyboard_key(&mut self, seat: &Seat, time: u32, key_code: u32, state: KeyState, count: u32);
+    /// Called when a new keyboard event was received.
```
vberger: This is a pointer moved event

src/backend/input.rs
```diff
@@ -0,0 +1,277 @@
...
+    ///
+    /// - `seat` - The `Seat` the event belongs to
+    /// - `time` - A upward counting variable useful for event ordering. Makes no gurantees about actual time passed between events.
+    /// - `to` - Absolute screen coordinates of the pointer moved to.
+    ///
+    /// # TODO:
+    /// - check if events can arrive out of order.
+    /// - Make stronger time guarantees
+    fn on_pointer_move(&mut self, seat: &Seat, time: u32, to: (u32, u32));
+    /// Called when a new keyboard event was received.
```
vberger: this is a pointer button event

src/backend/input.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// - `seat` - The `Seat` the event belongs to
+    /// - `time` - A upward counting variable useful for event ordering. Makes no gurantees about actual time passed between events.
+    /// - `button` - Which button was pressed..
+    /// - `state` - `MouseButtonState` of the event
+    ///
+    /// # TODO:
+    /// - check if events can arrive out of order.
+    /// - Make stronger time guarantees
+    fn on_pointer_button(&mut self, seat: &Seat, time: u32, button: MouseButton, state: MouseButtonState);
+    /// Called when a new keyboard event was received.
```
vberger: this is a pointer scroll event

src/backend/input.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// - `time` - A upward counting variable useful for event ordering. Makes no gurantees about actual time passed between events.
+    /// - `axis` - `Axis` this event was generated for.
+    /// - `source` - Source of the scroll event. Important for interpretation of `amount`.
+    /// - `amount` - Amount of scrolling on the given `Axis`. See `source` for interpretation.
+    ///
+    /// # TODO:
+    /// - check if events can arrive out of order.
+    /// - Make stronger time guarantees
+    fn on_pointer_scroll(&mut self, seat: &Seat, time: u32, axis: Axis, source: AxisSource, amount: f64);
+    /// Called when a new keyboard event was received.
```
vberger: this is a touch event

src/backend/glutin.rs
```diff
@@ -0,0 +1,359 @@
...
+        self.handler = Some(Box::new(handler));
+    }
+
+    fn get_handler(&mut self) -> Option<&mut InputHandler>
+    {
+        self.handler.as_mut().map(|handler| handler as &mut InputHandler)
+    }
+
+    fn clear_handler(&mut self) {
+        if let Some(ref mut handler) = self.handler {
```
vberger: `if let Some(handler) = self.handler.take() { ... }` and removing the `self.handler = None;`

src/backend/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+pub mod glutin;
+
+#[cfg(feature = "glium")]
+mod glium;
+#[cfg(feature = "glium")]
+pub use glium::*;
+
+trait NewIdType {
+    fn new(id: u32) -> Self;
+}
```
vberger: I'm not a huge fan of this trait, but I see the use. I've been using freestanding functions to do the same job in other projects, which is not really better.

Hopefully pub(restricted) will save us somewhere in the future.

vberger:
I like the structure of this overall.

The only thing in your list that I would like to see added before merging is the advertising of seat capability.

Simply something like adding a struct like this:

```
pub struct SeatCapability {
    pub pointer: bool,
    pub keyboard: bool,
    pub touch: bool
}
```

and a callback like this on the `InputHandler`, advertizing the new set of capabilities whenever it changes (and it never changes on the glutin backend, all booleans being set on `true`).

```rust
fn new_capability(&mut self, seat: Seat, capability: SeatCapability)
```

src/backend/glutin.rs
Drakulix: We don't at least for glutin. I have no idea, how to get that information. Also yes the name and description of the variable is an error, see below.

src/backend/graphics/opengl.rs
Drakulix: yep

src/backend/graphics/opengl.rs
Drakulix: Number of samples, I will make this more descriptive.

src/backend/graphics/opengl.rs
Drakulix: Good question. I would assume we should return "points" here (not pixels). The output should be scaled.

src/backend/graphics/opengl.rs
Drakulix: https://docs.rs/glutin/0.7.4/glutin/trait.GlContext.html#tymethod.make_current

^ Thats why and I don't have a better explaination.

src/backend/input.rs
Drakulix: I not sure this will to be an id type forever. E.g. even with a callback for capabilities, I still think it would make sense to expose the capabilities through `seat.capabilities()`.

src/backend/input.rs
Drakulix: Bad wording. That is the number of total keys pressed on the whole seat.
See https://wayland.freedesktop.org/libinput/doc/latest/group__event__keyboard.html#ga75a647f08ef9df9db7a6ebb9b297add4

src/backend/input.rs
Drakulix: Copy and paste ftw....

src/backend/mod.rs
Drakulix: Yes I hate that too, totally unreasonable position, purely dictated by the visibility rules.

Drakulix:
Still don't know, if I like this capability thing.
My preferred way to implement this would be to have these two:

- `seat.capabilties()`
- `on_seat_changed()`

Like this we could also add:
`seat.input_devices()` or something similar.

The trait is getting quite big already.

We (mostly you :smile: ) should decide, if we want to hide some less often used properties/events in the `Seat` struct or expose all of them (on_device_created is definitely also needed for configuration) and if we want to track all of that additionally in the `Seat` struct (see the reasoning on passing seat by-ref).

Also about the devices:

I expect that device configuration will be very different (or even not available) for all backends. How should be best deal with that? Maybe make the `InputBackend` generic over an `InputDevice` type? This would mean you cannot use a single `InputHandler` anymore for all different backends.

src/backend/input.rs
vberger: Adding this method instead of a callback, or as a complement? I think it's important to advertize the user via a callback if the capabilities change.

vberger:
Hmm, so you suggest we could have the `Seat` type give access to the seat characteristics, and a callback whenever they change, letting the user figure out what changed and if they need to do something? I guess it would work, I don't have any strong feeling about that.

Regarding device configuration, to be honest I don't understand what kind of configuration exists and should be exposed to the user... ?

Drakulix:
I just have the fear, that it is too annoying to figure out what exactly has changed. Then again, if you really care, you can most likely just do the whole initialization of the device configurations again anyway without caring about the exact differences. Similar things should be able for capabilities.

Most important configuration flags are e.g. for enabling tab_to_click on a track pad, or to control its sensitivity.
This is what sway does with it in it's configuration file: https://github.com/SirCmpwn/sway/blob/master/sway/sway-input.5.txt

src/backend/input.rs
Drakulix: As a complement is what I would suggest.

vberger:
Hmm, I see...

As you said, these configuration will very likely be dependant on the backend (actually, I think only the libinput backend will have them).

I think this kind of stuff could remain as inherent impls on the backend type, don't you think?

src/backend/graphics/opengl.rs
vberger: Hmm, given we are a compositor, I don't think the output should be scaled.

Rather, we should expose the actual size in pixels of the window/screen/etc... and let the user configure a DPI factor if necessary, to let the compositor scale its UI components and the clients if needed, but still be able to draw in HD.

So yes, the glutin backend will do scaling, and we cannot control that, but the others won't.

src/backend/graphics/opengl.rs
vberger: From what I gathered, it seems to be because an OpenGL context cannot be made current on multiple threads at the same time.

This should be written in the doc here I think.

src/backend/graphics/opengl.rs
Drakulix: Others won't scale because there is literally no way to get a scaling screen with something like a drm backend. One points always equals one pixel in that case. For `glutin` there is actually a way to work around this. OpenGL contexts are some of the only APIs where you can get pixel but also point accuracy. So when you say, that we should scale for nested compositors like when using the `glutin` backend, but not for other, "screen points" is actually the correct term to refer to in the docs, because "pixels" would mean we don't scale on `glutin`.

Scaling the clients correctly is actually always something the compositor needs to handle.

Drakulix:
> I think this kind of stuff could remain as inherent impls on the backend type, don't you think?

But you need callbacks for that stuff as well. Would that mean, that you have to register two different handlers on the libinput backend in that case?

src/backend/graphics/opengl.rs
vberger: Okay I see. Because glutin does some scaling without letting us choose... (see https://github.com/tomaka/winit/issues/105).

In this case, we should be explicit that what these "points" are is backend-dependant, and explicit in the docs of each backend what is its behaviour.

vberger:
Callbacks too here? What are the events that would be callback-ed on libinput? Is it just "a new device has been connected/disconnected", or something more complex?

vberger:
Also, as I've merged the inclusion of rustfmt & clippy in the travis build, could you rebase this PR on master?

Drakulix:
Just connect and disconnect

vberger:
In this case, maybe possibility would be to reproduce the EventLoop / EventLoopHandle pattern I've made in libwayland-server. So, something like:

```rust
pub trait InputBackend {
    type InputConfig;
    fn set_handler<H>(&mut self, handler: H) where H: InputHandler<Self> + 'static;
    fn get_handler(&mut self) -> Option<&mut InputHandler<Self>>;
    fn clear_handler(&mut self);
    fn config(&mut self) -> &mut Self::InputConfig;
}

pub trait InputHandler<B>  where B: InputBackend {
    fn on_seat_created(&mut self, config_handle: &mut B::InputConfig, seat: &Seat);
    fn on_seat_destroyed(&mut self, config_handle: &mut B::InputConfig, seat: &Seat);
    fn on_device_connected(&mut self, config_handle: &mut B::InputConfig, /* other needed args */);
    fn on_device_disconnected(&mut self, config_handle: &mut B::InputConfig, /* other needed args */);
    // other methods are not changed
}
```
And have methods relative to input configuration be inherent methods of the `InputConfig` associated type?

It forces to impl the InputHandler trait for each backend if the user requires specialized behavior, but specialization can help with that, with something like:

```rust
impl<B: InputBackend> InputHandler<B> for MyHandler {
    default fn on_device_connected(&mut self, config_handler: &mut B::InputConfig, ...) {}
    ...
}

impl InputHandler<LibInput> for MyHandler {
    fn on_device_connected(&mut self, config_handler: LibInputConfig, ...) { .... }
}
```

Also, I'm wondering how much the `dispatch_pending_events(&mut self)` method could be made into the `InputBackend` trait, too.

Drakulix:
This looks like a good pattern.

If we can find a generic `InputConfig` trait, compositors not caring about configuration could even do this:

```
impl<B: InputBackend<InputConfig=C>, C: InputConfig> InputHandler<B> for MyHandler {
    fn on_device_connected(&mut self, config_handler: &mut C, ...) {}
    ...
}
```

But I would actually name it `InputDevice`.


What I would like to do about the dispatch functions is to add the following two functions:

```
pub trait InputBackend {
    ...
    fn dispatch_new_events(&mut self);
    fn dispatch_on(self, event_loop: &mut EventLoop);
}
```

So essentially we provide helpers to drive every `InputBackend` either manually or on the wayland event loop, which is what most compositors will want to do anyway.

src/backend/graphics/opengl.rs
Drakulix: "Points" already is a general term for "scaled pixels". But I will add that to the documentation anyways to be super safe.

vberger:
Hmm, I don't see how the `InputConfig` trait is needed in your example. We can just do a

```rust
impl<B: InputBackend> InputHandler<B> for MyHandler {
    fn on_device_connected(&mut self, config_handler: &mut B::InputConfig, ...) { .. }
    ...
}
```
and never use the `config_handler` parameter... Or am I missing something?


Regarding the event InputBackend methods, given how the wayland event loop works, I don't see how the `dispatch_on` method could be implemented... The pattern i'd intuitively follow would rather to make the input backend possible to insert into the event loop as a secondary event source.

And actually, all backends expect the glutin one could profit of the "FD event source" already implemented in libwayland-server using epoll, to smartly wait on the socket of respectively the wayland connection (for the wayland backend), the X11 connection, or the libinput socket.

Drakulix:
I think I was just thinking to complicated.

Alright, I have not much of an idea, how the event loop will actually work, but we should try to integrate glutin as nicely as easily possible. `wlc` also makes it possible to schedule timers on the event loop which fire periodically, I am using this feature quite a bit in fireplace, and that would also help the glutin backend. So maybe its worth to take a look at `wlc`'s source code to see how they've done it.

vberger:
I'm currently finalizing the implementation of the timers in wayland-server, so we'll definitely be able to use them.

Cargo.toml
Drakulix: Any clue on how to do this without renaming the glutin feature?

Drakulix:
Playing with the implementation I notice, that we probably don't want a single `InputConfig`, but rather multiple `InputDevices`.

This makes designing the api to receive them without using the callback a bit tricky.

Current ideas:
```
pub trait InputBackend {
    type InputDevice;
    fn input_devices(&self) -> &[&Self::InputDevice] { .. }
    fn input_devices_mut(&mut self) -> &[&mut Self::InputDevice] { .. }
    ...
}
```

```
pub trait InputBackend {
    type InputDevice;
    fn input_devices(&self) -> Iter<Self::InputDevice> { .. }
    fn input_devices_mut(&mut self) -> IterMut<Self::InputDevice> { .. }
    ...
}
```

```
pub trait InputBackend {
    type Seat;
    fn seats(_mut); //same options as about
    ...
}

impl LibinputSeat {
    fn input_devices(&self) -> ...
}
```

I really don't know, what would be the best here.

Cargo.toml
vberger: Maybe we can accept to rename the feature into something explicit and adopt some convention, I'm thinking something like:

- for the backends: `backend_glutin`, `backend_x11`, `backend_wayland`, `backend_tty`
- for the opengl apis: `renderer_glium`, `renderer_gfx`, (maybe `renderer_vulkano` some time in the future ? ^^)

And in general, a pattern like `<function>_<implementation name>` for every feature of this kind.

src/backend/graphics/opengl.rs
vberger: Ok, I see. A pointer saying that the scaling behavior is backend dependant and documented in each backends docs would be perfect I think. 👍

vberger:
What about:

- the `on_connected_device` and `on_disconnected_device` provide you to an internal handle of the InputBackend (what I named `InputCOnfig` but could be renamed), as well as with a `DeviceId`.
- the input backend has a method `fn get_device(&mut self, DeviceId) -> &mut DeviceConfig`

What troubles me is that the notion of "input device" is a notion exclusive to the libinput backend, I think? (I'm not really sure about x11 to be honest). So it feels strange to put it in the generic part of the API.

Regarding timer events, they are here on wayland-server-0.8.7 : http://vberger.github.io/wayland-rs/wayland_server/struct.EventLoop.html#method.add_timer_event_source

Drakulix:
An `InputConfig` making devices accessible could also work out. That makes configuration for `glutin` or other backends, that don't have different devices easier, you are right.

But I don't think any id-based `get_device` can work. You can easily attach two devices with the same device and vendor id to your machine. No id that we can use is really unique. That is why I wanted to have a way to enumerate all devices.

vberger:
Ah, I see.

In this case, I'd say we stick with `InputConfig` for now (anyway, this does nothing on glutin), and we'll think about how to expose the devices of the libinput backend on the `LibinputConfig` struct when we'll actually be implementing the libinput backend.

This way, we can focus on making basic functionnality work with the glutin backend, while keeping the door open to easily integrate the other backends in the future.

Is that ok for you?

Drakulix:
Sure.

vberger:
I've been looking at weston's internals for #12, and I think we should find a better name for the graphics backend traits: they are not renderers, but rather the renderers will be built upon them (and may be part of smithay at some point too).

Drakulix:
What about "graphics backends" then?

vberger:
So `GlGraphicsBackend`, and `CpuGraphicsBackend` ? That can work yeah.

Drakulix:
Okay! Next round @vberger . I hope I have not forgotten anything.

src/backend/glium.rs
```diff
@@ -13,26 +15,41 @@ impl From<SwapBuffersError> for GliumSwapBuffersError
     }
 }
 
-impl<T: OpenglRenderer> Backend for T
+pub struct GliumGraphicBackend<T: OpenglGraphicsBackend>(T);
```
vberger: Why did you need to introduce this struct and trait?

vberger:
(Ugh, again github showing my review as "outdated".)

Anyway, I like it ! 👍 I just wonder why you had to introduce this wrapper for glium, but otherwise I think we can merge this as a good starting point.

We'll probably need to make these API change & evolve as we dig into the other functionalities and discover needs & constraints we had not anticipated anyway 😅

Drakulix:
I cannot do `impl<T: OpenglGraphicsBackend> Backend for T`, because then we implement an external trait for an exported trait. That means users of `smithay` could implement `OpenglGraphicsBackend` and get a free implementation of `Backend`, which might be unexpected, if they import `glium` as well. Thats why Rust does not allow that. We need to implement `Backend` for a specific struct, that is why I went with the wrapper approach, because I did not want to do the same implementation for both `glutin` headless and windowed structs and for every future implementation of `OpenglGraphicsBackend`.

vberger:
Oh right, I see. Forgot about coherence rules. All is right then. 👍

Drakulix:
Okay, let me fix the formatting and remove the WIP tag an then feel free to merge.

Drakulix:
Okay formatting is not fixed, because your travis config still uses rustfmt 0.7.*, which I did format with rustfmt 0.8. I guess we need to force install rustfmt or use something like [cargo-update](https://crates.io/crates/cargo-update).

vberger:
Oh. Yes, I guess we could add something like this to the travis.yml

```
- which cargo-install-update || cargo install cargo-update
- cargo install-update -a
```

Can you add a commit doing it in this PR?

vberger:
Hmm, I think we can remove the reporting of TODOs and FIXMEs. This is far too early for it... 😅

Drakulix:
Okay done. Lets hope that's it for this PR.

vberger:
The failure is clippy itself not compiling... welp, I'll merge.

Drakulix:
Maybe for the future (because this can happen with any new nightly) allow nightly builds to fail.

PR #14, opened 2017-04-09
keyboard: first draft of xkbcommon handling

vberger:
First draft of the keyboard handling with xkbcommon and forwarding events to the client.

The mechanism for choosing the keymap needs to be determined (`create_keyboard_handler` will change).

@Drakulix I'd like to have your opinion on the API and its implementation.

### DIFF
--- a/src/keyboard/mod.rs
+++ b/src/keyboard/mod.rs
@@ -1,0 +1,263 @@
+//! Utilities for keyboard handling
+//!
+//! This module provides utilities for keyboardand keymap handling: keymap interpretation
+//! and forwarding keystrokes to clients using xkbcommon.
+//!
+//! You can first create a `KbdHandle` using the `create_keyboard_handler` function in this module.
+//! The handle you obtained can be cloned to access this keyboard state from different places. It is
+//! expected that such a context is created for each keyboard the compositor has access to.
+//!
+//! This handle gives you 3 main way to interact with the keymap handling:
+//!
+//! - send the keymap information to a client using the `KbdHandle::send_keymap` method.
+//! - set the current focus for this keyboard: designing the client that will receive the key inputs
+//!   using the `KbdHandle::set_focus` method.
+//! - process key inputs from the input backend, allowing them to be catched at the compositor-level
+//!   or forwarded to the client. See the documentation of the `KbdHandle::input` method for
+//!   details.
+
+
+use backend::input::KeyState;
+use std::io::Write;
+use std::os::unix::io::AsRawFd;
+use std::sync::{Arc, Mutex};
+
+use tempfile::tempfile;
+
+use wayland_server::{Liveness, Resource};
+use wayland_server::protocol::{wl_keyboard, wl_surface};
+
+use xkbcommon::xkb;
+
+pub use xkbcommon::xkb::{Keysym, keysyms};
+
+/// Represents the current state of the keyboard modifiers
+///
+/// Each field of this struct represents a modifier and is `true` if this modifier is active.
+///
+/// For some modifiers, this means that the key is currently pressed, others are toggled
+/// (like caps lock).
+#[derive(Copy,Clone,Debug)]
+pub struct ModifiersState {
+    /// The "control" key
+    pub ctrl: bool,
+    /// The "alt" key
+    pub alt: bool,
+    /// The "shift" key
+    pub shift: bool,
+    /// The "Caps lock" key
+    pub caps_lock: bool,
+    /// The "logo" key
+    ///
+    /// Also known as the "windows" key on most keyboards
+    pub logo: bool,
+    /// The "Num lock" key
+    pub num_lock: bool,
+}
+
+impl ModifiersState {
+    fn new() -> ModifiersState {
+        ModifiersState {
+            ctrl: false,
+            alt: false,
+            shift: false,
+            caps_lock: false,
+            logo: false,
+            num_lock: false,
+        }
+    }
+
+    fn update_with(&mut self, state: &xkb::State) {
+        self.ctrl = state.mod_name_is_active(&xkb::MOD_NAME_CTRL, xkb::STATE_MODS_EFFECTIVE);
+        self.alt = state.mod_name_is_active(&xkb::MOD_NAME_ALT, xkb::STATE_MODS_EFFECTIVE);
+        self.shift = state.mod_name_is_active(&xkb::MOD_NAME_SHIFT, xkb::STATE_MODS_EFFECTIVE);
+        self.caps_lock = state.mod_name_is_active(&xkb::MOD_NAME_CAPS, xkb::STATE_MODS_EFFECTIVE);
+        self.logo = state.mod_name_is_active(&xkb::MOD_NAME_LOGO, xkb::STATE_MODS_EFFECTIVE);
+        self.num_lock = state.mod_name_is_active(&xkb::MOD_NAME_NUM, xkb::STATE_MODS_EFFECTIVE);
+    }
+}
+
+struct KbdInternal {
+    focus: Option<(wl_surface::WlSurface, wl_keyboard::WlKeyboard)>,
+    pressed_keys: Vec<u32>,
+    mods_state: ModifiersState,
+    _context: xkb::Context,
+    keymap: xkb::Keymap,
+    state: xkb::State,
+}
+
+impl KbdInternal {
+    fn new() -> Result<KbdInternal, ()> {
+        let context = xkb::Context::new(xkb::CONTEXT_NO_FLAGS);
+        // TODO: api to choose the keymap to load
+        let keymap = xkb::Keymap::new_from_names(&context, &"", &"", &"fr", &"oss", None, 0).ok_or(())?;
+        let state = xkb::State::new(&keymap);
+        Ok(KbdInternal {
+               focus: None,
+               pressed_keys: Vec::new(),
+               mods_state: ModifiersState::new(),
+               _context: context,
+               keymap: keymap,
+               state: state,
+           })
+    }
+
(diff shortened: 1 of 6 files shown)


src/keyboard/mod.rs
```diff
@@ -0,0 +1,258 @@
+//! Utilities for keyboard handling
+//!
+//! This module provides utilities for keyboardand keymap handling: keymap interpretation
+//! and forwarding keystrokes to clients using xkbcommon.
+//!
+//! You can first create a `KbdHandle` using the `create_keyboard_handler` function in this module.
+//! The handle you obtained can be cloned to access this keyboard state from different places. It is
+//! expected that such a context is created for each keyboard the compositor has access to.
```
Drakulix: So are we handling multiple keyboards with different layouts now?

src/keyboard/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+//!
+//! This handle gives you 3 main way to interact with the keymap handling:
```
Drakulix: *ways

src/keyboard/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// keyboard handler. It will internally track the state of the keymap.
+    ///
+    /// The `filter` argument is expected to be a closure which will peek at the generated input
+    /// as interpreted by the keymap befor it is forwarded to the focused client. If this closure
+    /// returns false, the input will not be sent to the client. This mechanism can be used to
+    /// implement compositor-level key bindings for example.
+    ///
+    /// The module `smithay::keyboard::keysyms` exposes definitions of all possible keysyms
+    /// to be compared against. This includes non-characted keysyms, such as XF86 special keys.
+    pub fn input<F>(&self, keycode: u32, state: KeyState, serial: u32, filter: F)
```
Drakulix: The filter function seems to be a little annoying. This will most likely be a normal function (not a closure) in the vast majority of use cases, because you will always send the same function/closure to handle key bindings.

I wonder if it would not be nicer to provide a single function on creation rather then passing a closure on every function call?

src/keyboard/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+            kbd.key(serial, 0, keycode, wl_state);
+        }
+    }
+
+    /// Set the current focus of this keyboard
+    ///
+    /// Any previous focus will be sent a `wl_keyboard::leave` event, and if the new focus
+    /// is not `None`, a `wl_keyboard::enter` event will be sent.
+    pub fn set_focus(&self, focus: Option<(wl_surface::WlSurface, wl_keyboard::WlKeyboard)>, serial: u32) {
+        // TODO: check surface and keyboard are from the same client
```
Drakulix: I am not sure I understand the concept of keyboard focus on wayland.
When and why do I have to call this function when implementing a compositor? When a window gains focus and should have keyboard focus as well? Also to map keyboards in a multi-seat environment to different windows?

src/keyboard/mod.rs
vberger: Well, each seat can have its own keyboard. So here, it's counting "logical keyboards" as the compositor sees them (as exposed by the input backend), not necessarily physical keyboard. It'd depend on the configuration of libinput (or the other input backends).

src/keyboard/mod.rs
vberger: Actually, I was thinking this `filter` would often be a closure: it'd handle the logic of compositor keybindings and to do that would likely need to capture some environment.

As I see it, having it provided at this level (rather than set once at creation) makes this capture easier and avoids the need to add more synchronisation primitives.

That's what I had in mind. Maybe it's not realistic from an usability point of view?

src/keyboard/mod.rs
vberger: By definition, keyboard focus sets where the keystrokes of the user go. So, the actual definition of when keyboard focus is switched may depend on the compositor.

On many desktop environments, the keyboard focus is moved by clicking on the target window, or by using some keybinding (often `alt+tab`). Others (often tiling WM) change the focus as the pointer moves on an other window.

Here, the `set_focus` method is to be called whenever the target surface for keystrokes should be changed (by the compositor's decision). It must be provided a tuple of the surface that gained focus, and the keyboard object (associated with the same client as this surface) that will be used for dispatching.

As such, the logic of tracking when keyboard focus changes is not provided in this PR. Only the logic for serializing keystrokes and keymap information to the clients.

src/keyboard/mod.rs
Drakulix: I see. When I did finish the libinput bindings, you will see that we maybe need to change some stuff about the `InputBackend` anyway, as libinput is in fact tracking single devices much better then seats (which is what we currently want to do in the `InputBackend`).

I guess we will figuring that out then.

src/keyboard/mod.rs
Drakulix: Well it does not seem to be efficient, because you would create a capturing closure for every call to `input` (not sure if the compiler may optimize that away).

But local environment is a good point. You will most likely call `input` from `on_keyboard_key` of your `InputHandler` implementation, so the filter closure will likely look like `|state, keysym| self.filter_keys(state, keysym)`, which would be much harder to do and would probably avoid lifetimes, if you would pass a static function.

src/keyboard/mod.rs
Drakulix: I see, so this is called as part of the whole direct-input-to-another-window procedure (also called "window focus") and especially targets the keyboard of a particular seat. (Or not if the compositor decides that window focus is an outdated concept :stuck_out_tongue: )

src/keyboard/mod.rs
vberger: Given it's a generic method here, we're going full static dispatch. I'd really expect the compiler to be able to optimize it properly.

Drakulix: Then I have zero objections to this one.

src/keyboard/mod.rs
vberger: The concept of "keyboard focus" is really part of the core wayland protocol, and a very classic behavior. Most wayland clients will expect this setup (for example, an app with multiple windows will likely want to know in which of them the user is typing).

If a compositor does not want to do it that way, well, they'd just no use this helper. That the whole philosophy I wand to put in smithay: provide helpers that are good for 90% of the use-cases, and the compositors built on top of it use only what they want/need. :)

src/keyboard/mod.rs
vberger: Not however that the "seat" concept is the way inputs are handled in the wayland protocol, so we cannot completely get rid of it.

For example, clients cannot handle two different keyboards i(with possibly different keymaps) if they are not on different seats.

Drakulix: Then I will need to do some work to track seat capabilities manually. Fun times :smile:

vberger: I must admit I'm surprised, from what I saw, it seemed like libinput was also built around this concept of seats... ?

Drakulix: It is. Every context is bound to one udev seat (physical seat), it even exposes seats as an object and it allows you to move devices between logical seats, which you can create for yourself. But a seat itself is nothing else then a group of devices sharing that seat and a matching libinput context.
(The only methods of a seat are `get_devices` and `get_context`.)


The are no seat capabilities, instead devices or device groups have certain capabilities and all those properties and you can only derive properties of a seat by receiving and merging the properties of all it's attached devices.

Drakulix: This also means libinput does not generate seat-events. I will have to track seats with each device-event. The library really seems to expect you to handle devices instead of seats like the wayland protocol.

vberger: I see... I guess it would be worth having a look at what WLC and Weston do to match libinput with the wayland seats then.

Anyway, this PR explicitly stops at the boundary of "one logical keyboard", be it a single one or multiple ones. So I guess we could move this discussion to a dedicated issue, as I think this is not blocking here?

Drakulix: Yes this is not blocking at all.

Drakulix: Quick heads up, it seems that wlc is not handling this at all. It simply tells clients, that one seat with keyboard, pointer and touch capabilities exists (like our glutin backend). I mean that works, but is not a very clean solution.

vberger: Weston on its side seems to just use the seats as they are defined by libinput: https://github.com/wayland-project/weston/blob/75b7197f4e072a4e2de124ddbe93b85cffb1c0f8/libweston/libinput-seat.c#L412-L423 (the logic is quite scattered around this file actually...)

My understanding is that it keeps an xkbcommon state for each seat, and fuses all keyboard of a seat into a single logical keyboard. That's quite closer to my initial understanding of how libinput works.

Drakulix: Yeah this is right and what we should do.
I was just pointing out, that to populate [this](https://github.com/vberger/smithay/blob/master/src/backend/input.rs#L18) correctly some work will be needed.

vberger: Oh, ok, I understand now.

That's right. on this point we can hardly take inspiration from weston. They're going full C-style here, with pointers all over the place.

Drakulix: Pointers everywhere

Drakulix:
So is this ready for merge? Or what are your plans on this?

vberger:
I want to fix the `create_keyboard_handler` function first, so that the keymap is not hardcoded.

Also, I'm attempting to see how alive xkbcommon-rs is: https://github.com/rtbo/xkbcommon-rs/pull/4

Drakulix:
Looks good :+1:

PR #15, opened 2017-04-14
WIP: libinput backend

Drakulix:
A first version of [Drakulix/input.rs](https://github.com/Drakulix/input.rs) is finally done and on crates.io.

So I started building the backend in smithay. It's not done yet, but feedback is already welcome and this should make tracking progress easier.


src/backend/libinput.rs
```diff
@@ -0,0 +1,147 @@
...
+                            let new_caps = SeatCapabilities {
+                                pointer: added.has_capability(DeviceCapability::Pointer),
+                                keyboard: added.has_capability(DeviceCapability::Keyboard),
+                                touch: added.has_capability(DeviceCapability::Touch),
+                            };
+
+                            let device_seat = added.seat();
+                            self.devices.push(added);
+
+                            let contains = self.seats.contains_key(&device_seat);
```
vberger: You could use the [entry API](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) here, to avoid redundant checks.

src/backend/libinput.rs
```diff
@@ -0,0 +1,147 @@
...
+
+                            // remove device
+                            self.devices.retain(|dev| *dev == removed);
+
+                            let device_seat = removed.seat();
+
+                            // update capabilities, so they appear correctly on `on_seat_changed` and `on_seat_destroyed`.
+                            if let Some(seat) = self.seats.get_mut(&device_seat) {
+                                let caps = seat.capabilities_mut();
+                                caps.pointer = self.devices.iter().any(|x| x.has_capability(DeviceCapability::Pointer));
```
vberger: Shouldn't you only count devices associated with this seat?

src/backend/libinput.rs
```diff
@@ -0,0 +1,147 @@
...
+                                caps.keyboard = self.devices.iter().any(|x| x.has_capability(DeviceCapability::Keyboard));
+                                caps.touch = self.devices.iter().any(|x| x.has_capability(DeviceCapability::Touch));
+                            } else {
+                                //TODO is this the best we can do?
```
vberger: Seems like the tight think to do in my opinion. This case would be a bug (be it in smithay, input.rs or libinput), and we want to catch it clearly rather than silently try to handle it.

src/backend/libinput.rs
```diff
@@ -0,0 +1,147 @@
...
+
+                            // check if the seat has any other devices
+                            if !self.devices.iter().any(|x| x.seat() == device_seat) {
+                                // it has not, lets destroy it
+                                if let Some(seat) = self.seats.remove(&device_seat) {
+                                    if let Some(ref mut handler) = self.handler {
+                                        handler.on_seat_destroyed(&seat);
+                                    }
+                                } else {
+                                    //TODO is this the best we can do?
```
vberger: Same here, this is the good thing to do.

vberger:
This starts quite well ! 😄 

A few general comments:

- I think it'd be good to integrate logging mechanism with slog (like I've started to do with shm handlers, and I'm currently working on adding it to other parts). It'll really help any future debugging. I have no real guideline about what to log when, so do as you feel would be most useful.
- This `HashMap` really feels like an attempt to do the same functionality as the `user_data` mechanism provided by libinput, but without using it. I think that should be okay, as it's accessed sufficiently rarely that it should not be a performance issue, I guess. And I understand why you prefer not to use it.

vberger:
Also, I've had a quick look at input.rs docs, the crate looks pretty neat! 👍

src/backend/libinput.rs
Drakulix: Good idea, but the thing is I also need a mutable reference to `self.seats` in line 91.

Otherwise I could also have done `if let Some(old_seat) = self.seats.get_mut(&device_seat) {` there.

Maybe I am missing something, but I don't think the Entry API will help here?

src/backend/libinput.rs
Drakulix: I could instead just count the number of devices that expose a certain capability.

Then again devices are grouped into `device_groups` in libinput and I did read in the libinput documentation (can't seem to find it now), that some devices may only expose certain capabilties if the complete group is available. Especially some combined physical devices (e.g. keyboard + touchpad combos) do this.

This means every removed device could change the capabilties of the remaining ones and this is also not as inefficient as it looks, as the amount of devices should not be very high.

I also think this would make the tracking code even uglier, so I would call your idea premature optimization. We should probably consider this at a later stage again and confirm what libinput actually does. (My thinkpad does some of this device_group stuff, so I should be able to test this.)

src/backend/libinput.rs
Drakulix: Yes that was my idea as well. I am going to remove these comments.

Drakulix:
> I think it'd be good to integrate logging mechanism with slog (like I've started to do with shm handlers, and I'm currently working on adding it to other parts). It'll really help any future debugging. I have no real guideline about what to log when, so do as you feel would be most useful.

Will do

> This HashMap really feels like an attempt to do the same functionality as the user_data mechanism provided by libinput, but without using it. I think that should be okay, as it's accessed sufficiently rarely that it should not be a performance issue, I guess. And I understand why you prefer not to use it.

It exactly is that attempt. There are also some other reasons, other then the userdata api being unsafe and flawed. Because the libinput `Device` type is exposed via the `InputConfig`, it is possible for the user of smithay to get the `Seat` and `Libinput` context of a `Device` as well. Using the userdata api would allow users to access this internal information and also disallow then to store their own.

Drakulix:
> Also, I've had a quick look at input.rs docs, the crate looks pretty neat! :+1:

Thanks! Although I have mostly copy&pasted libinput's own docs. ;)

src/backend/libinput.rs
vberger: No, I was not talking about the performance here.

I may be wrong, but it seems you are computing the capabilities of the seat using all devices, even the ones not associated to this seat? If it's the case, the result could be wrong.

Drakulix: Oh right, good catch!
I will need to filter these before.

src/backend/libinput.rs
vberger: The entry API allows you to insert a value at the key you provided.

Then, given the branches of this if are completely different, the gain would not be that notable. But if the creation event was separated from advertising capabilities, we could have done something like:

```rust
let seat = self.seats.entry(device_seat.clone()).or_insert_with(|| {
    // create the seat object adn return it
});
// compute capabilities and send update event  with seat
```

Drakulix:
Lots and lots of new stuff rebased on current master.

Feedback pls. :)

src/backend/libinput.rs
```diff
@@ -0,0 +1,329 @@
...
+                                trace!(self.logger, "Calling on_pointer_move with {:?}", desc.pointer);
+                                handler.on_pointer_move(&desc.seat, motion_event.time(), desc.pointer);
+                            }
+                        },
+                        PointerEvent::MotionAbsolute(motion_event) => {
+                            let device_seat = motion_event.device().seat();
+                            let desc = self.seats.get_mut(&device_seat).expect("Recieved pointer event of non existing Seat");
+                            desc.pointer = (
+                                motion_event.absolute_x_transformed(
+                                /*FIXME: global.get_focused_output_for_seat(&desc.seat).width() or something like that*/ 1280) as u32,
```
vberger: I think the easiest and least intrusive way to handle this would be a `set_output_metadata()` method on the `LibinputInputBackend`, that the compositor should call whenever the cursor changes screen, maybe?

Do we know in which cases this kind of event (`MotionAbsolute`) is generated?

src/backend/libinput.rs
```diff
@@ -0,0 +1,329 @@
...
+
+impl From<::input::event::touch::TouchEvent> for ::backend::input::TouchEvent {
+    fn from(libinput: ::input::event::touch::TouchEvent) -> Self {
+        use ::input::event::touch::{TouchEventSlot, TouchEventPosition};
+        use ::backend::TouchSlotInternal;
+
+        match libinput {
+            ::input::event::touch::TouchEvent::Down(down_event) => ::backend::input::TouchEvent::Down {
+                slot: down_event.slot().map(|x| ::backend::input::TouchSlot::new(x)),
+                x: down_event.x_transformed(/*FIXME: global.get_focused_output_for_seat(&desc.seat).width() or something like that*/ 1280),
```
vberger: Same here (`set_output_metadata()`).

Also, we'll probably want to handle cases when a touchscreen is associated with an output forever. For example in the case of a touchscreen. Not sure how we should handle that...

vberger:
Wow, you've been quite busy yesterday!

Other than the few comments I've made, I've not much to say, it sure looks good!

src/backend/libinput.rs
```diff
@@ -0,0 +1,329 @@
...
+        }
+    }
+
+    fn input_config(&mut self) -> &mut Self::InputConfig {
+        &mut self.devices
+    }
+
+    fn set_cursor_position(&mut self, _x: u32, _y: u32) -> Result<(), ()> {
+        // FIXME later.
+        // This will be doable with the hardware cursor api and probably some more cases
```
vberger: "hardware cursor API" ? Is it something provided by libinput?

src/backend/libinput.rs
Drakulix: No we sadly don't. Some pointer devices like a mouse or a trackpoint will always send a relative movement, while some devices like a touchscreen/tablet emulating a mouse (which is something some devices do in hardware) will always send an absolute event.

set_output_metadata sounds like a good idea. I wonder if this should be added to the `InputBackend` in general. A glutin input handler might also control non-glutin outputs/screens and therefor needs this information.

src/backend/libinput.rs
Drakulix: I would say this is something a concrete implementation should handle. Just don't call `set_output_metadata` in that case.

src/backend/libinput.rs
Drakulix: Nope. This is something provided by linux and I am not sure yet how to handle this. It will mostly likely only apply to the `DRM` backend, then again any `InputBackend` should not know anything about the underlying render implementation. (Also see http://stackoverflow.com/questions/6957039/what-is-hardware-cursor-and-how-does-it-work)

Maybe we should move `set_cursor_position` to the graphics backends?

Drakulix: So I think it would make sense to move the set_cursor api to the graphics backend,
as the hardware cursor api is provided via libdrm (https://github.com/rust-kms/libdrm-rs/blob/318339bb8fdce4d0838dba41d289899d65c3eb8e/src/ffi/xf86drm_mode.rs#L443).

I am just not sure, how this would effect the glutin backend. For x11 this should be no problem whats so ever, but for wayland it might be. Does the wayland implementation in glutin currently handle this? Is it possible for a client to move the cursor within its bounds?

Drakulix: If I interpret `wl_pointer::set_cursor` correctly we can make it look like the pointer is in a different position in our compositor? (https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_pointer)

I would argue that is good enough for windowed backends.

Drakulix: Also there is the [pointer constraints protocol](https://github.com/wayland-project/wayland-protocols/blob/master/unstable/pointer-constraints/pointer-constraints-unstable-v1.xml), at least supported by weston. If glutin does not support it right now, this is not a burden smithay should carry, but rather something that should be fixed in glutin instead.

Drakulix:
I went ahead and implemented my purposed changes. If you have reservations I can revert/change them, not a big deal.

I am thinking if `set_pointer_position` should be changed to `set_pointer` to set a given pointer image in addition to the position. Maybe even call it `draw_pointer` because that is essentially the purpose of this function now.

vberger:
I'm a little puzzled regarding pointer position.

On the one hand, I'd tend to think handling the pointer should be entirely done by the code using the input and graphics backends, and that they should be oblivious to it. The input backend should just forward events directly, be they relative of absolute (but tag them as such). And the graphics backend should not know about it, the using code should draw the cursor itself.

That's for example what weston does: when ran as an X11 client, it hides the X11 cursor and draws its own (you can see that when you take the cursor out of weston's window: the weston cursor remains stuck on the internal edge, plus it does not look like the X11 cursor on my setup). Regarding this, I think we should do the same thing (hide the cursor and the the compositor built on smithay draw it itself).

Now, taking this route means that some backends will always forward absolute events, which can obviously be a problem for some functionality. The only backend that would really play nice with that is libinput's. And tbh, I tend to think it's the most important one.

If I recap:

- libinput/tty backend should really not be aware of cursor position, and let the user code handle it completely
- wayland backend would not allow to teleport cursor at all
- x11 backend would allow to teleport cursor but require a special function to do it
- glutin backend can teleport cursor, but only when run on x11, and require a special function to do it

Tbh, I'd tend to think the cursor teleportation logic should not be in the backend traits, but specialized methods on the backends that require/support it. But this would break the idea of writing code generic and able to handle it. I don't see an obvious satisfactory solution here. :(

src/backend/input.rs
```diff
@@ -216,6 +231,9 @@ pub trait InputBackend: Sized {
     /// Get current `InputConfig`
     fn input_config(&mut self) -> &mut Self::InputConfig;
 
+    /// Called to inform the Input backend about a new focused Output for a `Seat`
+    fn set_output_metadata(&mut self, seat: &Seat, output: &Output);
```
vberger: If `Output` is a trait, can I'd rather have this method generic to avoid virtual dispatch.

But couldn't this method just be `fn set_output_metadata(&mut self, width: u32, height: u32)` ?

Also, the question of device giving absolute positions (such as touchscreens) and locked to one output, independently of pointer focus remains.

Drakulix:
You maybe missed my previous comments, as they are now hidden again.

> If I interpret wl_pointer::set_cursor correctly we can make it look like the pointer is in a different position in our compositor? (https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_pointer)
> 
> I would argue that is good enough for windowed backends.

> Also there is the pointer constraints protocol, at least supported by weston. If glutin does not support it right now, this is not a burden smithay should carry, but rather something that should be fixed in glutin instead.

>Tbh, I'd tend to think the cursor teleportation logic should not be in the backend traits, but specialized methods on the backends that require/support it. But this would break the idea of writing code generic and able to handle it. I don't see an obvious satisfactory solution here. :(

I think the we should not think about teleporting the cursor, but instead about "drawing" the cursor. In both X11 and Wayland (at some point / with some controls) we could just grab the cursor and draw a software cursor. In the tty/drm backend we just call [`drmModeSetCursor2`](https://github.com/rust-kms/libdrm-rs/blob/318339bb8fdce4d0838dba41d289899d65c3eb8e/src/ffi/xf86drm_mode.rs#L443), which also draws the cursor.

If wayland (or glutin on wayland) does currently not work like that (yet), I would just ignore that for now. This is the best backend agnostic api, I could come up with.

src/backend/input.rs
Drakulix: Generics would make `InputHandler` not object-safe, which is problematic for the `InputBackend` (just try it, this was the only solution I did found).

I though this would be more extensible and gives a nice API for the backends (you could just implement `Input::Output` for the `GraphicsBackend`s specific output type). I also think we have an API hole here, currently it looks like every `GraphicsBackend` is meant to be exactly one output, which will not work with DRM. I have some idea for that, but that belongs into a seperate design issue.

> Also, the question of device giving absolute positions (such as touchscreens) and locked to one output, independently of pointer focus remains.

I would say this is something a concrete implementation should handle. Just don't call set_output_metadata more then once for that output in that case.

vberger:
Hmm. Okay, I had missed the point with `dmrModeSetCursor2`. Didn't realize the drm modesetting backend of linux had a native support of a cursor. Sorry, it all makes sense now.

Yes, cursor grabbing would definitely be the solution in this case. But I still think the input backend itself should not hold any cursor-tracking logic. Absolute cursor moves and relative cursor moves should be forwarded to the handler directly, letting it decide what to do with them. And update the cursor position in the graphics backend if appropriate.

I would say it's fine to assume we will be able to do cursor grabbing in all cases (even if it's not yet supported with glutin on wayland, I expect it will be at some point).

The graphics backend would thus gain 2 methods, one to change the cursor image, and one to move the cursor (mimicking `dmrModeSetCursor2` and `drmModeMoveCursor`) ?

Drakulix:
Sounds very reasonable.
I will rework the pointer handling in regards to absolute and relative events and change the graphics backend.

src/backend/input.rs
vberger: Do we expect the input backend to need more information than just the dimensions of the output currently focused?

Also, we could go an other route: as the result of "transformed" values from libinput are floats, could we just use `(width, height) = (1,1)` and define the result from the backend to be exprimed as a ration value relative to the total size of the output, and let the handler deal with that ?

Drakulix: I expect to get wired behavior then like [some bugs](https://github.com/Cloudef/wlc/issues/181) in wlc, I did encounter, because of floating point rounding.

I don't know if any other information are necessary. I would just go with `fn set_output_metadata(&mut self, width: u32, height: u32)` then and just refactor it, if we ever need more information.

vberger: Ah, yes I see...
In this case, I'm thinking the output metadata should probably be a per-device property rather than per-seat. See, in a context of multi-output:

- pointer devices can change of "currently focused output", and as such the resolution to consider can change depending on where the pointer is in compositor space
- touchscreens on the other hand, are likely fixed to a single output, and their output resolution will not depend on where the pointer currently is
- tablet devices can, on x11, be configured to be either fixed with a single output, or cover the entire screen space covering all outputs. I'd like to keep this behavior.

Drakulix: For both touchscreen and tablet devices the screen size is still important in that case.

I would argue, that the usual expected behavior is to have one focused output per seat. When a tablet is used, wouldn't you just move the focus to the fixed output in that case? In that case the only API error is, that you cannot change the output metadata anymore, once you have received the event.

Maybe we should change the `on_pointer` callback then to receive an event like the `TouchEvent` we already have? Then those could implement something like `x_transformed` themselves instead of handling this internally in the backend.

vberger: If we can replicate libinput's API on the events, and make the transformation an event method, that would be the best in my opinion. This would completely eliminate the need for `set_output_metadata`, which is good. (I really would like the scope of the backends to be as small as possible, we can always build more elaborate abstractions on top of them later.)

It also would directly solve the question of devices linked to an output, as the inputbackend would not have to worry about it.

I guess to make it work, TouchEvent and PointerEvent would need to become traits, with concrete types being backend-dependant?

Drakulix: We could convert them, like I did for the `TouchEvent` already. That makes matching easier and I don't think the current `TouchEvent` is missing something critical. I could also model a backend-agnostic `PointerEvent` based on libinputs event.

I fear to many generics and traits will make it look much less platform-independently then it actually is.

Drakulix: Thinking about it, I would need to re-implement the transform methods in that case, as there is no generic representation.
I think we should use traits for these events.

Should I make everything event based in that case including tablet,keyboard and such?

vberger: Let's see how it gets yeah. The `InputBackend` trait would then gain 3 associated types, for pointer, touch and keyboard events?

Drakulix: Yes seems like it

Drakulix:
For `setCursorImage` I was thinking for adding a dependency on the `image` crate? It provides a very solid [`ImageBuffer`](https://docs.rs/image/0.12.3/image/struct.ImageBuffer.html) implementation, that would allow checking the format and all.

vberger:
I'm ok with it. If it proves cumbersome we'll still be able to easily change it afterwards (much more easily than the profound API design questions).

src/backend/input.rs
Drakulix: I am think about make much more then three types out of them.

Not being able to match the `PointerEvent` feels so wrong. Then every method would return an `Option`,  because e.g. `PointerAxisEvents` do not have a `motion` or `position` property.

So I would go with associated types for:
`KeyboardKeyEvent`
`PointerAxisEvent`
`PointerButtonEvent`
`PointerMotionEvent`
`PointerMotionAbsoluteEvent`
`TouchDownEvent`
`TouchUpEvent`
`TouchMotionEvent`
`TouchCancelEvent`
`TouchFrameEvent`

Drakulix: Do you think this is too much or do you maybe have a better idea, how this could be modeled?

vberger: I'm not sure. It seems to be quite a lot indeed, but on the other hand, I'm not sure what other alternative we could have to handle this properly...

Drakulix:
Ok there are a bunch of clippy errors remaining and I am too tired to read over all of the code again, like I would usually do, but I wanted to push this out, as my free time might be even more limited in the next days (although I don't know for sure).

I think api design wise, this is now "final" (until we encounter the next api hole we need to fix). If you could take a look at it, that would be nice.

Also I was thinking about removing the `SeatInternal` and `TouchSlotInternal` traits. These are the only missing functions to implement additional backends outside of smithay and I think that should be allowed. Maybe I should open another issue about that.

vberger:
I'll see if I can find some time to look into that. These three days will be quite busy for me, so this may have to wait a little.

Drakulix:
Alright, some small fixes here and there. I would feel comfortably seeing this getting merged now. Let me know what you think, once you got some free time. :)

vberger:
This is taking a really nice shape! 👍 

Just a few nits / questions:

- What's the reason for implementing all the events traits for `()`, and with `unreachable!()` as a content?
- The `TouchDownEvent` and `TouchMotionEvent` traits looks identical, would it make sens to fuse them? Same question for `TouchUpEvent` and `TouchCancelEvent`?
- (this one is secondary and can be handled afterwards) I think we could organize the `backend` module and its file differently. It feels weird how the graphics part is in its own subfolder and the input part is not...

Drakulix:
- Reasoning is that not every backend may issue all event types. E.g. glutin does not send `PointerMotionEvent`s just `PointerMotionAbsoluteEvent`s and just falls back to `()` in these cases (instead of providing it's own empty implementation of `PointerMotionEvent`). However accidentally sending `()` (e.g. by a misplaced `;`) is still considered an implementation error, which is why I used `unimplemented!()` to stub these unused functions, which will lead to a panic.
- I would not fuse them in case they get more diverse. These are separate types in c-libinput and possibly in other backends and refactoring these later to revert the fuse would cause more api breakage then just adding a new function to only one of these traits.
- Well these modules are just different. I am not sure how I could easily split the input module, while the graphics module has some a lot of types not depending on each other (which is my primary reason for splitting). For user-imports and documentation they appear equal right? So I think this is not a huge issue.

vberger:
- I see. This is the kind of things that would beg for the `!` type to be stabilized. In this case, I'd in favor of making this more explicit, like by creating an empty enum `enum NeverOccuringEvent { }`. This adds the interesting property that, as these kind of types are not instanciable, this gives a lot of static guarantees and allows a lot of dead-code elimination for rustc, as well as preventing accidentally passing a `()`.
- Seems reasonable. I've nothing more to say about this.
- This is not in scope of this PR, I'll think more about it.

Drakulix:
Alright, sounds like an even better solution. I will fix that asap (probably today).

Drakulix:
Done

src/backend/input.rs
```diff
@@ -77,7 +77,17 @@ pub trait Event {
...
-impl Event for () {
+/// Struct to mark events never emitted by an `InputBackend` implementation.
+///
+/// Implements all event types and can be used in place for any `Event` type,
+/// that is not used by an `InputBackend` implementation. Initialization is not
+/// possible, making accidential use impossible and enabling a lot of possible
+/// compiler optimizations.
+pub struct UnusedEvent {
+    _hidden_field: (),
+}
```
vberger: Just use `pub enum UnusedEvent {};`, an empty enum. This would be statically guaranteed to never be instanciable, and rustc understands this.

Also, all the impls of methods for this type would become:

```rust
fn foo(&self) -> ... {
    match *self {}
}

Drakulix: Any particular reason to use `match *self {}` instead of `unimplemented!()`?

vberger: Mostly it makes everything lighter by not pulling in the whole panic machinery for a panic that is in a dead code-path anyway.

Drakulix:
aaaaand done.

vberger:
Looks like you forgot a few `unreachable!()` in the impls for `UnusedEvent`. 😉 

Other than that, all is good for me.

Drakulix:
forgot to save...

vberger:
Alright, perfect!

When the `void_type` rustc feature gets stabilized, we'll be able to change `UnusedEvent` definition to `type UnusedEvent = !;` for maximum explicitness. 😄

Do you have anything to add, or can I merge?

Drakulix:
Nothing to add for this pull request :)

Issue #12, opened 2017-03-17
Handle client EGL-backed buffers


vberger:
This will require:

- [ ] Understanding the mesa<->wayland compositor integration
- [ ] Expose a way to fetch the contents of an EGL buffer in a way directly usable to render on OpenGL backends

vberger:
Ooookay, looks like we'll encounter `user_data` issues again, this time with mesa. It uses it in its `wl_buffer`s.

*sigh*

Timidger:
:( I do wish Wayland was more amendable to the Rust memory model. Doubly so because compositors control so much, I would feel much more secure letting my all-knowing compositor be written in something like Rust that is much less likely to have memory exploits than in C.

Drakulix:
Mesa puts data into wl_buffers user data and expects the compositor to read it? This sounds completely broken and like something that should be handled by a real protocol. How is the compositor supposed to use this "user data", if it has to rely on it for communication..

vberger:
No, mesa do not expect anyone to read its user data, but rather uses it internally.

However, `wayland_server` (and `wayland_client`) rely so much on the `user_data` field to maintain the safety of the bindings that I had to make the assumption that these libs would be the only source of object creation, and thus that I would exclusively control this field.

With mesa, this assumption is obviously false. I'm thinking about how I can relax it.

Drakulix:
This is still unreasonable. How are you expected to ever use the `user_data` field on a `wl_buffer`, if almost any compositor needs to work together with mesa.

vberger:
Hmm, actually I *think* I found a way to discriminate between objects managed by my crates vs other crates.

I'll need to refactor a bit wayland-client/server (and release a new breaking version I think), but this would solve it once and for all. And probably solve the compability with WLC issues, maybe. I need to dig into that more.

vberger:
I *think* this would work: https://github.com/vberger/wayland-rs/pull/98.

vberger:
Okay. This will be funky. From what I gathered from weston and WLC, the magic stuff comes from the `EGL_WL_bind_wayland_display` extension (implying that we require an EGL context).

From this, we can retreive the contents of a `wl_buffer` coming from an OpenGL client into an `EGLImage`.

Looks like this `EGLImage` must then be drawn onto an EGL texture using `glEGLImageTargetTexture2DOES`...

To be honest, I'm completely discovering all this :D

I think, as a first step, I'll setup an API to expose the EGLImages.

[...]

PR #17, opened 2017-05-18
EGL Rework

Drakulix:
Alright, it is late and I am not sure, if I did not miss anything important.
Docs are also largely missing.
Other then that, I think this is good for review/first feedback.

Partly addresses #12.


Cargo.toml
```diff
@@ -5,22 +5,27 @@ authors = ["Victor Berger <victor.berger@thalesgroup.com>"]
...
-wayland-server = "0.9.1"
+wayland-server = "0.9.4"
 nix = "0.7.0"
 xkbcommon = "0.2.1"
 tempfile = "2.1.5"
 slog = { version = "2.0.0" }
 slog-stdlog = "2.0.0-0.2"
-glutin = { version = "~0.7.4", optional = true }
+libloading = "0.4.0"
+wayland-client = { version = "~0.8.6", optional = true }
```
vberger: Having wayland-client 0.8 and wayland-server 0.9 actually compiles? O.o

Cargo.toml
```diff
@@ -5,22 +5,27 @@ authors = ["Victor Berger <victor.berger@thalesgroup.com>"]
...
+gl_generator = "0.5"
+
 [dev-dependencies]
 slog-term = "~1.5"
 
 [features]
-default = ["backend_glutin", "backend_libinput", "renderer_glium"]
-backend_glutin = ["glutin", "wayland-server/dlopen"]
+default = ["backend_winit", "backend_libinput", "renderer_glium"]
+backend_winit = ["winit", "wayland-server/dlopen", "wayland-client", "wayland-client/dlopen"]
```
vberger: I don't think we need both `wayland-client` and `wayland-client/dlopen` here, do we?

src/backend/glutin.rs
```diff
@@ -21,7 +21,7 @@ use std::rc::Rc;
 /// Create a new `GlutinHeadlessRenderer` which implements the `OpenglRenderer` graphics
 /// backend trait
 pub fn init_headless_renderer() -> Result<GlutinHeadlessRenderer, CreationError> {
-    init_headless_renderer_from_builder(HeadlessRendererBuilder::new(1024, 600))
+    init_headless_renderer_from_builder(HeadlessRendererBuilder::new(1024, 600)
```
vberger: unclosed delimiter `(` 😛

src/backend/graphics/egl.rs
```diff
@@ -0,0 +1,662 @@
+//! Common traits and types for opengl rendering on graphics backends
+
+/// Large parts of the following file are taken from
+/// https://github.com/tomaka/glutin/tree/master/src/api/egl at commit
+/// `044e651edf67a2029eecc650dd42546af1501414`
+///
+/// It therefor falls under glutin's Apache 2.0 license
+/// (see https://github.com/tomaka/glutin/blob/master/LICENSE)
```
vberger: "therefor**e**"

Also I guess this should be good enough. But to be safe, let's ask @tomaka his opinion on the matter.

src/backend/graphics/egl.rs
```diff
@@ -0,0 +1,662 @@
...
+
+        include!(concat!(env!("OUT_DIR"), "/egl_bindings.rs"));
+    }
+}
+
+#[derive(Clone, Copy)]
+pub enum Native {
+    X11(ffi::NativeDisplayType, ffi::NativeWindowType),
+    Wayland(ffi::NativeDisplayType, ffi::NativeWindowType),
+    Gbm(ffi::NativeDisplayType, ffi::NativeWindowType),
```
vberger: Is this what will be used for the native/modesetting backed?

src/backend/winit.rs
```diff
@@ -0,0 +1,655 @@
...
+    }
+
+    /// Processes new events of the underlying event loop to drive the set `InputHandler`.
+    ///
+    /// You need to periodically call this function to keep the underlying event loop and
+    /// `Window` active. Otherwise the window may no respond to user interaction and no
+    /// input events will be received by a set `InputHandler`.
+    ///
+    /// Returns an error if the `Window` the window has been closed. Calling
+    /// `process_new_events` again after the `Window` has been closed is considered an
```
vberger: `dispatch_new_events`

src/backend/glutin.rs
vberger: Actually, is this file needed at all now?

vberger:
I don't have much to say other than this. This is quite a lot of boilerplate to plug everything together, not the easiest thing to review. I'll try to give a few more passes on it.

Also, making the travis build pass would be a good thing to do as well. 😉

src/backend/graphics/egl.rs
tomaka: Well, yeah, I don't see a problem with this.

Cargo.toml
Drakulix: Yes when I set both to use dlopen.
And wayland-client 0.8 is sadly required, as winit is still expecting that.

Cargo.toml
Drakulix: See comment above.

src/backend/glutin.rs
Drakulix: No, I just forgot to remove it.

src/backend/graphics/egl.rs
Drakulix: Yes. This was the main motivation of doing this change now, as this is essential to make anything useful out of the drm implementation I am working on.

Cargo.toml
vberger: But, don't `wayland-client/dlopen` imply `wayland-client` ?

Drakulix:
Thanks for the first review. Typos, Docs and Travis are definitely my main priority today.

Cargo.toml
Drakulix: Not sure, good point. I will test that.

Drakulix:
Alright, I should have addressed most of the things.

The example is still not working right (immediately closes on wayland and panics on x11), but its getting somewhere.

vberger:
> The example is still not working right (immediately closes on wayland and panics on x11), but its getting somewhere.

Closes without panicking on wayland? Well, that's weird, is it some segfault or stuff like that?

src/backend/winit.rs
```diff
@@ -0,0 +1,652 @@
...
+    let events_loop = EventsLoop::new();
+    let window = Rc::new(builder.build(&events_loop)?);
+
+    let (native, surface) = if let (Some(conn), Some(window)) =
+        (get_x11_xconnection(), window.get_xlib_window()) {
+        (Native::X11(conn.display as *const _, window), None)
+    } else if let (Some(display), Some(surface)) =
+        (window.get_wayland_display(), window.get_wayland_client_surface()) {
+        let (w, h) = window.get_inner_size().unwrap();
+        let egl_surface = wegl::WlEglSurface::new(surface, w as i32, h as i32);
```
vberger: You'll want to keep this `egl_surface` around, as dropping it destroys the EGL_surface, and you'll need it to handle resizing of the surface.

src/backend/winit.rs
```diff
@@ -0,0 +1,652 @@
...
+            let surface = &self.surface;
+            let mut handler = self.handler.as_mut();
+
+            self.events_loop
+                .poll_events(move |event| if let Some(ref mut handler) = handler {
+                                 let Event::WindowEvent { event, .. } = event;
+                                 match event {
+                                     WindowEvent::Resized(x, y) => {
+                                         window.set_inner_size(x, y);
+                                         if let Some(wl_surface) = surface.as_ref() {
```
vberger: To avoid confusion, this is a `wl_egl_surface`, not a `wl_surface`

vberger:
I'd like to have the example running without crashing before merging, as this is likely some weird bug somewhere and we'll probably regret not hunting it if we merge it now.

Drakulix:
Alright, I will hunt it.
No segfault, it just closes and returns a bad error code. Maybe we should integrate `dispatch_new_events` into the `event_loop` somehow. Probably with a timer.

vberger:
I mead, given the source of this example, it finishing by an other way than panicking is very weird, and likely not a good sign at all...

Drakulix:
Well testing on wayland is quite painful rightnow for me, because I need to switch my DE (see https://github.com/Cloudef/wlc/issues/235), which is why I was hoping, that you might have an idea.

vberger:
What I do, is I run weston as an X11 client, and then start my wayland apps as weston clients.

I can try to run your code like this this evening on my computer, if you want.

Drakulix:
> What I do, is I run weston as an X11 client, and then start my wayland apps as weston clients.
Good idea.

> I can try to run your code like this this evening on my computer, if you want.
This will probably be a rather nasty bug, so if you are willing to help debugging (even just by testing), that would be highly appreciated.

src/backend/winit.rs
```diff
@@ -0,0 +1,652 @@
...
+
+    let context = unsafe {
+        EGLContext::new(native,
+                        attributes,
+                        PixelFormatRequirements {
+                            hardware_accelerated: Some(true),
+                            color_bits: Some(24),
+                            alpha_bits: Some(8),
+                            ..Default::default()
+                        })?
```
Drakulix: Okay I have debugged it. This line causes a segfault, if context creation fails (why it fails is a different story, see other comment). The reason is, that the `wl_egl_surface` is dropped, which causes the segfault in this specific case. Any thoughts @vberger ?

src/backend/winit.rs
```diff
@@ -0,0 +1,652 @@
...
+/// graphics backend trait, from a given `WindowBuilder` struct and a corresponding
+/// `WinitInputBackend`, which implements the `InputBackend` trait
+pub fn init_from_builder(builder: WindowBuilder)
+                         -> Result<(WinitGraphicsBackend, WinitInputBackend), CreationError> {
+    init_from_builder_with_gl_attr(builder,
+                                   GlAttributes {
+                                       version: None,
+                                       profile: None,
+                                       debug: cfg!(debug_assertions),
+                                       robustness: Robustness::TryRobustLoseContextOnReset,
```
Drakulix: This causes context creation to fail, because apparently my EGL implementation reports, that it supports robustness, but creation later fails with `BadAttribute`, if it is set to any other value then `NotRobust`.

vberger: Okay, this is far beyond my knowledge of EGL 😅

src/backend/winit.rs
vberger: if the `wl_egl_surface` is dropped, this destroys the EGL surface, and the pointer you retrieve line 82 becomes invalid. Could this be the reason?

Drakulix: The crash happens in `wl_egl_window_destroy` though.

src/backend/winit.rs
Drakulix: It just sounds like a typical driver bug. I have changed it to default to `NotRobust` now.

src/backend/winit.rs
vberger: Uh. So, that means the EGL surface ends up in an invalid state following the failed initialization, and then it blows up when it is destroyed, I'd guess?

src/backend/graphics/egl.rs
```diff
@@ -0,0 +1,746 @@
...
+
+        let surface = {
+            let surface = match native {
+                Native::X11(_, window) |
+                Native::Wayland(_, window) |
+                Native::Gbm(_, window) => egl.CreateWindowSurface(display, config_id, window, ptr::null()),
+            };
+
+            if surface.is_null() {
+                return Err(CreationError::OsError(String::from("eglCreateWindowSurface failed")));
```
vberger: This check should be done in the `Gbm` branch of the match, otherwise providing a null pointer in input with one of the others backend may trigger it...

An other assert checking that they are not null can be added to replace this one, too.

src/backend/winit.rs
Drakulix: Probably. I will check, what kind of cleanup should be done in this case.

src/backend/graphics/egl.rs
Drakulix: Are you aware, that the match statement, does the same for all backend types? The rustfmt format might have tricked you there.

vberger: Oh! Right, indeed. My bad. 😅

src/backend/graphics/egl.rs
```diff
@@ -0,0 +1,746 @@
...
+            }
+            surface
+        };
```
vberger: Good news, I found the cause for the segfault!

The reason is, because of the reentrancy of this function(see the match on line 146, it calls itself), this par ends up called twice. Especially, `egl.CreateWindowSurface()` is called twice on the same native surface, and apparently this is not good.

The simplest fix is to move this whole `let surface = { .... }` block down in the function, right before the end, until after the EGL context has been successfully created (so line 500 of this diff). :)

vberger: Note it only solves the segfault, not the panic 😅

vberger:
👍

PR #19, opened 2017-06-04
First draft of the compositor/subcompositor global handler

vberger:
There are still a some things to do before merging.

Two requests of `wl_surface` are missing:

- [x] `wl_surface.frame`
- [x] `wl_surface.commit`

I plan make the user provide a Handler to directly retrieve these, as they cannot be simply accumulated for double-buffering like the others.

Two requests of `wl_subsurface` are missing:

- [x] `wl_subsurface.place_above`
- [x] `wl_subsurface.place_below`

I'm not yet sure how to handle these, the current tree structure could handle ordering of the children of a surface, but we need to _also_ handle ordering the parent relative to each of its children.

EDIT: subsequent needs:

- [x] documentation
- [x] logging
- [ ] find a clean way to handle ordering of the children relative to the parent ? (I can think of ugly ways to do so, but...)

### DIFF
--- a/examples/helpers/glium.rs
+++ b/examples/helpers/glium.rs
@@ -1,0 +1,116 @@
+use glium;
+use glium::Surface;
+use glium::index::PrimitiveType;
+
+#[derive(Copy, Clone)]
+struct Vertex {
+    position: [f32; 2],
+    tex_coords: [f32; 2],
+}
+
+implement_vertex!(Vertex, position, tex_coords);
+
+pub struct GliumDrawer<'a, F: 'a> {
+    display: &'a F,
+    vertex_buffer: glium::VertexBuffer<Vertex>,
+    index_buffer: glium::IndexBuffer<u16>,
+    program: glium::Program,
+}
+
+impl<'a, F: glium::backend::Facade + 'a> GliumDrawer<'a, F> {
+    pub fn new(display: &'a F) -> GliumDrawer<'a, F>
+    {
+
+        // building the vertex buffer, which contains all the vertices that we will draw
+        let vertex_buffer = glium::VertexBuffer::new(display,
+                                                     &[Vertex {
+                                                           position: [0.0, 0.0],
+                                                           tex_coords: [0.0, 0.0],
+                                                       },
+                                                       Vertex {
+                                                           position: [0.0, 1.0],
+                                                           tex_coords: [0.0, 1.0],
+                                                       },
+                                                       Vertex {
+                                                           position: [1.0, 1.0],
+                                                           tex_coords: [1.0, 1.0],
+                                                       },
+                                                       Vertex {
+                                                           position: [1.0, 0.0],
+                                                           tex_coords: [1.0, 0.0],
+                                                       }])
+                .unwrap();
+
+        // building the index buffer
+        let index_buffer =
+            glium::IndexBuffer::new(display, PrimitiveType::TriangleStrip, &[1 as u16, 2, 0, 3]).unwrap();
+
+        // compiling shaders and linking them together
+        let program = program!(display,
+			100 => {
+				vertex: "
+					#version 100
+					uniform lowp mat4 matrix;
+					attribute lowp vec2 position;
+					attribute lowp vec2 tex_coords;
+					varying lowp vec2 v_tex_coords;
+					void main() {
+						gl_Position = matrix * vec4(position, 0.0, 1.0);
+						v_tex_coords = tex_coords;
+					}
+				",
+
+				fragment: "
+					#version 100
+					uniform lowp sampler2D tex;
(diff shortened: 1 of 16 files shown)


vberger:
For the record, this PR includes this: https://twitter.com/levansfr/status/874644019757936640 ! 😄

examples/helpers/glium.rs
```diff
@@ -0,0 +1,121 @@
...
+        GliumDrawer {
+            display,
+            vertex_buffer,
+            index_buffer,
+            program,
+        }
+    }
+
+    pub fn draw(&self, target: &mut glium::Frame, contents: &[u8], surface_dimensions: (u32, u32),
+                surface_location: (i32, i32), screen_size: (u32, u32)) {
```
Drakulix: While this will likely will never be a performance bottleneck in our code, given the amount of surfaces we will have at once on the screen, you would usually try to minimize the amount of draw calls and instead commit all vertices and textures at once.
So we would instead cache all contents to be drawn and then have a "finalize" call or something similar, that results in once draw call with all windows. However this would make this code unnecessarily complex, I think. Might be worth to benchmark this some time later.

examples/helpers/shell.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+unsafe impl<U, H> ::wayland_server::Handler<wl_shell::WlShell> for WlShellStubHandler<U, H>
+    where U: Send + 'static,
+          H: CompositorHandler<U> + Send + 'static
+{
+    unsafe fn message(&mut self, evq: &mut EventLoopHandle, client: &Client, resource: &wl_shell::WlShell,
+                      opcode: u32, args: *const ::wayland_server::sys::wl_argument)
+                      -> Result<(), ()> {
+        <WlShellStubHandler<U,H> as ::wayland_server::protocol::wl_shell::Handler>::__message(self, evq, client, resource, opcode, args)
```
Drakulix: What does this do, why do we need it and where is this documentated.

examples/simple.rs
```diff
@@ -1,46 +1,173 @@
...
+        // we retrieve the contents of the associated buffer and copy it
+        token.with_surface_data(surface, |attributes| {
+            match attributes.buffer.take() {
+                Some(Some((buffer, (x, y)))) => {
+                    self.shm_token.with_buffer_contents(&buffer, |slice, data| {
+                        let offset = data.offset as usize;
+                        let stride = data.stride as usize;
+                        let width = data.width as usize;
+                        let height = data.height as usize;
+                        let mut new_vec = Vec::with_capacity(width * height * 4);
```
Drakulix: Does this happen whenever the buffer contents to change?
We should try to optimize this in the future for sure.

examples/helpers/shell.rs
vberger: This should hopefully be an internal detail of wayland-server, but unfortunately the `declare_handler!()` macro does not handle generic types yet, I need to see if/how I can improve that...

vberger:
Regarding the efficiency of drawing: I voluntarily didn't try to optimize it and went for the most obvious way, in order to have it working quickly. I agree the current implementation is far from efficient, but I don't have the knowledge or experience to optimize it properly, so I suggest we do that later. :)

Regarding `get_shell_surface`, well, it's a request from the `wl_shell` global from the core protocol, it's used by the clients to signify that a given surface should be treated as a window by the compositor.

examples/helpers/shell.rs
vberger: https://docs.rs/wayland-server/0.9.7/wayland_server/macro.declare_handler.html

vberger:
Okay, in the last commit I've used the new version of the macros from wayland-server to hide all these horrible unsafe impls.

I'm willing to merge this now, to have a ground on top of which work, knowing that:

- the graphics handling in the examples is **very** naïve, I only focused on having something that works, it would definitely need a 2nd pass to optimize it
- I'm not 100% satisfied of the subsurface handling part, but I think it's worth starting to use it as it is now, in order to see exactly how it should be improved

What do you think @Drakulix ?

Drakulix:
I agree on both points, but just to clarify, are you handling subsurface ordering relative to the parent in any way right now or not?

vberger:
Right now, the parent is hard-coded to always be behind all its subsurface. This will need to be changed, but this will likely need a consequent refactor.

Unless you think it's an hard-requirement to handle this correctly right now, I'm thinking on opening an issue to track this and the other problems we'll certainly find with this implementation, and refactor the whole thing when we have a better view of the constraints.

Drakulix:
No I think this is totally fine for now, I was just interested in the status of this problem.

vberger:
Alright, here we go then. :)

PR #20, opened 2017-06-04
Fix/Add winit&egl logging

Drakulix:
- [x] Added logging to winit + egl
- [x] Fixed a small some small copy&paste message errors in the libinput backend
- [x] Fixed winit backend only sending horizontal pointer axis events
- [x] Added support for double buffering an egl context (I should do a glutin pr, if this works)

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -25,6 +25,7 @@ use winit::os::unix::{WindowExt, get_x11_xconnection};
 pub struct WinitGraphicsBackend {
     window: Rc<Window>,
     context: EGLContext,
+    logger: ::slog::Logger,
 }
 
 /// Abstracted event loop of a `winit` `Window` implementing the `InputBackend` trait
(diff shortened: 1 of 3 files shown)


src/backend/winit.rs
```diff
@@ -156,6 +178,7 @@ impl EGLGraphicsBackend for WinitGraphicsBackend {
     }
 
     unsafe fn make_current(&self) -> Result<(), SwapBuffersError> {
+        debug!(self.logger, "Making context the current one");
```
vberger: I can't parse this sentence, maybe a word is missing?

vberger:
LGTM modulo the small nit. 👍

src/backend/winit.rs
Drakulix: Yes that is kinda bad. Changed it.

vberger:
👍

PR #21, opened 2017-06-07
Update libinput bindings

Drakulix:
This update removes the bindgen/clang dependency when building with `backend_libinput`.


vberger:
Thanks !

PR #22, opened 2017-06-10
Decouple EGLSurface from EGLContext

Drakulix:
So this is hopefully the last piece refactored in the egl stuff to get drm working.

Most graphical backends will probably yield more then one surface to draw, which all need to be bound to the same egl context, which is why this is necessary.

This actually makes the winit backend more difficult to implement, because the surfaces need to be deconstructed before the context is, which is reflected by lifetimes. However storing an owned var (context) + a reference to that (surface) in a struct is not representable normally in rust. I had to introduce another dependency here to work around this problem, which only changes internal stuff (nothing, that is exposed) and I do think this is actually the best solution to the problem, because:

- This gets the dependency between `EGLSurface` and `EGLContext` right and safe in case we want to expose this as part of public api
- This will not be the case for most backends (when you have multiple outputs, all holding their own surface, this problem ist gone)
-  This implementation works already as an example for everybody running into the same problem, when implementing your own backends as a library user.

Let me know what you think. :)

### DIFF
--- a/src/backend/graphics/egl.rs
+++ b/src/backend/graphics/egl.rs
@@ -521,7 +521,7 @@ impl EGLContext {
         debug!(log, "EGL context successfully created");
 
         let surface_attributes = {
-            let mut out: Vec<c_int> = Vec::with_capacity(2);
+            let mut out: Vec<c_int> = Vec::with_capacity(3);
 
             match reqs.double_buffer {
                 Some(true) => {
@@ -537,6 +537,7 @@ impl EGLContext {
                 None => {}
             }
 
+            out.push(ffi::egl::NONE as i32);
             out
         };
 
--- a/src/backend/graphics/egl.rs
+++ b/src/backend/graphics/egl.rs
@@ -168,10 +168,11 @@ impl EGLContext {
     ///
     /// This method is marked unsafe, because the contents of `NativeDisplay` cannot be verified and may
     /// contain dangling pointers or similar unsafe content
-    pub unsafe fn new<L>(native: NativeDisplay, mut attributes: GlAttributes,
-                         reqs: PixelFormatRequirements, logger: L)
+    pub unsafe fn new<L>(native: NativeDisplay, mut attributes: GlAttributes, reqs: PixelFormatRequirements,
+                         logger: L)
                          -> Result<EGLContext, CreationError>
-        where L: Into<Option<::slog::Logger>>
+    where
+        L: Into<Option<::slog::Logger>>,
     {
         let logger = logger.into();
         let log = ::slog_or_stdlog(logger.clone()).new(o!("smithay_module" => "renderer_egl"));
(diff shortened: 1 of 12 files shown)


src/backend/graphics/egl.rs
```diff
@@ -497,49 +525,56 @@ impl EGLContext {
...
-        let ret = unsafe {
-            self.egl
-                .SwapBuffers(self.display as *const _, self.surface as *const _)
-        };
+    /// Creates a surface bound to the given egl context for rendering
+    ///
+    /// # Unsafety
+    ///
+    /// This method is marked unsafe, because the contents of `NativeSurface` cannot be verified and msy
+    /// contain dangeling pointers are similar unsafe content
```
vberger: s/msy/may/
s/dangeling/dangling/
s/are/or/

I believe?

src/backend/graphics/egl.rs
```diff
@@ -497,49 +525,56 @@ impl EGLContext {
...
 
-        if ret == 0 {
-            match unsafe { self.egl.GetError() } as u32 {
-                ffi::egl::CONTEXT_LOST => Err(SwapBuffersError::ContextLost),
-                err => panic!("eglSwapBuffers failed (eglGetError returned 0x{:x})", err),
+        let surface = {
+            let surface = match native {
+                NativeSurface::X11(window) |
+                NativeSurface::Wayland(window) |
+                NativeSurface::Gbm(window) => {
```
vberger: So, we do not even check if the provided native surface is the same kind (X11/Wayland/Gbm) as the display of the context?...

src/backend/graphics/egl.rs
```diff
@@ -497,49 +525,56 @@ impl EGLContext {
             out
```
vberger: In this block, a `out.push(ffi::egl::NONE as i32);` should be added I believe, so that the attribute list is terminated.

Without this, surface creation fails on my computer with `EGL_BAD_ATTRIBUTE`, but it works fine with it.

(Note that this was already missing prior to this PR.)

Drakulix: Good point. Forgot that

src/backend/graphics/egl.rs
Drakulix: I think I copy&pasted that part. I will check the docs again.

src/backend/graphics/egl.rs
Drakulix: What an oversight, thanks. I will add a run time check (the methods are unsafe anyway and all variants contain the same type, compile time checks through new types don't make a lot of sense).

Drakulix:
Okay I think this is ready now.

src/backend/graphics/egl.rs
vberger: We may do this in a more compact way, by doing the two matches at the same time:

```rust
let surface = match (native, self.backend_type) {
    (NativeSurface::X11(window), NativeType::X11) |
    (NativeSurface::Wayland(window), NativeType::Wayland) |
    (NativeSurface::Gbm(window), NativeType::Gbm) => {
        // create window surface
    },
    _ => return Err(CreationError::NonMatchingSurfaceType)
};
```

What do you think about that?

vberger:
Other than my stylistic nit, this looks nice and clean! 👍 

Looks like rustfmt is again getting crazy all over the place though...

Drakulix:
rustfmt got another huge update (0.9), which is why this was failing again. I adjusted the code for your suggestion and then applied the new rustfmt version. If you don't want that or want to change the rustfmt config at first, feel free to remove thelast commit.

vberger:
This is good this way. 👍

PR #30, opened 2017-08-28
[WIP] Draft of the drm backend

### LINKED ISSUE #18: Design of the drm backend
So this will be a small write-up about thoughts I had about writing a drm backend for smithay
mostly for @nshp and for further discussion with everyone.

A drm backend will be the first to yield more then one graphics backends, because there may be multiple outputs for one device and those may also be attached and detached at run time.
Which makes it difficult to do this correctly, as an output is theoretically not owned by anyone and as no specified lifetime, I have no idea, what the best way to handle this in rust is, but I bet there are many solutions (winit suffers from the same problem, as a window might get closed at any time).

Opening the devices will require some session code in the future (`systemd`, `ConsoleKit`, etc) to allow launching the compositor as a user process. Right now I would like to handle this case, like the libinput backend does right now, just don't open the device yourself. The drm backend could just assume that the device is already opened. The `simple` example could just open the device using `libc::open` and thus require root privileges for now until we provide a nice way to deal with sessions.

This also leaves the problem of choosing the card out of scope for now. In the future we maybe also want to have a handler that uses `udev` to select the most appropriate graphics card available.

Both of this suggestions flow nicely with smithay's idea of exposing all low-level functionality directly. We can always plug more logic on top of it.

My initial idea was to do something like this:
```
#[derive(Debug, PartialEq, Eq, Hash)]
pub struct DrmOutput {}

impl EglGraphicsBackend for DrmOutput { .. }

pub trait DrmHandler {
    pub fn on_output_connected(&mut self, output: DrmOutput);
    pub fn on_output_disconnected(&mut self, output: &DrmOutput);
}

pub struct DrmBackend<'a, H: DrmHandler> {
    gbm: Device,
    handler: H,
}

impl<'a, H: Handler> DrmBackend<'a, H> {
    fn new_from_fd(fd: RawFd, handler: H) -> IoResult<DrmBackend<'a, H>> {
        let gbm = Device::new(fd)?;
        ...
        DrmBackend {
            gbm: Device,
            handler: handler,
        }
    }
}
```

Some inspiration for the implementation can be gathered from:
- [`wlroots`](https://github.com/SirCmpwn/wlroots/blob/master/backend/drm/drm.c)
- [`wlc`](https://github.com/Cloudef/wlc/blob/master/src/platform/backend/drm.c)
- [`weston`](https://github.com/wayland-project/weston/blob/master/libweston/compositor-drm.c)

Please note that `wlroots` is very new and largely untested in a productive environment
and that `wlc` still suffers up to this day from [issues when attaching/detaching monitors at runtime](https://github.com/Cloudef/wlc/issues/87#issuecomment-158879094), possibly related to some race conditions, so it is certainly not bug-free.

Still other implementations can provide some insight and weston's implementation is certainly the most difficult to read (probably for a reason).

Providing a stable solution is our goal here, which will require reading through a lot of docs and probably some suffering during debugging and which also brings us to the next topic: drm bindings.

The are many different `libdrm` bindings out there:
- [`modesetting`](https://crates.io/crates/modesetting)
- [`drm`](https://crates.io/crates/drm)
- [`drm-rs`](https://crates.io/crates/drm-rs)
- [`libdrm-sweet`](https://crates.io/crates/libdrm-sweet)
- [`libdrm-sys`](https://crates.io/crates/libdrm-sys)
- [`drm-sys`](https://crates.io/crates/libdrm-sys)

Last time I checked non of the rusty ones were complete, which is most likely the reason, why they are so many. While we already have some unsafe c-bindings in rust (egl code) I would like to see a safe wrapper been used.

[rusty-desktop](https://github.com/rusty-desktop) will be trying to solve this problem, so it looks like we should try using [`drm-rs`](https://github.com/rusty-desktop/drm-rs) at first and contribute to that repository, when functionality is lacking behind what we need.

Drakulix:
I don't have a lot time to program in the next days, so I though I might put this very early version up for initial review and testing.

This contains a working drm backend (at least on my machine) with an example.

This also still contains:
    - a lot of debug code
    - a good chunk of unrelated changes, like trying to get the libinput backend to work
    - almost no error handling

This is definitely not ready for merge and a lot of cleanup has to happen, but it displays something (libinput is not working).

This will eventually fix #18


Drakulix:
@Slabity This is using your drm-bindings successfully in combination with my gbm bindings.

I would like to ask you for a first release on crates.io, so I can follow that up with a release of my gbm bindings instead of using git dependencies.

Also this requires Rust Beta currently, because of some features in use by drm-rs. (only associated consts, I think)

Slabity:
> I would like to ask you for a first release on crates.io, so I can follow that up with a release of my gbm bindings instead of using git dependencies.

Done. You can use the new version with `drm = "0.2.0"`. I've also added you as an owner to the `gbm` crate on crates.io so that you can overwrite my deprecated bindings.

> Also this requires Rust Beta currently, because of some features in use by drm-rs. (only associated consts, I think)

Unfortunately yes. This will probably be the case for a while until a better solution can be made.

PR #32, opened 2017-09-05
First draft of shell backend

vberger:
Unifying abstraction over `wl_shell` and `wdg_shell`.

Still to do before merging:

- [x] update example to use it (and thus test it)
- [x] write all the docs


Drakulix:
:tada:

PR #34, opened 2017-09-13
Drm Backend

Drakulix:
Rough list of changes:

- new backend rendering via egl via gbm directly on a drm device
- refine EGLContext and EGLSurface dependencies through lifetimes and surface types (super safe now, no unsafe code for initializing is required anymore outside of egl.rs itself).
- add new example using the drm backend instead
- change GliumDrawer to be static for the drm example
- fixup the old winit backend and example to work with these changes

I am sorry this turned out to be one commit (ouch), but I reworked this for so many times and had to rebase this on your re-formatted code, so you might be able to imagine that mess.

This is ready for a first review. ~It is only missing a lot of documentation and more logging.~ It appears to work just fine on my machine and does not seem to have broken anything.

### DIFF
--- a/examples/drm.rs
+++ b/examples/drm.rs
@@ -4,7 +4,6 @@ extern crate glium;
 extern crate rand;
 #[macro_use(define_roles)]
 extern crate smithay;
-extern crate wayland_protocols;
 extern crate wayland_server;
 
 #[macro_use]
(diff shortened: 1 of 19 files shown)


Drakulix:
Also merge of this still requires the following possibilities in `wayland-server`:

- [ ] Ability to access other handlers from the `EventLoopHandle`
- [ ] Ability to create a single-threaded `EventLoop` that does not require `Handler`s to be `Send`

Drakulix:
Also blocked by https://github.com/rusty-desktop/drm-rs/pull/12

Drakulix:
Also this fixed #18

src/backend/drm/backend.rs
```diff
@@ -0,0 +1,503 @@
...
+        });
+    }
+
+    pub(crate) fn is_crtc(&self, crtc: crtc::Handle) -> bool {
+        crtc == self.crtc
+    }
+}
+
+impl DrmBackend {
+    /// Add a connector to backend
```
vberger: It wold be good to have either a basic explanation of what a "connector" is, or at least a link to the relevant documentation of the drm crate.

I for one I absolutely no idea of what it is. :)

src/backend/drm/mod.rs
```diff
@@ -0,0 +1,502 @@
+//! Drm/Kms types and backend implementations
+//!
+//! This module provide a `DrmDevice` which acts as a reprensentation for any drm
+//! device and can be used to create the second provided structure a `DrmBackend`.
+//!
+//! The latter represents a crtc of the graphics card you can render to.
+//!
```
vberger: Also here, a general explanation of how drm works, or a link to the relevant documentation in the drm crate.

(what is a "crtc"? a "connector"? ..)

src/backend/drm/mod.rs
```diff
@@ -0,0 +1,502 @@
...
+        L: Into<Option<::slog::Logger>>,
+    {
+        let log = ::slog_or_stdlog(logger).new(o!("smithay_module" => "backend_drm"));
+
+        /* GBM will load a dri driver, but even though they need symbols from
+         * libglapi, in some version of Mesa they are not linked to it. Since
+         * only the gl-renderer module links to it,  these symbols won't be globally available,
+         * and loading the DRI driver fails.
+         * Workaround this by dlopen()'ing libglapi with RTLD_GLOBAL.
+         */
```
vberger: Wow, this looks like quite an hack. Are Weston/wlc/wlroots doing that as well?

vberger:
I don't have much to say, my knownledge of drm being wuite limited.

I think the module-level docs of the `drm` module would deserve some explanations on the drm structure, mainly:

- what are "crtcs" and "connectors", and how do they relate with "physical monitors"?
- hints about how it would be done to handle multiple screens that are not just mirrors of each other

Also, how does your code interact with TTY switching?

examples/drm.rs
```diff
@@ -0,0 +1,350 @@
...
+use std::io::Error as IoError;
+use std::os::unix::io::AsRawFd;
+use std::time::Duration;
+use wayland_protocols::unstable::xdg_shell::server::{zxdg_shell_v6, zxdg_toplevel_v6};
+use wayland_server::{Client, EventLoopHandle};
+use wayland_server::protocol::{wl_callback, wl_compositor, wl_output, wl_seat, wl_shell, wl_shm,
+                               wl_subcompositor, wl_surface};
+use wayland_server::sources::READ;
+
+define_roles!(Roles => [ ShellSurface, ShellSurfaceRole ] );
```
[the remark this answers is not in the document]
Drakulix: Yeah probably, already thought about that. I will fix that.

src/backend/drm/mod.rs
Drakulix: I will most likely go with a link and do a PR with more explanations on the drm crate.

Drakulix:
I know I need to add these explanations to the docs, but for you to understand whats going on:

> what are "crtcs" and "connectors", and how do they relate with "physical monitors"?

A crtc is basically a unit of your graphics card, that can render stuff. The amount is limited, if you have 2 crtcs, but 4 available monitor ports you can only use 2 at a time (in most cases, mirroring might work around that).

The second entity is an encoder. An encoder translates what the crtc produces to the signal a connector needs. You don't need to manage these in code, but you need to make sure your configuration is possible to fulfill by the kernel. A crtc is usually connected to a bunch of encoders (but most likely not all of them) and each encoder can drive a limited set of connectors. E.g. You might have a separate analog-encoder for a VGA port, but a digital one for an HDMI port.

Last entity is the connector, that just represents a port on your computer, possibly with a connected monitor.

What you need to provide is the crtc you want to use and the connectors. The user is responsible to check if there are encoders for all these connectors that are connected to the given crtc, or we throw an error.

The problem of how you want to set up your configuration is NP-complete. If there is not a crtc serving both connectors you want to mirror, you might want to fallback to do that in software, etc, but obviously you want to you hardware-mirroring where possible and might need to re-assign the connectors at run-time to find the best solution.

I have given all that responsibility in the hand of the user for now. A new udev-supported structure will support hot-plugging in the future and probably create these backends for you building up on this code.

> hints about how it would be done to handle multiple screens that are not just mirrors of each other

Just create multiple backends, like you can create multiple winit backends to get more windows. But yes, I will hint that somewhere.

> Also, how does your code interact with TTY switching?

Not at all right now.
I will tackle this once we have proper session support and can therefor start it from a display manager.
With that setup tty-switching should be a lot easier to actually implement.

src/backend/drm/backend.rs
Drakulix: Noted. A "connector" is just what it sounds like. A port on your computer possibly with a connected monitor.

src/backend/drm/mod.rs
vberger: That sounds good 👍

src/backend/drm/mod.rs
vberger: I see, alright then.

vberger:
Thanks for the explanation, that's much clearer now.

> Just create multiple backends, like you can create multiple winit backends to get more windows. But yes, I will hint that somewhere.

So, just to check I'm understanding correctly. a DrmBackend links one crtc to one or more connectors, all of these connectors being thus setup to mirror each other. Thus you create a backend for each different "screen" you want to render. That's correct?

> Not at all right now.
> I will tackle this once we have proper session support and can therefor start it from a display manager.
With that setup tty-switching should be a lot easier to actually implement.

I see, okay.

vberger:
I'll start working on the rework of wayland-server's event loop with notably multi-access to the state, as I believe this is blocking your PR, right?

Drakulix:
> So, just to check I'm understanding correctly. a DrmBackend links one crtc to one or more connectors, all of these connectors being thus setup to mirror each other. Thus you create a backend for each different "screen" you want to render. That's correct?

Yes exactly.

> I'll start working on the rework of wayland-server's event loop with notably multi-access to the state, as I believe this is blocking your PR, right?

Right. :)

Drakulix:
> So, just to check I'm understanding correctly. a DrmBackend links one crtc to one or more connectors, all of these connectors being thus setup to mirror each other. Thus you create a backend for each different "screen" you want to render. That's correct?

Exactly.

PR #56, opened 2017-11-07
Udev backend and session api

Drakulix:
Okay this is ready for testing and first review.

The new udev example needs root permissions currently, this will change once a logind session is available. Also the pointer location can become negative with interesting side-effects in the new example.

The other big bug is that restoring the tty still does not work correctly, it looks like some `Drop` implementations are not called correctly, which might be related to the `EventLoop`, but I am not yet sure, this needs some more debugging (help here would be greatly appreciated, all calls are there, it really *should* work).

Apart from the missing logind support a lot of docs and some workarounds/hacks need to be addressed before this is ready for merge (and it needs to be rebased as well), but the rough idea of how everything might work should be clear now.

Any input is very welcome.


examples/udev.rs
```diff
@@ -0,0 +1,472 @@
...
+        debug!(self.log, "key"; "keycode" => keycode, "state" => format!("{:?}", state));
+
+        match (keycode, state) {
+            (1 /*ESC*/, KeyState::Pressed) => {
+                self.running.store(false, Ordering::SeqCst);
+            },
+            (keycode, state) => {
+                let serial = self.next_serial();
+                self.keyboard.input(keycode, state, serial, |_, _| true);
+            }
```
vberger: We may use a more traditional shortcut to quit (I'm thinking `CTRL`+`ALT`+`BACKSPACE`) by replacing this match block with:

```rust
let serial = self.next_serial();
self.keyboard.input(keycode, state, serial, |modifiers, keysym| {
    if modifiers.ctrl && modifiers.alt && keysym == xkb::KEY_BackSpace {
        self.running.store(false, ordering::SeqCst);
        false
    } else {
        true
    }
});
```

This is after all exactly how the `keyboard.input(..)` API has been designed, so lets use it in our examples, shall we ? 😉

examples/udev.rs
```diff
@@ -0,0 +1,472 @@
...
+    }
+    fn on_touch_frame(&mut self, _: &input::Seat, _: event::touch::TouchFrameEvent) {
+        /* not done in this example */
+    }
+    fn on_input_config_changed(&mut self, _: &mut [LibinputDevice]) {
+        /* not done in this example */
+    }
+}
+
+fn main() {
```
vberger: General comment about this example : given it is single-threaded, is it really necessary to use thread-safe primitives (mutexes) everywhere, instead of simple RefCells?

vberger:
This is a lot to digest on this PR, I'll review it in several passes

vberger:
Hmm, could you give here an high-level explanation of what are the different modules you added and how the interact with each other?

examples/udev.rs
Drakulix: That looks way easier.

examples/udev.rs
Drakulix: Most likely not, I will change that.

Drakulix:
Alright.

`src/backend/session` is a module that provides the session api. Sessions are active interfaces between a user and the kernel via a program, in this context our compositor. Multiple running compositors are all different sessions not necessarily by the same user.

Sessions are responsible for opening devices and passing requests to the session manager (like switching the tty) and handling requests by that manager (like pausing/activating), so the management may switch between the open sessions.

`src/backend/session/direct` is an implementation that uses the kernels tty interface for that functionality. That api is outdated, but still available and functional and mostly used as a fallback in case no better api is available. But it is considered problematic, because a zombie compositor cannot be forced to swap the vt/tty and the whole api can cause to race conditions.

`src/backend/session/logind` will be an implementation not dealing with these problems as logind is a dbus api usually provided by systemd, which is considered the better implementation. It also will not need root permissions to open system devices.

For that reason session do require that when signaled the open file descriptors may not be used, because an other active compositor may be in charge. This is what the `SessionObserver` trait is for.

`src/backend/udev` is a backend that automatically enumerates all available drm devices and creates a `DrmDevice` for all of them. The user may then choose to create `DrmBackends` for all of them via the `UdevHandler` interface.

Because udev will be responsible for opening and closes these devices as they will become available, change and get removed (hot-plugging drm devices is a thing) we need the `Session` trait to be able to use different (possibly user provided) mechanisms to open/close these devices. With the old plain `DrmDevice` and `DrmBackend` apis (which are of course still available) the user was always in charge of opening those devices.

Another nice thing the udev backend does allow us, is to be notified when we should scan for new connectors/monitors again. The `changed` event is what does notify our `UdevBackend` and results in the example in a re-scan. Theoretically this functionality could be provided without requiring the session api, but that has proven to be very difficult in the implementation and there is not much use to it.
Most people will want to use sessions and the udev stack anyways.

We also need sessions for the libinput backend as libinput also expects to be provided an interface (`LibinputInterface`) for opening and closing devices for that very reason. Without that we would also always need root permissions as a normal user is not allowed to open input devices.

All changes to the libinput backend, the drm code and the egl code are motivated by either the use of sessions (so we can be started by a login manager in future) or by monitor hot-swap support (which may throw some errors during `eglSwapBuffer`).

A very good introduction to that topic is this https://dvdhrm.wordpress.com/2013/08/24/session-management-on-linux/ + the two follow up articles, I highly recommend to read those to get a basic idea of what is happening.

Side node: The only reason we were able to open a `DrmDevice` previously is that you become the drm master process automatically if no other process is claiming the current tty, but while documentated this is nothing we can really rely on as the same does not hold true for libinput.

vberger:
Okay, thanks for the explanation, this is a very clear one. 👍 

I don't think I can tell much about the code itself, but the architecture as you described it seems pretty good.

It's good to separate logind session vs. direct session. While we absolutely want logind support, I'd rather not have it be a hard dependency of smithay. On this direction, if we want to support consolekit too, how would it be done? Would it go through the direct interface, or would we need to make a `session_consolekit` feature too?

Drakulix:
logind was basically modeled after the ConsoleKit api or the other way around. I am not sure about their exact relation, so I don't want to spread misinformation, but I am quite sure the api's are mostly identical.

We may need to provide a `session_consolekit` feature, but I think we can support both with the logind implementation.

vberger:
Yes indeed, logind is an intented replacement for consolekit, which is much older, though I don't know how similar are the APIs. Anyway, as long as we're not closing any door regarding its support, all is good for me. The `Session` abstraction is pretty neat on this regard. 👍

examples/udev.rs
```diff
@@ -0,0 +1,472 @@
...
+        window_map: window_map.clone(),
+        pointer_location,
+        screen_size: (w, h),
+        serial: 0,
+        running: running.clone(),
+    });
+    let libinput_event_source = libinput_bind(libinput_backend, &mut event_loop).unwrap();
+
+    let session_event_source = direct_session_bind(notifier, &mut event_loop, log.clone()).unwrap();
+    let udev_event_source = udev_backend_bind(&mut event_loop, udev_token).unwrap();
```
vberger: Is it necessary to have this "bind" step, separated from the creation of these backends? Couldn't the backend just take the event loop as argument at creating and do this internally?

examples/udev.rs
```diff
@@ -0,0 +1,472 @@
...
+            pointer_location: self.pointer_location.clone(),
+            logger: self.logger.clone(),
+        })
+    }
+
+    fn device_changed<'a, S: Into<StateProxy<'a>>>(&mut self, state: S, device: &StateToken<DrmDevice<GliumDrawer<DrmBackend>>>) {
+        //quick and dirty
+        let mut state = state.into();
+        self.device_removed(&mut state, device);
+        state.with_value(device, |state, device| self.scan_connectors(state, device));
```
vberger: So this goes the "quick & dirty way" and removes then re-init the device, right? (nothing wring with that, just checking I'm understanding correctly)

src/backend/drm/mod.rs
```diff
@@ -462,61 +517,99 @@ pub trait DrmHandler<B: Borrow<DrmBackend> + 'static> {
     ///
     /// The `id` argument is the `Id` of the `DrmBackend` that finished rendering,
     /// check using `DrmBackend::is`.
-    fn ready(&mut self, evlh: &mut EventLoopHandle, device: &mut DrmDevice<B>, backend: &StateToken<B>,
-             frame: u32, duration: Duration);
+    fn ready<'a, S: Into<StateProxy<'a>>>(&mut self, state: S, device: &mut DrmDevice<B>, backend: &StateToken<B>,
```
vberger: In general for the methods providing an `Into<StateProxy>` rather than the full `EventLoopHandle`, it would be good to document if some state keys known to the user are susceptible to be already borrowed (and thus trying to borrow them again would result in a panic).

If the only already-borrowed state keys are implementation details (not known to smithay's user), no need to obviously.

examples/udev.rs
vberger: Coming back to this after reading more of the PR, I can elaborate a little. In my opinion, processes like "get this value, then store it in the state event loop, then give me back the token" are not very ergonomic:

- either the value must absolutely be in the event loop state (as in there is no reasonable way to use them in any other way), in which case the init functions should insert it themselves
- or the value could possibly be used in an other way, in which case we should not force the state token use.

But overall, if the user cannot do anything with this value other than store it in the state and give me back the token, they should not see the value at all.

This is what I've been leaning to with the implementation/implementation data schemes in the wayland-side implementation: if my implementation data needs to be accessed by other things outside of my module (and specifically I don't know in advance what or how), I shove it in the state and use a state token. But if it is only used internally by the module, I try to keep it as implementation data, hidden from the user (note that using an internal state token to ease sharing works too, but it's not clear how resource freeing would be handled in this case).

Though, I'm just opening a discussion here, not saying you should change everything. I lack most of the context for how you designed this whole API.

src/backend/session/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+//! Abstraction of different session apis.
+//!
+//! Sessions provide a way for multiple graphical systems to run in parallel by providing
+//! mechanisms to switch between and handle device access and permissions for every running
+//! instance.
+//!
+//! They are crutial to allow unpriviledged processes to use graphical or input devices.
+//!
+//! The following mechanisms are currently provided:
+//!     - direct - legacy tty / virtual terminal kernel api
```
vberger: Insert a newline before the list or the markdown doesn't render correctly

src/backend/session/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+    ///
+    /// Returns an id of the inserted observer, can be used to remove it again.
+    fn register<S: SessionObserver + 'static>(&mut self, signal: S) -> usize;
+    /// Removes an observer by its given id from `SessionNotifier::register`.
+    fn unregister(&mut self, signal: usize);
+
+    /// Check if this session is currently active
+    fn is_active(&self) -> bool;
+    /// Which seat this session is on
+    fn seat(&self) -> &str;
```
vberger: Is there a reason why `Session::seat()` returns `String` but `SessionNotifier::seat()` returns `&str`?

src/backend/session/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Part of the session api which allows to get notified, when the given session
+/// gets paused or becomes active again. Any object implementing the `SessionObserver` trait
+/// may be registered.
+pub trait SessionNotifier {
+    /// Registers a given `SessionObserver`.
+    ///
+    /// Returns an id of the inserted observer, can be used to remove it again.
+    fn register<S: SessionObserver + 'static>(&mut self, signal: S) -> usize;
+    /// Removes an observer by its given id from `SessionNotifier::register`.
+    fn unregister(&mut self, signal: usize);
```
vberger: How does it behave if a bad id is given as argument?

src/backend/session/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Trait describing the ability to be notified when the session pauses or becomes active again.
+///
+/// It might be impossible to interact with devices while the session is disabled.
+/// This interface provides callbacks for when that happens.
+pub trait SessionObserver {
+    /// Session is about to be paused.
+    ///
+    /// In case the implementor is a `StateToken` the state of the `EventLoop`
+    /// is provided via a `StateProxy`.
+    fn pause<'a>(&mut self, state: &mut StateProxy<'a>);
```
vberger: Is there a reason this receives a `&mut StateProxy` specifically, and not a `&mut EventLoopHandle`?

src/backend/udev.rs
```diff
@@ -0,0 +1,402 @@
...
+
+/// Handler for the `UdevBackend`, allows to open, close and update drm devices as they change during runtime.
+pub trait UdevHandler<B: Borrow<DrmBackend> + 'static, H: DrmHandler<B> + 'static> {
+    /// Called on initialization for every known device and when a new device is detected.
+    ///
+    /// Returning a `DrmHandler` will initialize the device, returning `None` will ignore the device.
+    ///
+    /// ## Panics
+    /// Panics if you try to borrow the token of the belonging `UdevBackend` using this `StateProxy`.
+    fn device_added<'a, S: Into<StateProxy<'a>>>(&mut self, state: S, device: &mut DrmDevice<B>) -> Option<H>;
```
vberger: If I understand correctly, this `S: Into<StateProxy>` is needed because the `UdevHandler` instance will be stored in a value that will itself be stored in the EventLoop's State.

Is storing this into the EventLoop's state absolutely necessary, or could using the implementation data be an option too? This would make it possible to not have to borrow the state to retrieve the handler, and as such probably allow to directly provide a `&mut EventLoopHandle`, I think?

examples/udev.rs
vberger: I think this whole comment is still valid overall.

src/backend/session/mod.rs
Drakulix: Yes, because of the implementations of `Rc<RefCell<Session>>` and `Arc<Mutex<Session>>`. It is just way easier to clone the String in that case. Do you have a better idea?

src/backend/udev.rs
Drakulix: It is necessary to register the `UdevBackend` to the `SessionNotifier`, which is also the reason, why I have a separate `bind`-function.

I have already evaluated, if I could remove any bind calls and add them to the constructor, but for all remaining values I did not find a way to do that with the current api.

`examples/udev.rs` really shows what needs to be done with all values at the minimum, if you have concerns about the general api or about any specific case fine, but the overall structure of `new` -> `register`/user-calls -> `bind` is kinda necessary.

Also we were talking about building a extensible and composable library. I would rather not want to force the user to use the `State` or the `EventLoop` at all, where easily possible.

src/backend/session/mod.rs
Drakulix: Should be fixed by the latest commit. It is now impossible to create different Ids. Passing the same Id twice is safe.

src/backend/session/mod.rs
Drakulix: Not in the current implementation. I mostly wanted to keep us as much freedom as possible until we got more session implementations.

Drakulix:
Feature-wise this should be done now.
Working on getting everything upstream now:

drm: https://github.com/Smithay/drm-rs/pull/18 https://github.com/Smithay/drm-rs/pull/19 https://github.com/Smithay/drm-rs/pull/20 
udev: https://github.com/Stebalien/udev-rs/issues/1
libinput: Blocked by udev

Drakulix:
@vberger Its done!

examples/udev.rs
```diff
@@ -0,0 +1,477 @@
...
+        debug!(self.log, "key"; "keycode" => keycode, "state" => format!("{:?}", state));
+
+        let serial = self.next_serial();
+        self.keyboard.input(keycode, state, serial, |modifiers, keysym| {
+            if modifiers.ctrl && modifiers.alt && keysym == xkb::KEY_BackSpace {
+                self.running.store(false, Ordering::SeqCst);
+                false
+            } else if modifiers.logo && keysym == xkb::KEY_Return {
+                let _ = Command::new("weston-terminal").spawn();
+                false
```
vberger: This should only trigger if `state == Pressed`, otherwise a terminal is spawned on key press *and* on key release, so two instances are spawned each time

.travis.yml
```diff
@@ -1,40 +1,32 @@
...
-  only:
-    - master
+      - libcurl4-openssl-dev
+      - libelf-dev
+      - libdw-dev
 
 env:
-  global:
-    - RUST_BACKTRACE=1
   matrix:
```
vberger: We can maybe add a few lines to this features matrix, given what you added?

vberger:
🎉 🎉 

Just two small last things, and we're good to merge I think! 👍

Drakulix:
@vberger Should be done, I think. Lets see what travis thinks about these new features :)

vberger:
🎉

PR #57, opened 2017-12-28
Feature: EGL Buffer support / Fix #12

### LINKED ISSUE #12: Handle client EGL-backed buffers

Drakulix:
Fix for #12, hardware acceleration via EGL is working now.

Easiest way to test this is the weston example binary `weston-simple-egl` which draws a rotating triangle on a semi-transcluent background.

I also fixed seat 1 clients, as `weston-simple-egl` is one of these.
I also fixed our rendering for transcluent pixels.

I reworked a lot of the EGL module to use `Weak` pointers to their parent structures to allow easy storage of `EGLImages`. This also removed the rental crate completely, but made some changes to gbm.rs necessary.

Missing:
- [x] drm example support
- [x] Release new gbm version
- [x] Documentation


src/backend/graphics/egl/wayland.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// to utilize hardware-accelerated surface via EGL.
+    ///
+    /// ## Errors
+    ///
+    /// This might return `WlExtensionNotSupported` if binding is not supported
+    /// by the EGL implementation.
+    ///
+    /// This might return `OtherEGLDisplayAlreadyBound` if called for the same
+    /// `Display` multiple times, as only one context may be bound at any given time.
+    fn unbind_wl_display(&self, display: &Display) -> Result<()>;
```
vberger: Shouldn't be a `fn(self)` method on `EGLDisplay` rather?

If I understand correctly, calling this method will invalidate a previously created `EGLDisplay`, what happens if I try to continue to use it?

src/backend/graphics/egl/wayland.rs
```diff
@@ -0,0 +1,322 @@
...
+}
+
+pub struct EGLDisplay(Weak<ffi::egl::types::EGLDisplay>);
+
+impl EGLDisplay {
+    pub fn new<B: native::Backend, N: native::NativeDisplay<B>>(context: &EGLContext<B, N>) -> EGLDisplay {
+        EGLDisplay(Rc::downgrade(&context.display))
+    }
```
vberger: This looks like it is an implementation detail of `bind_wl_display` and should be `pub(crate)`, no?

src/backend/graphics/egl/wayland.rs
```diff
@@ -0,0 +1,322 @@
...
+{
+    fn bind_wl_display(&self, display: &Display) -> Result<EGLDisplay> {
+        (**self).bind_wl_display(display)
+    }
+    fn unbind_wl_display(&self, display: &Display) -> Result<()> {
+        (**self).unbind_wl_display(display)
+    }
+}
+
+impl<B: native::Backend, N: native::NativeDisplay<B>> EGLWaylandExtensions for EGLContext<B, N> {
```
vberger: I see you've added a lot of implementations of this traits to various types that embed an `EGLContext`, which seem redundant, what's the reason of that?

examples/helpers/implementations.rs
```diff
@@ -13,35 +16,68 @@ define_roles!(Roles => [ ShellSurface, ShellSurfaceRole ] );
...
+    pub texture: Option<Texture2d>,
 }
 
-pub fn surface_implementation() -> SurfaceUserImplementation<SurfaceData, Roles, ()> {
+pub enum Buffer {
+    Egl { images: EGLImages },
+    Shm { data: Vec<u8>, size: (u32, u32) },
+}
+
+unsafe impl Send for Buffer {}
```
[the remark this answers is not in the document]
Drakulix: You are right, it is not needed.

src/backend/graphics/egl/wayland.rs
Drakulix: Changed it into `EGLDisplay`s Drop implementation

src/backend/graphics/egl/wayland.rs
Drakulix: Does not even need to be `pub(crate)`, changed.

src/backend/graphics/egl/wayland.rs
Drakulix: I could make this a `pub(crate)` function of `EGLContext` instead, that every other implementation calls into. But I think this is unnecessary and just hides more options from the user. I just implemented the trait for every struct it made sense for.

vberger:
🎉

PR #65, opened 2018-01-31
Winit handler

Drakulix:
This adds an additional `WinitEventsHandler` to our `WinitBackend`. This allows a real compositor to receive additional window events not available in the more restricted `InputHandler`, that it might be interested in.

This also exposes the underlying `winit::Window` object to allow further customization like the window title or to receive properties of the window, like its hidpi factor, if the compositor wants to take that into account.

I figured giving the user more control is way easier then adding a bunch of functions to the backend for that purpose and follows smithay's philosophy.

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -71,6 +71,7 @@ pub struct WinitGraphicsBackend {
 /// You need to call `dispatch_new_events` periodically to receive any events.
 pub struct WinitInputBackend {
     events_loop: EventsLoop,
+    events_handler: Option<Box<WinitEventsHandler>>,
     window: Rc<Window>,
     time_counter: u32,
     key_counter: u32,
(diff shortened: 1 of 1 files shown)


src/backend/winit.rs
```diff
@@ -178,6 +180,27 @@ where
...
+    fn focus_changed(&mut self, evlh: &mut EventLoopHandle, focused: bool);
+    /// The window needs to be redrawn
+    fn refresh(&mut self, evlh: &mut EventLoopHandle);
+    /// The window's hidpi factor changed
+    fn hidpi_changed(&mut self, evlh: &mut EventLoopHandle, scale: f32);
+}
+
+impl WinitGraphicsBackend {
+    /// Get a reference to the internally used `winit::Window`
+    pub fn window(&self) -> &WinitWindow {
```
vberger: In favor of full explicitness, I'd rather name this `.winit_window()`.

Drakulix: done

PR #70, opened 2018-02-15
Add names to backend seats

Drakulix:
Otherwise this information gets lost in the `InputHandler` and cannot be passed onto `smithay::wayland::seat::Seat::new`.


Drakulix:
My bad, should be fixed. Lets see what travis says.

PR #73, opened 2018-02-20
Allow `DrmBackend` to act as `drm::control::Device`

Drakulix:
Allows for easier handling of drm related stuff in a compositor. Nothing really special, this is safe, because you could just receive the `RawFd` of the `DrmDevice` and build your own clonable struct that implements `drm::Device` and `drm::control::Device`, this is just for usability reasons.


src/backend/drm/backend.rs
```diff
@@ -528,6 +530,16 @@ impl<A: Device + 'static> EGLGraphicsBackend for DrmBackend<A> {
     }
 }
 
+// for users convinience
```
vberger: typo, "conv_e_nience"

PR #82, opened 2018-03-17
Replace time stub with proper timestamps

### LINKED ISSUE #58: winit backend has stubbed timestamps for events
Right now, the winit backend stubs timestamps of events using a counter. This is obviously incorrect and needs to be fixed.

The issue though is that winit does not provide timestamps. So:

- do we emulate timestamps by just using the system time at the moment winit dispatches the event to us?
- do we try to get winit to provide timestamps?

Drakulix:
Fixes #58

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -16,6 +16,7 @@ use std::cmp;
 use std::error;
 use std::fmt;
 use std::rc::Rc;
+use std::time::Instant;
 use wayland_client::egl as wegl;
 use wayland_server::{Display, EventLoopHandle};
 use winit::{ElementState, Event, EventsLoop, KeyboardInput, MouseButton as WinitMouseButton, MouseCursor,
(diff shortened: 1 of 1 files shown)


vberger:
Just making sure I understand correctly: you are setting the time origin to be the instant of creation of the winit backend, and all times are reported relative to this instant, is that correct ?

Is there a reason to not just use the current time ?

Drakulix:
> Just making sure I understand correctly: you are setting the time origin to be the instant of creation of the winit backend, and all times are reported relative to this instant, is that correct ?

Yes

> Is there a reason to not just use the current time ?

To quote rustdoc:

```
Instants are always guaranteed to be greater than any previously measured instant when created, and are often useful for tasks such as measuring benchmarks or timing how long an operation takes.

Note, however, that instants are not guaranteed to be steady. In other words, each tick of the underlying clock may not be the same length (e.g. some seconds may be longer than others). An instant may jump forwards or experience time dilation (slow down or speed up), but it will never go backwards.
```

Current system time in milliseconds might jump backwards triggering unexpected behavior or even a protocol error.
Duration relative to two Instants will always be positive. Also the wayland protocol explicitly states, that time values just need to be comparable and in millisecond granularity, but may have an unknown base.

PR #83, opened 2018-03-20
Upgrade to wl_seat version 5

Drakulix:
Alright this is a first attempt to update our wl_seat implementation to version 5.

This implementation is work in progress, not meant to merged in this state and open to discussion.

- This removes support for wl_seat v4. I do not know, if that is a problem, I do not know any wayland library still relying on v4 support.

- This is still not enough to get GTK-apps to receive input. I do not know why this is the case, mutter (gnome-shell) does also only advertise wl_seat v5 (and not v6). `WAYLAND_DEBUG=1` shows that e.g. `gedit` is just not binding the advertised seat.

*EDIT*: https://github.com/swaywm/wlroots/issues/103#issuecomment-323360784
So this is step one to get GTK to work.
 
- This has shown a major problem for our `PointerAxisEvent` implementation. It separates the original event by axis, while the user might want to submit multiple axis at the same time. Our event should be changes to include both values, if available.

- I am not sure the builder pattern for the `PointerAxisHandle` is the best solution to group all axis events into one frame.

- While the winit-example implementation is perfect (except for that axis handling as stated above), the libinput implementation in the udev example is not (it looses information). Without relying the libinput specific event functions, there is no way to get the `amount` in pixels for `Wheel` or `WheelTilt` events although libinput provides them. Also `Finger` is currently not handled correctly, because I was to lazy to implement stop events in the example.
Question is, if we should also adjust the `PointerAxisEvent` to provide the "pixel-amount" on discrete sources? We cannot guarantee this to be available all the time, winit might only send `LineDelta` values instead of `PixelDelta` values.


vberger:
Before I look at the implementation itself :

> This removes support for wl_seat v4. I do not know, if that is a problem, I do not know any wayland library still relying on v4 support.

We can't do that. Backward-compatibility is imposed by the protocol, and a compositor advertising "wl_seat v5" actually means "wl_seat: all versions up to v5".

However, if the implementations for v4 and v5 are so different, we can have two separate implementations internally and dispatch to the appropriate one in the binding callback.

Drakulix:
> We can't do that. Backward-compatibility is imposed by the protocol, and a compositor advertising "wl_seat v5" actually means "wl_seat: all versions up to v5".

Okay, thats not a big deal, we would just need to avoid sending the new axis events (everything expect 'axis' itself) to v4 clients. How would I know, if a client is supporting v4 or v5, while only advertising v5 in the registry?
Or do I need to register two globals for both versions?

vberger:
My general feeling after looking at the code:

I like the API. I don't know if the builder pattern is the best possible, but in my opinion it looks good enough, at least to start experimenting. :+1: 

I'd just add a `fn done(self) { }` method. It'd just trigger the drop implementation, but allows in the code to clearly show that it is the end of this event group, and that there was no oversight.

------

Regarding the management of version, the server only advertises in the registry the highest supported version, and the the client can request any version between 1 and the one advertised (inclusive).

We can check the version of any wayland object in smithay using the `.version()` method provided by the `Resource` trait from wayland-server.

My approach regarding multi-version interfaces in general is to design the rust API to match the most recent version, then integrate internal code that adapts or ignore some calls to retrofit them to the lower versions.

Drakulix:
Alright, that last commit should fix v4 clients.
I will rework the `PointerAxisEvent` later for merge.

src/wayland/seat/pointer.rs
```diff
@@ -5,7 +5,7 @@ use wayland_server::protocol::{wl_pointer, wl_surface};
 // TODO: handle pointer surface role
 
 struct PointerInternal {
-    known_pointers: Vec<wl_pointer::WlPointer>,
+    known_pointers: Vec<(wl_pointer::WlPointer, i32)>,
```
vberger: Is there a big need to to track the versions here, given we can just to `pointer.version()` whenever we need ?

The pointer automatically inherit the same version as the seat that created them.

Drakulix: Oh I did not know that. `pointer.version()` would be sufficient in that case.

Drakulix:
@vberger `PointerAxisEvent` is refactored. I am now happy with this PR, so if you could give this another  look before merge, that would be nice.

PR #85, opened 2018-03-25
Pin slog version

Drakulix:
Slog 2.2.1 breaks smithay due to https://github.com/slog-rs/slog/issues/164

Lets pin the latest working version for now.


Drakulix:
Bug is gone, 2.2.2 (which is compatible with our current semver version) fixes the problem. Closing

PR #88, opened 2018-04-13
[WIP] Upgrade to wayland-rs 0.20

vberger:
This is an on-going work to upgrade smithay to the new incoming [wayland-rs 0.20](https://github.com/Smithay/wayland-rs/pull/165) API. Doing this work allows me to find out what is bad/missing from wayland-rs API, notably.

TODO:

- [x] wayland
  - [x] SHM
  - [x] Compositor
  - [x] Shell
  - [x] Output
  - [x] Seat
- [x] backend
  - [x] winit
  - [x] libinput
  - [x] drm
  - [x] session


vberger:
Okay, with this the porting should be complete.

r? @Drakulix :smile:

vberger:
With this last commit, we can even work around the rustc bug, and retrieve all functionality :tada:

examples/udev.rs
```diff
@@ -435,14 +421,15 @@ fn main() {
     libinput_event_source.remove();
 
     // destroy the udev backend freeing the drm devices
-    udev_event_source.remove().close(&mut event_loop)
+    udev_backend = *(downcast_impl(udev_event_source.remove()).unwrap_or_else(|_| unreachable!()));
```
Drakulix: Maybe put a small explanation here, what this does?

src/backend/udev.rs
```diff
@@ -48,11 +49,13 @@ pub struct UdevBackend<
     S: Session + 'static,
     T: UdevHandler<H> + 'static,
 > {
-    devices: Rc<RefCell<HashMap<dev_t, FdEventSource<(DrmDevice<SessionFdDrmDevice>, H)>>>>,
+    _handler: ::std::marker::PhantomData<H>,
```
Drakulix: Is there a good reason to keep this type parameter? Otherwise lets just drop it, if it is not required anymore.

src/wayland/shell/legacy/mod.rs
```diff
@@ -0,0 +1,355 @@
+//! Utilities for handling shell surfaces with the `wl_shell` protocol
+//!
+//! This module provides automatic handling of shell surfaces objects, by being registered
+//! as a global handler for `wl_shell`. This protocol is deprecated in favor of `xdg_shell`,
+//! thus this module is provided as a compatibility layer with older clients. As a consequence,
+//! you can as a compositor-writer decide to only support its fonctionnality in a best-effort
```
Drakulix: fonctionnality

src/backend/udev.rs
vberger: The reason to keep it is because it appears in the `T: UdevHandler<H>` constraint.

It may be possible to remove it, but I expect it'd require a significant refactor.

Drakulix: Well than lets skip that for now.

Drakulix:
Looks good for merge!

vberger:
Alright then, let's go ! :shipit:

PR #89, opened 2018-04-30
xwayland: core infrastructure

vberger:
This is the very core of an XWayland infrastructure. Notably it does not provide an X11 WM implementation.

The full exposed API is:

```rust
/// Trait to be implemented by you WM for Xwayland
///
/// This is a very low-level trait, only notifying you
/// when the connection with XWayland is up, or when
/// it terminates.
///
/// You WM must be able handle the XWayland server connecting
/// then disconnecting several time in a row, but only a single
/// connection will be active at any given time.
pub trait XWindowManager {
    /// The XWayland server is ready
    ///
    /// Your previlegied connection to it is this `RawFd`
    fn xwayland_ready(&mut self, fd: UnixStream, client: Client);
    /// The XWayland server has exited
    fn xwayland_exited(&mut self);
}

pub struct XWayland<WM: XWindowManager> { ... }

impl<WM: XWindowManager> XWayland<WM> {
    /// Start the XWayland server
    pub fn init(wm: WM, token: LoopToken, display: Rc<RefCell<Display>>) -> Result<XWayland<WM>,()> { ... }

}
```

### DIFF
--- a/src/xwayland/x11_sockets.rs
+++ b/src/xwayland/x11_sockets.rs
@@ -7,19 +7,15 @@ use nix::errno::Errno;
 use nix::sys::socket;
 
 /// Find a free X11 display slot and setup
-pub(crate) fn prepare_x11_sockets() -> Result<(u32, [UnixStream; 2]), ()> {
+pub(crate) fn prepare_x11_sockets() -> Result<(X11Lock, [UnixStream; 2]), ()> {
     for d in 0..33 {
         // if fails, try the next one
-        if let Err(()) = grab_lockfile(d) {
-            continue;
+        if let Ok(lock) = X11Lock::grab(d) {
+            // we got a lockfile, try and create the socket
+            if let Ok(sockets) = open_x11_sockets_for_display(d) {
+                return Ok((lock, sockets));
+            }
         }
-        // we got a lockfile, try and create the socket
-        if let Ok(fds) = open_x11_sockets_for_display(d) {
-            return Ok((d, fds));
-        }
-        // creating the sockets failed, for some readon ?
-        // release the lockfile and try with the next
-        release_lockfile(d);
     }
     // If we reach here, all values from 0 to 32 failed
     // we need to stop trying at some point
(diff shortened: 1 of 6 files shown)


src/xwayland/x11_sockets.rs
```diff
@@ -0,0 +1,126 @@
...
+                ::std::mem::drop(file);
+                let pid = ::nix::unistd::Pid::from_raw(::std::str::from_utf8(&spid)
+                    .map_err(|_| ())?
+                    .trim()
+                    .parse::<i32>()
+                    .map_err(|_| ())?);
+                if let Err(NixError::Sys(Errno::ESRCH)) = ::nix::sys::signal::kill(pid, None) {
+                    // no process whose pid equals the contents of the lockfile exists
+                    // remove the lockfile and try grabbing it again
+                    let _ = ::std::fs::remove_file(filename);
```
Drakulix: You should not discard this result. What if the file is read-only for our process? `create_new` will fail, reading will succeed, deleting will not and we start over. Possibly an endless recursion?

src/xwayland/x11_sockets.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub(crate) fn display(&self) -> u32 {
+        self.display
+    }
+}
+
+impl Drop for X11Lock {
+    fn drop(&mut self) {
+        // Cleanup all the X11 files
+        let _ = ::std::fs::remove_file(format!("/tmp/.X11-unix/X{}", self.display));
+        let _ = ::std::fs::remove_file(format!("/tmp/.X{}-lock", self.display));
```
Drakulix: At least print a warning with the Err if it fails. `if let Err(err) = ... { warn!(this.logger, "Err deleting file: {}", err); }` or something.

src/xwayland/x11_sockets.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+}
+
+/// Open the two unix sockets an X server listens on
+///
+/// SHould only be done after the associated lockfile is aquired!
```
Drakulix: Typo

src/xwayland/xserver.rs
```diff
@@ -0,0 +1,358 @@
...
+                // something went wrong :(
+                success = false;
+                break;
+            }
+        }
+    }
+
+    if success {
+        // signal the WM
+        wm.xwayland_ready(
```
Drakulix: So if this does not succeed, how would a user of smithay know when to retry starting XWayland?

src/xwayland/xserver.rs
```diff
@@ -0,0 +1,358 @@
...
+        CString::new("-rootless").unwrap(),
+        CString::new("-terminate").unwrap(),
+        CString::new("-wm").unwrap(),
+        CString::new(format!("{}", wm_socket.as_raw_fd())).unwrap(),
+    ];
+    for socket in listen_sockets {
+        args.push(CString::new("-listen").unwrap());
+        args.push(CString::new(format!("{}", socket.as_raw_fd())).unwrap());
+    }
+    // setup the environment: clear everything except PATH and XDG_RUNTIME_DIR
```
Drakulix: Why do we do this? That might be problematic for some XKB_ env variables and other stuff.

src/xwayland/xserver.rs
```diff
@@ -0,0 +1,358 @@
...
+    for (key, _) in env::vars_os() {
+        if key.to_str() == Some("PATH") || key.to_str() == Some("XDG_RUNTIME_DIR") {
+            continue;
+        }
+        env::remove_var(key);
+    }
+    // the WAYLAND_SOCKET var tells Xwayland where to connect as a wayland client
+    env::set_var("WAYLAND_SOCKET", format!("{}", wayland_socket.as_raw_fd()));
```
Drakulix: And environment variable with the number of our file descriptor? This is really how this works?

vberger: Well, yeah: https://github.com/wayland-project/wayland/blob/master/src/wayland-client.c#L1128

src/xwayland/xserver.rs
vberger: Well, the reasoning if that if XWayland crashed at startup, there is no real point restarting it later, it's unlikely to change anything.

It may be worth logging, though. Event if the Xorg logs will already contain a lot of things probably.

src/xwayland/x11_sockets.rs
vberger: Fair point, I didn't consider that, i'll take that into account.

src/xwayland/x11_sockets.rs
vberger: Yes, this is right, will add extensive logging.

PR #90, opened 2018-05-07
move examples into our own compositor

### LINKED ISSUE #59: Have a reference compositor
Like libweston has weston and wlroots has rootson, we could have a reference basic compositor built using smithay. This would help on several aspects:

- ensure all modules in smithays play smoothly together
- easily test smithay
- provide a kind-of-real-life example of how smithay is used

vberger:
This removes the examples of smithay and merges them into a basic compositor, which we'll one day be able to use as a testbed for smithay (notably to allow consistent repro of bugs). This would be the equivalent of rootson for wlroots.

I named it anvil because it's in the theme and I didn't find a better idea. Please bikeshed if you wish to.

I brutally merged the examples into a single program, and didn't try to refactor this further than "making it work". Leaving this for future work to actually make a compositor from it.

Fixes #59


vberger:
In the end, I have done some refactoring actually. I have factored the input-handling code for both winit and tty backend, deduplicating around 200 lines. :)

vberger:
Ah, this is a good point, didn't take that into account.

I'll bring the drm example back.

vberger:
@Drakulix do you have anything to add, or can I merge?

Drakulix:
I dont have the time to do a careful review today, but your last commit look good enough, that I would merge them anyway.

vberger:
Alright, :shipit: then. We'll be able to correct it afterwards anyway, this is only the beginning!

PR #93, opened 2018-05-13
anvil: fix shm buffer loading

vberger:
Refactor the SHM-buffer loading code in anvil to properly load buffers and convert them to RGBA that is internally used by glium.

This removes the hack in the fragment shader that caused the EGL buffers to have wrong colors (as it was translating ARGB to RGBA).

This is not perfect, as loading SHM buffers is now suboptimal, but ideally we would take advantage of OpenGL various loading capabilities if they were exposed by glium...

As a side note, with this we can run alacritty in anvil without any visual glitch! :tada:

### DIFF
--- a/anvil/src/glium_drawer.rs
+++ b/anvil/src/glium_drawer.rs
@@ -70,32 +70,32 @@ impl<T: Into<GliumGraphicsBackend<T>> + EGLGraphicsBackend + 'static> From<T> fo
 
         // compiling shaders and linking them together
         let program = program!(&display,
-			100 => {
-				vertex: "
-					#version 100
-					uniform lowp mat4 matrix;
-					attribute lowp vec2 position;
-					attribute lowp vec2 tex_coords;
-					varying lowp vec2 v_tex_coords;
-					void main() {
-						gl_Position = matrix * vec4(position, 0.0, 1.0);
-						v_tex_coords = tex_coords;
-					}
-				",
+            100 => {
+                vertex: "
+                    #version 100
+                    uniform lowp mat4 matrix;
+                    attribute lowp vec2 position;
+                    attribute lowp vec2 tex_coords;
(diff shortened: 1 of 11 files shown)


anvil/src/glium_drawer.rs
```diff
@@ -70,32 +70,32 @@ impl<T: Into<GliumGraphicsBackend<T>> + EGLGraphicsBackend + 'static> From<T> fo
...
+                    #version 100
+                    uniform lowp sampler2D tex;
+                    varying lowp vec2 v_tex_coords;
+                    void main() {
+                        lowp vec4 color = texture2D(tex, v_tex_coords);
+                        gl_FragColor.r = color.x;
                         gl_FragColor.g = color.y;
-                        gl_FragColor.b = color.x;
+                        gl_FragColor.b = color.z;
                         gl_FragColor.a = color.w;
```
Drakulix: Can't we fix the ARGB -> RGBA problem in the shader (much faster) instead of converting the buffer itself?
Just store the data like we receive it and assign color.y/z/w to gl_FragColor.r/g/b and x to a?

vberger: Well, the issue is that EGL clients are already in RGBA format.

Now, the good thing to do would be to load the SHM buffers in EGL directly specifying the correct format to OpenGL, [like weston does](https://github.com/wayland-project/weston/blob/925788fc76499290329863c0954275b74955b3a5/libweston/gl-renderer.c#L1568)..

Or possibly define a shader for each buffer type actually, that's maybe the cleanest indeed. I'll look into that.

vberger:
Okay, I rather seriously refactored the buffer loading logic in anvil, decoupling it from the protocol handling part as much as possible.

I also introduced a set of specialized fragment shaders to handle different buffer kinds.

PR #102, opened 2018-09-26
Upgrade winit backend to 0.17

trimental:
This PR upgrades the winit backend to 0.17. I couldn't test this in detail since I can't open applications in anvil because of an error about seat manager support but I would assume none of these changes would have much affect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


vberger:
> seat manager support

If is about data device manager? Smithay does not support it yet, while it's in the core protocol.

`weston-terminal` should work in anvil, though.

vberger:
For the record, this looks good. I just want to sort #101 before merging it. :+1:

PR #105, opened 2018-09-30
[travis] Only do cargo check for incomplete feature-sets

Drakulix:
As discussed on matrix, this is a try to speed up travis builds.

The reasoning is, that we just want to make sure all features do build independently,
but testing is only required once, so as long as tests are run for the "all" target, this should be enough.


Drakulix:
Time looks like it was reduced by roughly 1-2 minutes per job compared to the last #101 build.

vberger:
:+1:

PR #107, opened 2018-10-01
Refactor Seat::add_keyboard

jplatte:
With the new XkbConfig struct, it is now easier to just use xkbcommon's
default configuration, by moving the xkbcommon specific options into a
struct with according documentation.

Additionally, anvil now uses xkbcommon's defaults with all backends
(previously, the winit backend had a hardcoded french keyboard layout).


jplatte:
Clippy fails, but with an unrelated error. I think fixing that doesn't belong in this PR. I'll do that tomorrow evening, if noone else does.

vberger:
Clippy build is not blocking for PRs for now.

vberger:
Looks good, thanks! :+1:

PR #113, opened 2018-10-28
Spelling and capitalisation fixes

yzhs:
I ran a spell checker and fixed what I could clearly identify as typos. I also capitalised some fixed terms like "API" or "EGL" and enclosed identifiers in backticks in the comments.


vberger:
Thanks at lot!

PR #114, opened 2018-11-14
Fix compilation with calloop 0.4

pablostebler:
I also needed to add a few Debug `impl`s in calloop. https://github.com/Smithay/calloop/pull/5

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ tempfile = "2.1.5"
 slog = "2.1.1"
 slog-stdlog = "3.0.2"
 libloading = "0.4.0"
-wayland-client = { version = "0.20.5", optional = true }
+wayland-client = { version = "0.21.1", features = ["egl"], optional = true }
 winit = { version = "0.17.0", optional = true }
 drm = { version = "^0.3.1", optional = true }
 gbm = { version = "^0.4.0", optional = true, default-features = false, features = ["drm-support"] }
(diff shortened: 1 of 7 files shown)


vberger:
Hi, thanks for the pull request!

However, `InsertError` implements `Into<IoError>`, and I'd rather do the conversion rather than change the prototype for these functions.

There is work underway to significantly refactor this code, so I'd like to change it as little as possible in the meantime.

Cargo.toml
```diff
@@ -18,7 +18,7 @@ tempfile = "2.1.5"
 slog = "2.1.1"
 slog-stdlog = "3.0.2"
 libloading = "0.4.0"
-wayland-client = { version = "0.20.5", optional = true }
+wayland-client = { version = "0.21.1", features = ["egl"], optional = true }
```
vberger: I don't understand, why did you need to add this feature here?

pablostebler: It's no longer a default feature since 0.21. But in fact I suppose that there is a good reason not to use 0.21 so I should probably remove this change from my PR.

vberger: Ah right, I forgot about that. Okay! :+1: 

Upgrading wayland-client is completely fine, but in this case, could you also upgrade winit to 0.18 ? I believe the upgrade should be trivial.

pablostebler: Yes, it just compiles and anvil seems to work normally.

vberger:
From the travis build, it looks like there is some code in the logind backend that needs updating, too.

vberger:
Hi, thanks this looks good! :+1: 

There is just a small fmt issue before I can merge, see the travis build.

vberger:
Perfect, thanks!

PR #115, opened 2018-11-18
Data device implementation

### LINKED ISSUE #28: Data device handler
We'll need to do it at some point.

For now, I don't really know what it is and how it works, I've never studied it yet. AFAIK it's linked to drag'n'drop behavior, possibly with copy/paste too.

All is still do be done here.

vberger:
closes #28 

- [x] client to client selection
- [x] internal compositor access to the selection
- [x] client to client drag'n'drop
  - [x] let the compositor customize the action selection policy
- [x] internal compositor access to the drag'n'drop
- [x] write module-level doc


anvil/src/shell.rs
```diff
@@ -68,9 +68,9 @@ pub fn init_shell(
         compositor_token,
         move |shell_event| match shell_event {
             XdgRequest::NewToplevel { surface } => {
-                // place the window at a random location in the [0;300]x[0;300] square
+                // place the window at a random location in the [0;800]x[0;800] square
                 use rand::distributions::{IndependentSample, Range};
-                let range = Range::new(0, 300);
+                let range = Range::new(0, 800);
```
Drakulix: Drag'n'Drop does not work well on a small space, right? :laughing:

src/wayland/data_device/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+    }
+}
+
+/// Initialize the data device global
+///
+/// You need to provide a `(DndAction, DndAction) -> DndAction` closure that will arbitrate
+/// the choice of action resulting from a drag'n'drop session. Its first argument is the set of
+/// available actions (which is the intersection of the actions supported by the source and targets)
+/// and the second argument is the preferred action reported by the target. If no action should be
+/// chosen (and thus the drag'n'drop should abort on drop), return `DndAction::empty()`.
```
Drakulix: The callback argument does not seem to be explained?
And in the anvil implementation it is empty.

I assume this is this callback:
`- You can provide a callback closure to `init_data_device` to peek into the the actions of your clients`
from the module description?

Is this also where I would handle a server drag'n'drop?
Could maybe document how that would work exactly somewhere?

anvil/src/misc.rs
```diff
@@ -0,0 +1,19 @@
+use smithay::wayland_server::protocol::wl_data_device_manager::DndAction;
+
+pub fn dnd_action_chooser(available: DndAction, preferred: DndAction) -> DndAction {
```
Drakulix: Is there a reason to do anything different, then this function is doing?
Maybe we can provide this function as a default directly in smithay?

anvil/src/shell.rs
vberger: Yeah, testing was getting difficult with all the windows stacking onto each other, as anvil does not yet support moving windows

src/wayland/data_device/mod.rs
vberger: That's what you get for writing docs past midnight I guess. I'll detail all this more.

anvil/src/misc.rs
vberger: Yes, the wayland protocol files gives examples such as a compositor changing its behavior depending on key presses (possibly overriding the client preference). For example stuff like "holding `shift` for copy and holding `control` for move".

We could provide this function in smithay for users that don't want to customize it, though.

vberger:
Okay, the docs should be more clear now, if you want to give it a second pass @Drakulix ?

PR #116, opened 2018-11-21
Refactor drm backend (v3)

Drakulix:
Still work in progress (docs and anvil support as well as some new examples are missing).

Ready for a first review.

The biggest change is the split of the `DrmDevice` and `DrmBackend` into
- `LegacyDrmDevice` and `LegacyDrmSurface`
- `GbmDevice` and `GbmSurface`
- `EglDevice` and `EglSurface`

To be able to do this clean seperation new traits have been added.
Most importantly `Device` and `Surface`, describing drm devices in general
+ `RawDevice` and `RawSurface` providing the lowest layer and modesetting support, while the higher-level devices rely on other means of swapping buffers.

- `LegacyDrm*` implements the `Raw*`-traits, `Gbm*` requires a `Raw*`-object as it provides buffer management, but itself does not provide `Raw*`-implementations anymore, but instead the more high-level `page_flip`, `recreate` and `unlock_buffer` functions.

- `EglDevice` can wrap any `Device` that also implements egl's `NativeDisplay` trait (and where it's surface also implements `NativeSurface`), which is provided by the `GbmDevice` (and `GbmSurface`), through the methods listed above.

This provides very high composability, as the usual (and currently only viable) chain of an egl-gbm-drm-device, is not dictated anymore by the implementation, but by public-accessible traits.

In future this will allow us to add the following implementations easily:
- `AtomicDrmDevice` to support atomic modesetting (once [drm-rs current refactor](https://github.com/Smithay/drm-rs/issues/25) is done)
- `EglStreamDevice` to support nVidia's buffer mangement
- `VulkanDevice` to support Vulkan instead of OpenGL as a rendering api
- ... (probably some more)

+ this also allows users of the library to provide their own implementations for more exotic devices.

During the refactor some additional changes have been made:
- Usage of the calloop 0.4 update to utilize `EventedFd` instead of `EventedRawFd`
- Seperate egl and gl modules
- Provide a raw GL loader additionally to the mostly deprecated (although still working) glium wrapper
- Renaming the `GraphicsBackend` to the more fitting `CursorBackend`
- Don't let any graphical backend depend on `CursorBackend` anymore
- Rebuild the udev-backend, so that it no longer creates devices (move this responsibility to the library user). This greatly simplifies the backend (removing most of it's weird api quirks). Additionally this makes working with the new drm-devices much nicer, as the concrete type of the device is not dictated by the udev backend anymore.


src/backend/drm/gbm/surface.rs
```diff
@@ -0,0 +1,319 @@
...
+    <D as Device>::Return: ::std::borrow::Borrow<<D as RawDevice>::Surface>,
+    <D as RawDevice>::Surface: CursorBackend<'a>,
+    <<D as RawDevice>::Surface as CursorBackend<'a>>::CursorFormat=&'a Buffer,
+    <<D as RawDevice>::Surface as CursorBackend<'a>>::Error: ::std::error::Error + Send 
+{
+*/
+// But for now got to do this:
+
+impl<'a, A: AsRawFd + 'static> CursorBackend<'a> for GbmSurface<LegacyDrmDevice<A>> {
+    type CursorFormat = &'a ImageBuffer<Rgba<u8>, Vec<u8>>;
```
vberger: Does that mean that it will not be possible to efficiently handle clients that use opengl to draw their cursor images? (Tbh I have no idea if that can actually happen, but in general I'd think that the ability to use OpenGL on a surface should be orthogonal to its role)

Drakulix: Actually this does hardware compositing (draw the cursor via a seperate drm plane instead of drawing it via GL in software), which should be way faster (and does not require a redraw of the GL-Surface to be visible). So this is (when possible) the fastest implementation.

Converting certain formats (like the xcursor) into an ImageBuffer *could* be slow, but that should really be a problem for very frequent cursor changes.

In generally I would like to allow drawing a software cursor via GL as well, but I believe I need the `specialization` feature to be able to provide multiple stacking implementations and do this right. Maybe we could also move `CursorFormat` from an associated type to generic parameters of the trait, so we can actually provide multiple implementations depending on the actual format.

Right now every surface does provide an implementation that should be as fast as possible (although not very flexible).
Except for `GbmSurface`, which does only provide an implementation for `GbmSurface<LegacyDrmDevice<A>>`, if you provide your own `RawDevice` you will need to implement this yourself (for now).

vberger: Okay I see. So if a client decides to use OpenGL to draw its cursor image, the compositor will have to extract the `EGLImage` into an `ImageBuffer` to then composite it on the hardware, right? And the performance penalty should not be big (assuming the client doesn't set up ridiculously big cursor images)?

I agree this is a pretty niche question anyway, we can absolutely come back to it at a later time (or never if it's not needed).

Drakulix: Well the client could also decide not to use the `CursorBackend` (or clear the cursor by providing a clear image (maybe we should provide an api for that)) and just render a software cursor itself.

That might actually be the best way to handle this case. Provide an implementation for hardware compositing and let clients do their own software cursor, if they want to.

vberger: Yeah, lets forget about that for now, I'm just being needlessly nitpicking about hypothetical edge-case uses.

Drakulix:
TODO (notes to myself):

- [ ] Remove arguments from `Device::create_surface`. We do not modeset on creation anyway and this way a commit is always pending after creation to actually set the given state. Instead read-out the current state and let the user figure out any changes, potentially avoiding modeset operations all together!
- [ ] Documentation of everything
- [ ] At least a `LegacyDrmDevice` example
- [ ] More comments on anvil code
- [ ] More comments on the new drm code
- [ ] Update comments to new module locations
- [ ] Check links in doc-comments
- [ ] Make GbmSurface rendering functions `unsafe`
  - [ ] Also fix this in the `gbm` crate, those functions can trigger a SIGSEV inside mesa way too easily if missused
- [ ] Fix lib`input` crate to require the use of `unsafe` to obtain a File Descriptor and implement `AsRawFd` instead

src/backend/graphics/cursor.rs
```diff
@@ -0,0 +1,34 @@
+/// Functions to render cursors on any graphics backend independently from it's rendering techique.
+pub trait CursorBackend<'a> {
+    /// Format representing the image drawn for the cursor.
+    type CursorFormat: 'a;
+
+    /// Error the underlying backend throws if operations fail
+    type Error;
+
+    /// Sets the cursor position and therefor updates the drawn cursors position.
```
vberger:
```suggestion
    /// Sets the cursor position and therefore updates the drawn cursors position.
```

src/backend/graphics/gl.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Returns the address of an OpenGL function.
+    ///
+    /// Supposes that the context has been made current before this function is called.
+    unsafe fn get_proc_address(&self, symbol: &str) -> *const c_void;
+
+    /// Returns the dimensions of the window, or screen, etc in points.
+    ///
+    /// That are the scaled pixels of the underlying graphics backend.
+    /// For nested compositors this will respect the scaling of the root compositor.
```
vberger: What do you mean by "respect the scaling"? AFAIK the winit backend transposes back the pixels to expose the "physical pixels" (as winit calls them), is that what you have in mind? This formulation could be clarified.

src/backend/graphics/gl.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Returns the pixel format of the main framebuffer of the context.
+    fn get_pixel_format(&self) -> PixelFormat;
+}
+
+/// Loads a Raw GLES Interface for a given `GLGraphicsBackend`
+///
+/// This remains valid as long as the underlying `GLGraphicsBackend` is alive
+/// and may only be used in combination with the backend. Using this with any
+/// other gl context *may* cause undefined behavior.
```
vberger: Can't there be some link between the backend and the returned `Gles2` ? Maybe adding a `PhantomData` lifetime to `Gles2` ?

src/backend/drm/gbm/egl.rs
```diff
@@ -0,0 +1,95 @@
...
+        if let Err(err) = GbmSurface::recreate(self) {
+            error!(self.0.logger, "Failure recreating internal resources: {:?}", err);
+            false
+        } else {
+            true
+        }
+    }
+
+    fn swap_buffers(&self) -> ::std::result::Result<(), SwapBuffersError> {
+        // this is save since `eglSwapBuffers` will have been called exactly once
```
vberger:
```suggestion
        // this is safe since `eglSwapBuffers` will have been called exactly once
```

src/backend/drm/gbm/surface.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+}
+
+/// Gbm surface for rendering
+pub struct GbmSurface<D: RawDevice + 'static>(pub(super) Rc<GbmSurfaceInternal<D>>);
+
+impl<D: RawDevice + 'static> GbmSurface<D> {
+    /// Flips the underlying buffers.
+    ///
+    /// The surface will reported being already flipped until the matching events
```
vberger:
```suggestion
    /// The surface will report being already flipped until the matching event
```

src/backend/graphics/gl.rs
Drakulix: Sadly not. This is auto-generated code from `gl_generator` and it would probably also cause a lot of inconveniences for the user. GL Contexts are weird and many gl functions are unsafe anyway. If you want a safe alternative there is `glium`.

PR #117, opened 2018-11-22
Protocol support for client-provided cursor surfaces

### LINKED ISSUE #49: Implement `wl_pointer.set_cursor`

vberger:
depends on #115 

fixes #49 

This introduces the necessary code in the wayland protocol handlers to correctly process requests from clients to set custom cursor images, both in general and in a DnD context.

This does *not* implement any of the necessary logic in anvil to actually display them.

### DIFF
--- a/src/wayland/data_device/mod.rs
+++ b/src/wayland/data_device/mod.rs
@@ -24,23 +24,33 @@
 //! - the freestanding function `start_dnd` allows you to initiate a drag'n'drop event from the compositor
 //!   itself and receive interactions of clients with it via an other dedicated callback.
 //!
+//! The module also defines the `DnDIconRole` that you need to insert into your compositor roles enum, to
+//! represent surfaces that are used as a DnD icon.
+//!
 //! ## Initialization
 //!
 //! ```
 //! # extern crate wayland_server;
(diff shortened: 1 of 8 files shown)


src/wayland/data_device/mod.rs
```diff
@@ -71,7 +84,22 @@ pub enum DataDeviceEvent {
...
-    DnDStarted(Option<Resource<wl_data_source::WlDataSource>>),
+    DnDStarted {
+        /// The data source provided by the client
+        ///
+        /// If it is `None`, this means the DnD is restricted to surfaces of the
+        /// same client and the client will manage data transfert by itself.
+        source: Option<Resource<wl_data_source::WlDataSource>>,
+        /// The icon the client requested to be used to replace the cursor icon
+        /// during the drag'n'drop.
+        icon: Option<Resource<wl_surface::WlSurface>>,
```
emersion: Note that the icon doesn't replace necessarily the cursor. It's usually an additional surface drawn underneath the cursor (which is usually set by the compositor depending on the current drag-and-drop action).

vberger: Thanks for the clarification! I'll correct the wording.

PR #124, opened 2018-12-15
Migrate to Rust 2018 Edition

Drakulix:
So this a run of `cargo fix --edition` for all our feature combinations + actually setting `edition = "2018"` for our crates + adding the `rust_2018_idioms` lint to our warnings + fixing those + formatting.

That lint is currently not doing very much. It is supposed to fix "old-style" code and is what is running, when using `cargo fix --edition-idioms`. But right now that only seems to include usage of the `dyn` keyword for trait objects, and explicit anonymous lifetimes as well as useless `extern crate` definitions.

A lot more changes are recommend, like importing macros instead of using `macro_use`. But especially for macros this requires a hack to support all editions for libraries in the moment ([e.g.](https://github.com/rust-lang-nursery/log/pull/288)) and I am not sure, if all our dependencies are prepared for that right now.

So instead lets go with the pace of ecosystem and fix up new lints as they popup in the `rust_2018_idioms` group in the next rust updates, instead of trying the whole code base now.

Depends on #122 & #123  (as everything needed to be compiling to run `cargo fix`, which is how I noticed those errors and decided to put them into our travis build).

### DIFF
--- a/anvil/src/glium_drawer.rs
+++ b/anvil/src/glium_drawer.rs
@@ -18,15 +18,15 @@ use smithay::{
         egl::{BufferAccessError, EGLImages, Format},
         graphics::{gl::GLGraphicsBackend, glium::GliumGraphicsBackend},
     },
+    reexports::wayland_server::{
+        protocol::{wl_buffer, wl_surface},
+        Resource,
+    },
     wayland::{
         compositor::{roles::Role, SubsurfaceRole, TraversalAction},
         data_device::DnDIconRole,
         seat::CursorImageRole,
         shm::with_buffer_contents as shm_buffer_contents,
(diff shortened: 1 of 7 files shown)


vberger:
Do we still need all these `pub extern crate` in the 2018 edition?

Drakulix:
> Do we still need all these `pub extern crate` in the 2018 edition?

Tested it and we can indeed get rid of `pub extern crate` without losing `#[doc(hidden)]` for all cases we do not need `#[macro_use]` for (until we fixed our macro imports sometime later).

vberger:
Apparently there are still some fmt issues. LGTM other than that :+1:

PR #125, opened 2018-12-16
Removed --tty-raw from README

buffet:
Anvil doesn't seem to have the --tty-raw backend anymore, it was still being mentioned in the README though. I just removed it.


vberger:
Thanks! Indeed we removed the backend very recently during a refactor and the README was not update.

PR #138, opened 2019-05-03
Wayland dmabuf handler

vberger:
This is a first attempt at a handler for the `linux-dmabuf` protocol.

I still need to:

- [x] write the module-level documentation
- [x] improve logging

What do you think of the overall API @Drakulix ?

### DIFF
--- a/src/wayland/dmabuf/mod.rs
+++ b/src/wayland/dmabuf/mod.rs
@@ -1,0 +1,377 @@
+use std::{cell::RefCell, os::unix::io::RawFd, rc::Rc};
+
+pub use wayland_protocols::unstable::linux_dmabuf::v1::server::zwp_linux_buffer_params_v1::Flags;
+use wayland_protocols::unstable::linux_dmabuf::v1::server::{
+    zwp_linux_buffer_params_v1::{
+        Error as ParamError, RequestHandler as ParamRequestHandler, ZwpLinuxBufferParamsV1 as BufferParams,
+    },
+    zwp_linux_dmabuf_v1,
+};
+use wayland_server::{protocol::wl_buffer, Display, Global, NewResource};
+
+/// Representation of a Dmabuf format, as advertized to the client
+pub struct Format {
+    /// The format identifier
+    pub format: u32,
+    /// High part of the supported modifiers
+    pub modifier_hi: u32,
+    /// Low part of the supported modifiers
+    pub modifier_lo: u32,
+    /// Number of planes used by this format
+    pub plane_count: u32,
+}
+
+/// A plane send by the client
+pub struct Plane {
+    /// The file descriptor
+    pub fd: RawFd,
+    /// The plane index
+    pub plane_idx: u32,
+    /// Offset from the start of the Fd
+    pub offset: u32,
+    /// Stride for this plane
+    pub stride: u32,
+    /// High part of the modifiers for this plane
+    pub modifier_hi: u32,
+    /// Low part of the modifiers for this plane
+    pub modifier_lo: u32,
+}
+
+/// The complete information provided by the client to create a dmabuf buffer
+pub struct BufferInfo {
+    /// The submitted planes
+    pub planes: Vec<Plane>,
+    /// The width of this buffer
+    pub width: i32,
+    /// The height of this buffer
+    pub height: i32,
+    /// The format in use
+    pub format: u32,
+    /// The flags applied to it
+    ///
+    /// This is a bitflag, to be compared with the `Flags` enum reexported by this module.
+    pub flags: u32,
+}
+
+/// Handler trait for dmabuf validation
+///
+/// You need to provide an implementation of this trait that will validate the parameters provided by the
+/// client and import it as a dmabuf.
+pub trait DmabufHandler {
+    /// The data of a successfully imported dmabuf.
+    ///
+    /// This will be stored as the `user_data` of the `WlBuffer` associated with this dmabuf.
+    type BufferData: 'static;
+    /// Validate a dmabuf
+    ///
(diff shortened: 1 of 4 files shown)


src/wayland/dmabuf/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+//! To setup the dmabuf global, you will need to provide 2 things:
+//!
+//! - a list of the dmabuf formats you wish to support
+//! - an implementation of the `DmabufHandler` trait
+//!
+//! The list of supported format is just a `Vec<Format>`, where you will enter all the (format, modifier)
+//! couples you support.
+//!
+//! The implementation of the `DmabufHandler` trait will be called whenever a client has finished setting up
+//! a dma buffer. You will be handled the full details of the sclient's submission as a `BufferInfo` struct,
```
Drakulix: Typo: **s**client

src/wayland/dmabuf/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+//! and you need to validate it and maybe import it into your rendered. The `BufferData` associated type
```
Drakulix: Typo: rendere**d** (I guess you meant rendere**r**?)

src/wayland/dmabuf/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+//! allows you to store any metadata of handle to the resource you need into the created `wl_buffer`,
```
Drakulix: Typo: Probably 'metadata **or** handle'?

src/wayland/dmabuf/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+//!
+//! struct MyBufferData {
+//!     /* ... */
+//! }
+//!
+//! impl Drop for MyBufferData {
+//!     fn drop(&mut self) {
+//!         // This is called when all handles to this buffer have been dropped,
+//!         // both client-side and server side.
+//!         // You can now free the associated resource in your renderer.
```
Drakulix: Typo: resource**s**

src/wayland/dmabuf/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+    },
+    zwp_linux_dmabuf_v1,
+};
+use wayland_server::{protocol::wl_buffer, Display, Global, NewResource};
+
+/// Representation of a Dmabuf format, as advertized to the client
+pub struct Format {
+    /// The format identifier.
+    ///
+    /// It must be a `DRM_FORMAT` code, as defined by the libdrm's drm_fourcc.h. The Linux kernel's DRM
```
Drakulix: I think there is a typed variant of this in drm-rs. We probably want to use it. You cannot use dmabufs with the drm-system anyway. So dmabuf-support would need to depend on the `backend_drm` feature or pull in drm-rs at least.

src/wayland/dmabuf/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+    pub planes: Vec<Plane>,
+    /// The width of this buffer
+    pub width: i32,
+    /// The height of this buffer
+    pub height: i32,
+    /// The format in use
+    pub format: u32,
+    /// The flags applied to it
+    ///
+    /// This is a bitflag, to be compared with the `Flags` enum reexported by this module.
```
Drakulix: Isn't there a nice crate to handle bitmaps a little bit better?

src/wayland/dmabuf/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+            ParamError::Incomplete as u32,
+            format!(
+                "Format {:x} requires {} planes but got {}.",
+                format.format, format.plane_count, max_plane_set
+            ),
+        );
+        return false;
+    }
+    // Width and height must be positivie
+    if width < 1 || height < 1 {
```
Drakulix: Does it need to be strictly greater? Isn't a 1x1 buffer valid as well?

src/wayland/dmabuf/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+            None => {
+                params.as_ref().post_error(
+                    ParamError::OutOfBounds as u32,
+                    format!("Size overflow for plane {}.", plane.plane_idx),
+                );
+                return false;
+            }
+            Some(e) => e,
+        };
+        if let Ok(size) = ::nix::unistd::lseek(plane.fd, 0, ::nix::unistd::Whence::SeekEnd) {
```
Drakulix: Won't we want to reset the Seek ptr? Otherwise this might break, if the user later converts the fd to a `File` or similar.

src/wayland/dmabuf/mod.rs
vberger: Oh, good point, do you mean https://docs.rs/drm/0.3.4/drm/buffer/enum.PixelFormat.html ?

Drakulix: Yes, that should be it

src/wayland/dmabuf/mod.rs
vberger: The correct thing to do would be to add the indication that this is a bitflag to the protocol file, so that wayland-scanner would directly generate it as a bitflag rather than an enum...

src/wayland/dmabuf/mod.rs
vberger: This code allows a 1x1 buffer, it errors if either width or height is strictly smaller than 1

src/wayland/dmabuf/mod.rs
vberger: Interesting point. I assumed it was not necessary for drm import later, as weston leaves it as it. I guess it cannot hurt to reset it though.

PR #143, opened 2019-08-23
drm-rs 0.4

Drakulix:
Work-in-progress branch for the next iteration of `drm-rs`'s api (now supporting atomic-modesetting and planes much better!).


Drakulix:
Looks good to me.

@Slabity How far off would be a 0.4 release (including my remaining PRs) for smithay? It would be nice to avoid having a git-dependency.

PR #149, opened 2019-12-31
Upgrade to dbus 0.7, change imports to match.

christophebiocca:
This is step one of upgrading: no real code changes, still using the old API.
Will follow this up with an actual upgrade to the recommended new way of doing things.


vberger:
Seems pretty straightforward, thanks!

Issue #94, opened 2018-05-13
Implement shell interactions (move, resize) on anvil


vberger:
Right now, anvil ignores the `move` and `resize` events from the clients, we need to implement them at some point.

PR #153, opened 2020-02-01
Implement the Move request

YaLTeR:
Based on #152 for my convenience. New commits start from Feb 1.

cc #94

This is WIP as in it works, but it hard-codes a few things that shouldn't be hard-coded. I'm opening this draft PR to ask for comments on the approach and if I'm doing this right.


anvil/src/shell.rs
```diff
@@ -45,6 +46,56 @@ pub type MyWindowMap = WindowMap<
...
+
+    fn button(
+        &mut self,
+        handle: &mut PointerInnerHandle<'_>,
+        button: u32,
+        state: ButtonState,
+        serial: u32,
+        time: u32,
+    ) {
+        // TODO: should use the button that initiated the grab, not hardcode left.
```
vberger: Indeed, and this will require some changes in Smithay to expose this from the implicit grab that is generated whenever a button is pressed, so that that implicit `ClickGrab` can be upgraded into a `MoveSurfaceGrab`.

anvil/src/shell.rs
```diff
@@ -45,6 +46,56 @@ pub type MyWindowMap = WindowMap<
...
+        // TODO: where does this magic number come from?
```
vberger: This is the code for this button click. It's standard, likely inherited from X11. Not much to do apart from defining a constant somewhere I guess.

anvil/src/shell.rs
```diff
@@ -45,6 +46,56 @@ pub type MyWindowMap = WindowMap<
...
+        if state == ButtonState::Released && button == 0x110 {
+            // TODO: this should use a new serial, right? Because the original one
+            // is used below.
+            handle.unset_grab(SCOUNTER.next_serial(), time);
```
vberger: Hmm, I think it should rather use the same serial for both.

anvil/src/shell.rs
```diff
@@ -100,6 +151,45 @@ pub fn init_shell(
...
+
+                // If the focus was for a different surface, ignore the request.
+                if start_data.focus.is_none()
+                    || !compositor_token.contains_in_tree(
+                        surface.get_surface().unwrap(),
+                        &start_data.focus.as_ref().unwrap().0,
+                    )
+                {
+                    return;
+                }
```
vberger: Why do we need this check? I'd say checking that the serial matches the grab & that the focused surface belongs to the correct client should be enough.

src/wayland/data_device/server_dnd_grab.rs
```diff
@@ -212,6 +212,10 @@ where
         // we just forward the axis events as is
         handle.axis(details);
     }
+
+    fn start_data(&self) -> &GrabStartData {
+        unimplemented!()
+    }
```
vberger: This will need to be implemented

src/wayland/seat/pointer.rs
```diff
@@ -219,6 +228,22 @@ impl PointerHandle {
...
+    /// Access the current location of this pointer in the global space
+    pub fn current_location(&self) -> (f64, f64) {
+        self.inner.borrow().location
+    }
+}
+
+/// Data about the event that started the grab.
+#[derive(Clone)]
+pub struct GrabStartData {
+    /// The focused surface and its location, if any, at the start of the grab.
```
vberger: I think this is the location of the pointer relative to the surface, not the location of the surface itself. This module has no knowledge of the surface locations.

anvil/src/shell.rs
```diff
@@ -111,21 +201,62 @@ pub fn init_shell(
...
+                    }
+
+                    let toplevel = SurfaceKind::Wl(surface);
+                    let initial_window_location = shell_window_map.borrow().location(&toplevel).unwrap();
+
+                    let grab = MoveSurfaceGrab {
+                        start_data,
+                        window_map: shell_window_map.clone(),
+                        toplevel,
+                        // TODO: use location at the time of the event with the given serial.
```
vberger: Tbh I don't think this is worth doing in anvil, or at all.

We don't want to store a large history of what occurred just because some client decided to wait a long time between the start of its implicit grab and the moment it sent a `move` request...

anvil/src/shell.rs
YaLTeR: Well, this is the "focused surface belongs to the correct client" part of the check. It verifies that the grab's focused surface is the surface that send Move, or one of its children.

src/wayland/data_device/server_dnd_grab.rs
YaLTeR: Where should I get the grab start data? The `start_dnd` function only accepts a serial.

src/wayland/seat/pointer.rs
YaLTeR: This location comes from the motion handler, which (in anvil) gets it from `WindowMap::get_surface_under`, which returns the surface location. But, as I said in the IRC, I don't actually see this field used for any purpose anywhere, so it might as well be not what was intended.

anvil/src/shell.rs
YaLTeR: Maybe we can re-purpose the surface location part of the focus to be the toplevel location and then just use that?

anvil/src/shell.rs
vberger: I think a looser check like would be sufficient:

```rust
let same_client = start_data.focus().as_ref().map(
    |sdata| sdata.as_ref().same_client_as(surface.get_surface())
).unwrap_or(false);
if !same_client {
    return;
}
```

I don't see why we should prevent a client from moving ones of its windows using a grab initiated on an other window, as long as both windows belong to the same client.

src/wayland/data_device/server_dnd_grab.rs
vberger: We can change the `start_dnd` function to take more arguments to fill out the grab data if needed.

src/wayland/seat/pointer.rs
vberger: Ah right, it's been too long I touched at this part, sorry.

anvil/src/shell.rs
vberger: The "location of the toplevel surface to which this surface belongs at the start of the drag" is imo way too specific to be something stored at this level. Would it be of any use to anything other than the move grab?

YaLTeR: Yeah, I'm not sure about this. I suppose that focus location can be just removed for now, and, should it become really useful, a better mechanism can be made.

YaLTeR: Ah, actually it's used in `motion` to get the surface-relative coordinates. My bad, didn't notice.

src/wayland/seat/pointer.rs
```diff
@@ -226,6 +235,17 @@ impl PointerHandle {
...
+    pub focus: Option<(WlSurface, (f64, f64))>,
+    /// The button that initiated the grab.
+    pub button: u32,
+    /// The location of the click that initiated the grab.
+    pub location: (f64, f64),
+}
+
```
vberger: One last thing, I think it'd be best to make explicit in the doc that the locations here are expressed in *compositor space* (rather than relative to some surface).

PR #155, opened 2020-02-03
anvil: retrieve buffer dimensions on commit

YaLTeR:
This fixes the issue outlined in the IRC where moving the pointer around the surface would frequently result in the surface behind it getting focus. The problem was that when the buffer is committed, the texture (used for size calculations) gets set to None, and is updated only on the next draw, which can happen after `self_update`. `self_update` would thus think the surface was zero-sized.

An easy way to test this is to open something with sctk decorations and mouse over the buttons. Before the patch the buttons randomly lose color (because they become unfocused due to this bug), after the patch they work correctly.

### DIFF
--- a/anvil/src/buffer_utils.rs
+++ b/anvil/src/buffer_utils.rs
@@ -1,0 +1,59 @@
+use std::{cell::RefCell, rc::Rc};
+
+use slog::Logger;
+
+#[cfg(feature = "egl")]
+use smithay::backend::egl::{BufferAccessError, EGLDisplay};
+use smithay::{
+    reexports::wayland_server::protocol::wl_buffer::WlBuffer,
+    wayland::shm::with_buffer_contents as shm_buffer_contents,
+};
+
+/// Utilities for working with `WlBuffer`s.
+#[derive(Clone)]
+pub struct BufferUtils {
+    #[cfg(feature = "egl")]
+    egl_display: Rc<RefCell<Option<EGLDisplay>>>,
+    log: Logger,
+}
+
+impl BufferUtils {
+    /// Creates a new `BufferUtils`.
+    #[cfg(feature = "egl")]
+    pub fn new(egl_display: Rc<RefCell<Option<EGLDisplay>>>, log: Logger) -> Self {
+        Self { egl_display, log }
+    }
+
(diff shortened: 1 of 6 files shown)


src/backend/egl/mod.rs
```diff
@@ -472,6 +472,46 @@ impl EGLDisplay {
...
+
+    /// Try to receive the dimensions of a given [`WlBuffer`].
+    ///
+    /// In case the buffer is not managed by EGL (but e.g. the [`wayland::shm` module](::wayland::shm))
+    /// a [`BufferAccessError::NotManaged`](::backend::egl::BufferAccessError::NotManaged) is returned with the original buffer
+    /// to render it another way.
+    pub fn egl_buffer_dimensions(
+        &self,
+        buffer: WlBuffer,
+    ) -> ::std::result::Result<(i32, i32), BufferAccessError> {
```
vberger: I think this function can just be `fn(&self, &WlBuffer) -> Option<(i32, i32)>` tbh. The reuse of `BufferAccessError` here makes thing more complicated than they need to be.

anvil/src/shell.rs
```diff
@@ -157,26 +166,27 @@ fn surface_commit(surface: &wl_surface::WlSurface, token: CompositorToken<Roles>
                     old_buffer.release();
                 }
                 data.texture = None;
+                // Should the buffer be discarded if this fails?
```
vberger:
```suggestion
                // If this fail, the buffer will be discarded later by the drawing code
```

anvil/src/buffer_utils.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// Creates a new `BufferUtils`.
+    #[cfg(not(feature = "egl"))]
+    pub fn new(log: Logger) -> Self {
+        Self { log }
+    }
+
+    /// Returns the dimensions of an image stored in the buffer.
+    #[cfg(feature = "egl")]
+    pub fn dimensions(&self, buffer: WlBuffer) -> Option<(i32, i32)> {
```
vberger: I think this method (and `shm_buffer_dimensions`) can be changed to only require a `&WlBuffer` as well, and avoid unneeded clones.

PR #156, opened 2020-02-03
anvil: implement the Resize request

### LINKED ISSUE #94: Implement shell interactions (move, resize) on anvil

YaLTeR:
Everything works, the geometry is computed as in Weston.

[Video!](https://streamable.com/52jd8) (the graphical glitches during gnome-calculator resizing are some OBS issue with capturing OpenGL that I've been seeing in other applications too)

Closes #94 

I tested the xdg-shell resizing on weston-terminal, alacritty and evince. I tested the wl-shell resize on the old weston-terminal ([this](https://gitlab.freedesktop.org/wayland/weston/commit/c0f17aba5e007821bf08e58b5033e7e545f6ab10) is the last commit before the xdg-shell switch) and on the sctk clients with xdg-shell disabled in anvil; the sctk clients seem to be computing the initial size incorrectly so the size jumps a little, but otherwise work fine.

The alacritty decorations being off-sync during a left resize seems like a bug in alacritty because it's not present in sctk clients and also it's not present when alacritty is running some program like `evince` and waiting for its output.

### DIFF
--- a/anvil/Cargo.toml
+++ b/anvil/Cargo.toml
@@ -14,6 +14,7 @@ rand = "0.6"
 glium = { version = "0.23.0", default-features = false }
 wayland-server = "0.23"
 xkbcommon = "0.4.0"
+bitflags = "1.2.1"
 
 [dependencies.smithay]
 path = ".."
--- a/anvil/src/shell.rs
+++ b/anvil/src/shell.rs
@@ -7,9 +7,12 @@ use std::{
 use rand;
 
 use smithay::{
-    reexports::wayland_server::{
-        protocol::{wl_buffer, wl_pointer::ButtonState, wl_shell_surface, wl_surface},
-        Display,
+    reexports::{
+        wayland_protocols::xdg_shell::server::xdg_toplevel,
+        wayland_server::{
+            protocol::{wl_buffer, wl_pointer::ButtonState, wl_shell_surface, wl_surface},
+            Display,
+        },
     },
     utils::Rectangle,
     wayland::{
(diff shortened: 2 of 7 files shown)


vberger:
LGTM overall, but I'll need to make an other pass on it as it is quite a large patch.

A few things:

- I'm not fan of using the `wl_shell_surface` types internally, as `wl_shell` is basically deprecated, it feels awkward to use its types as the backbone of the logic
- Some of the awkwardness about state sharing will be solved with the update to wayland-rs 0.25 and the use of `DispatchData`, so I'm fine with keeping it like that for now

YaLTeR:
> * I'm not fan of using the `wl_shell_surface` types internally, as `wl_shell` is basically deprecated, it feels awkward to use its types as the backbone of the logic

I just used it because it's a bitflags over the resize edges, more convenient to work with. The one in xdg is unfortunately not a bitflags (and [was intentionally changed that way](https://gitlab.freedesktop.org/wayland/wayland-protocols/merge_requests/13#note_390799)). I suppose I could make an exact duplicate of that bitflags somewhere (where?).

vberger:
> I suppose I could make an exact duplicate of that bitflags somewhere (where?).

I think I'd prefer that yes. Maybe in anvil for now, and we can see afterwards. My guess is that at least some part of the ResizeGrab and MoveGrab can be moved into smithay itself, but I think this would better be done after the wayland-rs 0.25 update, once things have cleared up.

PR #158, opened 2020-02-08
anvil.shell: don't send first configure to WlShell

YaLTeR:
It doesn't need it, and it also takes the (0, 0) size literally, which makes everything try to be the smallest size possible initially.

Notably fixes sctk clients being one pixel wide initially if wl_shell is used.


vberger:
Thanks!

PR #165, opened 2020-02-13
[WIP] Integration tests for Anvil

YaLTeR:
This PR adds some basic infrastructure for Anvil integration tests, namely helper methods for launching Anvil and feeding it simulated input. It also adds a basic Move test.

TODO:
- How do I do headless EGL?
- How do I properly wait for requests, in this case the Move request? For this particular test ideally I only need to wait for the Move request, but for a Resize test I'll need to wait for a Resize request, then potentially for an Ack, and then for a surface commit or something, and I'm not sure how to handle all of this. Maybe just dispatch with a 100 ms timeout lol.
- Do the logging in such a way that it's only visible when the test fails or something, so there's no spam.
- It would be nice to spawn SCTK clients right in the test, but requires more effort.

### DIFF
--- a/anvil/src/lib.rs
+++ b/anvil/src/lib.rs
@@ -1,0 +1,21 @@
+#![warn(rust_2018_idioms)]
+
+#[macro_use]
+extern crate glium;
+#[macro_use]
+extern crate slog;
+#[macro_use(define_roles)]
+extern crate smithay;
+
+#[macro_use]
+pub mod shaders;
+pub mod buffer_utils;
+pub mod glium_drawer;
+pub mod input_handler;
+pub mod shell;
+pub mod shm_load;
+#[cfg(feature = "udev")]
+pub mod udev;
+pub mod window_map;
+#[cfg(feature = "winit")]
+pub mod winit;
--- a/anvil/src/main.rs
+++ b/anvil/src/main.rs
@@ -1,27 +1,15 @@
 #![warn(rust_2018_idioms)]
 
 #[macro_use]
-extern crate glium;
-#[macro_use]
 extern crate slog;
-#[macro_use(define_roles)]
-extern crate smithay;
 
 use slog::Drain;
 use smithay::reexports::{calloop::EventLoop, wayland_server::Display};
 
-#[macro_use]
-mod shaders;
-mod buffer_utils;
-mod glium_drawer;
-mod input_handler;
-mod shell;
-mod shm_load;
 #[cfg(feature = "udev")]
-mod udev;
-mod window_map;
+use anvil::udev;
 #[cfg(feature = "winit")]
-mod winit;
+use anvil::winit;
 
 static POSSIBLE_BACKENDS: &[&str] = &[
     #[cfg(feature = "winit")]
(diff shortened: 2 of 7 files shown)


vberger:
Okay so, first of all I'm impressed what you managed here, that's quite cool already! :smiley: 

Now, to answer your questions, my current point of view is:

> How do I do headless EGL?

I'm fine with the tests not compatible with the `egl` feature for now. Maybe it'd be worth adding some clear error message using [`compile_error!`](https://doc.rust-lang.org/std/macro.compile_error.html) for now.

In the long term, I think it'll be worth adding a proper headless EGL backend for Smithay (to support things like purely virtual displays for example), and once this is done anvil will be able to use that for its tests.

> How do I properly wait for requests, in this case the Move request? For this particular test ideally I only need to wait for the Move request, but for a Resize test I'll need to wait for a Resize request, then potentially for an Ack, and then for a surface commit or something, and I'm not sure how to handle all of this. Maybe just dispatch with a 100 ms timeout lol.

In the case of move, maybe it can be worth adding more introspection methods on `PointerHandle` to check more precisely what kind of grab is current. For example a `fn name(&self) -> &'static str` method to the `PointerGrab` trait allowing grabs to identify themselves, and  symmetric method on `PointerHandle` to check the kind of grab. With that, it's only a matter of looping on flush/dispatch until the expected kind of grab has started.

> Do the logging in such a way that it's only visible when the test fails or something, so there's no spam.

Doesn't `cargo test` already capture test output unless they fail?

> It would be nice to spawn SCTK clients right in the test, but requires more effort.

Yeah, I think we can keep this for latter.

One last point, I think it'd actually be worth adding a proper `VirtualInputBackend` to smithay itself, rather than improvising one in anvil. It could be used for tests, but also for things like virtual input devices.

YaLTeR:
> Maybe it'd be worth adding some clear error message using [`compile_error!`](https://doc.rust-lang.org/std/macro.compile_error.html) for now.

Do you mean to `tests/common/mod.rs`, so that you can't forget to add `#![cfg(not(feature = "egl"))]` to the actual test cases?

> In the case of move, maybe it can be worth adding more introspection methods on `PointerHandle` to check more precisely what kind of grab is current. For example a `fn name(&self) -> &'static str` method to the `PointerGrab` trait allowing grabs to identify themselves, and symmetric method on `PointerHandle` to check the kind of grab. With that, it's only a matter of looping on flush/dispatch until the expected kind of grab has started.

What about waiting for acks or surface commits? The resize grab ends earlier than the commit, but the final size should be tested after the commit.

> Doesn't `cargo test` already capture test output unless they fail?

For some reason it didn't capture the slog output.

vberger:
> Do you mean to `tests/common/mod.rs`, so that you can't forget to add `#![cfg(not(feature = "egl"))]` to the actual test cases?

That sounds right!

> What about waiting for acks or surface commits? The resize grab ends earlier than the commit, but the final size should be tested after the commit.

That's a good question. This might require some deeper integration into anvil maybe. :/

> For some reason it didn't capture the slog output.

Indeed, apparently cargo only captures stdout, not stderr. Maybe slog can be configured to output to stdout?

vberger:
Given how smithay has diverged, I'm going to close this PR and leave #265 as a tracking issue for this kind of testing functionality.

PR #166, opened 2020-02-13
wayland: fix deadlock in SurfaceData::cleanup

YaLTeR:
If the surface is not alive and is present among its own children, the code would deadlock.

### DIFF
--- a/src/wayland/compositor/tree.rs
+++ b/src/wayland/compositor/tree.rs
@@ -74,11 +74,12 @@ where
         }
         // orphan all our children
         for child in &my_data.children {
-            // don't do anything if this child is ourselves
-            if child.as_ref().equals(surface.as_ref()) {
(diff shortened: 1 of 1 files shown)


src/wayland/compositor/tree.rs
```diff
@@ -74,11 +74,12 @@ where
         }
         // orphan all our children
         for child in &my_data.children {
-            // don't do anything if this child is ourselves
-            if child.as_ref().equals(surface.as_ref()) {
+            let child_mutex = child.as_ref().user_data::<Mutex<SurfaceData<R>>>().unwrap();
+            if child_mutex as *const _ == my_data_mutex as *const _ {
```
vberger:
```suggestion
            if std::ptr::eq(child_mutex, my_data_mutex) {
```

I think this would be better style, while keeping the intent clear.

vberger:
Thanks!

PR #167, opened 2020-03-21
Update winit dependency to version 22.0

Siborgium:
Summary:
* use PhysicalSize as winit seems to prefer it to LogicalSize everywhere
* run_return instead of poll_events, explicit ControlFlow synchronization
* other changes are mostly related to naming and changes in winit's internal structure

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -381,8 +381,7 @@ impl KeyboardKeyEvent for WinitKeyboardInputEvent {
 pub struct WinitMouseMovedEvent {
     size: Rc<RefCell<WindowSize>>,
     time: u32,
-    x: f64,
-    y: f64,
+    logical_position: LogicalPosition<f64>,
 }
 
 impl BackendEvent for WinitMouseMovedEvent {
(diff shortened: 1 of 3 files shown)


src/backend/winit.rs
```diff
@@ -692,170 +701,181 @@ impl InputBackend for WinitInputBackend {
...
-                                surface.resize(physical_size.width as i32, physical_size.height as i32, 0, 0);
+                    },
+                    Event::WindowEvent { event, .. } => {
+                        let duration = Instant::now().duration_since(*time);
+                        let nanos = duration.subsec_nanos() as u64;
+                        let time = ((1000 * duration.as_secs()) + (nanos / 1_000_000)) as u32;
+                        match (event, handler.as_mut(), events_handler.as_mut()) {
+                            (WindowEvent::Resized(psize), _, events_handler) => {
+                                trace!(logger, "Resizing window to {:?}", psize);
+                                window.window().set_inner_size(psize);
```
vberger: IIRC when we receive a `Resized` event from winit, it now means that the window has been resized, so I think this line should be removed. It might be worth double checking though.

Siborgium: Done.

vberger:
I'll clone your branch and test it further, but overall this looks like. Is there something more you want to add before marking this as a non-draft PR?

vberger:
So, after some testing, there is an issue with dpi handling. When I run anvil on winit on my scale 2 screen, the pointer is not at the correct location.

Siborgium:
I'll look into it.

Siborgium:
Could you give me some insight on how to reproduce the issue? I tried running anvil with all combinations of xrandr set to {1.0, 2.0} and WINIT_X11_SCALE_FACTOR {1.0, 2.0}, and all of them seem to work fine.

vberger:
Hmm, that is weird, I cannot reproduce it again. Must have been a mistake from my side, sorry.

In that case, your PR looks good as is. Is there something more you want to add before removing the "draft" marker ?

vberger:
Pushed to master, but Github doesn't seem to understand it. :man_shrugging:

PR #171, opened 2020-04-04
Update wayland-rs to 0.25

### LINKED ISSUE #169: Update smithay to wayland-server 0.25
A lot of refactoring to do, as the wayland-server API changed significantly.

jplatte:
Will resolve #169

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -263,7 +263,6 @@ pub fn run_udev(mut display: Display, mut event_loop: EventLoop<AnvilState>, log
         {
             running.store(false, Ordering::SeqCst);
         } else {
-            // FIXME: should we supply non-() data?
             if state.need_wayland_dispatch {
                 display.borrow_mut().dispatch(std::time::Duration::from_millis(0), &mut state);
             }
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -35,6 +35,10 @@ use smithay::{
         udev::{primary_gpu, udev_backend_bind, UdevBackend, UdevHandler},
     },
     reexports::{
+        calloop::{
+            generic::{EventedFd, Generic},
+            EventLoop, LoopHandle, Source,
+        },
         drm::control::{
             connector::{Info as ConnectorInfo, State as ConnectorState},
             crtc,
@@ -44,10 +48,6 @@ use smithay::{
         input::Libinput,
         nix::{fcntl::OFlag, sys::stat::dev_t},
         wayland_server::{
-            calloop::{
-                generic::{EventedFd, Generic},
-                EventLoop, LoopHandle, Source,
-            },
             protocol::{wl_output, wl_surface},
             Display,
         },
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -36,7 +36,7 @@ use smithay::{
     },
     reexports::{
         calloop::{
-            generic::{EventedFd, Generic},
+            generic::{SourceFd, Generic},
             EventLoop, LoopHandle, Source,
         },
         drm::control::{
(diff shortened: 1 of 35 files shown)


src/xwayland/xserver.rs
```diff
@@ -159,10 +159,10 @@ fn launch<WM: XWindowManager + 'static>(inner: &Rc<RefCell<Inner<WM>>>) -> Resul
         guard
             .wayland_display
             .borrow_mut()
-            .create_client(wl_me.into_raw_fd())
+            .create_client(wl_me.into_raw_fd(), &mut ())
```
jplatte, line 162: No idea what the extra argument is for, but it wanted a`&mut impl Any` and I gave it one.

vberger: This is what will be used as `DispatchData`, as such it should be provided by the user and forwarded here.

src/wayland/compositor/handlers.rs
```diff
@@ -118,24 +115,24 @@ where
...
+    surface.quick_assign({
+        let mut implem = SurfaceImplem::make(log, implem);
+        move |surface, req, _| implem.receive_surface_request(req, surface.deref().clone())
+    });
+    surface.assign_destructor(Filter::new(|surface, _, _| SurfaceData::<R>::cleanup(&surface)));
     surface
+        .as_ref()
+        .user_data()
+        .set(|| SurfaceData::<R>::init(&surface));
+    surface.deref().clone()
```
jplatte, line 135: Is this translation about right?

src/xwayland/xserver.rs
jplatte: Is this at all related to `XWayland::init`s `Data` type parameter?

src/wayland/compositor/handlers.rs
vberger: Hmm, no. The `user_data()` should be `set_threadsafe(|| SurfaceData::<R>::new())`, and then you should keep the `SurfaceData::<R>::init(&surface)`.

jplatte: Okay, I was thinking since there was previously `implement_closure` and `implement_threadsafe`, only the latter would result in `user_data().set_threadsafe()`. How do I determine which set fn to use?

jplatte: Also, this will require `R: Send`. Is that an expected change?

jplatte: I've pushed another commit assuming that `R: Send` is okay and all `.set`s need to be `.set_threadsafe`s.

vberger: I'm not entirely sure it's good to require `R: Send`, but it's more in line with the internal implementation of `SurfaceData` which is done in terms of mutexes. Let's go with that for now.

anvil/src/winit.rs
```diff
@@ -193,7 +193,8 @@ pub fn run_winit(display: &mut Display, event_loop: &mut EventLoop<()>, log: Log
         {
             running.store(false, Ordering::SeqCst);
         } else {
-            display.flush_clients();
+            // FIXME: should we supply non-() data?
```
vberger, line 196: No, currently anvil does not use the `DispatchData`, this can be left for a future refactor.

anvil/src/udev.rs
```diff
@@ -253,15 +254,20 @@ pub fn run_udev(mut display: Display, mut event_loop: EventLoop<()>, log: Logger
...
+
     while running.load(Ordering::SeqCst) {
         if event_loop
-            .dispatch(Some(::std::time::Duration::from_millis(16)), &mut ())
+            .dispatch(Some(::std::time::Duration::from_millis(16)), &mut state)
             .is_err()
         {
             running.store(false, Ordering::SeqCst);
         } else {
             // FIXME: should we supply non-() data?
```
jplatte, line 266: You forgot to remove this comment.

PR #172, opened 2020-04-05
winit: Don't resize in resize handler

dcz-self:
The resize handler with a window resize request causes oscillations between two window sizes (presumably the inner and the outer). While it's not clear what part of the stack causes it, checking `window.get_inner_size()` already presents the correct size.

Either way, changing the same property in its change handler is a recipe for feedback loops as this one and should be avoided whenever possible, and then carefully fortified too.


vberger:
Thanks! Looks like at some point the behavior of winit changed and we forgot to update this part.

PR #177, opened 2020-04-11
Update to udev 0.4

Drakulix:
This upgrades our udev dependency to 0.4, which does not require a context anymore, therefor simplifying our code.

This also updates input-rs in one go, as it also depends on udev and the old version still expects a context, which we are now unable to provide.

The new input-rs version also enables us to use functions introduced in newer libinput versions. But because we are not doing so now, I have for now disabled its features.

But we are re-exporing libinput and users might have use-cases for the later versions, so we should likely make this configurable in the future (through more features....).


vberger:
Looks like the travis failures are legit

vberger:
> But we are re-exporing libinput and users might have use-cases for the later versions, so we should likely make this configurable in the future (through more features....).

Users can still enable libinput features by explicitly depending on the crate with these features enabled. Cargo unifies the features set of crates across the dependency graph.

Siborgium:
Anvil build fails with these changes. It uses `Libinput::new_with_udev`, but Smithay does not have `input/udev` because it imports `input` disabling its default features.

Siborgium:
I fixed it in #179 
It compiles and works fine on my machine.

PR #179, opened 2020-04-12
Fix Anvil build

Siborgium:
Add `backend_libinput_udev` feature to Smithay
This feature enables backend_libinput, backend_udev and input/udev features.
Use `backend_libinput_udev` feature in Anvil to fix build

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,14 +44,13 @@ slog-term = "2.3"
 gl_generator = { version = "0.10", optional = true }
 
 [features]
-default = ["backend_winit", "backend_drm_legacy", "backend_drm_gbm", "backend_drm_egl", "backend_libinput_udev", "backend_session_logind", "renderer_glium", "xwayland", "wayland_frontend"]
+default = ["backend_winit", "backend_drm_legacy", "backend_drm_gbm", "backend_drm_egl", "backend_libinput", "backend_udev", "backend_session_logind", "renderer_glium", "xwayland", "wayland_frontend"]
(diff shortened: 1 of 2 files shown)


Drakulix:
Could you change this PR to make anvil depends on `input` and `input/udev` instead? We do not need the feature in smithay to expose any new functionality, it would be just there for the re-export. So in my eyes this is a problem the compositor needs to deal with instead.

Siborgium:
However, there is little sense in importing another crate just for the sake of one feature. If user needs functionality behind "input/udev", they'd be required to dig into Smithay manifest and look up what features are enabled. Exposing guts to everyone isn't good imo, especially since input and udev are expected to be frequently combined together.

That's what I thought.

vberger:
> However, there is little sense in importing another crate just for the sake of one feature. If user needs functionality behind "input/udev", they'd be required to dig into Smithay manifest and look up what features are enabled.

Currently, the user is responsible for initializing the `Libinput` context anyway, so *given the current status*, I think having the user enable the feature itself makes sense. Documentation should made be clear about that though.

That said, I think it would make sense to provide in Smithay some glue code to automatically initialize a libinput context and link it to a `Session` and the `UdevBackend`. In such case, a `backend_libinput_udev` would definitely make sense, as we cannot simply trigger the `input/udev` feature whenever both `backend_udev` and `backend_libinput` are enabled, due to cargo limitations: https://github.com/rust-lang/cargo/issues/3494

Though this would probably be easier done once we revamp the signalling framework.

Siborgium:
Fine then, I'll push patch in a second.

PR #181, opened 2020-04-12
Update wayland-client & wayland-egl

Siborgium:
Update wayland-client dependency to 0.25.0
Add wayland-egl 0.25.0 dependency


Cargo.toml
```diff
@@ -50,7 +51,7 @@ backend_drm = ["drm", "failure"]
 backend_drm_legacy = ["backend_drm"]
 backend_drm_gbm = ["backend_drm", "gbm", "image"]
 backend_drm_egl = ["backend_drm", "backend_egl"]
-backend_egl = ["gl_generator"]
+backend_egl = ["gl_generator", "wayland-egl"]
```
vberger, line 54: I think `waykand-egl` should instead be introduced as a dependency to `backend_winit`, since this is the only part of Smithay that needs it (and it already depends on `backend_egl`).

Siborgium: Well, why not? Done.

Cargo.toml
```diff
@@ -22,7 +22,8 @@ tempfile = "3.0"
 slog = "2.1.1"
 slog-stdlog = "3.0.2"
 libloading = "0.5.0"
-wayland-client = { version = "0.23.4", features = ["egl"], optional = true }
+wayland-client = { version = "0.23.4", optional = true }
```
jplatte, line 25: This is still the old wayland-client...

vberger: Ah right.

Actually, since this change, `smithay` no longer depends on `wayland-client`, only `wayland-egl`. So this line can be removed altogether :sweat_smile:

PR #182, opened 2020-04-12
Ignore removal events for unrelated sessions

csnewman:
If a session is terminated in a different VT, all other sessions are notified of its removal.

Removal events are now filtered to ensure its about this session.


vberger:
Thanks! The travis failures are spurious, merging.

PR #183, opened 2020-04-12
Replace `failure` and `error-chain` by `thiserror`

vberger:
`error-chain` is basically abandoned, `failure` is in the process of being to.

This PR migrates Smithay to `thiserror`, which has the good side of only using the `std::error::Error` trait, meaning no one will have to deal with foreign types in the API, and errors adapters and whatever.

It currently fails to compile because `drm::SystemError` does not implement `std::error::Error`, but I suspect this can easily be fixed.


Drakulix:
:+1:

vberger:
So, I've temporarily restored the `failure` dependency to use `Compat<_>` on `drm::SystemError` in the meantime. And also converted the remaining manual implementations of `Error` to `thiserror`.

Drakulix:
> So, I've temporarily restored the `failure` dependency to use `Compat<_>` on `drm::SystemError` in the meantime. And also converted the remaining manual implementations of `Error` to `thiserror`.

Well then let's merge it and open an issue on the drm-rs repo, I guess?

vberger:
Yeah, I think it's the easiest way forward. :)

PR #184, opened 2020-04-14
Remove dependency on wayland-client

jplatte:
No longer required as noted in https://github.com/Smithay/smithay/pull/181#discussion_r408170484.

Should I also sort the dependencies alphabetically, to make oversights like the one that happened with wayland-client less likely in the future?


vberger:
The `winit` feature needs to be updated to depend on `wayland-egl` instead of `wayland-client`.

(note that `wayland-egl` already depends on `wayland-client/dlopen`)

jplatte:
The feature already depends on `wayland-egl`. I removed `wayland-client/dlopen`.

PR #185, opened 2020-04-14
Fix most clippy warnings

vberger:
The few remainings are complexity warnings on large functions, which will require more intense refactoring to address.


src/backend/libinput.rs
```diff
@@ -614,17 +614,17 @@ impl AsRawFd for LibinputInputBackend {
...
 /// Automatically feeds the backend with incoming events without any manual calls to
 /// [`dispatch_new_events`](InputBackend::dispatch_new_events). Should be used to achieve the smallest possible latency.
 pub fn libinput_bind<Data: 'static>(
     backend: LibinputInputBackend,
     handle: LoopHandle<Data>,
-) -> ::std::result::Result<
-    Source<Generic<SourceFd<LibinputInputBackend>>>,
-    InsertError<Generic<SourceFd<LibinputInputBackend>>>,
-> {
+) -> ::std::result::Result<Source<LibinputSource>, InsertError<LibinputSource>> {
```
jplatte, line 627: Doesn't seem necessary to use the fully qualified path to `Result` here.

src/backend/session/direct.rs
```diff
@@ -403,9 +403,10 @@ impl BoundDirectSession {
     pub fn unbind(self) -> DirectSessionNotifier {
         let BoundDirectSession { source, notifier } = self;
         source.remove();
-        match Rc::try_unwrap(notifier) {
```
jplatte, line 406: Did clippy suggest this? I don't actually find the new code more readable than the old.

src/wayland/compositor/mod.rs
```diff
@@ -105,6 +105,19 @@ struct Marker<R> {
     _r: ::std::marker::PhantomData<R>,
 }
 
+/// New buffer assignation for a surface
+pub enum BufferAssignation {
```
jplatte, line 109: Apparently "assignation" really is a word, but I would expect "assignment". English is not my first language, so I'm not entirely sure.

src/backend/libinput.rs
vberger: Yeah indeed, it's no longer necessary after #183

src/backend/session/direct.rs
Drakulix: Probably, because we are not using the Err(_)-contents. But it is fine in my eyes.

src/wayland/compositor/mod.rs
Drakulix: Sounds weird for sure, but might be a *german* bias. :laughing:

src/wayland/dmabuf/mod.rs
```diff
@@ -269,8 +275,7 @@ impl<H: DmabufHandler> ParamsHandler<H> {
         plane_idx: u32,
         offset: u32,
         stride: u32,
-        modifier_hi: u32,
-        modifier_lo: u32,
+        modifier: (u32, u32),
```
Drakulix, line 278: is it completely obvious, whats high and low, or should we add a little comment here?
Also below you are only using this as a u64. Is this two part really necessary?

src/backend/session/direct.rs
jplatte: Yeah, it's not like it's become clearly worse, but I'd say it's also not clearly better.

src/wayland/dmabuf/mod.rs
vberger: The two parts are how it is transmitted from the protocol. But maybe the conversion can be done before this function, indeed.

src/wayland/compositor/mod.rs
vberger: Ah well, I guess it's possibly a frenchism from me. :sweat_smile:

src/backend/session/direct.rs
jplatte: Oh, it complained that you're `panic!`ing with a general message, ignoring what's inside the `Err`?

vberger: Yes, it was hitting [`match_wild_err_arm`](https://rust-lang.github.io/rust-clippy/master/#match_wild_err_arm)

Drakulix: Well, if is not like the `if let` is much better. The clippy warning would be justified, if we could do something else then ignoring the error or panic'ing here, but that is not the case.

So I don't care, if we keep the change or add an explicit `#[allow(...)]` for `match_will_err_arm` here. Whatever you like better @vberger.

vberger: I guess a third alternative would be

```rust
.unwrap_or_else(|| panic!("....."))
```

Might be a little better?

I'm not really fan of sparkling a lot of `#[allow(...)]` everywhere tbqh.

vberger:
I found a few other easy fixes in anvil that I missed previously. The rest requires more significant refactor though.

PR #188, opened 2020-04-15
Refactor EGL backend

csnewman:
Refactors the `backend:egl` to more closely represent how EGL works under the hood.

I have kept the exposed `EglSurface` api the same, Now it represents a `EGLContext` and `EGLSurface` pair. Allowing the `GLBackend` to still be implemented without the creation of a new type. Similar has happened for `winit`.

The old `EGLDisplay` has now became `WaylandEGLDisplay`, which can be created from either a `EglDevice`, `EGLDisplay, or `WinitGraphicsBackend`.

These changes should then make exposing shared contexts, and improving thread safety in the future a lot easier.

### DIFF
--- a/src/backend/drm/egl/mod.rs
+++ b/src/backend/drm/egl/mod.rs
@@ -23,7 +23,7 @@ use crate::backend::egl::{display::EGLBufferReader, EGLGraphicsBackend};
 
 mod surface;
 pub use self::surface::*;
-use crate::backend::egl::context::GlAttributes;
+use crate::backend::egl::context::{GlAttributes, PixelFormatRequirements};
 use crate::backend::egl::display::EGLDisplay;
 
 #[cfg(feature = "backend_session")]
@@ -49,6 +49,8 @@ where
 {
     dev: EGLDisplay<B, D>,
     logger: ::slog::Logger,
+    default_attributes: GlAttributes,
+    default_requirements: PixelFormatRequirements,
 }
 
 impl<B, D> AsRawFd for EglDevice<B, D>
@@ -72,7 +74,34 @@ where
     ///
     /// Returns an error if the file is no valid device or context
     /// creation was not successful.
-    pub fn new<L>(mut dev: D, logger: L) -> Result<Self, Error<<<D as Device>::Surface as Surface>::Error>>
+    pub fn new<L>(dev: D, logger: L) -> Result<Self, Error<<<D as Device>::Surface as Surface>::Error>>
+    where
+        L: Into<Option<::slog::Logger>>,
+    {
+        EglDevice::new_with_defaults(
+            dev,
+            GlAttributes {
+                version: None,
+                profile: None,
+                debug: cfg!(debug_assertions),
+                vsync: true,
+            },
+            Default::default(),
+            logger,
+        )
+    }
+
+    /// Try to create a new [`EglDevice`] from an open device with the given attributes and
+    /// requirements as defaults for new surfaces.
+    ///
+    /// Returns an error if the file is no valid device or context
+    /// creation was not successful.
+    pub fn new_with_defaults<L>(
+        mut dev: D,
+        default_attributes: GlAttributes,
+        default_requirements: PixelFormatRequirements,
+        logger: L,
+    ) -> Result<Self, Error<<<D as Device>::Surface as Surface>::Error>>
     where
         L: Into<Option<::slog::Logger>>,
     {
@@ -83,6 +112,8 @@ where
         debug!(log, "Creating egl context from device");
         Ok(EglDevice {
             dev: EGLDisplay::new(dev, log.clone()).map_err(Error::EGL)?,
+            default_attributes,
+            default_requirements,
             logger: log,
         })
     }
@@ -141,21 +172,15 @@ where
     ) -> Result<Self::Surface, <Self::Surface as Surface>::Error> {
         info!(self.logger, "Initializing EglSurface");
 
-        // Device trait is unaware of opengl, so using sensible defaults
-        let attributes = GlAttributes {
-            version: None,
-            profile: None,
-            debug: cfg!(debug_assertions),
-            vsync: true,
-        };
-        let reqs = Default::default();
-
-        let context = self.dev.create_context(attributes, reqs).map_err(Error::EGL)?;
+        let context = self
+            .dev
+            .create_context(self.default_attributes, self.default_requirements)
+            .map_err(Error::EGL)?;
         let surface = self
             .dev
             .create_surface(
                 context.get_pixel_format(),
-                reqs.double_buffer,
+                self.default_requirements.double_buffer,
                 context.get_config_id(),
                 crtc,
             )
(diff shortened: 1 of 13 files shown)


src/backend/egl/context.rs
```diff
@@ -385,116 +93,125 @@ impl<B: native::Backend, N: native::NativeDisplay<B>> EGLContext<B, N> {
...
-impl<B: native::Backend, N: native::NativeDisplay<B>> Drop for EGLContext<B, N> {
+impl Drop for EGLContext {
     fn drop(&mut self) {
         unsafe {
             // we don't call MakeCurrent(0, 0) because we are not sure that the context
             // is still the current one
-            ffi::egl::DestroyContext((*self.display) as *const _, (*self.context) as *const _);
-            ffi::egl::Terminate((*self.display) as *const _);
+            if let Some(display) = self.display.upgrade() {
+                ffi::egl::DestroyContext((*display) as *const _, (*self.context) as *const _);
```
Drakulix, line 210: Something is fishy here:
- If we cannot cleanup the context without the display, the display should be no `Weak`-ptr.
- If the context is automatically destroyed, when the Display is, this is a dangling ptr. Wouldn't make_current fail in that case?

Whatever it is, I think I would prefer `EGLContext` to have a strong reference to `ffi:EGLDisplay`. Which means we need another slim wrapper around `ffi:EGLDisplay`, that is not user exposed and just implements `Deref` and `Drop` to probably destroy the display.

Same goes for the other types like `EGLSurface` or `EGLImages`.

src/backend/egl/display.rs
```diff
@@ -0,0 +1,618 @@
...
+        } else {
+            None
+        }
+    }
+}
+
+#[cfg(feature = "use_system_lib")]
+impl Drop for WaylandEGLDisplay {
+    fn drop(&mut self) {
+        if let Some(display) = self.display.upgrade() {
```
Drakulix, line 614: See `EGLContext` drop. We definitely always want to be able to unbind. Otherwise we might end up leaving `wl_drm` in a bad state, which is bad for long running compositors, which might swap their devices, when new graphic systems appear.

src/backend/egl/context.rs
Drakulix: I am noticing right now, that this is equivalent to the old behavior.
And I also notice, that I was not smart enough to make a comment on `EGLContext`s `Drop`-implementation on how destructing in EGL works.

But while we are at it, we may as well fix that. So the open question remains.

src/backend/egl/mod.rs
```diff
@@ -65,10 +65,12 @@ impl ::std::error::Error for EglExtensionNotSupportedError {}
...
-pub unsafe fn get_proc_address(symbol: &str) -> *const c_void {
-    let addr = CString::new(symbol.as_bytes()).unwrap();
-    let addr = addr.as_ptr();
-    ffi::egl::GetProcAddress(addr) as *const _
+pub fn get_proc_address(symbol: &str) -> *const c_void {
+    unsafe {
+        let addr = CString::new(symbol.as_bytes()).unwrap();
+        let addr = addr.as_ptr();
+        ffi::egl::GetProcAddress(addr) as *const _
+    }
```
Drakulix, line 73: Is it though? No way this can error?

src/backend/egl/display.rs
```diff
@@ -0,0 +1,622 @@
...
+            display: Arc::new(display as *const _),
+            egl_version,
+            extensions,
+            logger: log,
+            _backend: PhantomData,
+        })
+    }
+
+    /// Finds a compatible [`EGLConfig`] for a given set of requirements
+    pub fn choose_config(
```
Drakulix, line 138: I noticed, that clippy is still complaining about the `cognitive complexityy (35/25)` (or rather length) of this function. While we are it, we could indeed cut it down into some smaller and more manageable chunks. What do you think?

src/backend/egl/mod.rs
csnewman: The return type is void, and the doc do not mention any possible errors (Nor can I find any mention through googling). Apart from some sort of undocumented thread killing error handling within that I'm unaware of, its as safe as any of our other "safe" wrappers.

Drakulix: :+1:

src/backend/egl/display.rs
csnewman: Moving some of the reqs mapping onto some sort of helper on `PixelFormatRequirements` might help with the length

src/backend/egl/context.rs
csnewman: The `ffi::EGLDisplay` is kinda like a memory allocation group. Such that, when the display is destroyed, all resources associated with it are automatically cleaned up.

Drakulix: Well then we need to use strong-pointers and a drop-wrapper. Good you are refactoring this. Could have been a long time until I would have noticed this otherwise.

src/backend/drm/egl/mod.rs
```diff
@@ -73,31 +72,7 @@ where
...
-    /// Create a new [`EglDevice`] from an open device and given [`GlAttributes`]
-    ///
-    /// Returns an error if the file is no valid device or context
-    /// creation was not successful.
-    pub fn new_with_gl_attr<L>(
-        mut dev: D,
-        attributes: GlAttributes,
-        logger: L,
-    ) -> Result<Self, Error<<<D as Device>::Surface as Surface>::Error>>
+    pub fn new<L>(mut dev: D, logger: L) -> Result<Self, Error<<<D as Device>::Surface as Surface>::Error>>
```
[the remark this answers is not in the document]
csnewman: Yeah, I'll probably call the fields something like `default_attributes`

src/backend/egl/display.rs
csnewman: I'm unsure what would happen if the device that is providing the egl context dies/is removed. If the hardware device disappears, is the `ffi::EGLDisplay` is automatically killed driver side? meaning we'd be unable to call `UnbindWaylandDisplayWL` as the display ptr would be invalid. But, if the display ptr is still valid, then we would be able to call unbind. I'm unsure what the behaviour is around this.

It is also possible that the driver automatically removes the interfaces it has added.

Drakulix: This is actually a very interesting question, because I was not yet able to forcefully remove an initialized graphics card from smithay (and I bet it would crash). I guess we would notice this in the udev-handler or through drm-errors (which all should be not critical) very quickly and start cleaning up the device.

But if we are in the middle of a render, GL or EGL might just break and I am pretty confident, that there is an unwrap or a panic somewhere in the rendering code to make smithay crash.

But I guess I cannot reasonably debug this without something like an external thunderbolt gpu.

What I meant instead is if the compositor writer adds a way to remove and EGL- and all underlying devices during runtime (something like an `eject` command for e.g. a thunderbold gpu). The `EGLDisplay` would be destroyed with the `EGLDevice` and the `EGLSurface` would also long be gone, but the `EGLImages` or the `EGLBufferReader` might still be alive until it gets cleaned up in the next rendering step.

And until that `wl_drm` would maybe be in a bad state depending how mesa handles destruction of a bound `EGLDisplay`. Which is why I would also add a strong pointer to those two types to ensure the `EGLDisplay` always dies at last.

PR #189, opened 2020-04-15
Signaling framework

vberger:
This is the implementation of the Signaling framework discussed a few days ago in the chatroom.

This currently only brings the implementation itself. I'll add other commits migrating the different parts of the codebase to using it.


vberger:
Actually, migrating the rest of the codebase to the signaling framework will require significant refactor, so this may probably be better done as separate PRs.

PR #194, opened 2020-04-18
Atomic modesetting support

Drakulix:
This is finally ready for a first review...

TODO:
- [x] split up massive first commit
- [x] actually add a `raw_atomic_drm` example
- [x] add a `FallbackDevice<D1, D2>`, that tries one platform and falls back to another to make handling easier
- [x] adjust anvil to use it instead of hard-coding it to atomic

### DIFF
--- a/src/backend/drm/atomic/mod.rs
+++ b/src/backend/drm/atomic/mod.rs
@@ -1,3 +1,13 @@
+//!
+//! [`RawDevice`](RawDevice) and [`RawSurface`](RawSurface)
+//! implementations using the atomic mode-setting infrastructure.
+//!
+//! Usually this implementation will wrapped into a [`GbmDevice`](::backend::drm::gbm::GbmDevice).
+//! Take a look at `anvil`s source code for an example of this.
+//!
+//! For an example how to use this standalone, take a look at the raw_atomic_drm example.
+//!
+
 use std::cell::RefCell;
 use std::collections::HashMap;
 use std::os::unix::io::{AsRawFd, RawFd};
@@ -26,6 +36,7 @@ use self::surface::AtomicDrmSurfaceInternal;
 #[cfg(feature = "backend_session")]
 pub mod session;
 
+/// Open raw drm device utilizing atomic mode-setting
 pub struct AtomicDrmDevice<A: AsRawFd + 'static> {
     dev: Rc<Dev<A>>,
     dev_id: dev_t,
--- a/src/backend/drm/atomic/mod.rs
+++ b/src/backend/drm/atomic/mod.rs
@@ -196,9 +196,7 @@ impl<A: AsRawFd + 'static> AtomicDrmDevice<A> {
         let log = crate::slog_or_stdlog(logger).new(o!("smithay_module" => "backend_drm"));
         info!(log, "AtomicDrmDevice initializing");
 
-        let dev_id = fstat(fd.as_raw_fd())
-            .map_err(|source| Error::UnableToGetDeviceId(source))?
-            .st_rdev;
+        let dev_id = fstat(fd.as_raw_fd()).map_err(Error::UnableToGetDeviceId)?.st_rdev;
 
         let active = Arc::new(AtomicBool::new(true));
         let mut dev = Dev {
--- a/src/backend/drm/atomic/mod.rs
+++ b/src/backend/drm/atomic/mod.rs
@@ -115,6 +115,75 @@ impl<A: AsRawFd + 'static> Drop for Dev<A> {
     }
 }
 
+impl<A: AsRawFd + 'static> Dev<A> {
+    // Add all properties of given handles to a given drm resource type to state.
+    // You may use this to snapshot the current state of the drm device (fully or partially).
+    fn add_props<T>(&self, handles: &[T], state: &mut Vec<(T, PropertyValueSet)>) -> Result<(), Error>
+    where
+        A: AsRawFd + 'static,
+        T: ResourceHandle,
+    {
+        let iter = handles.iter().map(|x| (x, self.get_properties(*x)));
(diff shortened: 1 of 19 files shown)


src/backend/drm/atomic/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+        self.backends.borrow_mut().insert(crtc, Rc::downgrade(&backend));
+        Ok(AtomicDrmSurface(backend))
+    }
+
+    fn process_events(&mut self) {
+        match self.receive_events() {
+            Ok(events) => {
+                for event in events {
+                    if let Event::PageFlip(event) = event {
+                        trace!(self.logger, "Got event!");
```
vberger, line 326: Maybe this `trace!` can be made a little more explicit? I suspect this will not be very helpful in a trace-log.

src/backend/drm/atomic/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+                            if self
+                                .backends
+                                .borrow()
+                                .get(&event.crtc)
+                                .iter()
+                                .flat_map(|x| x.upgrade())
+                                .next()
+                                .is_some()
+                            {
+                                trace!(self.logger, "Handling event for backend {:?}", event.crtc);
```
vberger, line 337: Same here, given we know the event is a page-flip, the message should say it.

src/backend/drm/atomic/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+                                    handler.borrow_mut().vblank(event.crtc);
+                                }
+                            } else {
+                                self.backends.borrow_mut().remove(&event.crtc);
+                            }
+                        } else {
+                            debug!(self.logger, "Device not active. Ignoring PageFlip");
+                        }
+                    } else {
+                        trace!(self.logger, "Unrelated event");
```
vberger, line 348: I think this one can be made more explicit (unrelated to what?)

src/backend/drm/atomic/session.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl<A: AsRawFd + 'static> SessionObserver for AtomicDrmDeviceObserver<A> {
+    fn pause(&mut self, devnum: Option<(u32, u32)>) {
+        if let Some((major, minor)) = devnum {
+            if major as u64 != stat::major(self.dev_id) || minor as u64 != stat::minor(self.dev_id) {
+                return;
+            }
+        }
+
+        // TODO: Clear overlay planes (if we ever use them)
```
vberger, line 53: Do I understand correctly that this TODO is destinated to our future selves when we'll implement overlay planes support?

src/backend/drm/common/fallback.rs
```diff
@@ -0,0 +1,347 @@
+//!
+//! Types to make fallback device initialization easier
+//!
```
vberger, line 3: This module is quite a lot of boilerplate... Though I'm not really clear how this could be improved.

src/backend/drm/legacy/mod.rs
```diff
@@ -273,6 +233,9 @@ impl<A: AsRawFd + 'static> Device for LegacyDrmDevice<A> {
             }
         }
 
+        // Now try to do the least possible amount of changes to set this to the state the users requested
+        // TODO!
```
vberger, line 237: Does this `TODO` means that the current code is simply not as efficient as it could be, or is it something that needs to be fixed for merge?

src/backend/drm/atomic/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+            })?;
+
+        let plane_handles = dev.plane_handles().compat().map_err(|source| Error::Access {
+            errmsg: "Error loading planes",
+            dev: dev.dev_path(),
+            source,
+        })?;
+        let planes = plane_handles.planes();
+
+        fn add_props<A, T>(
```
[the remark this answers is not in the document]
Drakulix: sure thing. once I understand better, what atomic-helpers almost every compositor will implement anyway, I also want to add some more public helpers. These would be good candidates, maybe I should add a todo.

src/backend/drm/atomic/mod.rs
Drakulix: Oh yeah, that is just a left-over from debugging, but I guess it does not really hurt to leave in there and give it more meaning.

src/backend/drm/atomic/mod.rs
Drakulix: same

src/backend/drm/atomic/mod.rs
Drakulix: and again :sweat_smile:

src/backend/drm/atomic/session.rs
Drakulix: Yes. So I do not forget, where I needed to do that. :sweat_smile:

src/backend/drm/common/fallback.rs
Drakulix: I have tried to skip a lot of code by using macros.

I bet I could safe some more, but I wanted it to be a reasonable time investment then just optimizing for code size. If this code ever gets touched at all, it likely gets replaced, so I did not want to spend too much time with it.

The only thing that might be changed is adding some more trait implementations and the macros will help with that as well.

But it is not single purpose code at least. I want to re-use it for `EGLStreamDevice` vs `EGLDevice`, when nvidia support lands.

src/backend/drm/legacy/mod.rs
Drakulix: This should be a left-over, I have implemented this differently. Good catch though, I will clean this up.

PR #195, opened 2020-04-21
Implement Error trait for KeyboardError using thiserror

Siborgium:


Siborgium:
I'll split it in a second

PR #200, opened 2020-04-21
anvil: set saner values for key repeat rate

vberger:


kchibisov:
maybe this docs should also use saner values https://github.com/Smithay/smithay/blob/b54d95c15d7bef05c49ea6743495d6d6cfadf3c2/src/wayland/seat/mod.rs#L266 ?

vberger:
Oh, good point, thanks.

PR #202, opened 2020-04-25
Update wayland-rs and calloop

vberger:
Updates wayland-rs to 0.26 and calloop to 0.6.

This is a minimal update without any refactoring. I'll rather do the actual refactoring on a per-module basis in subsequent PRs.


vberger:
Merged in #204

Issue #203, opened 2020-04-26
Rework the internal structure of Smithay


vberger:
For tracking the refactoring work I'm planning:

- Rework the various backends into calloop `EventSource`s generating special events rather than using their handler traits
  - [ ] `xwayland`
  - [x] `backend.udev`
  - [x] `backend.libinput`
  - [x] `backend.drm`

- [x] Rework the `session` backends around the use of `Signaler`s instead of the currents observers

- [x] Update anvil to make extensive use of calloop's `Data` for state sharing

PR #204, opened 2020-04-27
Rework the udev backend as a calloop event source

vberger:
depends on #202 
cc #203 

This only reworks the Smithay part, anvil is simply minimally modified to adapt by keeping the `UdevHandler` and just making a plumbing closure. Final goal would be to fuse the `UdevHandler` into the main `AnvilState`, but I'd like feedback on the Smithay API before attacking that.

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -32,7 +32,7 @@ use smithay::{
             auto::{auto_session_bind, AutoSession},
             notify_multiplexer, AsSessionObserver, Session, SessionNotifier,
         },
-        udev::{primary_gpu, udev_backend_bind, UdevBackend, UdevHandler},
+        udev::{primary_gpu, UdevBackend, UdevEvent},
     },
     reexports::{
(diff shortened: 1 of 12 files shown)


anvil/src/udev.rs
```diff
@@ -226,7 +223,18 @@ pub fn run_udev(
...
+        udev_handler.device_added(dev, path.into())
+    }
+
+    let udev_event_source = event_loop
+        .handle()
+        .insert_source(udev_backend, move |event, _, _state| match event {
+            UdevEvent::Added { device_id, path } => udev_handler.device_added(device_id, path),
+            UdevEvent::Changed { device_id } => udev_handler.device_changed(device_id),
+            UdevEvent::Removed { device_id } => udev_handler.device_removed(device_id),
+        })
```
Drakulix, line 237: I really like, that the `UdevHandler` is now captured by a closure instead of being completely gone into the backend.
This style should be much more flexible to use. :+1:

PR #207, opened 2020-05-02
Clippy fixes

vberger:
These should finally fix the last remaining clippy warnings.


vberger:
At last, a green build status! :laughing:

PR #215, opened 2020-05-19
Rework session backend

### LINKED ISSUE #127: elogind support
As an alternative to systemd-logind, we need to support [elogind](https://github.com/elogind/elogind). It uses the exact same dbus API as logind, so it is just a matter of abstracting this part of the code over the 2 different system libs:

https://github.com/Smithay/smithay/blob/48264926e65d8eb4c05352ca38fb3fef5ce38f87/src/backend/session/dbus/logind.rs#L90-L93

vberger:
- [x] **Rework as calloop event sources**

  Rework the Session Notifiers so that they are calloop event sources by themselves, allowing them to be inserted by the user without the `bind_session` dance.

  Also update the logind backend to use the current dbus-rs API, rather than the deprecated one.
- [x] **use pkg-config to find logind lib**

  Introduce the `backend_session_elogind` cargo feature which pulls `backend_session_logind` and makes the logind session backend seek `libelogind.so` instead of `libsystemd.so`.

  Fixes #127
- [x] **Migrate to using Signaler**

  Change the session backend to rely on `Signaler` to propagate its signals. Also introduce the `Linkable` trait to allow generic composition of objects needing to listen for signals.

  cc #203


src/backend/session/dbus/logind.rs
```diff
@@ -611,3 +555,91 @@ impl AsErrno for Error {
...
+            Library::new("libsystemd.so").or_else(|_| Library::new("libelogind.so")).ok().and_then(|lib| {
+                unsafe {
+                    let sd_pid_get_session = match lib.get::<unsafe extern "C" fn(pid_t, *mut *mut c_char) -> c_int>(b"sd_pid_get_session\0") { Ok(s) => s.into_raw(), Err(_) => return None };
+                    let sd_session_get_seat = match lib.get::<unsafe extern "C" fn(*const c_char, *mut *mut c_char) -> c_int>(b"sd_session_get_seat\0") { Ok(s) =>  s.into_raw(), Err(_) => return None };
+                    let sd_session_get_vt = match lib.get::<unsafe extern "C" fn(*const c_char, *mut c_uint) -> c_int>(b"sd_session_get_vt\0") { Ok(s) => s.into_raw(), Err(_) => return None };
+                    Some(LibLogind {
+                        _lib: lib,
+                        sd_pid_get_session,
+                        sd_session_get_seat,
+                        sd_session_get_vt
```
kchibisov, line 639: Is there any benefit of doing that instead of normal dynamic linking? IIRC those libraries could be ABI incompatible and there's no way they could be in the same time on the same system. I feel like just addign pkg config at build.rs could make sense.

anvil/out.log
```diff
@@ -0,0 +1,160 @@
+unable to lock lockfile /run/user/1000/wayland-0.lock, maybe another compositor is running
```
kchibisov, line 1: I guess someone had intense debug session :thinking:

kchibisov:
note, I've tested this PR on logind system, and it seems to work fine for me. I don't have elogind system right now, but I've already tested similar changes when I wrote a patch to rust-systemd( I mean just using elogind for this 3 functions).

src/backend/drm/egl/session.rs
```diff
@@ -15,44 +14,47 @@ use crate::backend::egl::{
...
-        > + AsSessionObserver<S>
+        > + Linkable<SessionSignal>
         + 'static,
     <D as Device>::Surface: NativeSurface,
 {
-    fn observer(&mut self) -> EglDeviceObserver<S, <D as Device>::Surface> {
-        EglDeviceObserver {
-            observer: self.dev.borrow_mut().observer(),
+    fn link(&mut self, signaler: Signaler<SessionSignal>) {
+        self.dev.borrow_mut().link(signaler.clone());
```
Drakulix, line 42: What determines the order of the activate/pause calls of the different stacked devices, if we just pass down the signaler?

The order is important for some of the operations done in the different observers. I would prefer to call them explicitly.

vberger: Callbacks are always invoked in the order in which they have been registered, so calling `link()` on the underlying device before doing anything ensures it'll always receive the notification before.

PR #220, opened 2020-06-06
drm: Make surfaces `Send`

Drakulix:
This PR tries to make the drm `Surface`-implementations all `Send` to allow the rendering to happen in another thread.

@csnewman I would especially like you to test this branch with your flutter compositor.
You should be able to remove your hacks and move the surfaces without any problems.
Please report all errors, crashes, deadlock you encounter.
This PR is work-in-progress, because I have not given this much testing.
But since it was already working for your use-case and is only safer now, I do not expect this to cause much fuzz.

That said, I do not think it is reasonable to assume, that I did not miss anything.
I actually expect this to contain some rare and hard to debug errors, but I have no idea, how we should verify this approach without testing it in the wild. The documentation of thread-safety of the mesa-libs is very sparse, especially for gbm (egl is way better).

Also note that this is a compromise between usability, maintainability and performance constraints.
Making the devices `Send` is technically an option, but the code is already complicated enough and I do not think you would gain much. They need to be connected to an event loop anyway and you can just create them on their target thread.
This also tries to minimize thread-synchronization like mutexes on frequent calls, like page_flips, but is very naively implemented for anything else. Even in page_flip there are potential performance improvements by using more atomic types instead of locks, but I would say we profile first, before any premature optimization takes place.

### DIFF
--- a/src/backend/drm/atomic/surface.rs
+++ b/src/backend/drm/atomic/surface.rs
@@ -6,11 +6,9 @@ use drm::control::{
 };
 use drm::Device as BasicDevice;
 
-use std::cell::Cell;
 use std::collections::HashSet;
 use std::os::unix::io::{AsRawFd, RawFd};
-use std::rc::Rc;
-use std::sync::{atomic::Ordering, RwLock};
+use std::sync::{atomic::Ordering, Arc, Mutex, RwLock};
 
 use failure::ResultExt as FailureResultExt;
 
@@ -20,9 +18,9 @@ use crate::backend::graphics::CursorBackend;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CursorState {
-    pub position: Cell<Option<(u32, u32)>>,
-    pub hotspot: Cell<(u32, u32)>,
-    pub framebuffer: Cell<Option<framebuffer::Handle>>,
+    pub position: Option<(u32, u32)>,
+    pub hotspot: (u32, u32),
+    pub framebuffer: Option<framebuffer::Handle>,
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
@@ -39,14 +37,14 @@ pub struct Planes {
 }
 
 pub(in crate::backend::drm) struct AtomicDrmSurfaceInternal<A: AsRawFd + 'static> {
-    pub(super) dev: Rc<Dev<A>>,
+    pub(super) dev: Arc<Dev<A>>,
     pub(in crate::backend::drm) crtc: crtc::Handle,
-    pub(super) cursor: CursorState,
+    pub(super) cursor: Mutex<CursorState>,
     pub(in crate::backend::drm) planes: Planes,
     pub(super) state: RwLock<State>,
     pub(super) pending: RwLock<State>,
     pub(super) logger: ::slog::Logger,
-    pub(super) test_buffer: Cell<Option<(DumbBuffer, framebuffer::Handle)>>,
+    pub(super) test_buffer: Mutex<Option<(DumbBuffer, framebuffer::Handle)>>,
 }
 
 impl<A: AsRawFd + 'static> AsRawFd for AtomicDrmSurfaceInternal<A> {
@@ -60,7 +58,7 @@ impl<A: AsRawFd + 'static> ControlDevice for AtomicDrmSurfaceInternal<A> {}
 
 impl<A: AsRawFd + 'static> AtomicDrmSurfaceInternal<A> {
     pub(crate) fn new(
-        dev: Rc<Dev<A>>,
+        dev: Arc<Dev<A>>,
         crtc: crtc::Handle,
         mode: Mode,
         connectors: &[connector::Handle],
@@ -167,16 +165,16 @@ impl<A: AsRawFd + 'static> AtomicDrmSurfaceInternal<A> {
         let surface = AtomicDrmSurfaceInternal {
             dev,
             crtc,
-            cursor: CursorState {
-                position: Cell::new(None),
-                framebuffer: Cell::new(None),
-                hotspot: Cell::new((0, 0)),
-            },
+            cursor: Mutex::new(CursorState {
+                position: None,
+                framebuffer: None,
+                hotspot: (0, 0),
+            }),
             planes: Planes { primary, cursor },
             state: RwLock::new(state),
             pending: RwLock::new(pending),
             logger,
-            test_buffer: Cell::new(None),
+            test_buffer: Mutex::new(None),
         };
 
         Ok(surface)
@@ -202,10 +200,13 @@ impl<A: AsRawFd + 'static> AtomicDrmSurfaceInternal<A> {
                 dev: self.dev_path(),
                 source,
             })?;
-        if let Some((old_db, old_fb)) = self.test_buffer.replace(Some((db, fb))) {
+
+        let mut test_buffer = self.test_buffer.lock().unwrap();
+        if let Some((old_db, old_fb)) = test_buffer.take() {
             let _ = self.destroy_framebuffer(old_fb);
             let _ = self.destroy_dumb_buffer(old_db);
         };
+        *test_buffer = Some((db, fb));
 
         Ok(fb)
     }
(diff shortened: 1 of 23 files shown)


src/backend/drm/atomic/surface.rs
```diff
@@ -949,7 +949,7 @@ impl<A: AsRawFd + 'static> AtomicDrmSurfaceInternal<A> {
 
 /// Open raw crtc utilizing atomic mode-setting
 pub struct AtomicDrmSurface<A: AsRawFd + 'static>(
-    pub(in crate::backend::drm) Rc<AtomicDrmSurfaceInternal<A>>,
+    pub(in crate::backend::drm) Arc<AtomicDrmSurfaceInternal<A>>,
```
Drakulix, line 952: Our current structure of a shared internal `Surface` actually requires our internal surface structs not only to be `Send`, but `Sync` (this is mainly, because they are accessed by the session-callback, but that means we have true parallel access here).

All changes in `session.rs` or `device.rs` files usually stem from this.

src/backend/drm/atomic/surface.rs
```diff
@@ -39,14 +37,14 @@ pub struct Planes {
...
+    pub(super) dev: Arc<Dev<A>>,
     pub(in crate::backend::drm) crtc: crtc::Handle,
-    pub(super) cursor: CursorState,
+    pub(super) cursor: Mutex<CursorState>,
     pub(in crate::backend::drm) planes: Planes,
     pub(super) state: RwLock<State>,
     pub(super) pending: RwLock<State>,
     pub(super) logger: ::slog::Logger,
-    pub(super) test_buffer: Cell<Option<(DumbBuffer, framebuffer::Handle)>>,
+    pub(super) test_buffer: Mutex<Option<(DumbBuffer, framebuffer::Handle)>>,
```
Drakulix, line 47: Because our Surfaces need to be `Sync` (see below) all mutable access to surface internal data must be locked or be read-only.

src/backend/drm/atomic/surface.rs
Drakulix: Note that the session-handlers usually do not access every field of the surface, usually they only unlock buffers.
This means we can in theory cheat with our implementation as long, as we make sure, the functionality used by the session-code is thread-safe or cannot cause undefined behavior or bad state.

The user is actually unable to obtain two references to the same surface, as it is wrapped by the external surface type.
This means even the usage of a `Cell` could be fine, if we know for sure, we do not call the function from the weak-reference, that the device holds.

But this is mostly a very conservative implementation. If we would ever want to exploit this fact, we should probably split the internal-struct further to give only parts to the device and thus only require that part to be `Sync`.

src/backend/drm/atomic/surface.rs
```diff
@@ -1032,3 +1032,16 @@ impl<A: AsRawFd + 'static> RawSurface for AtomicDrmSurface<A> {
...
+    use super::AtomicDrmSurface;
+    use std::fs::File;
+
+    fn is_send<S: Send>() {}
+
+    #[test]
+    fn surface_is_send() {
+        is_send::<AtomicDrmSurface<File>>();
+    }
+}
```
Drakulix, line 1047: These tests make sure, that we do not violate our new `Send` feature unknowingly during any refactoring or feature development. It makes sure, that for a type we know is `Send` (e.g. File), the Surface is `Send` as well.

src/backend/drm/gbm/mod.rs
```diff
@@ -74,8 +73,9 @@ static LOAD: Once = Once::new();
 
 /// Representation of an open gbm device to create rendering surfaces
 pub struct GbmDevice<D: RawDevice + ControlDevice + 'static> {
-    pub(self) dev: Rc<RefCell<gbm::Device<D>>>,
-    backends: Rc<RefCell<HashMap<crtc::Handle, Weak<GbmSurfaceInternal<D>>>>>,
+    pub(self) dev: Arc<Mutex<gbm::Device<gbm::FdWrapper>>>,
```
Drakulix, line 76: We do not wrap the underlying device in the `gbm::Device` anymore, because we need to share it with the surface. If we would also wrap the `D`, we would also share the `D`, which would mean the device would need to be `Send`, which it isn't.

Note, that this does not violate our drop order. We can now drop the underlying device, while surfaces are still active, BUT a `GbmSurface` still holds a `Legacy/AtomicDrmSurface`, which holds a `Dev`, which is the struct that actually closes the device file descriptor on *Drop* and resets the device. So nothing to worry about here.

src/backend/drm/gbm/surface.rs
```diff
@@ -1,97 +1,146 @@
...
-    pub(super) front_buffer: Cell<Option<SurfaceBufferHandle<framebuffer::Handle>>>,
-    pub(super) next_buffer: Cell<Option<SurfaceBufferHandle<framebuffer::Handle>>>,
-    pub(super) recreated: Cell<bool>,
+pub struct Buffers {
+    pub(super) current_frame_buffer: Option<framebuffer::Handle>,
+    pub(super) front_buffer: Option<BufferObject<framebuffer::Handle>>,
+    pub(super) next_buffer: Option<BufferObject<framebuffer::Handle>>,
+}
+
+pub(super) struct GbmSurfaceInternal<S: RawSurface + 'static> {
```
Drakulix, line 19: Because we do not share the underlying device with the surface anymore (see the changes to gbm/mod.rs), we do not need the Device as a generic anymore, but only the Surface. This makes the types easier to understand imo, so I took the chance to refactor this as well. (Same applies to `EGLStreamSurface`).

src/backend/drm/gbm/surface.rs
```diff
@@ -1,97 +1,146 @@
...
         match result {
             Ok(_) => {
-                self.recreated.set(false);
-                self.current_frame_buffer.set(Some(fb));
+                self.recreated.store(false, Ordering::SeqCst);
+                let mut buffers = self.buffers.lock().unwrap();
+                buffers.current_frame_buffer = Some(fb);
                 Ok(())
             }
             Err(err) => {
```
Drakulix, line 146: `unlock_buffer` also locks `.buffers`, so we need to make sure that the scope above is closed already, before entering this function, otherwise we deadlock. (std::sync::Mutex is not re-entrant safe.)

src/backend/egl/context.rs
```diff
@@ -16,6 +16,8 @@ pub struct EGLContext {
     config_id: ffi::egl::types::EGLConfig,
     pixel_format: PixelFormat,
 }
+unsafe impl Send for EGLContext {}
+unsafe impl Sync for EGLContext {}
```
Drakulix, line 20: We need this for `EGLSurface` and the egl api does support it (it is internally locked).

src/backend/egl/surface.rs
```diff
@@ -4,22 +4,21 @@ use super::{ffi, native, EGLError, SurfaceCreationError, SwapBuffersError};
...
-};
+use std::sync::{Arc, atomic::{AtomicPtr, Ordering}};
+use std::ops::{Deref, DerefMut};
 
 /// EGL surface of a given EGL context for rendering
 pub struct EGLSurface<N: native::NativeSurface> {
     pub(crate) display: Arc<EGLDisplayHandle>,
     native: N,
-    pub(crate) surface: Cell<ffi::egl::types::EGLSurface>,
+    pub(crate) surface: AtomicPtr<nix::libc::c_void>,
```
Drakulix, line 14: We access the surface quite often, this should be much faster then a Mutex.
A little more difficult to handle, but manageable.

src/backend/egl/surface.rs
```diff
@@ -119,8 +115,11 @@ impl<N: native::NativeSurface> EGLSurface<N> {
                         SurfaceCreationError::NativeSurfaceCreationFailed(err) => {
                             SwapBuffersError::Underlying(err)
                         }
-                    })?
-            });
+                    })? as *mut _
+            }, Ordering::SeqCst);
+            if previous == surface && !surface.is_null() {
```
Drakulix, line 120: Lets swap the order around to make sure, the old pointer cannot be accessed by anyone else anymore, before we destroy it. (`compare_and_swap` is perfect here and also makes sure, that parallel recreations don't override each other.)

Drakulix:
Rebased on latest master without further changes.
Made some tests on my local system to confirm, that this still works like it is supposed to.
No noticeable performance regressions were found, even the debug build consumes roughly the same cpu percentage on my system. I do not consider this WIP anymore, but ready for review und later merge.

PR #221, opened 2020-06-12
anvil: support multiple outputs

### LINKED ISSUE #151: Mouse in anvil is limit to 1920x1080 square at the top left on 4k screen
When starting anvil from a tty with `cargo run --features logind -- --tty-udev` the mouse is limited to 1080p area in top left corner.

Drakulix:
~~Depends on #220 as that fixes a rather stupid bug for gbm cursor uploads.~~ merged.

This PR fixes #151 and allows anvil to make proper use of multiple monitors (instead of doing our weird mirroring).

This is best reviewed on a commit-by-commit basis.

Approach:

- [x] Adjust `GliumDrawer::draw_windows` to draw a subset of windows on the window map.
- [x] Create a "Output"-map. This is currently just a simple vector of outputs. The order of the outputs in the vector determines the layout. Every new output is to the right of the previously most right, aligned on the top border.
- [x] Clamp absolute input to the size of the current monitor and relative input to the total x of all monitors and y of the current.
- [x] Additionally allow monitor swapping via shortcuts (Mod + 1-9)

Issues left:

- [x] ~~Fix the cursor state. Currently the cursor state swaps between hardware and software cursor mode all the time, resulting in a flickering cursor on empty space. On a window the hardware cursor mode is activated all the time. I cannot spot the underlying problem, although it likely has to be in `anvil/src/udev.rs` line 673-735.~~ This was only an issue on multi-monitor setups as DrmHandler is reused for different renders. Cursor logic is now moved to GliumDrawer instead:  31e9205.
- [x] ~~Currently a copied shm-buffer is removed, when any texture is initialized. This results in windows becoming visible on a new output only after forcing a new buffer (e.g. through resizing). It *should* be enough to remove `anvil/src/glium_drawer.rs` line 333-343, as we are still releasing the buffer in `anvil/src/shell.rs` line 706. But if I do that, weston-terminal crashes shortly after startup, likely because it is running out of buffers. I have no good idea, how to best debug this.~~ Fixed in bcd96d6

Possible improvements (worth discussing):

- [ ] The output map could be made into it's own type, as some functionality is re-implemented in slightly different ways across anvil. The new functions in `AnvilState` (input_handler.rs) to clamp input could also be moved into this new type. This type could also later evolve into supporting more layout configuration and possibly become a new helper in smithay.
- [x] ~~Add a way in smithays `CursorBackend` to clear the current cursor. Currently this is done rather hacky through assigning an empty cursor buffer.~~ Through cursor-visibility this is now also possible on the winit-backend, fixing double cursors as a direct result.


vberger:
Regarding the `Textures` question, the fact is that now the buffer are released:

- for SHM buffers, as soon as the `Texture2d` was created
- for EGL buffers, when the associated `EGLImages` is dropped.

Now that we are creating more than one `EGLImages` per buffer, releasing the buffer on drop of the `EGLImages` is problematic, as we may release it several times. In practice this is no a problem, as `wayland-server` will ignore the subsequent calls to `release()` on a dead object, but still.

My suggestion here would be the following:

- no longer release the buffer in `EGLImages` drop implementation
- Create some `BufferTextures` abstraction which takes ownership of the `WlBuffer`, and internally handles the the loading of the buffer contents (be it a SHM or an EGL buffer) into appropriate `Texture2d` for each context (it'll thus hold several `Texture2d` and `EGLImages` if necessary), and releases the `WlBuffer` on drop.

This would allows to have all buffers released at the same time : when the `BufferTextures` is dropped.

The buffer release in `anvil/src/shell.rs` line 706 only applies to buffers which are overwritten by the client *before* our drawer loads them, hence just removing the release in `anvil/src/glium_drawer.rs` line 333-343 makes us no longer release the SHM buffer.

Drakulix:
@vberger I consider this now ready for merge.

PR #222, opened 2020-06-13
wayland.seat: Send modifiers event right after enter event

kchibisov:
wl_keyboard.modifiers must be sent after wl_keyboard.enter event,
otherwise it's a protocol violation.


kchibisov:
relevant spec part https://gitlab.freedesktop.org/wayland/wayland/-/blob/master/protocol/wayland.xml#L2206

PR #225, opened 2020-07-10
Lighten some dependencies

vberger:
- make `slog-stdlog` optional, but activated by default. This dependency has a significant dependency tree, but is not used at all if the downstream crate uses slog as well, rather than `log`.
- disable the default features of `image`, which we don't use at all
- make `tempfile` an optional dependency, it is only required by the wayland frontend.


vberger:
Indeed it is, though at the moment Smithay is doing breaking changes every two weeks more or less so... :shrug: 

Though this feature should be better documented in the crate root at least. I'll add that.

PR #227, opened 2020-07-12
backend.input: keep sub-pixel precision in events

### LINKED ISSUE #224: Anvil: Janky mouse movement
When runnig Anvil with `cargo run --features logind -- --tty-udev`, the mouse cursor acts as if my mouse has horribly inconsistent tracking. I've already tried running Weston on the same system, and it didn't experience the same issue (although Weston had a hard time displaying anything on the drm backend, so kudos to Anvil on that one). Sway has none of the above issues.

vberger:
libinput provides sub-pixel precision for pointer motion & touch events.
Keep this precision by switching all coordinates values from input
events to f64 (rather than i32 or u32). Otherwise, values are rounded
and part of the movment is lost.

Potentially fixes #224


LiHRaM:
Can confirm that this fixes janky mouse behavior for me, especially in precision movements.

maroider:
This seems to fix the issue for me.

Issue #161, opened 2020-02-08
Smithay assumes monotonically increasing serials


YaLTeR:
...however that's not specified anywhere in the docs, and it's not even the case for the default serial counter which is said to wrap around on overflow. Example where the assumption takes place:
https://github.com/Smithay/smithay/blob/7fa7fe03be52e35bf7a7b1dbaa947bfe700dfdc6/src/wayland/shell/xdg/xdg_handlers.rs#L320-L325

vberger:
Indeed, serials should "only" be locally monotonically increasing.

Probably we should add a `Serial` abstraction that takes the wrapping around into account correctly.

mcmfb:
> the default serial counter which is said to wrap around on overflow

I'd also add that next_serial() returns a `u32` even when `usize` has 64 bits, so the wrap-around (generated serial is smaller than the previous one) may happen even before that.

Is it reasonable to assume that at most `u32` serials have been generated between the ones being compared? If so, it could be solved storing a `bool` that toggles when the wrap-around happens. Then, if two serials are compared and that flag is different, we can just invert the comparison (e.g. `s1 <= s2` instead of `s1 > s2`.) Because at most u32 serials were generated, the inverted condition should hold.

vberger:
I had in mind making an implementation of a serial counting taking wrap-around into account based on the fact that serials are overall quite short-lived. So even though a long-lived compositor may wrap-around the `u32` in the serials it generates, the actual set of "live" serials should be much closer to each other than that.

So my idea was that if the distance between the two serials is greater than `u32::MAX / 2`, then assume that a wrap-around occurred between the two.

So, making a `Serial(u32)` type, with a custom `PartialOrd` impl taking that into account.

mcmfb:
Fair enough. I've made a pull request to do that: https://github.com/Smithay/smithay/pull/230

It's my first time using git to contribute to someone else's project, so please let me know if I screwed up.

vberger:
Fixed in #230

PR #230, opened 2020-08-08
Solve issue #161

mcmfb:
Implement a Serial type whose comparison takes the underlying counter's wrap-around into account, and use this type instead of u32 for all serials.


mcmfb:
> I think It would be worth adding `From` impls for converting between `Serial` and `u32` (so `impl From<u32> for Serial` and `impl From<Serial> for u32`), and then use `.into()` for conversions between the two, rather than making the inner field public. Seems cleaner to me that way.

Indeed it does. Will do!

> Also, it looks like the CI is unhappy about formatting, please run `cargo fmt` on your commits to make it happy. :)

Oops, my bad. I'll fix that too.

vberger:
Thanks, this looks quite good!

The clippy failure is unrelated to your PR, so no need to worry about it.

Thinking a little more about it, if you could add a few unit tests for the `PartialOrd` impl, just to future-proof the whole thing in case we come back to modify it at some point, that'd be perfect.

mcmfb:
> The clippy failure is unrelated to your PR, so no need to worry about it.

Oh, ok, I was wondering about that too.
 
> Thinking a little more about it, if you could add a few unit tests for the `PartialOrd` impl, just to future-proof the whole thing in case we come back to modify it at some point, that'd be perfect.

Done. If there's any other test cases I should include, just let me know.
Since tests in Rust are run in parallel, I've assumed I can't use the global SERIAL_COUNTER, so I've made a new SerialCounter for each test.

vberger:
Sorry for the late reply, this is good from my point of view, thanks!

PR #237, opened 2020-10-31
Fix nix build issues in top level project

anholt:
I was curious to check out the project, and was surprised to find that it didn't build.  Later I found I was supposed to build and run from anvil/, but it seems like we should keep the top level green, too.


vberger:
Ah, thanks! It looks like our nix dependency was automatically bumped via our git dependencies. We'll need to fix that.

PR #244, opened 2020-12-06
XWayland: Use _exit() in the fork() child

psychon:
The XWayland code does a double fork. The grandchild then exec's
Xwayland. The intermediate child watches the grandchild and reports its
status to the main process. I've been trying to get this to work, but
things kept crashing. Valgrind says:
```
 Process terminating with default action of signal 11 (SIGSEGV)
  Access not within mapped region at address 0x101609D0
    at 0x4A4B297: __pthread_clockjoin_ex (pthread_join_common.c:89)
    by 0x738B106: ??? (in /usr/lib/x86_64-linux-gnu/dri/iris_dri.so)
    by 0x738B1BB: ??? (in /usr/lib/x86_64-linux-gnu/dri/iris_dri.so)
    by 0x4ABB436: __run_exit_handlers (exit.c:108)
    by 0x4ABB5D9: exit (exit.c:139)
    by 0x567D3D: smithay::xwayland::xserver::launch (xserver.rs:215)
    by 0x5665BA: smithay::xwayland::xserver::XWayland<WM>::init (xserver.rs:113)
    by 0x2597C1: anvil::winit::run_winit (winit.rs:125)
    by 0x338DBC: anvil::main (main.rs:53)
    by 0x4EFACA: core::ops::function::FnOnce::call_once (function.rs:227)
    by 0x549B1D: std::sys_common::backtrace::__rust_begin_short_backtrace (backtrace.rs:137)
    by 0x6A11B0: std::rt::lang_start::{{closure}} (rt.rs:66)
```
I certainly do not understand this issue, but since this intermediate child
does not do much, I do not think it should run atexit handlers. Thus, this
commit just uses _exit() instead of exit() to skip these handlers.

Long-term, it would be better to fork() before starting any threads. I think
calling fork() in a threaded process is even undefined. Short-term, this fix
should be good enough.

Signed-off-by: Uli Schlachter <psychon@znc.in>

### DIFF
--- a/src/xwayland/xserver.rs
+++ b/src/xwayland/xserver.rs
@@ -206,9 +206,13 @@ fn launch<WM: XWindowManager + 'static, T: Any>(
                     if let Ok(signal::Signal::SIGCHLD) = sig {
                         // XWayland has exited before being ready
                         let _ = ::nix::sys::wait::waitpid(child, None);
-                        unsafe { ::nix::libc::exit(1) };
+                        // See below for exit vs _exit
+                        unsafe { ::nix::libc::_exit(1) };
                     }
-                    unsafe { ::nix::libc::exit(0) };
+                    // Use _exit so that atexit handlers are not run. We certainly did not set any,
+                    // but we might have inherited some from the main process and they have been
+                    // seen causing segfaults.
+                    unsafe { ::nix::libc::_exit(0) };
                 }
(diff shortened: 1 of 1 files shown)


psychon:
Superseded by #250 which actually fixes the underlying problem instead of papering over the symptom.

Issue #246, opened 2020-12-06
The XWayland integration kills the process via SIGUSR1


psychon:
I've been trying to get the existing Xwayland code to work, but anvil just kept dying. valgrind pointed out:
```
==116398== Process terminating with default action of signal 10 (SIGUSR1)
==116398==    at 0x4A507B2: futex_wait_cancelable (futex-internal.h:183)
==116398==    by 0x4A507B2: __pthread_cond_wait_common (pthread_cond_wait.c:508)
==116398==    by 0x4A507B2: pthread_cond_wait@@GLIBC_2.3.2 (pthread_cond_wait.c:638)
==116398==    by 0xFE92A4: std::thread::park (in /home/psychon/projects/smithay/target/debug/anvil)
==116398==    by 0x33D5FB: crossbeam_channel::context::Context::wait_until (context.rs:175)
==116398==    by 0x5ADEFA: crossbeam_channel::flavors::array::Channel<T>::recv::{{closure}} (array.rs:442)
==116398==    by 0x42882F: crossbeam_channel::context::Context::with::{{closure}} (context.rs:50)
==116398==    by 0x428613: crossbeam_channel::context::Context::with::{{closure}} (context.rs:58)
==116398==    by 0x2353FC: std::thread::local::LocalKey<T>::try_with (local.rs:265)
==116398==    by 0x427FDA: crossbeam_channel::context::Context::with (context.rs:53)
==116398==    by 0x5ADDCC: crossbeam_channel::flavors::array::Channel<T>::recv (array.rs:431)
==116398==    by 0x6A3B3B: crossbeam_channel::channel::Receiver<T>::recv (channel.rs:719)
==116398==    by 0x35BFDD: slog_async::AsyncCoreBuilder<D>::spawn_thread::{{closure}} (lib.rs:281)
==116398==    by 0x549AD2: std::sys_common::backtrace::__rust_begin_short_backtrace (backtrace.rs:137)
```
What seems to happen:
- At the beginning of `main()`, slog is initialised and starts a thread.
  - This thread has no signals blocked.
- I call `smithay::xwayland::XWayland::init()` to set up `Xwayland`.
  - This forks a child process that watches XWayland startup. Once XWayland started, the main process is informed via the equivalent of `kill(getppid(), SIGUSR1)`.
- Since this sends a signal to the process, all threads can receive it
- The main thread has `SIGUSR1` blocked by `smithay::xwayland::XWayland::init()`, thus the slog thread gets to handle the signal.
[...]

PR #248, opened 2020-12-25
xwayland: Use pipe instead of signal

psychon:
Handling SIGUSR1 requires setting a signal handler, which is a global
resource. This can also cause problems in the presence of threads, since
every thread needs to set up a signal handler. This commit instead
changes the code to use a pipe instead (actually a UnixStream). When the
child closes the pipe, the parent wakes up and knows "to do something".

Fixes: https://github.com/Smithay/smithay/issues/246
Signed-off-by: Uli Schlachter <psychon@znc.in>

### DIFF
--- a/src/xwayland/xserver.rs
+++ b/src/xwayland/xserver.rs
@@ -37,10 +37,7 @@ use std::{
     sync::Arc,
 };
 
-use calloop::{
-    signals::{Signal, Signals},
-    LoopHandle, Source,
-};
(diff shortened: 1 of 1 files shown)


psychon:
I didn't cause the remaining clippy warnings and don't think I should fix them in this PR. Should I?

vberger:
Thanks for this PR! It looks quite good. :+1:

Indeed, these clippy fixes are unrelated, you can ignore them.

vberger (approved):

PR #250, opened 2020-12-29
XWayland: Use a fork helper process to launch Xwayland

psychon:
Calling `fork` in a multi-threaded program is dangerous. According to `man fork`, you are only allowed to call async-signal-safe functions afterwards. `exit` is not async-signal-safe. Hence in #244 I reported seeing crashes from some exit handlers when the forked child process tries to exit.

This PR adds a new `ForkHelper` to the public API. It can be created via `ForkHelper::fork()`, which is `unsafe`, because it calls `fork`. The `ForkHelper` can then later be used to actually launch XWayland.

(This PR also fixes the problem from #244 by simply not calling `exit`, because the fork helper process keeps running after starting `Xwayland`... However, I now get a double panic when closing the winit window because of a failed unwrap. At least no running processes seem to remain.)

This PR was tested via the following diff. The log following messages showed up, so apparently `Xwayland` managed to start:
```
Dec 29 13:50:14.960 INFO XWayland is ready, signaling the WM., smithay_module: XWayland
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Unsupported maximum keycode 569, clipping.
>                   X11 cannot support keycodes above 255.
Errors from xkbcomp are not fatal to the X server
```
```diff
diff --git a/anvil/Cargo.toml b/anvil/Cargo.toml
index 4dbfd7c1..9d56d058 100644
--- a/anvil/Cargo.toml
+++ b/anvil/Cargo.toml
@@ -19,7 +19,7 @@ xkbcommon = "0.4.0"
 [dependencies.smithay]
 path = ".."
 default-features = false
-features = [ "renderer_glium", "backend_egl", "wayland_frontend" ]
+features = [ "renderer_glium", "backend_egl", "wayland_frontend", "xwayland" ]
 
 [build-dependencies]
 gl_generator = "0.14"
diff --git a/anvil/src/main.rs b/anvil/src/main.rs
index f6d28ccd..6e97f8de 100644
--- a/anvil/src/main.rs
+++ b/anvil/src/main.rs
@@ -11,6 +11,7 @@ use std::{cell::RefCell, rc::Rc};
 
 use slog::Drain;
 use smithay::reexports::{calloop::EventLoop, wayland_server::Display};
+use smithay::xwayland::LaunchHelper;
 
 #[macro_use]
 mod shaders;
@@ -36,6 +37,8 @@ static POSSIBLE_BACKENDS: &[&str] = &[
 ];
 
 fn main() {
+    let helper = unsafe { LaunchHelper::fork() }.unwrap();
+
     // A logger facility, here we use the terminal here
     let log = slog::Logger::root(
         slog_async::Async::default(slog_term::term_full().fuse()).fuse(),
@@ -50,7 +53,7 @@ fn main() {
         #[cfg(feature = "winit")]
         Some("--winit") => {
             info!(log, "Starting anvil with winit backend");
-            if let Err(()) = winit::run_winit(display, &mut event_loop, log.clone()) {
+            if let Err(()) = winit::run_winit(display, &mut event_loop, log.clone(), helper) {
                 crit!(log, "Failed to initialize winit backend.");
             }
         }
diff --git a/anvil/src/winit.rs b/anvil/src/winit.rs
index eafbfcf2..bed38c7c 100644
--- a/anvil/src/winit.rs
+++ b/anvil/src/winit.rs
@@ -1,4 +1,4 @@
-use std::{cell::RefCell, rc::Rc, sync::atomic::Ordering, time::Duration};
+use std::{cell::RefCell, rc::Rc, sync::atomic::Ordering, time::Duration, os::unix::net::UnixStream};
 
 #[cfg(feature = "egl")]
 use smithay::backend::egl::EGLGraphicsBackend;
@@ -6,13 +6,14 @@ use smithay::{
     backend::{graphics::gl::GLGraphicsBackend, input::InputBackend, winit},
     reexports::{
         calloop::EventLoop,
-        wayland_server::{protocol::wl_output, Display},
+        wayland_server::{protocol::wl_output, Client, Display},
         winit::window::CursorIcon,
     },
     wayland::{
         output::{Mode, Output, PhysicalProperties},
         seat::CursorImageStatus,
     },
+    xwayland::{LaunchHelper, XWayland, XWindowManager}
 };
 
 use slog::Logger;
@@ -21,10 +22,24 @@ use crate::buffer_utils::BufferUtils;
 use crate::glium_drawer::GliumDrawer;
 use crate::state::AnvilState;
 
+#[derive(Default)]
+struct XWm;
+
+impl XWindowManager for XWm {
+    fn xwayland_ready(&mut self, _connection: UnixStream, _client: Client) {
+        println!("Hey Uli: XWayland is ready, DISPLAY={}", std::env::var("DISPLAY").unwrap_or("NONE".to_string()));
+    }
+
+    fn xwayland_exited(&mut self) {
+        println!("Hey Uli: XWayland exited");
+    }
+}
+
 pub fn run_winit(
     display: Rc<RefCell<Display>>,
     event_loop: &mut EventLoop<AnvilState>,
     log: Logger,
+    helper: LaunchHelper,
 ) -> Result<(), ()> {
     let (renderer, mut input) = winit::init(log.clone()).map_err(|err| {
         slog::crit!(log, "Failed to initialize Winit backend: {}", err);
@@ -89,6 +104,11 @@ pub fn run_winit(
         refresh: 60_000,
     });
 
+    info!(log, "Hey Uli: Starting XWayland");
+    let wm = XWm::default();
+    let mut data = ();
+    let _xwayland = XWayland::init(wm, event_loop.handle(), display.clone(), &mut data, log.clone(), helper);
+
     let start_time = std::time::Instant::now();
 
     info!(log, "Initialization completed, starting the main loop.");
diff --git a/src/backend/winit.rs b/src/backend/winit.rs
index 8bd2fce8..76f606d7 100644
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -252,6 +252,7 @@ impl CursorBackend for WinitGraphicsBackend {
     type Error = ();
 
     fn set_cursor_position(&self, x: u32, y: u32) -> ::std::result::Result<(), ()> {
+        return Ok(());
         debug!(self.logger, "Setting cursor position to {:?}", (x, y));
         self.window
             .window()
```

### DIFF
--- a/src/xwayland/launch_helper.rs
+++ b/src/xwayland/launch_helper.rs
@@ -38,6 +38,29 @@ impl LaunchHelper {
         }
     }
 
+    /// Check the status of a previous launch.
+    pub(crate) fn was_launch_succesful(&self, pid: Pid) -> IOResult<bool> {
+        // find out if the launch was a success by waiting on the intermediate child
+        use nix::{errno::Errno, sys::wait};
+        loop {
+            match wait::waitpid(pid, None) {
+                Ok(wait::WaitStatus::Exited(_, 0)) => {
+                    // XWayland was correctly started :)
+                    return Ok(true);
+                }
+                Ok(_) => {
+                    // Something weird happened, most likely WaitStatus::Signaled
+                    return Ok(false);
+                }
+                Err(NixError::Sys(Errno::EINTR)) => {
+                    // interrupted, retry
+                    continue;
+                }
+                Err(e) => return Err(nix_error_to_io(e)),
+            }
+        }
+    }
+
     /// Fork a child and call exec_xwayland() in it.
     pub(crate) fn launch(
         &self,
--- a/src/xwayland/launch_helper.rs
+++ b/src/xwayland/launch_helper.rs
@@ -5,9 +5,12 @@ use std::{
 
 use nix::{
     libc::exit,
-    unistd::{fork, ForkResult},
+    sys::signal,
+    unistd::{fork, ForkResult, Pid},
     Error as NixError, Result as NixResult
 };
+
+use super::xserver::exec_xwayland;
 
 /// A handle to a `fork()`'d child process that is used for starting XWayland
 pub struct LaunchHelper(UnixStream);
@@ -32,6 +35,68 @@ impl LaunchHelper {
                 }
             }
             ForkResult::Parent { child: _ } => Ok(Self(me))
+        }
+    }
+
+    /// Fork a child and call exec_xwayland() in it.
+    pub(crate) fn launch(
+        &self,
+        display: u32,
+        wayland_socket: UnixStream,
+        wm_socket: UnixStream,
+        listen_sockets: &[UnixStream],
+        log: &::slog::Logger,
+    ) -> Result<(Pid, UnixStream), ()> {
+        let (status_child, status_me) = UnixStream::pair().map_err(|_| ())?;
+        match unsafe { fork() } {
+            Ok(ForkResult::Parent { child }) => {
+                // we are the main smithay process
+                Ok((child, status_me))
+            }
+            Ok(ForkResult::Child) => {
+                // we are the first child
+                let mut set = signal::SigSet::empty();
+                set.add(signal::Signal::SIGUSR1);
+                set.add(signal::Signal::SIGCHLD);
+                // we can't handle errors here anyway
+                let _ = signal::sigprocmask(signal::SigmaskHow::SIG_BLOCK, Some(&set), None);
+                match unsafe { fork() } {
+                    Ok(ForkResult::Parent { child }) => {
+                        // When we exit(), we will close() this which wakes up the main process.
+                        let _status_child = status_child;
+                        // we are still the first child
+                        let sig = set.wait();
+                        // Parent will wait for us and know from out
+                        // exit status if XWayland launch was a success or not =)
+                        if let Ok(signal::Signal::SIGCHLD) = sig {
+                            // XWayland has exited before being ready
+                            let _ = ::nix::sys::wait::waitpid(child, None);
+                            unsafe { ::nix::libc::exit(1) };
+                        }
+                        unsafe { ::nix::libc::exit(0) };
+                    }
+                    Ok(ForkResult::Child) => {
+                        // we are the second child, we exec xwayland
+                        match exec_xwayland(display, wayland_socket, wm_socket, listen_sockets) {
+                            Ok(x) => match x {},
+                            Err(e) => {
+                                // well, what can we do ?
+                                error!(log, "exec XWayland failed"; "err" => format!("{:?}", e));
+                                unsafe { ::nix::libc::exit(1) };
+                            }
+                        }
+                    }
+                    Err(e) => {
+                        // well, what can we do ?
+                        error!(log, "XWayland second fork failed"; "err" => format!("{:?}", e));
+                        unsafe { ::nix::libc::exit(1) };
+                    }
+                }
+            }
+            Err(e) => {
+                error!(log, "XWayland first fork failed"; "err" => format!("{:?}", e));
+                Err(())
+            }
         }
     }
 }
--- a/src/xwayland/launch_helper.rs
+++ b/src/xwayland/launch_helper.rs
@@ -24,6 +24,8 @@ pub struct LaunchHelper(UnixStream);
 impl LaunchHelper {
     /// Start a new launch helper process.
     ///
+    /// # Safety
+    ///
     /// This function calls [`nix::unistd::fork`] to create a new process. This function is unsafe
     /// because `fork` is unsafe. Most importantly: Calling `fork` in a multi-threaded process has
     /// lots of limitations. Thus, you must call this function before any threads are created.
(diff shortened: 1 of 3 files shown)


nbdd0121:
I personally think it is better to use the approach in #244, because it is less restrictive to use and make the API cleaner. I had a look at the approach in #244 and believe that it could be made safe with some modifications.

According to `man signal-safety`, `fork`, `exec`, fd-related ops and signal-related ops are all signal-safe. This means that we could have the job done, but we just need to be extra careful. One thing is that we cannot do any memory allocation, so all CString-related stuff needs to be completed beforehand. We also shouldn't use `println!` or `log!` families, so we will need to use return codes to pass error to parent process. But apart from that, other stuffs under `nix` should be safe.

I believe we could rewrite the code so that it prepares everything beforehand, then do a double fork with minimal code executed in the child process and the grandchild process before `exec/_exit`.

psychon:
From a quick look at the code: The only "not allowed" stuff that `exec_xwayland` currently does is allocation (`Vec` and `CString`) and messing with the environment. Since the environment is cleaned up, this can be done by switching to `execvpe`.

The old code that this PR removed also called `nix::signal::SigSet::wait`. This calls `libc::sigwait()`, which is not async-signal-safe (surprise!). I don't see a good way around that, but I did not really check. This all feels too `unsafe` to me.

The mentioned FD leak is still real and unavoidable, I guess.

nbdd0121:
> what restrictions would it remove / how would it make the API cleaner?

If a user of smithay is producing a library, the library would need to expose LaunchHelper further up. Only the outmost binary can be certain that no thread has been spawned prior to create the helper. Furthermore, user shouldn't care about the caveats around SIGUSR1 and fork, because that's should just be an implementation detail...

> avoiding interference with the other threads (like the one spawned by the logger)

This is not really an issue. Fork in a MT program will only fork the current thread, and if only async-signal-safe functions are used, it is safe (aka just apply the same caution needed to write a signal handler)

> This calls libc::sigwait(), which is not async-signal-safe (surprise!). I don't see a good way around that, but I did not really check. This all feels too unsafe to me.

Even though sigwait is not async-signal-safe, sigsuspend is (surprise!).

> solves the question of not leaking unrelated open file descriptors (like the DRM node or the input devices) to XWayland

This is indeed an issue that I haven't considered. A workaround is to close fds by looking at `/proc/self/fd`, which can be done safety using getdents (safe as it's syscall, but it is not POSIX) or readdir_r (strictly speaking not listed as signal-safe but is, but this is getting deprecated). But this seems to hacky.

The best way IMO is probably just to use a helper binary. This binary would be very small, so we can just `include_bytes!` it, and put it in a temporary file and run it.

Drakulix:
> > what restrictions would it remove / how would it make the API cleaner?
> 
> If a user of smithay is producing a library, the library would need to expose LaunchHelper further up. Only the outmost binary can be certain that no thread has been spawned prior to create the helper. Furthermore, user shouldn't care about the caveats around SIGUSR1 and fork, because that's should just be an implementation detail...

Well but writing a compositor is always a difficult task. Even if you write a library to abstracts a lot of it, you cannot really hide the nature of the underlying complex code.

Smithay is currently preferring to make its apis and whats happening in the background very explicit and I personally would rather keep it that way. `LaunchHelper` gives the user full control about when forks are happening.

> > avoiding interference with the other threads (like the one spawned by the logger)
> 
> This is not really an issue. Fork in a MT program will only fork the current thread, and if only async-signal-safe functions are used, it is safe (aka just apply the same caution needed to write a signal handler)

Assuming we would change this and use the approach seen in #244 and that you are writing said library. What prevents the end-user from using any non async-signal-safe function? How do you communicate, that the user is not supposed to use certain functions? Is it not easier to communicate, that there is a certain function (which internally calls `LaunchHelper`), that absolutely needs to be called before anything else?
(Or does this only apply to the fork-code?)

Also are you sure all of this is defined behavior? Smithay also aims to support BSD systems for example. Does all of this also apply there and is documented as such?

psychon:
Answering both of you:

> The best way IMO is probably just to use a helper binary. This binary would be very small, so we can just include_bytes! it, and put it in a temporary file and run it.

How would that be made to work with `cargo`? Have a `build.rs` that spawns another `cargo` instance to build the binary? That sounds like it will not really work well.

Also: Where would you put the temporary file? It needs a place that is writable and executable. I bet somewhere out there mounts `/tmp` as `noexec`.

> A workaround is to close fds by looking at /proc/self/fd, which can be done safety using getdents (safe as it's syscall, but it is not POSIX) or readdir_r (strictly speaking not listed as signal-safe but is, but this is getting deprecated). But this seems to hacky.

That's Linux specific. See https://gitlab.gnome.org/GNOME/glib/-/issues/1638 for "some of the horrors". glib (the C library) is slow on BSD with its `g_spawn_*` API, because it tries to close all FDs and the only way to do that is a loop that tries to `close()` each number. And some OS version raised the maximum FD number, making the code in question awfully slow (I think the time was measured in seconds).

> Even though sigwait is not async-signal-safe, sigsuspend is (surprise!).

From a quick look:
- No nice wrapper in `nix`, but that's not a big issue.
- That function doesn't tell us which signal arrived. So this needs to actually set up a signal handler that then sets some global(?) variable with the signal number that came in(?).

My "surprise" was not meant as being an attack, but more like "it's not a big surprise that some necessary functions are missing".

> (Or does this only apply to the fork-code?)

This only happens to the fork-code. From `man fork`, emphasis mine:

> *  After  a fork() in a multithreaded program, *the child* can safely call only async-signal-safe functions (see signal-safety(7)) until such time as it calls execve(2).

nbdd0121:
I think I find a even better way: just use sh to handle the signal.
```sh
sh -c 'trap ":" USR1; (trap "" USR1; exec Xwayland <args>) & wait; exit $?'
```
If wait gives non-zero value, it indicates USR1 is triggered. This is the same logic as the double fork, but by using sh it does not have many limitations like `fork`, while is still portable.

PR #255, opened 2021-01-19
Use sh to handle SIGUSR1 and replace fork+exec with spawn

nbdd0121:
Idea from #250, but instead of using exit codes, this uses a pipe to communicate between smithay and sh. Comments yet to be fixed.

cc @psychon @Drakulix @vberger

Issue #229, opened 2020-08-06
Unable to build project from source unless bumping up nix to version `0.18.0`


mjarkk:
When i try to build the project from source i get this error:
```
 smithay master cargo build
   Compiling smithay v0.2.0 (/home/mark/Documents/smithay)
warning: allow(rust_2018_idioms) is ignored unless specified at crate level
 --> src/backend/graphics/gl.rs:7:22
  |
7 | #[allow(clippy::all, rust_2018_idioms, missing_docs)]
  |                      ^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_attributes)]` on by default

warning: allow(rust_2018_idioms) is ignored unless specified at crate level
  --> src/backend/egl/ffi.rs:45:22
   |
45 | #[allow(clippy::all, rust_2018_idioms)]
   |                      ^^^^^^^^^^^^^^^^

warning: allow(rust_2018_idioms) is ignored unless specified at crate level
 --> src/backend/graphics/gl.rs:7:22
  |
7 | #[allow(clippy::all, rust_2018_idioms, missing_docs)]
  |                      ^^^^^^^^^^^^^^^^

error[E0308]: mismatched types
  --> src/backend/drm/common/mod.rs:84:28
   |
81 |             } if match source.get_ref() {
   |                        ---------------- this expression has type `&drm::SystemError`
...
84 |                     errno: nix::errno::Errno::EBUSY,
   |                            ^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `nix::errno::consts::Errno`, found enum `nix::errno::Errno`
   |
   = note: perhaps two different versions of crate `nix` are being used?

error[E0308]: mismatched types
  --> src/backend/drm/common/mod.rs:87:28
   |
81 |             } if match source.get_ref() {
   |                        ---------------- this expression has type `&drm::SystemError`
...
87 |                     errno: nix::errno::Errno::EINTR,
   |                            ^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `nix::errno::consts::Errno`, found enum `nix::errno::Errno`
   |
   = note: perhaps two different versions of crate `nix` are being used?

error[E0308]: mismatched types
   --> src/backend/drm/eglstream/mod.rs:356:32
    |
354 |                 if match x.get_ref() {
    |                          ----------- this expression has type `&drm::SystemError`
355 |                     drm::SystemError::Unknown {
356 |                         errno: nix::errno::Errno::EBUSY,
[...]

PR #259, opened 2021-02-09
Implement Debug trait for some public types

PolyMeilex:
List of reviewed modules/files:
- [ ] smithay
    - [ ] backend
        - [ ] drm
            - [x] atomic
            - [x] common
            - [x] egl
            - [x] eglstream
            - [ ] gbm
            - [ ] legacy
            - [x] mod.rs
        - [x] egl
        - [x] graphics
        - [x] libinput
        - [x] session
        - [x] input.rs
        - [x] mod.rs
        - [x] udev.rs
        - [x] winit.rs
    - [x] utils
    - [x] wayland
    - [x] xwayland
    - [x] lib.rs
    - [x] reexports.rs
    - [x] signaling.rs


psychon:
Would it make sense to enable warnings on the `missing-debug-implementations` lint? Otherwise, this issue will likely come up again and again.

PolyMeilex:
> Would it make sense to enable warnings on the `missing-debug-implementations` lint? Otherwise, this issue will likely come up again and again.

In some cases implementing Debug does not make too much sense, but still I would advocate for `missing-debug-implementations` to be enabled by default, so it would have to be explicitly disabled every time you think Debug is useless.
TLDR: I think implementing Debug should be opt-out instead of opt-in

vberger:
Ah yes, it should definitely be done in calloop as well you're right.

PolyMeilex:
With those 2 commits I think I'm done with this PR, as I'm not too keen on working on remaining `drm/gbm` and `drm/legacy` Debug traits.
According to `missing-debug-implementations` lint, after this PR most of the crate is covered (with only a few exceptions)

vberger:
Hi, thanks for this PR, and sorry for the late answer.

It looks pretty good overall, but in the name of future maintenance, I think it'd be nice to add a comment explaining why a manual debug impl was necessary when it was. Because some of them are not obvious to me, as they seem to be mostly doing exactly what `derive(Debug)` would have done...

Also, looks like this PR needs to be rebased, (#255 removed the `xwayland/launch_helper.rs` file).

PolyMeilex:
Sure, I'll review all of them once again, to make sure that there are no unnecessary manual impls left over, and I'll add comments as I go.

PolyMeilex:
Oh good, I believe that I did not cause those build errors, it looks like it cheapens on master branch as well. 
Looks like `nix` crate does not follow semver properly. (or maybe I'm wrong, no idea)

vberger:
Agh, again a `nix` update. This is #229 yet again. I'll update nix on the repo, so that you can rebase on top of that.

vberger:
Okay, I've done a pass of dependencies-updating, that should be good now.

PolyMeilex:
Done, I believe that I removed all unnecessary manual impls (sorry about that, It happened because I did not add all of them in one go, and some of the impls became obsolete after the whole PR was done, and I just did not notice), and the PR is successfully rebased onto master.

src/backend/libinput/mod.rs
```diff
@@ -423,6 +427,26 @@ impl From<event::pointer::ButtonState> for backend::MouseButtonState {
...
+                    .finish()
+            }
+        }
+
+        f.debug_tuple("LibinputSessionInterface")
+            .field(&SessionDebug(&self.0))
+            .finish()
+    }
+}
+
```
vberger, line 449: Same here

src/backend/session/direct.rs
```diff
@@ -160,6 +162,25 @@ pub struct DirectSessionNotifier {
...
+                "source",
+                &match self.source {
+                    Some(_) => "Some(..)",
+                    None => "None",
+                },
+            )
+            .finish()
+    }
+}
+
```
vberger, line 183: This one is missing a comment, I guess it's because `Signals` does not implement `Debug`?

src/backend/drm/common/fallback.rs
```diff
@@ -49,6 +50,22 @@ pub enum FallbackDevice<D1: Device + 'static, D2: Device + 'static> {
...
+                .field("device_id", &d.device_id())
+                .finish(),
+            FallbackDevice::Fallback(d) => f
+                .debug_struct("FallbackDevice::Preference")
+                .field("device_id", &d.device_id())
+                .finish(),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
PolyMeilex: Sure, I'll stick to derive than

src/backend/session/direct.rs
PolyMeilex: yep, I added a comment for that

vberger (approved):

vberger:
Thank you!

PR #261, opened 2021-04-07
Rework backend/rendering code... again (v4)

Drakulix:
# v4 - hopefully the last big overhaul

It is best to go through this commit by commit, so I am paraphrasing my commit messages here.
(Heading for commits, quote for the original message, free text for additional info and checkboxes for open questions and tasks.)

## Add allocator module 

> - Add a module that deals with different kinds of buffers (memory and external),
>   their allocation and usage for rendering. Also try to properly support modifiers this time.
> - Describe gbm functionality as an allocator (instead of a rendering device/surface).
> - Also create a quick-and-dirty dumb buffer allocator for tesing / simpler tasks.
> - Add a (current untested) wrapper for dmabufs and some code for converting from gbm for now.
> - (also untested) Swapchain helper to manage front/backbuffers for surfaceless rendering.

TODO/Questions:
- [x] Use the allocator and swapchain apis with the drm-backend and test their usability
- [ ] Implement more dmabuf conversions, currently we can only import gbm buffers and convert dmabufs into EGLImages for rendering
- [x] ~Implement the linux-dmabuf protocol to further verify our dmabuf wrapper is functional~ Different PR, but works

### (The biggest offender for too large commits in this PR..)

> - Remove the `Device`, `Surface`, `RawDevice` and `RawSurface` traits.
>   We are not trying to merge the different responsibilities between...
>   - Drm-Surfaces: crtcs and planes for displaying buffers
>   - Gbm-Surfaces: allocating and managing front- and back-buffers
>   - EGL-Surfaces: Load drawing apis.
>   ..anymore.
> - Instead unify the legacy and atomic backends into one `DrmDevice` and
>   a `DrmSurface`, that under the hood dispatches onto the different modules.
> - Do not keep a list of surfaces inside the devices. These are reference-counting
>   nightmares and a frequent issue due to the read-only nature of `Rc` and `Arc` and
>   thread-safety related problems.
> - Surfaces are mostly doing state-tracking independently of each other
>   and contain some more helper functions.
> - Simplify session code, because we do not need to pass the events
>   through three layers of devices and surfaces anymore.
> - Each surface now represents a plane. Rendering to a plane is just like
>   any other rendering step, including the Cursor-plane.
>   (Legacy interfaces only have one to avoid dealing with short-comings of the api).

TODO/Questions:

- [ ] Decide if we want to support any kind of surface-based rendering on our drm-backend. This approach sets a hard requirement on PRIME for gpu drivers. This leaves only some very old devices (like intel GMA 500, which only support dumbbuffers) behind. I personally would just drop it in favor of the new api.
- [ ] The new restriction on the primary surface for the legacy api removes explicit support for hardware cursors from the old backend. The drm `set_cursor` ioctls  - we were using - are now officially deprecated in favor of cursor planes, which *can* be used with the legacy api. But I would really not like to unify the plane support for atomic and legacy, because although you can use e.g. overlay planes on the legacy api, their experience is very bad, as it is un-synchronized. That does not really matter for cursors, but allowing planes to be used for every `DrmDevice` could mean that the same plane optimization works on one system (that picks the atomic backend) and performs horribly on the other (legacy) with the same smithay-calls. I would argue, that users of smithay that want hardware cursors on a legacy platform can work-around that by using drm-rs directly. `is_atomic` provides the necessary information to support this. But that is a purposefully cumbersome api, where an alternative approach could make things easier, but also allows easier "miss"-use... meh..

## Simplify egl platform code

> - Remove EGLStream code, nvidia support will re-introduced,
>   when 470 hopefully lands with GBM support this summer.
> - Greately simplify the native code by setting exclusively on
>   `EGL_EXT_platform_base` and its extensions for initialization.
> - Remove generic parameters for the underlying objects on `EGLDisplay`
>   and `EGLSurface` by using trait objects instead.
> - Allow creation of `EGLContext`s without a config for surface-less usage.

TODO/Questions:
- [x] ~Implement logging like for GL~ Later
- [x] ~Try to request high-priority contexts~ Later

## Introduce a new OpenGL renderer

> This pulls a lot of code from anvil/src/glium_drawer and replaces glium
> as glium has too many assumptions about the backend.
> (Mostly that a fixed framebuffer exists, see fix bullet point.)
> 
> Depending on how picky glium is, we could try to re-introduce glium support
> later with a some workarounds, but for now this is actually more
> straight-forward to support and test.
> 
> - Add a new GL renderer, that only depends on egl as an initialization platform.
>   In particular do not depend on a surface being available.
>   Renderers share some basic drawing functions and may bind objects to render upon.
>   E.g. surfaces or buffers, textures, etc for the gles2 renderer.
> - Be explicit about extensions we require and use. Use some very very common ones
>   to make our lives easier (e.g. BGRA and unpack to read in manditory shm formats).
> - Enable GL debug output
> - Allow more flexible rendering (e.g. 3D compositors) by allowing user-provided
>   projection matrices. Also provide helper functions to allow easy-ish handling
>   of surface and output transformations.
> - Add API for renderers to tell the wayland-frontend about supported buffer-types.
> - Also incoperate code from anvil/src/shm_load to handle buffer loading in the renderer
>   (as this is a renderer dependant operation).

TODO/Questions:
- [x] ~Explore if we can support glium by ignoring its "clean-context" requirement (*The OpenGL context must be newly-created. If you make modifications to the context before passing it to this function, glium's state cache may mismatch the actual one.*) as EGL-state is not GL-state and building a flexible backend, that for example returns different [framebuffer dimensions](https://docs.rs/glium/0.29.0/glium/backend/trait.Backend.html#tymethod.get_framebuffer_dimensions) based on the last `bind`-call. Using [exec_in_context](https://docs.rs/glium/0.29.0/glium/backend/struct.Context.html#method.exec_in_context) might even help us work-around short-comings for buffer/texture import and similar.~ TBD later
- [x] ~Try to support robustness~ Later
- [x] Add additional import/format functions for egl buffers
- [x] ~Add additional import/format functions for dmabuf buffers~ later
- [x] ~Provide our own exec_in_context function on the trait or similar helpers on the actual gles-structs to allow users, who want to depend on a particular rendering api to do more complex rendering on their own.~ later
- [x] ~Maybe provide more rendering helpers? (colored quad? actual text rendering?)~ later
- [x] ~Should the renderer keep track of imported buffers? I do not think so, maybe multiple renderers are in use and free-ing the buffer from one, should not free it in general. I fear we might need some kind of BufferWrapper, that handles reference-couting or similar. (`Dmabuf` already does something like that internally.) I would like to aid users here, shm buffers can be freed after importing (to all possible renderers that is) and need to be to prevent going out of space. Currently the user is responsible to do that, but that is something I would like to improve. Suggestions?~ It should not. Importing buffers and invalidation + tracking and freeing textures is hard and very dependent on the actual buffer api. Lets not complicate the API now and instead wire up a helper later, when we have a better understanding on the subject through usage in anvil.

## Remove the generic "graphics" module.

> - Cursor is not necessary anymore as rendering hardware vs software
>   cursors just depends on "where" you render the cursor.
> - Format is replaced by drm-fourcc and more related to buffers and
>   therefor the allocator module anyway.
> - GL/Glium are replaced by the renderer module.
> - Software was unused and very small anyway and can be implemented
>   with the new renderer api.
> - SwapBuffersError is (for now) just moved into the top-level backend module.

TODO/Questions:
- [x] ~Is this a good place for `SwapBuffersError`?~ Good enough imo.

## Migrate the winit backend to the new egl and renderer apis.

TODO/Questions:
- [x] ~Winit now implements `Renderer` itself and proxies that to the internal `Gles2Renderer`, while handling surface binding itself. This is likely fine for winit (and such tightly coupled apis in general), but is this a good general approach? This is the same kind of wrapping, that got us into problems with the `Surface`-API. Maybe explore different apis to handle the window and the renderer as separate objects from the user perspective?~ Good to keep in mind, but winit currently has a hard dependency on EGL and therefor GL. It does not make much sense to expose the `Renderer` for winit at least.

## Fixup anvil for winit rendering 

> - Since a lot of functionality is now in smithay's rendering module
>   we can get rid of shm_load, shaders and the glium dependency.
> - glium_drawer becomes drawing and only features some code to get from surfaces
>   to textures for any(!) renderer.
>   (Should probably moved into smithay at some point as well.)
> - buffer_utils is only necessary to query the buffer size anymore.
> - disable egl-buffer support temporarily

TODO/Questions:
- [x] ~Decide if we want to move more of the drawing-code into smithay or do that later? (Maybe together with the window-map?)~ Currently the drawing code is using Surface user-data and custom texture tracking, that cannot be easily moved out of anvil. This is good enough for now.
- [x] ~Likely provide a similar helper to buffer_utils in the allocator module, once the other buffer types are supported (again)~ later

## General TODOS:

- [x] Implement the `DrmRenderSurface`-API in the drm-backend. This will provide a similar experience to previous smithay releases, allowing us to port anvil quickly and test the new API. Anything extra from the old API is out-of-scope, compositors *should* instead look into implementing that code on their own to support things like direct-scanout more easily, but `DrmRenderSurface` might prove a nice tool to get started quickly.
- [x] Fixup anvils udev backend using `DrmRenderSurface`
- [ ] Test(!) the new approach.
  - [x] Works on `i915`
     - [x] with and without `SMITHAY_USE_LEGACY=1`
     - [x] EGL Buffers work / SHM Buffers work
     - [x] tty-switching works (untested)
  - [x] Works on `nouveau`
    - [x] ~with and without `SMITHAY_USE_LEGACY=1` (please upload logs for both runs in any case to check how format negotiation works out on this driver)~ noveau does not support atomic
    - [x] EGL Buffers work / SHM Buffers work (e.g. weston-simple-egl and weston-terminal)
    - [x] tty-switching works
  - [x] Test with `nvidia`
    - [x] Successfully starts with software rendering  
  - [x] Works on `amdgpu`
  - [x] Multi-GPU (should work as good/bad as before, SHM clients should be no problem, egl buffer might not display on other gpus or straight up crash anvil) - egl buffers crash the compositor, needs fixing later.
- [x] Copy and adjust documentation of the old modules where possible and write up more for the new ones.
- [x] Fixup/Reimplement egl-buffer support
- [x] ~Implement linux-dmabuf (maybe in a later PR?)~ Later.

## Closing thoughts

Feel free to discuss and share your thoughts.

I really think the new drm-api is simplifying the code a lot and I am super happy with it.
User might need to manage a few more objects (allocators, renderers, devices, surfaces), but can do so more freely as they are not entangled (as much) with each other anymore.

On top of that I am also pretty happy with the new allocator and rendering apis and I think they will serve us well for e.g. vulkan rendering. (That of course will need to be tested and I expect smaller changes to these api to happen later down the line, but that should be fine.) However I do not want to take much credit for that, because I took a lot of inspiration from wlroots api's here and also for a good chunk of the actual rendering code. (There are only so many ways to render a flat surface and I wanted a flexible and battle-tested one, as I am not (yet) much of an OpenGL-wizard.)


src/backend/renderer/gles2/mod.rs
```diff
@@ -0,0 +1,568 @@
...
+    BindBufferEGLError(#[source] crate::backend::egl::Error),
+    #[error("Unsupported pixel format: {0:?}")]
+    #[cfg(feature = "wayland_frontend")]
+    UnsupportedPixelFormat(wl_shm::Format),
+    #[error("Error accessing the buffer ({0:?})")]
+    #[cfg(feature = "wayland_frontend")]
+    BufferAccessError(crate::wayland::shm::BufferAccessError),
+}
+
+extern "system" fn gl_debug_log(_source: ffi::types::GLenum,
```
vberger, line 106: Why `extern "system"` rather than `extern "C"` ?

src/backend/renderer/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            (height, width)
+        } else {
+            (width, height)
+        }
+    }
+}
+
+#[cfg(feature = "wayland-frontend")]
+impl From<wayland_server::protocol::wl_output::Transform> for Transform {
```
vberger, line 96: I suspect we'll likely need the conversion the other way around as well.

Drakulix:
Alright. @vberger I have written a good bunch of docs, this is ready for a first proper review round and for testing. :)

I expect tty-switching to be broken in subtile ways on some configuration, because this is currently missing some kind of state resetting, if the session pauses. I plan to fix this later.


I am also not sure, if the compatibility layer, that is `DrmRenderSurface` is a helpful abstraction. It again binds a single renderer exclusively, which required some interesting hacks in anvil by using shared EGLContexts (e.g. the `pointer_image`).

I think the internal `Buffers` helper might be its primary use. Maybe I should instead provide a layer (something like `SurfaceSwapchain`?) that just handles allocating, swapping, but no rendering and allow users to `Bind` that to a renderer.
Similar things (decoupling the rendering part) could be done to the `WinitGraphicsBackend`, which would also make it easy to introduce vulkan support for that later. The render target abstraction is probably the better abstraction instead of introducing wrappers again (and passing through a lot of functions in the render surface implementation).

But maybe this is good enough for now. It is certainly nice for quickly migrating from the last stable smithay release (although I doubt we have many users and those we do have are likely aware of our frequent api breakage). I will play around with alternative abstractions later and if they avoid some of the `DrmRenderSurface` current problems, I will quickly mark this abstraction as deprecated. By smithay 0.4 or 0.5 (whenever this is) I expect to remove this again completely.

Drakulix:
I will remove the draft status, once new drm-rs and gbm-rs versions are released.
I want to move to stable dependencies again with this PR, as both are very close to release.
I have just prioritized getting this in a working state before.

joeftiger:
I have a system with an AMD card so I can test it. What exactly do I need to do for it? I never tried to write some code with smithay yet.

Drakulix:
> I have a system with an AMD card so I can test it. What exactly do I need to do for it? I never tried to write some code with smithay yet.

Thanks! There is not really much code to do, but just try to run anvil on a tty with this branch.
So it would be very helpful if you could:
- Start anvil: `env RUST_BACKTRACE=1 cargo run -p anvil -- --tty-udev 2>anvil.log`
- Test the following clients:
  - `weston-terminal` (pressing Cmd+Enter)
  - `weston-simple-shm`
  - `weston-simple-egl`
  - `weston-simple-dmabuf-egl`
  - `glxgears -info` (and verify, that this prints an AMD renderer and not uses llvmpipe)
- Close anvil (Cmd+Q) and upload the resulting log here.

joeftiger:
First log (crash) when switching between tty1 and 2.
[anvil.log](https://github.com/Smithay/smithay/files/6487679/anvil.log)

joeftiger:
Second log (clean) without switching.
[anvil2.log](https://github.com/Smithay/smithay/files/6487690/anvil2.log)

I noticed that when running `glxgears -info` it didn't print out all the stuff it does on weston.

joeftiger:
@Drakulix It crashed immediately.

[anvil.log](https://github.com/Smithay/smithay/files/6488152/anvil.log)

Drakulix:
> @Drakulix It crashed immediately.
> 
> [anvil.log](https://github.com/Smithay/smithay/files/6488152/anvil.log)

It seems you do not have XWayland installed, but anvil tries to set that up and crashes later. Can you try running without the xwayland feature? Like: `cd anvil; cargo run --no-default-features --features "udev logind egl" -- --tty-udev 2>anvil.log`

joeftiger:
> [...] It seems you do not have XWayland installed [...]`

I'll try once with once without

joeftiger:
### Without `xwayland`
Compiled and started. However, `Super + Enter` did nothing and after trying some things my pc hangs. 
Have to restart.

Drakulix:
> ### Without `xwayland`
> 
> Compiled and started. However, `Super + Enter` did nothing and after trying some things my pc hangs.
> Have to restart.
> 
> EDIT: [anvil.log](https://github.com/Smithay/smithay/files/6488166/anvil.log)

You do not have the required clients installed, most importantly not `weston-terminal`, which is the default terminal anvil starts on Cmd+Enter. So nothing happened on your system. Usually these test clients are installed alongside weston.

joeftiger:
> You do not have the required clients installed, most importantly not `weston-terminal`, which is the default terminal anvil starts on Cmd+Enter. So nothing happened on your system. Usually these test clients are installed alongside weston.

That can well be. On my pc I am still on xorg as I am missing conky on wayland :D And I wanted to replace my Manjaro install with Arch as I did on my laptop already. 
Installing `weston` will probably fix it, so I'll try again :)

joeftiger:
@Drakulix 

On laptop with `DPI_PRIME=1` to use the  gpu.
[anvil.log](https://github.com/Smithay/smithay/files/6488231/anvil.log)

joeftiger:
On pc with gpu (only)

[anvil.log](https://github.com/Smithay/smithay/files/6488236/anvil.log)

Drakulix:
> On pc with gpu (only)
> 
> [anvil.log](https://github.com/Smithay/smithay/files/6488236/anvil.log)

Thanks!
Both logs do look good, so the amdgpu driver seems to play nicely with our code! :)

src/backend/allocator/gbm.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl<T> std::convert::TryFrom<GbmBuffer<T>> for Dmabuf {
+    type Error = GbmConvertError;
+
+    fn try_from(buf: GbmBuffer<T>) -> Result<Self, GbmConvertError> {
+        let planes = buf.plane_count()? as i32;
+
+        //TODO switch to gbm_bo_get_plane_fd when it lands
+        let mut iter = (0i32..planes).map(|i| buf.handle_for_plane(i));
+        let first = iter.next().expect("Encountered a buffer with zero planes");
```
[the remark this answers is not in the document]
Drakulix: A buffer with zero planes would have no contents. I think there is no way to create a buffer like that in gbm, because the planes depend on the format and there is no "empty" format.

src/backend/allocator/gbm.rs
```diff
@@ -0,0 +1,83 @@
...
+                    return Some(Ok(first));
+                }
+            }
+            None
+        }).is_none() {
+            // GBM is lacking a function to get a FD for a given plane. Instead,
+            // check all planes have the same handle. We can't use
+            // drmPrimeHandleToFD because that messes up handle ref'counting in
+            // the user-space driver.
+            return Err(GbmConvertError::UnsupportedBuffer); //TODO
```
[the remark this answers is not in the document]
Drakulix: To remove that error once the above mentioned gbm function is widely available.

src/backend/drm/device/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+            .map(|x| {
+                x == "1" || x.to_lowercase() == "true" || x.to_lowercase() == "yes" || x.to_lowercase() == "y"
+            })
+            .unwrap_or(false);
+
+        if force_legacy {
+            info!(log, "SMITHAY_USE_LEGACY is set. Forcing LegacyDrmDevice.");
+        };
+        
+        Ok(if dev.set_client_capability(ClientCapability::Atomic, true).is_ok() && !force_legacy {
```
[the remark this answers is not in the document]
Drakulix: good catch.

src/backend/renderer/gles2/mod.rs
Drakulix: Because that is what "gl_generator" generates and the api therefor expects (and to Rust "system" is different from "C", although this code will likely never run on anything else then linux anyway).

src/backend/renderer/mod.rs
Drakulix: At some point, yes. ^^

src/backend/drm/device/mod.rs
```diff
@@ -0,0 +1,586 @@
...
+    /// - [`crtc`](drm::control::crtc)s represent scanout engines of the device pointing to one framebuffer. \
+    ///     Their responsibility is to read the data of the framebuffer and export it into an "Encoder". \
+    ///     The number of crtc's represent the number of independant output devices the hardware may handle.
+    /// - [`plane`](drm::control::plane)s represent a single plane on a crtc, which is composite together with
+    ///     other planes on the same crtc to present the final image.
+    /// - [`mode`](drm::control::Mode) describes the resolution and rate of images produced by the crtc and \
+    ///     has to be compatible with the provided `connectors`.
+    /// - [`connectors`] - List of connectors driven by the crtc. At least one(!) connector needs to be \
+    ///     attached to a crtc in smithay.
+    pub fn create_surface(
```
vberger, line 338: So, from this doc I gather that if one is to use multiple planes, then a `DrmSurface` needs to be created for each.

Does the user have to provide the same set of `connector`s each time? What happens if multiple surfaces are created one the same crtc for different planes but a different set of connectors is provided each time?

And btw, what happens if I try to create the same surface (so for the same pair of crtc+plane) multiple times?

src/backend/allocator/swapchain.rs
```diff
@@ -0,0 +1,164 @@
...
+/// introduce, which is where `Swapchain` comes into play.
+///
+/// `Swapchain` allocates buffers for you and transparently re-created them, e.g. when resizing.
+/// All you tell the swapchain is: *"Give me the next free buffer"* (by calling [`acquire`](Swapchain::acquire)).
+/// You then hold on to the returned buffer during rendering and swapping and free it once it is displayed.
+/// Efficient re-use of the buffers is done by the swapchain.
+///
+/// If you have associated resources for each buffer, that can also be re-used (e.g. framebuffer Handles for a `DrmDevice`),
+/// you can store then in the buffer slots userdata, where it gets freed, if the buffer gets allocated, but
+/// is still valid, if the buffer was just re-used. So instead of creating a framebuffer handle for each new
```
vberger, line 37: This sentence is not very clear, I had to re-read it 3 times to get what it means. What about something like this?

> If you have resources associated with a buffer that can be reused (e.g. feamebuffer `Handle`s for a `DrmDevice`), you can store them in the `Slot`s `userdata` field. If a buffer is re-used, its `userdata` is preserved for the next time it is returned by `acquire()`.

src/backend/allocator/gbm.rs
```diff
@@ -0,0 +1,140 @@
+//! Module for Buffers created using [libgbm](reexports::gbm)
```
vberger, line 1: It's pretty weird in the documentation to have this module existing while containing only the `GbmConvertError` enum. At least the documentation should explain that the user needs to directly use the `gbm::Device` type as its allocator, an maybe a few relevant types from `gbm` could be re-exported?

src/backend/allocator/dmabuf.rs
```diff
@@ -0,0 +1,177 @@
...
+pub trait AsDmabuf {
+    /// Error type returned, if exporting fails
+    type Error;
+
+    /// Export this buffer as a new Dmabuf
+    fn export(&self) -> Result<Dmabuf, Self::Error>;
+}
+
+impl AsDmabuf for Dmabuf {
+    type Error = ();
```
vberger, line 172: Maybe, waiting for `!` to be stabilized, we could introduce our own `Never` empty enum for use in cases where the error condition can never happen? That could simplify matches & such down the line.

src/backend/renderer/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+        mat = mat * transform.invert().matrix();
+        mat = mat * Matrix3::from_translation(Vector2::new(-0.5, -0.5));
+
+        self.render_texture(texture, mat, alpha)
+    }
+}
+
+/// Returns the dimensions of a wl_buffer
+#[cfg(all(feature = "wayland_frontend", feature = "backend_egl"))]
+pub fn buffer_dimensions(
```
vberger, line 257: Maybe this function should be in `backend::egl`, no?

src/backend/egl/mod.rs
```diff
@@ -18,20 +18,20 @@
 //! You may then use the resulting [`EGLDisplay`](::backend::egl::EGLDisplay) to receive [`EGLImages`](::backend::egl::EGLImages)
 //! of an EGL-based [`WlBuffer`](wayland_server::protocol::wl_buffer::WlBuffer) for rendering.
```
vberger, line 19: All intra-docs links of this doc block are broken.

src/backend/egl/display.rs
```diff
@@ -446,39 +542,136 @@ impl<B: native::Backend, N: native::NativeDisplay<B>> EGLGraphicsBackend for EGL
...
+
+    trace!(log, "Supported dmabuf import formats: {:?}", texture_formats);
+    trace!(log, "Supported dmabuf render formats: {:?}", render_formats);
+
+    Ok((texture_formats, render_formats))
+}
+
 /// Type to receive [`EGLImages`] for EGL-based [`WlBuffer`]s.
 ///
 /// Can be created by using [`EGLGraphicsBackend::bind_wl_display`].
```
vberger, line 661: These links don't render in rustdoc.

src/backend/egl/display.rs
```diff
@@ -490,7 +683,7 @@ impl EGLBufferReader {
     pub fn egl_buffer_contents(
```
vberger, line 683: Looks like this doc is outdated (the original buffer is not returned, given it is only passed by reference).

src/backend/drm/device/mod.rs
Drakulix: Oh yeah, that is a good point. This is not only lacking on a documentation level, but probably also on an API level.

> So, from this doc I gather that if one is to use multiple planes, then a `DrmSurface` needs to be created for each.

yes.

> Does the user have to provide the same set of `connector`s each time? What happens if multiple surfaces are created one the same crtc for different planes but a different set of connectors is provided each time?

Different planes on the same crtc with the same connector will simply not work, which also means adding and removing connectors will be problematic, when you have multiple planes, because there is no way to commit them together and the atomic check would likely always fail.

This needs an API redesign. We probably need just one surface per crtc and another `plane` type that can be used e.g. for rendering to the different planes, but has nothing todo with managing surface-global stuff like connectors.
(Which makes the `DrmRenderSurface` an even worse abstraction, because that could not really handle multiple planes at all.)

> And btw, what happens if I try to create the same surface (so for the same pair of crtc+plane) multiple times?

The Creation for the second one should also succeed and modify the drm state without updating the old one of course. So they would both probably be able to commit and alter the state to some degree, but stumble upon each other.

And I think I do not really want to fix this, because that is the user doing stupid stuff. I mean we could track used crtc/plane combinations inside the device, but I will probably just document that this may cause undefined behavior. What do you think?

src/backend/allocator/dmabuf.rs
Drakulix: yes, probably a good idea. Do you recall other similar cases of this in smithay?

joeftiger: Rust provides the error [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) for errors that should/will never happen.

src/backend/drm/device/mod.rs
Drakulix: Should be fixed by a4f66da.

src/backend/allocator/dmabuf.rs
vberger: Ah, I wasn't aware it was added in `std`, that's perfect!

src/backend/allocator/swapchain.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// Retrieve userdata for this slot.
+    pub fn userdata(&self) -> MutexGuard<'_, Option<U>> {
+        self.userdata.lock().unwrap()
+    }
+
+    /// Clear userdata contained in this slot.
+    pub fn clear_userdata(&self) -> Option<U> {
+        self.userdata.lock().unwrap().take()
+    }
```
vberger, line 75: Ok nitpick time, there is a lot of redundancy between these functions, mostly exposing only `userdata()` would be enough, no? The two others are trivially implemented in terms of it, to the point that maybe the user could just do it?

src/backend/allocator/dumb.rs
```diff
@@ -0,0 +1,79 @@
...
+    format: Format,
+}
+
+impl<A: AsRawFd + 'static> fmt::Debug for DumbBuffer<A> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("DumbBuffer")
+            .field("handle", &self.handle)
+            .field("format", &self.format)
+            .finish()
+    }
```
vberger, line 26: Why not print the fd number as well?

src/backend/allocator/dmabuf.rs
```diff
@@ -0,0 +1,177 @@
+//! Module for [dmabuf](https://01.org/linuxgraphics/gfx-docs/drm/driver-api/dma-buf.html) buffers.
```
vberger, line 1: Also, this would be nice to have an example (even if just a sentence explaining it) of why and when weak references to dmabufs are useful.

src/backend/allocator/dmabuf.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    fn format(&self) -> Format {
+        self.0.format
+    }
+}
+
+impl Dmabuf {
+    pub(crate) fn new(
+        src: &impl Buffer,
```
vberger, line 69: Could be worth adding a comment explaining that the `src` buffer will only be used as a reference for buffer size & format, but not its actual content.

src/backend/allocator/gbm.rs
```diff
@@ -0,0 +1,140 @@
...
+        self.width().unwrap_or(0)
+    }
+
+    fn height(&self) -> u32 {
+        self.height().unwrap_or(0)
+    }
+
+    fn format(&self) -> Format {
+        Format {
+            code: self.format().unwrap_or(Fourcc::Argb8888), // we got to return something, but this should never happen anyway
```
vberger, line 42: I have a feeling that if this even happens, it'll be a nightmare to debug...

src/backend/allocator/gbm.rs
```diff
@@ -0,0 +1,140 @@
...
+    InvalidFD,
+}
+
+impl<T> AsDmabuf for GbmBuffer<T> {
+    type Error = GbmConvertError;
+
+    fn export(&self) -> Result<Dmabuf, GbmConvertError> {
+        let planes = self.plane_count()? as i32;
+
+        //TODO switch to gbm_bo_get_plane_fd when it lands
```
vberger, line 68: Also, once the PR is merged, we should probably open an issue to track this TODO and not forget it.

src/backend/allocator/gbm.rs
```diff
@@ -0,0 +1,140 @@
...
+        if handle.is_none() {
+            // GBM is lacking a function to get a FD for a given plane. Instead,
+            // check all planes have the same handle. We can't use
+            // drmPrimeHandleToFD because that messes up handle ref'counting in
+            // the user-space driver.
+            return Err(GbmConvertError::UnsupportedBuffer); //TODO
+        }
+
+        let fds = [self.fd()?, 0, 0, 0];
+        //if fds.iter().any(|fd| fd == 0) {
```
vberger, line 89: Should this comment be removed?

src/backend/allocator/gbm.rs
```diff
@@ -0,0 +1,140 @@
...
+            .map(|i| self.stride_for_plane(i))
+            .collect::<Result<Vec<u32>, gbm::DeviceDestroyedError>>()?;
+
+        Ok(Dmabuf::new(self, planes as usize, &offsets, &strides, &fds).unwrap())
+    }
+}
+
+impl Dmabuf {
+    /// Import a Dmabuf using libgbm, creating a gbm Buffer Object to the same underlying data.
+    pub fn import<A: AsRawFd + 'static, T>(
```
vberger, line 107: given this is a method of `Dmabuf` and not `GbmDevice` I think it'd be better named `import_to`, so that it stands out that it is `self` that is being imported into the first argument, rather than the reverse (that would be what is naturally expected for a method named `import`).

src/backend/drm/mod.rs
```diff
@@ -57,335 +50,118 @@
...
+use drm::control::{plane, crtc, Device as ControlDevice, PlaneType};
+
+/// A set of planes as supported by a crtc
+pub struct Planes {
+    /// The primary plane of the crtc (automatically selected for [DrmDevice::create_surface])
+    pub primary: plane::Handle,
+    /// The cursor plane of the crtc, if available
+    pub cursor: Option<plane::Handle>,
+    /// Overlay planes supported by the crtc, if available
+    pub overlay: Option<Vec<plane::Handle>>,
```
vberger, line 88: Is the `Option<_>` wrapping necessary here? I't think that an empty `Vec` could play the role of `None`. :thinking:

src/backend/renderer/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+            Flipped180 => Transform::Flipped180,
+            Flipped270 => Transform::Flipped270,
+        }
+    }
+}
+
+/// Abstraction for Renderers, that can render into different targets
+pub trait Bind<Target>: Unbind {
+    /// Bind a given rendering target, which will contain the rendering results until `unbind` is called.
+    fn bind(&mut self, target: Target) -> Result<(), <Self as Renderer>::Error>;
```
vberger, line 111: This should document what behavior is expected if `bind()` is called twice with no `unbind()` in between. Is this an error, or will the latest target just replace the previous one?

src/backend/renderer/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// - The given Transformation is not supported by the renderer (`Transform::Normal` is always supported).
+    /// - There was a previous `begin`-call, which was not terminated by `finish`.
+    /// - This renderer implements `Bind`, no target was bound *and* has no default target.
+    /// - (Renderers not implementing `Bind` always have a default target.)
+    fn begin(
+        &mut self,
+        width: u32,
+        height: u32,
+        transform: Transform,
+    ) -> Result<(), <Self as Renderer>::Error>;
```
vberger, line 200: General impression about this begin/end API. Would it make sense to instead build something with a closure, so that it cannot be misused?

Something like:

```rust
rendered.draw(|render_handle: &mut _| {
    // render_handle is some rendered-specific type that has the clear/render_at/etc... methods
});
```

src/backend/renderer/gles2/mod.rs
```diff
@@ -0,0 +1,1079 @@
...
+use wayland_server::protocol::{wl_buffer, wl_shm};
+#[cfg(feature = "wayland_frontend")]
+use wayland_commons::user_data::UserDataMap;
+
+#[allow(clippy::all, missing_docs)]
+pub mod ffi {
+    include!(concat!(env!("OUT_DIR"), "/gl_bindings.rs"));
+}
+
+static RENDERER_COUNTER: AtomicUsize = AtomicUsize::new(0);
```
vberger, line 48: I'd add a comment explaining the role of this static.

src/backend/allocator/swapchain.rs
Drakulix: Yeah, you are probably right.

src/backend/renderer/mod.rs
Drakulix: I have no idea, how this will work with future renderers, but for gles2 the current target is just replaced.

So the question is: Do we wanna be cautious here and define that as an `Error` or renderer-dependent or do we roll with gles2 behavior and potentially fix that up later? (e.g. when we introduce a vulkan renderer.)

src/backend/renderer/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    ///
+    /// If not otherwise defined by the implementation, this texture id is only valid for the renderer, that created it,
+    /// and needs to be freed by calling `destroy_texture` on this renderer to avoid a resource leak.
+    ///
+    /// This operation needs no bound or default rendering target.
+    #[cfg(feature = "image")]
+    fn import_bitmap<C: std::ops::Deref<Target = [u8]>>(
+        &mut self,
+        image: &image::ImageBuffer<image::Rgba<u8>, C>,
+    ) -> Result<Self::TextureId, Self::Error>;
```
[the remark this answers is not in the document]
Drakulix: Do to this more generically we would be to do a mapping between `image::Pixels` and GL formats.
Likely by implementing `Into` or something like it. Also we potentially need to check more GL extensions.

And that covers only the changes for `Gles2Renderer`.
We probably need another associated type on the rendering trait for the `PixelFormat`.

Totally doable, but I just didn't bother right now. Do you think we should fix this, before introducing the trait to master?

src/backend/renderer/mod.rs
Drakulix: Might be a good idea.

Previously this trait featured an associated `Frame`-type, which was returned by `begin`, would `end` on drop and implemented the clear/render_at/etc functions. But that was an absolute lifetime nightmare.

I will see if the closure api works better and is flexible enough for anvil.

src/backend/renderer/mod.rs
vberger: If we don't want to commit to any behavior, we can document that doing it is generally unadvised and that the actual behavior is renderer-dependent, but it should be documented clearly I think.

src/backend/renderer/mod.rs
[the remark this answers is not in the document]
vberger: I don't think this is blocking this PR, but if this is doable in a reasonable way, I think it'd be worth opening an issue to track it.

src/backend/renderer/mod.rs
vberger: That's the uncertainty I have with the closure yes, it's possible it'd not be flexible enough for the actual uses.

src/backend/allocator/dumb.rs
Drakulix: Because thats the fd of the drm device. Not sure, if that is actually helpful or more confusing.

src/backend/allocator/gbm.rs
Drakulix: Not really. If this happens the file descriptor of the gbm device has become invalid. So we cannot actually submit any rendering, that could be based on this format and the chain will fail sometime later anyway.

PR #263, opened 2021-04-27
Fix clippy warnings

joeftiger:
I fixed some clippy warnings by
- globally allowing upper case acronyms. There are a lot of structs/enum variants like *EGL*, therefore I allowed it globally. Should this be handled manually everywhere? There were many cases.
- globally allowing returning `Resulty<(), ()>` where clippy would warn about empty error types.
- locally allowing `match` that clippy would deem suitable to use `matches!()` instead. However, the code would not be more readable in my opinion, so I allowed those locally.
- updating the usage of a deprecated std function (since Rust version 1.50) to stable (since Rust version 1.10)
- using `vec![]` instead of manually creating mutable `Vec<>`
- using `From<>` instead of `Into<>`

For me the question remains whether the global allowing of acronyms and empty results is okay. I am not sure about standard practice here as I am only using Rust since end of last year.

### DIFF
--- a/anvil/src/glium_drawer.rs
+++ b/anvil/src/glium_drawer.rs
@@ -88,7 +88,7 @@ impl<T: Into<GliumGraphicsBackend<T>> + GLGraphicsBackend + 'static> GliumDrawer
 
         // building the index buffer
         let index_buffer =
-            glium::IndexBuffer::new(&display, PrimitiveType::TriangleStrip, &[1 as u16, 2, 0, 3]).unwrap();
+            glium::IndexBuffer::new(&display, PrimitiveType::TriangleStrip, &[1, 2, 0, 3]).unwrap();
 
         let programs = opengl_programs!(&display);
 
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -283,6 +283,7 @@ enum KeyAction {
     None,
 }
 
+#[allow(clippy::manual_range_contains)]
 fn process_keyboard_shortcut(modifiers: ModifiersState, keysym: Keysym) -> KeyAction {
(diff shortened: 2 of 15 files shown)


vberger:
Hi, thanks for the PR!

However, it appears that your PR has a very large conflict overlap with #261, which is mostly a complete rewrite of the graphic backend code, so we should probably postpone these clippy fixes until after that pr is finished (cc @Drakulix, what do you think?).

Drakulix:
> However, it appears that your PR has a very large conflict overlap with #261, which is mostly a complete rewrite of the graphic backend code, so we should probably postpone these clippy fixes until after that pr is finished (cc @Drakulix, what do you think?).

I agree, still this is a very useful reference, because parts of this are just my habits how I write rust. I will try to avoid those in #261 (and just run clippy before merging...).

Afterwards we should take a look, if parts of this still apply.

Drakulix:
Some of these are addressed by https://github.com/Smithay/smithay/pull/261/commits/7518db3dc2e5c45b9951652b63dc5701cc360212 . Thanks, this is a good reference.

@vberger Most offenders of `clippy::result_unit_err` (using `Result<_, ()>` instead of a proper error type) are inside our wayland frontend. You should be the judge of that.

joeftiger:
Alright, I'll wait for the big overhaul of #261 and see whether there's still something left ;-)

Should this be closed then?

vberger:
> @vberger Most offenders of `clippy::result_unit_err` (using `Result<_, ()>` instead of a proper error type) are inside our wayland frontend. You should be the judge of that.

Regarding that, the code should be fixed by introducing proper error types.

> Should this be closed then?

I guess, it can still be reopened later anyway.

PR #270, opened 2021-05-13
check for gl version in gles renderer...

### LINKED ISSUE #269: Failed to launch anvil with GLES 3.1 due to "missing" extension
## Expected Behavior

Launching anvil with drm backend on gles 3.1 should work.

## Actual Behavior

anvil fails to start due to missing gles extension "GL_EXT_unpack_subimage"

## Steps to Reproduce the Problem

  1. anvil --tty-udev

## Specifications

  - Version: renderer/v4
  - Platform: aarch64 i.MX8M
  - Subsystem: gles2_renderer

cmeissl:
Fixes #269 

Starting with Version 3.0 `UNPACK_ROW_LENGTH` is included in GLES and there is no need to check for
the `GL_EXT_unpack_subimage` extension. 
This allows to run anvil on platforms that do not explicit announce the `GL_EXT_unpack_subimage` extension like
the i.MX8M Processor.

## Proposed Changes

  - Check for gl version in gles renderer and only test for `GL_EXT_unpack_subimage` if the version is less than 3.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ wayland-server = { version = "0.28.3", optional = true }
 wayland-sys = { version = "0.28", optional = true }
 winit = { version = "0.24.0", optional = true }
 xkbcommon = "0.4.0"
-scan_fmt = "0.2"
+scan_fmt = { version = "0.2", default-features = false }
 
 [dev-dependencies]
 slog-term = "2.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ wayland-server = { version = "0.28.3", optional = true }
 wayland-sys = { version = "0.28", optional = true }
 winit = { version = "0.24.0", optional = true }
 xkbcommon = "0.4.0"
+scan_fmt = "0.2"
 
 [dev-dependencies]
 slog-term = "2.3"
--- a/src/backend/renderer/gles2/mod.rs
+++ b/src/backend/renderer/gles2/mod.rs
@@ -10,10 +10,13 @@ use std::sync::{
     mpsc::{channel, Receiver, Sender},
 };
 use std::{collections::HashSet, os::raw::c_char};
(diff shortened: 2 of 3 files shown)


Drakulix:
Thanks, that looks very useful on first sight, but it will take another day or two for proper review the new version code.

cmeissl:
Sounds good, thanks!

Cargo.toml
```diff
@@ -37,6 +37,7 @@ wayland-server = { version = "0.28.3", optional = true }
 wayland-sys = { version = "0.28", optional = true }
 winit = { version = "0.24.0", optional = true }
 xkbcommon = "0.4.0"
+scan_fmt = "0.2"
```
Drakulix, line 40: Since we do not need it, could you disable the regex feature of scan_fmt?

Drakulix (approved):
Small nitpick in an effort to keep the list of our dependencies (relatively) small.
Otherwise the code looks good! Thanks!

(And reminds me, I really need to do a rustfmt run against this PR. @cmeissl Would you prefer me to format the code before or after #272? If I format this before, I would prefer a rebase on your PR.)

Cargo.toml
cmeissl: > Small nitpick in an effort to keep the list of our dependencies (relatively) small.

This was one of my concers too, i tried to avoid it, but string parsing is not that trivial and i tried to find a small lib that can
free me from that burden :)

> Since we do not need it, could you disable the regex feature of scan_fmt?

I will check it out and update the PR

> (And reminds me, I really need to do a rustfmt run against this PR. @cmeissl Would you prefer me to format the code before or after #272? If I format this before, I would prefer a rebase on your PR.)

You can format the code before #272 if you like to, no problem.

cmeissl:
I have rebased the pr to the latest version of renderer/v4 and disabled the default features of scan_fmt

PR #272, opened 2021-05-13
Dynamically select the EGL platform display

### LINKED ISSUE #271: EGL Platform Display can not be initiaized due to missing egl extension
## Description

At the moment only `EGL_MESA_platform_gbm` is supported which prevents launching
anvil on platforms that only support `EGL_KHR_platform_gbm`.

## Expected Behavior

The EGL backend works with the EGL platform `EGL_KHR_platform_gbm`.

## Actual Behavior

At the moment only `EGL_MESA_platform_gbm` is supported and no
EGL platform display can be initialized on platforms that only support
`EGL_KHR_platform_gbm`.

## Steps to Reproduce the Problem

  1. start anvil on a platform that does not support `EGL_MESA_platform_gbm` with `anvil --tty-udev`

## Specifications

  - Version: renderer/v4
  - Platform: aarch64/i.MX8M
  - Subsystem: backend/egl

cmeissl:
Fixes #271 

## Proposed Changes

  - Dynamically select the egl platform


Drakulix (commented):
Nothing really to complain about the code.

I noticed a few more `Vec`, then I would personally like, but because we neither can return an `impl Iterator` in traits nor treat the list of platforms as `'static` (because of `native_display`), there is not much we can do about it. Additionally this is not a performance critical code path, so code-wise this is good to go.

But I would appreciate trying to support as many extensions as possible with this approach, if we are doing this anyway.

This would mean to add at least the following to the list, where possible:
- `EGL_KHR_platform_wayland`
- `EGL_KHR_platform_x11`
- `EGL_EXT_platform_gbm`

cmeissl:
Great, i will try to add the platforms.

cmeissl:
Platforms added:
- `EGL_KHR_platform_wayland`
- `EGL_KHR_platform_x11`

I could not find information about `EGL_EXT_platform_gbm` and it is not included in the egl header files on my laptop.

There are two additional extensions but i don't think they are relevant:
- [`EGL_EXT_platform_device`](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_platform_device.txt)
- [`EGL_MESA_platform_surfaceless`](https://www.khronos.org/registry/EGL/extensions/MESA/EGL_MESA_platform_surfaceless.txt)

At least not for the drm and winit backends.

I tried to document the new EGLPlatform class and added a macro for easier initialization and to get the `platform_name` directly from the platform ffi ident. I did not like the error prone repetition.

src/backend/egl/display.rs
```diff
@@ -61,6 +63,85 @@ pub struct EGLDisplay {
...
+        };
+
+        info!(
+            log,
+            "Successfully selected EGL platform: {}", platform.platform_name
+        );
+        return Ok(display);
+    }
+
+    crit!(log, "Unable to find suitable EGL platform");
```
Drakulix, line 141: Because the user might have multiple gpus or different backends running at the same time, I would not describe this error as *critical*. smithay might very well be functional, so I would suggest using `error!` here instead.

Drakulix (approved):
>
> I could not find information about `EGL_EXT_platform_gbm` and it is not included in the egl header files on my laptop.
>

I might have miss-remembered that.
 
> There are two additional extensions but i don't think they are relevant:
> 
>     * [`EGL_EXT_platform_device`](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_platform_device.txt)
> 
> At least not for the drm and winit backends.

EGL_EXT_platform_device might be necessary to provide proper hardware acceleration on the nvidia driver, but adding that now would cause other shenanigans with egl-buffers, so better leave it out for now. I might do tests with that platform later.

> I tried to document the new EGLPlatform class and added a macro for easier initialization and to get the `platform_name` directly from the platform ffi ident. I did not like the error prone repetition.

Nicely done. I really like, that users can still add platforms out-of-tree, if they implement `NativeDisplay`.

src/backend/egl/display.rs
cmeissl: Right, makes sense :)
Done!

PR #274, opened 2021-05-18
ci: Test specific features without default set

Drakulix:
Currently we always run tests for specific features with the default set enable, which is really pointless as the default set is current equivalent to "all". Lets fix that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --features "${{ matrix.features }}"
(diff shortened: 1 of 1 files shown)


Drakulix:
Nice, we have already broken everything except for default on master :rofl:

vberger:
If the plan to also fix the compilation failures in this PR?

Drakulix:
> If the plan to also fix the compilation failures in this PR?

Yes, otherwise the CI-experience would be even worse.
Just had some other priorities. :sweat_smile:

Drakulix:
Rebased on top of #288 to check, if that indeed solves our CI problems and it does! :)

PR #277, opened 2021-05-24
Fix damage handling to support multiple rectangles

vberger:


Drakulix:
Pulled into #261 as 6bfe6e1.

PR #278, opened 2021-05-24
Fix frame callbacks for firefox

cmeissl:
This PR changes the `frame_callback` to a `Vec` of `frame_callbacks`.
A client is allowed to register multiple frame callbacks with a single commit.
This will fix the startup of firefox-wayland which does use more than a single frame callback.


Drakulix (approved):
:+1:

PR #280, opened 2021-05-29
allocator: Let the allocator select the best modifier out of a given set

Drakulix:
This change fixes modifier selection by delegating this to the allocators
and thus to libgbm, which can ask the driver for an appropriate modifier
for scanout, that results in the best possible performance.

We do not have this information, the order in which modifiers are returned
by EGL has no meaning and this is far better then testing modifiers
non-deterministically at random and choosing the first one, that does
not error out...

### DIFF
--- a/src/backend/allocator/mod.rs
+++ b/src/backend/allocator/mod.rs
@@ -49,5 +49,11 @@ pub trait Allocator<B: Buffer> {
     type Error: std::error::Error;
 
     /// Try to create a buffer with the given dimensions and pixel format
-    fn create_buffer(&mut self, width: u32, height: u32, format: Format) -> Result<B, Self::Error>;
+    fn create_buffer(
+        &mut self,
+        width: u32,
+        height: u32,
+        fourcc: Fourcc,
+        modifiers: &[Modifier],
+    ) -> Result<B, Self::Error>;
 }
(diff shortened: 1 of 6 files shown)


src/backend/allocator/mod.rs
```diff
@@ -49,5 +49,11 @@ pub trait Allocator<B: Buffer> {
     type Error: std::error::Error;
 
     /// Try to create a buffer with the given dimensions and pixel format
-    fn create_buffer(&mut self, width: u32, height: u32, format: Format) -> Result<B, Self::Error>;
+    fn create_buffer(
+        &mut self,
+        width: u32,
+        height: u32,
+        fourcc: Fourcc,
+        modifiers: impl Iterator<Item = Modifier>,
```
vberger, line 57: Why an `impl Iterator` rather than a `&[Modifier]`? It seems that your main implementation starts by collecting in into a `Vec` anyway. Giving a `&[Modifier]` as argument would allow to have lees copying around, no? Or is there something more I'm missing?

Drakulix: Just to be more flexible. The `gbm`-api also takes an Iterator, but since we might need to call multiple gbm functions, we cannot really use it.
So I guess, I can change this, good idea.

Drakulix: done

vberger (approved):

PR #281, opened 2021-05-30
Reorganize anvil around AnvilState

vberger:
- Update calloop to `0.8`
- Change the `wayland::compositor` to make `DispatchData` available in the callbacks
- Reorganize `AnvilState` to separate backend-specific types into a generic field
- Flatten `UdevHandler` and `DrmRenderer` into `AnvilState`


vberger:
I think the CI failure is contingent to #274, I don't think I'm introducing *more* failure. Also, I've tested that anvil still works fine on TTY on my computer.

vberger:
Also, this significantly advances #203 :)

Drakulix (commented):
Well I found two nitpicks, a typo and likely an oversight.
The rest makes sense and is definitely an improvement for anvil. So all that's left to say is: Good Job!

Issue #97, opened 2018-05-20
Improve xdg_shell abstraction to better reflect the protocol


vberger:
The xdg_shell protocol associates a specific meaning to its elaborate configure dance. To honor it, our abstraction should keep track of the contents of the different configure sent to the client, and when the client acks one, discard all the older ones, and mark this one as the "current" state of the client.

As such, smithay's users should not assume that the client automatically is in the state they last configured it to, but rather smithay should expose some API to allow for accessing the last state that was acked by the client.

Also, other protocol extensions tend to piggy-back more state on the xdg_shell configure, so we need to allow associating arbitrary data with a state, possibly with a similar mechanism as what we currently do with the role data, via traits (but as a product type rather than a sum type in this case).

So, a seemingly generic API t expose on Toplevel/Popup surfaces would be something like:

```rust
/// Mutably access the current pending state of this surface, to modify its contents
///
/// This can be used to integrate data from external protocols into the xdg_shell
/// configure mechanism
pub fn with_pending_state<T, F>(&self, F) -> T 
    where F: FnOnce(&mut ConfigureState) -> T
{
    /* ... */
}

/// Send the current pending state to the client
///
/// A copy of the pending state is stored as sent to the client, and will become the
/// current state when the client acks it.
pub fn send_configure(&self) {
    /* ... */
}

/// Retrieve last ackend state of this surface
pub fn with_current_state<T, F>(&self, F) -> T
    where F: FnOnce(&ConfigureState) -> T
{
    /* ... */
}
```

YaLTeR:
But the pending state is applied on commit, and the client acking a configure event doesn't actually tell us anything about it because the configure event does not contain any info that's present in the pending state? So I think it should rather be a "last committed state", similar to what I'm currently adding to `SurfaceData` in anvil.

vberger:
These are two different pending states. The one sent by the server and acked by the client (which is the one this issue is concerned with), and the one sent by the client and commited on surfaces.

So here, the last acked state is the state the server should use as a reference wrt to what it has sent and the client is supposed to follow (for example about the configure size for a maximized surface, for which not obeying is a protocol error).

vberger:
Done in #286, thanks @cmeissl !

Issue #283, opened 2021-06-01
The subsurface state handling needs to be rethought


vberger:
Current handling of subsurface state caching in anvil is simplistic. If we follow the protocol strictly, a subsurface can have several intermediate cached states, in addition of its current and pending states. The whole subcompositor design needs to be redone with that in mind.

In addition, roles and some other protocol extensions piggy-back on the `wl_surface.commit` mechanism, so we need and interface that allows other smithay modules and the user to insert new caching-compatible state (as well as non-cached state) associated to a `wl_surface`.

Finally, the new design should be made with the upcoming [`wp_transaction` protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/26) in mind, so that when it lands we are able to integrate it seamlessly. There will also probably be some interaction with `explicit_synchronization`.

Also, this should address #260 as well

vberger:
Done in #304

PR #286, opened 2021-06-03
rework xdg_shell

cmeissl:
# Motivation

Based on the ideas of #97 

At the moment there is one surface role, `XdgSurfaceRole`, defined for the `xdg_shell` which is
assigned to the surface during processing the request `shell.get_xdg_surface`.
The `xdg_shell` protocol specification defines that the `xdg_surface` itself is not a role for `wl_surface`.
While it could be used as a simplification to represent it as a role in the compositor i believe it comes with
a few downsides. One is that it does not reflect the specification, the second one is that i find it cumberstone to
match on the `XdgSurfaceRole.pending_state` when processing the surface state. If a surface has been for example
assigned a role of `xdg_toplevel`, which is defined as a surface role in the specification, there should be no need to check if the
`XdgSurfaceRole.pending_state` is in fact a `XdgSurfacePendingState::Toplevel`.

Additionaly i would like put most of the state/protocol handling into smithay so that it is easier to integrate
the `xdg_shell` into a compositor.

## Proposed Changes

  - Use distinct roles for `xdg_toplevel` and `xdg_popup` and only assign them on `xdg_surface.get_toplevel` and `xdg_surface.get_popup`
  - Handle most of state/protocol handling inside the `xdg_handler` and/or `ToplevelSurface`/`PopupSurface`
  - Represent that a `xdg_surface` role has been destroyed by using an `Option` on the role `Attributes`
  
### Split `XdgSurfaceRole` into two distinct roles

In the proposed solution the roles are generated with the macro `xdg_role!`, the macro will generate
a struct for the role attributes unifying the base attributes defined in `xdg_surface` and the role specific
attributes. This way there is no need for a generic role or to add the base attributes as a field to the role.

Alternatives i considered are to use a generic `XdgSurfaceRole<A>` where `A` represents the role specific attributes
in an `Option` field. Or to have a `XdgToplevelSurfaceRole` with a `base: XdgSurfaceAttributes` field holding the base data.
Both solutions felt cumberstone to use in anvil when accessing the role attributes.

Do not only store the serial of the configures, but also the state when `configure` was called.

### State/Protocol Handling

Add the base protocol implemention for configure/ack_configure into the `xdg_handlers` and/or `ToplevelSurface`/`PopupSurface`. The `configure` function has been changed to use a saved `server_pending` state for `xdg_toplevel` which can be accessed by using `ToplevelSurface::with_pending_state`. The `ack_configure` request is mostly handled inside the `xdg_handler` and the existing `XdgRequest::AckConfigure` can be used as an integration point for protocol extensions and pointer/resize integration.

## Open questions

How much to put in the `xdg_handler` or in the `ToplevelSurface`/`PopupSurface` implementation.
For example the handling of `xdg_surface.ack_configure` contains a lot of duplicate code at the moment. I thought about putting it into  `ToplevelSurface`/`PopupSurface` and that the `user_impl` needs to call `ack_configure` on `ToplevelSurface`/`PopupSurface`. But i do not favor to do so because than the call to `ack_configure` would be mandatory in the compositor and not an integration point. Not calling `ack_configure` would break the protocol. I would try to add some private helper functions to reduce the code duplication.

How much of the role specific `commit` logic should reside in smithay or be handled external. For the moment i have added
a  `commit` function on `ToplevelSurface`/`PopupSurface` which is called during `surface_commit` in anvil. This is a simplification at the moment as commit should respect `subsurface` commit handling.

In wlroots and weston `configure` will schedule an idle task on the event loop instead of sending it directly. The exact reason
to do so are unclear, but it would be a good optimization to not flood the client with configures. Handling this outside of smithay could be challenging. For the moment i left it out of the implementation until there is a prove that this is more than just an optimization.

What to do with popups that have no parent. This has been changed between the `zxdg_shell_v6` and the stable version.
In the stable version the protocol allows for a `NULL` parent during `xdg_surface.get_popup` and specifies that the parent
can be specified out of band.

## Issues/Unsolved Problems

When handling the `ack_configure` in the `xdg_handler` the `user_impl` is called for the previous not acked serials and the current serial. The impl should not be called while holding a lock on the surface role data as this could result in a deadlock when the user code again tries to use the role data. I use a `Vec` at the moment to collect all configures which i would like to get rid of.

## Additional things

A simple experiment was added for supporting `XdgPopup`. Popups show up on the screen, but on the wrong position as the position is hard coded, i just wanted to check protocol handling correctness.

### DIFF
--- a/src/wayland/shell/mod.rs
+++ b/src/wayland/shell/mod.rs
@@ -14,5 +14,29 @@
 //! - The [`legacy`](legacy/index.html) module provides handlers for the `wl_shell` protocol, which
 //!   is now deprecated. You only need it if you want to support apps predating `xdg_shell`.
 
+use super::Serial;
+use thiserror::Error;
+
 pub mod legacy;
 pub mod xdg;
+
+/// Represents the possible errors returned from
+/// a surface operation
+#[derive(Debug, Error)]
+pub enum SurfaceError {
+    /// The underlying `WlSurface` is no longer alive
+    #[error("the surface is not alive")]
+    SurfaceNotAlive,
+}
+
+/// Represents the possible errors returned from
+/// a surface ping
+#[derive(Debug, Error)]
+pub enum PingError {
+    /// The operation failed because the underlying surface has an error
+    #[error("the ping failed cause the underlying surface has an error")]
+    SurfaceError(#[from] SurfaceError),
+    /// There is already a pending ping
+    #[error("there is already a ping pending `{0:?}`")]
+    PingAlreadyPending(Serial),
+}
--- a/src/wayland/shell/mod.rs
+++ b/src/wayland/shell/mod.rs
@@ -21,21 +21,12 @@ pub mod legacy;
 pub mod xdg;
 
 /// Represents the possible errors returned from
-/// a surface operation
-#[derive(Debug, Error)]
-pub enum SurfaceError {
-    /// The underlying `WlSurface` is no longer alive
-    #[error("the surface is not alive")]
-    SurfaceNotAlive,
-}
-
-/// Represents the possible errors returned from
 /// a surface ping
 #[derive(Debug, Error)]
 pub enum PingError {
-    /// The operation failed because the underlying surface has an error
-    #[error("the ping failed cause the underlying surface has an error")]
-    SurfaceError(#[from] SurfaceError),
+    /// The operation failed because the underlying surface has been destroyed
+    #[error("the ping failed cause the underlying surface has been destroyed")]
+    DeadSurface,
     /// There is already a pending ping
     #[error("there is already a ping pending `{0:?}`")]
     PingAlreadyPending(Serial),
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -110,7 +110,6 @@ use wayland_server::{
 };
 
 use super::PingError;
-use super::SurfaceError;
 
 // handlers for the xdg_shell protocol
 mod xdg_handlers;
@@ -249,13 +248,13 @@ where
 {
     fn set_window_geometry(&mut self, geometry: Rectangle) {
         self.as_mut()
-            .expect("xdg_toplevel exists but role has been destroyed?!")
+            .expect("xdg_surface exists but role has been destroyed?!")
             .set_window_geometry(geometry)
     }
 
     fn ack_configure(&mut self, serial: Serial) -> Result<Configure, ConfigureError> {
         self.as_mut()
-            .expect("xdg_toplevel exists but role has been destroyed?!")
+            .expect("xdg_surface exists but role has been destroyed?!")
             .ack_configure(serial)
     }
 }
@@ -596,7 +595,7 @@ pub struct ToplevelState {
     pub size: Option<(i32, i32)>,
 
     /// The states for this surface
-    pub states: ToplevelStates,
+    pub states: ToplevelStateSet,
 
     /// The output for a fullscreen display
     pub fullscreen_output: Option<wl_output::WlOutput>,
@@ -629,11 +628,11 @@ impl Clone for ToplevelState {
 /// and simplifies setting and un-setting a particularly
 /// `xdg_toplevel::State`
 #[derive(Debug, Clone, PartialEq)]
-pub struct ToplevelStates {
+pub struct ToplevelStateSet {
     states: Vec<xdg_toplevel::State>,
 }
 
-impl ToplevelStates {
+impl ToplevelStateSet {
     /// Returns `true` if the states contains a state.
     pub fn contains(&self, state: xdg_toplevel::State) -> bool {
         self.states.iter().any(|s| *s == state)
@@ -665,13 +664,13 @@ impl ToplevelStates {
     }
 }
 
-impl Default for ToplevelStates {
+impl Default for ToplevelStateSet {
     fn default() -> Self {
         Self { states: Vec::new() }
     }
 }
 
-impl IntoIterator for ToplevelStates {
+impl IntoIterator for ToplevelStateSet {
     type Item = xdg_toplevel::State;
     type IntoIter = std::vec::IntoIter<Self::Item>;
 
@@ -680,8 +679,8 @@ impl IntoIterator for ToplevelStates {
     }
 }
 
-impl From<ToplevelStates> for Vec<xdg_toplevel::State> {
-    fn from(states: ToplevelStates) -> Self {
+impl From<ToplevelStateSet> for Vec<xdg_toplevel::State> {
+    fn from(states: ToplevelStateSet) -> Self {
         states.states
     }
 }
@@ -712,6 +711,19 @@ impl Default for ToplevelClientPending {
             max_size: None,
         }
     }
+}
+
+/// Represents the possible errors
+/// returned from `ToplevelSurface::with_pending_state`
+/// and `PopupSurface::with_pending_state`
+#[derive(Debug, Error)]
+pub enum PendingStateError {
+    /// The operation failed because the underlying surface has been destroyed
+    #[error("could not retrieve the pending state because the underlying surface has been destroyed")]
+    DeadSurface,
+    /// The role of the xdg_surface has been destroyed
+    #[error("the role of the xdg_surface has been destroyed")]
+    RoleDestroyed,
 }
 
 pub(crate) struct ShellData<R> {
@@ -878,7 +890,7 @@ where
     /// Fails if this shell client already has a pending ping or is already dead.
     pub fn send_ping(&self, serial: Serial) -> Result<(), PingError> {
         if !self.alive() {
-            return Err(PingError::SurfaceError(SurfaceError::SurfaceNotAlive));
+            return Err(PingError::DeadSurface);
         }
         match self.kind {
             ShellClientKind::Xdg(ref shell) => {
@@ -912,12 +924,12 @@ where
     }
 
     /// Access the user data associated with this shell client
-    pub fn with_data<F, T>(&self, f: F) -> Result<T, SurfaceError>
+    pub fn with_data<F, T>(&self, f: F) -> Result<T, crate::utils::DeadResource>
     where
         F: FnOnce(&mut UserDataMap) -> T,
     {
         if !self.alive() {
-            return Err(SurfaceError::SurfaceNotAlive);
+            return Err(crate::utils::DeadResource);
         }
         match self.kind {
             ShellClientKind::Xdg(ref shell) => {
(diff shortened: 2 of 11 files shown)


src/wayland/shell/xdg/mod.rs
```diff
@@ -108,36 +113,270 @@ mod xdg_handlers;
...
+                /// Defines if the surface has received at least one
+                /// xdg_surface.ack_configure from the client
+                pub configured: bool,
+
+                /// The serial of the last acked configure
+                pub configure_serial: Option<Serial>,
+
+                /// Holds the next serial that will be used for
+                /// xdg_surface.configure
+                pub configure_next_serial: Serial,
```
vberger, line 144: This should just be using the global serial counter, we don't need to have independent serials for each surface, quite the contrary.

src/wayland/shell/xdg/mod.rs
```diff
@@ -108,36 +113,270 @@ mod xdg_handlers;
...
+    };
+}
+
+trait CompositorTokenExt {
+    fn with_xdg_role<F, T>(&self, surface: &wl_surface::WlSurface, f: F) -> Result<T, WrongRole>
+    where
+        F: FnOnce(&mut dyn XdgSurface) -> T;
+}
+
+impl<R> CompositorTokenExt for CompositorToken<R>
```
vberger, line 225: `CompositorToken` is defined in the same crate, we don't need an extension trait to add new methods to it.

src/wayland/shell/xdg/mod.rs
```diff
@@ -175,88 +427,150 @@ impl PositionerState {
...
-    /// This is a popup surface
-    ///
-    /// This corresponds to the `xdg_popup` role
-    ///
-    /// This are mostly for small tooltips and similar short-lived
-    /// surfaces.
-    Popup(PopupState),
-    /// This surface was not yet assigned a kind
-    None,
+impl From<&PositionerState> for Rectangle {
```
vberger, line 484: Could you add a comment explaining what this conversion does exactly?

Given it's a pretty non-obvious conversion, it should probably be a method of `PositionerState`, rather than a `From<_>` impl.

src/wayland/shell/xdg/mod.rs
```diff
@@ -175,88 +427,150 @@ impl PositionerState {
...
-/// The pending state of a popup surface
-#[derive(Debug)]
-pub struct PopupState {
-    /// Parent of this popup surface
-    pub parent: Option<wl_surface::WlSurface>,
-    /// The positioner specifying how this tooltip should
-    /// be placed relative to its parent.
-    pub positioner: PositionerState,
+    /// The states for this surface
+    // TODO: I would like to use a HashSet here, but the state
```
vberger, line 545: Why a HashSet?

src/wayland/shell/xdg/mod.rs
```diff
@@ -567,13 +881,81 @@ where
     /// Send a configure event to this toplevel surface to suggest it a new configuration
     ///
     /// The serial of this configure will be tracked waiting for the client to ACK it.
-    pub fn send_configure(&self, cfg: ToplevelConfigure) {
-        if !self.alive() {
-            return;
+    pub fn send_configure(&self) {
+        // TODO: Handle the idle scheduling of the configure here?
```
vberger, line 885: No. The method `send_configure` should do the actual sending. It's up to the user to schedule invoking this method from an idle callback if they desire so.

src/wayland/shell/xdg/mod.rs
```diff
@@ -108,36 +113,270 @@ mod xdg_handlers;
...
 mod zxdgv6_handlers;
 
-/// Metadata associated with the `xdg_surface` role
+macro_rules! xdg_role {
+    ($(#[$role_attr:meta])* $role:ident,
+     $configure:ty,
+     $(#[$attr:meta])* $element:ident {$($(#[$field_attr:meta])* $vis:vis$field:ident:$type:ty),*},
+     $role_ack_configure:expr) => {
+        $(#[$role_attr])*
+        pub type $role = Option<$element>;
```
vberger, line 122: Why is this an Option?

vberger (commented):
So, the general design you are proposing is pretty interesting, I just left a few questions/comments regarding it.

There is one thing I'm unsure about though, it's that in the process of doing #283, I'll very likely completely rework the whole role-handling logic, and I'm not clear at all how this with your work. There will be significant overlap, though.

src/wayland/shell/xdg/mod.rs
cmeissl: I have removed this field and moved most of the configure logic to the `ToplevelSurface` and `PopupSurface`. This has the nice side effect of reducing code duplication between `xdg_wm_base` and `zgdxv6`.

src/wayland/shell/xdg/mod.rs
cmeissl: The option represents the destruction of the `xdg_role`. This is again checked in the destruction of the `xdg_surface` which raises a protocol error if the role has not be destroyed before destroying the surface.
Before I changed the code this was represented as a `PendingState` enum with `Toplevel`, `Popup` and `None`.
As I have removed the enum I used a `Option` to represent the `None` in both roles.
An alternative could be to have a `is_valid` or `is_destroyed` field in the base attributes and check against that.

src/wayland/shell/xdg/mod.rs
cmeissl: That would add a reference to the `XdgSurface` trait inside the `compositor` mod which i assumed would be against the modular design of smithay. I could just use a helper function here, i just did not want to repeat the checks for both roles
all over the xdg implementation.

src/wayland/shell/xdg/mod.rs
cmeissl: With a HashSet there could only be at most one "instance" of each `xdg_toplevel::State` variant. At the moment i check if the state is already in the vec before adding it or otherwise some variant could be inserted multiple times.

src/wayland/shell/xdg/mod.rs
vberger: The impl would not need to be inside the compositor mod, you can add methods from here, unless that has recently been changed in rustc?

cmeissl: Right, sorry for the fuzz, i was mistakenly thinking the token was a trait 😅

src/wayland/shell/xdg/mod.rs
vberger: I have a general intuition that using a hashset for such a small collection (there are rarely more than 3 or 4 states) would induce more overhead than gains, but I admit I may be wrong on that. Do you have a general idea of how it would compare?

If the cost is negligible, I'm ok with using a HashSet.

src/wayland/shell/xdg/mod.rs
vberger: Right, the protocol specifies the order of destruction. Okay then. :+1:

anvil/src/drawing.rs
```diff
@@ -178,6 +178,78 @@ where
...
 
+fn draw_popup<R, E, F, T>(
+    renderer: &mut R,
+    frame: &mut F,
+    egl_buffer_reader: Option<&EGLBufferReader>,
+    compositor_token: MyCompositorToken,
+    surface: &wl_surface::WlSurface,
+    log: &Logger,
+) -> Result<(), SwapBuffersError>
+where
```
vberger, line 189: This function does not handle potential subsurfaces of popup surfaces. Is there a specific reason for not using `draw_surface_tree` for drawing the popups as well?

src/wayland/shell/xdg/mod.rs
cmeissl: I have removed the trait and implemented it directly.

anvil/src/drawing.rs
cmeissl: Popup drawing has been removed from the PR as discussed on matrix. I will try to implement drawing in a separete PR.

src/wayland/shell/xdg/mod.rs
cmeissl: Like discussed on matrix i have added a separate container for handling the states.

src/wayland/shell/xdg/mod.rs
cmeissl: I added the relevant parts of the protocol specification to the function.

src/wayland/shell/xdg/mod.rs
```diff
@@ -175,88 +463,254 @@ impl PositionerState {
...
-    /// be placed relative to its parent.
-    pub positioner: PositionerState,
+/// Container holding the states for a `XdgToplevel`
+///
+/// This container will prevent the `XdgToplevel` from
+/// having the same `xdg_toplevel::State` multiple times
+/// and simplifies setting and un-setting a particularly
+/// `xdg_toplevel::State`
+#[derive(Debug, Clone, PartialEq)]
+pub struct ToplevelStates {
```
vberger, line 632: Maybe we can find a better name for this, it's very easy to confuse with `ToplevelState`... Maybe something like `StateSet` ?

src/wayland/shell/mod.rs
```diff
@@ -14,5 +14,29 @@
...
+use super::Serial;
+use thiserror::Error;
+
 pub mod legacy;
 pub mod xdg;
+
+/// Represents the possible errors returned from
+/// a surface operation
+#[derive(Debug, Error)]
+pub enum SurfaceError {
```
vberger, line 26: Given this Error can only have a single variant, I'm not convinced of the need to make it an enum, rather than an unit struct. Furthermore, it may be worth making a generic error `DeadResource` alongside `utils::UnmanagedResource` and use that, rather than create a specific error for surfaces.

src/wayland/shell/mod.rs
```diff
@@ -14,5 +14,29 @@
...
+    SurfaceNotAlive,
+}
+
+/// Represents the possible errors returned from
+/// a surface ping
+#[derive(Debug, Error)]
+pub enum PingError {
+    /// The operation failed because the underlying surface has an error
+    #[error("the ping failed cause the underlying surface has an error")]
+    SurfaceError(#[from] SurfaceError),
```
vberger, line 38: Given the underlying error has only 1 variant (and is not expected to gain any), this really should just be a `DeadSurface` error, without any nesting.

vberger (changes requested):
That looks pretty good, thanks! :+1: Just a few minor comments.

src/wayland/shell/xdg/mod.rs
cmeissl: Renamed to ToplevelStateSet

src/wayland/shell/mod.rs
cmeissl: I added a DeadResource error to utils and changed the shell to use it.

src/wayland/shell/mod.rs
cmeissl: Removed the nesting, now it is just a DeadResource variant.

vberger (approved):

PR #288, opened 2021-06-05
Split off `import_buffer` of `Renderer` into buffer type depending sub-traits

Drakulix:


src/backend/renderer/mod.rs
```diff
@@ -268,8 +267,61 @@ pub trait Renderer {
...
+        Some(BufferType::Shm)
+    } else {
+        None
+    }
+}
+
+/// Returns the *type* of a wl_buffer
+/// 
+/// Returns `None` if the type is not recognized by smithay or otherwise not supported.
+#[cfg(all(feature = "wayland_frontend", not(feature = "backend_egl"), not(feature = "use_system_lib")))]
```
vberger: The way these cfg block are defined, if we have the `use_system_lib` feature enabled but not the `backend_egl`, this function does not exist.

src/backend/renderer/mod.rs
```diff
@@ -289,7 +341,9 @@ pub fn buffer_dimensions(
 }
 
 /// Returns the dimensions of a wl_buffer
-#[cfg(all(feature = "wayland_frontend", not(feature = "backend_egl")))]
+///
+/// *Note*: This will only return dimensions for buffer types known to smithay (see [`buffer_type`])
+#[cfg(all(feature = "wayland_frontend", not(feature = "backend_egl"), not(feature = "use_system_lib")))]
```
vberger: Same for this function, it is undefined if we have `use_system_lib` but not `backend_egl`.

PR #289, opened 2021-06-07
xwayland: rework as an EventSource

### LINKED ISSUE #203: Rework the internal structure of Smithay

### LINKED ISSUE #245: Xwayland API is fundamentally broken
Sorry for the title. Feel free to edit this. ;-)

I've been trying to get `Xwayland` to work. The latest issue is this one (after `Xwayland` started successfully):
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', [snip]/smithay/src/xwayland/xserver.rs:182:48
```
That line number points here:
https://github.com/Smithay/smithay/blob/36e11284c27c452557f8eb7799060c0b8e6a3ea7/src/xwayland/xserver.rs#L177-L179

I certainly do not understand this, but Levanx seems to do:
```
<Levanx> the `client_destroy::<WM, T>(&e, data.get().unwrap())` is a deeper problem though
<Levanx> I missed an important assumption when I wrote this code, and the automatic-restart of XWayland just cannot work like this
<Levanx> We'll need to change the API, probably by making the compositor process restart xwayland, rather than having it restart automatically
```

vberger:
Reorganize the XWayland abstraction into a calloop EventLoop.

Fixes #245
Fixes #203


src/xwayland/xserver.rs
```diff
@@ -65,167 +66,225 @@ use wayland_server::{Client, Display, Filter};
...
+    /// Attempt to start the XWayland instance
+    ///
+    /// If it succeeds, you'll eventually receive an `XWaylandEvent::Ready`
+    /// through the source provided by `XWayland::new()` containing an
+    /// `UnixStream` representing your WM connection to XWayland, and the
+    /// wayland `Client` for XWayland.
+    ///
+    /// Does nothing if XWayland is already started or starting.
+    ///
+    /// You need to provide this function the same `DispatchData` you pass to the main
```
Drakulix, line 125: Where do I provide this function the `DispatchData`?

Drakulix (approved):
Noticed a weird doc comment, but otherwise this looks good! :+1:

src/xwayland/xserver.rs
vberger: Ah indeed, thanks, this is a leftover from an intermediate state of the refactor. :sweat_smile:

PR #292, opened 2021-06-11
Initial libseat support

PolyMeilex:
Implements initial [libseat](https://sr.ht/~kennylevinsen/seatd/) support, as of now it is possible to launch and use anvil with it.
To test with anvil, simply build anvil with `libseat` flag instead of `logind` flag.

Notes:
- I'm not super happy with how `src/backend/session/auto.rs` looks now after adding `libseat` to it, `libseat` itself if responsible for choosing correct backend, so it's a little bit counterintuitive to have it inside `AutoSession`. Ideally all of this should be handled by `libseat` but I'm not confident enough yet, to drop all of `Smithay` session code in favor of `libseat`, I think it needs a lot more testing before we can relay on it.
- I'm the owner of [libseat](https://crates.io/crates/libseat/) crate, but I'm open to transfer the ownership if needed, just let me know.


src/backend/session/auto.rs
```diff
@@ -61,11 +66,47 @@ pub enum AutoSessionNotifier {
...
+    /// LibSeat session notifier
+    #[cfg(feature = "backend_session_libseat")]
+    LibSeat(LibSeatSessionNotifier),
 }
 
 impl AutoSession {
     /// Tries to create a new session via the best available interface.
-    #[cfg(feature = "backend_session_logind")]
+    #[cfg(feature = "backend_session_libseat")]
+    pub fn new<L>(logger: L) -> Option<(AutoSession, AutoSessionNotifier)>
```
vberger, line 77: Maybe we can flatten all this into a single method that does early return with internal `#[cfg(..)]` blocks, instead of having 3 different `new()` methods, that would probably be easier to follow and reduce code duplication.

Something like this:

```rust
pub fn new<L>(logger: L) -> Option<(AutoSession, AutoSessionNotifier)> {
    #[cfg(feature = "backend_session_libseat")]
    {
        /* try to init libseat backend, early return on success */
    }
    #[cfg(feature = "backend_session_logind")]
    {
        /* try to init logind backend, early return on success */
    }
    /* try to init direct backend */
}
```

src/backend/session/libseat.rs
```diff
@@ -0,0 +1,261 @@
...
+                let logger = logger.clone();
+                move |seat: &mut libseat::SeatRef| {
+                    debug!(logger, "Disable callback called");
+                    active.store(false, Ordering::SeqCst);
+                    seat.disable().unwrap();
+                    signaler.signal(SessionSignal::PauseSession);
+                }
+            };
+
+            Seat::open(enable, disable)
```
vberger, line 91: This is nitpicking, but would it make sense to have, in the API of your libseat bindings, this init function just take a single closure and pass it an enum argument to distinguish enable/disable events?

That could help with reducing code duplication a little.

src/backend/session/libseat.rs
```diff
@@ -0,0 +1,261 @@
...
+
+            session
+                .seat
+                .borrow_mut()
+                .open_device(&path)
+                .map(|(id, fd)| {
+                    session.devices.borrow_mut().insert(fd, id);
+                    fd
+                })
+                .map_err(|_| Error::Unknown)
```
vberger, line 135: I think we'd prefer to forward the error, rather than completely drop it into an `Unknown` enum variant.

vberger (changes requested):
Thanks a lot for this implementation, this looks pretty good! :+1:

I think it's fine to have libseat as an alternate session backend at the moment, for experimentation and testing, and once it has been battle-tested a little and is sufficiently packaged in distributions, we can just drop the whole `AutoSession` thing and only use libseat.

Regarding ownership of the libseat crate, I'd say it depends on what your plans are. If you want to host it under the Smithay organization, that's good with us. If you plan to maintain it yourself in the long run, that's good as well.

Last question, libseat exposes a `libseat_set_log_handler` function. Do you think it'd be feasible to use it to redirect its logging output into rust's `log`, so that it can be handled in that framework?

PolyMeilex:
I was playing around with `libseat_set_log_handler` yesterday, because I also wanted to use typical rust logging with libseat. But I gave up, maybe it's because of my lack of `c` knowledge. 

Log handler function accepts format string, and args in `va_list`, in `c` printing this is as simple as writing:
```c
vfprintf(stderr, format, args);
```
but in rust we don't have `va_list` and we don't have c-like string formatting, so I got stuck with logs that look like this:
`[%s:%d] Could not connect to socket %s: No such file or directory`
I wanted to implement my own formatter in rust, but I have no idea how to read a string pointer from `va_list`
there is [va_list](https://docs.rs/va_list/0.1.3/va_list/trait.VaPrimitive.html) crate that can easily read numbers from `va_list` but still I have no idea what to do about string pointers 

My best bet for implementing this is using `cstdio` functions in rust to format a string, and then convert it to rust string afterwards. Do you have some better ideas?

src/backend/session/libseat.rs
```diff
@@ -0,0 +1,261 @@
...
+            };
+            let disable = {
+                let active = active.clone();
+                let signaler = signaler.clone();
+                let logger = logger.clone();
+                move |seat: &mut libseat::SeatRef| {
+                    debug!(logger, "Disable callback called");
+                    active.store(false, Ordering::SeqCst);
+                    seat.disable().unwrap();
+                    signaler.signal(SessionSignal::PauseSession);
```
Drakulix, line 87: I *think* we have some cleanup functions in the drm-backend, that *try* to run on `PauseSignal` if the file descriptors are still available.

So while this works, we think we should emit the signal first and then disable the seat.

Drakulix (commented):
Really nice!

I am also in favor of removing `AutoSession` eventually, but for now, this is good enough imo.

Drakulix:
> My best bet for implementing this is using `cstdio` functions in rust to format a string, and then convert it to rust string afterwards. Do you have some better ideas?

Yeah, unfortunately that sounds like the best solution. It might be even easier to write a little c-shim to link with the rust code, that does the c-formatting and then returns a proper string.

src/backend/session/auto.rs
PolyMeilex: https://github.com/Smithay/smithay/pull/292/commits/4832951d0e1ed4900ea7d53c0d8af97839501884
Indeed, looks a lot cleaner.

src/backend/session/libseat.rs
PolyMeilex: Sure, I don't like it either, so I'm happy to change that. I will work on that tomorrow

src/backend/session/libseat.rs
PolyMeilex: https://github.com/Smithay/smithay/pull/292/commits/ace7a27d908b7ac7c21113ba2d60b08cdcb1dbe7
Now all methods have custom error types and are forwarding errno values properly.

anvil/build.rs
```diff
@@ -1,7 +1,7 @@
 use std::env::var;
 
 fn main() {
-    if var("CARGO_FEATURE_LOGIND").ok().is_none() {
+    if var("CARGO_FEATURE_LOGIND").ok().is_none() && var("CARGO_FEATURE_LIBSEAT").ok().is_none() {
         println!("cargo:warning=You are compiling anvil without logind support.");
```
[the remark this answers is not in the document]
PolyMeilex: Addresed in https://github.com/Smithay/smithay/pull/292/commits/6acf5b8e90efd2fe35641d5b2fcd982c03e10eab

src/backend/session/libseat.rs
PolyMeilex: Addressed in https://github.com/Smithay/smithay/pull/292/commits/395ef0f8920e84d146d3cfa6e0fd6030641a7231

src/backend/session/libseat.rs
PolyMeilex: Done https://github.com/Smithay/smithay/pull/292/commits/7b746483e43e6ae859449a4d1ae70a369f824036

PolyMeilex:
`Seat::open()` now accepts `slog::Logger` as a second argument, and all `libseat` events are redirected to it

Drakulix (approved):
:+1:

src/backend/session/libseat.rs
```diff
@@ -0,0 +1,279 @@
...
+    #[error("Failed to change vt: {0}")]
+    FailedToChangeVt(Errno),
+
+    /// Session is already closed,
+    #[error("Session is already closed")]
+    SessionLost,
+
+    /// Unknown
+    #[error("Unknown")]
+    Unknown,
```
vberger, line 272: It looks like this variant is no longer used, so it may as well be removed.

src/backend/session/libseat.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+impl AsErrno for Error {
+    fn as_errno(&self) -> Option<i32> {
+        None
+    }
+}
```
vberger, line 279: What is this impl used for?

vberger (changes requested):
Thanks a lot, that is very clean!

I just have two minor nits, and then we can merge this. :+1:

src/backend/session/libseat.rs
PolyMeilex: As far as I know, it is not used anywhere, I simply saw that dbus, direct and auto backends all have one, so I added one here as well, I just forgot to return proper errnos with it.

It's worth noting that both dbus and auto always return None when `as_errno` is called, so auto will never forward libseat errnos to caller

Drakulix: I think that was previously used to convert most session errors into `std::io::Error`s or `nix::Error`s. This got possibly removed when we switched from `error-chain` to `thiserror`.

vberger (approved):

vberger:
Thanks a lot @PolyMeilex !

PR #294, opened 2021-06-17
Fix Anvil panic when resizing dead window

PolyMeilex:
When Wayland client crashes during a resize, the whole compositor goes down with it. It was quite unexpected behavior when I first witnessed it. 

The fix is not ideal, but it's the best I could come up with, with my current knowledge of Anvil.

### DIFF
--- a/anvil/src/shell.rs
+++ b/anvil/src/shell.rs
@@ -162,12 +162,18 @@ struct ResizeSurfaceGrab {
 impl PointerGrab for ResizeSurfaceGrab {
     fn motion(
         &mut self,
-        _handle: &mut PointerInnerHandle<'_>,
+        handle: &mut PointerInnerHandle<'_>,
         location: (f64, f64),
         _focus: Option<(wl_surface::WlSurface, (f64, f64))>,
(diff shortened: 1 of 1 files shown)


anvil/src/shell.rs
```diff
@@ -168,6 +168,11 @@ impl PointerGrab for ResizeSurfaceGrab {
         _serial: Serial,
         _time: u32,
     ) {
+        // It is impossible to get `min_size` and `max_size` of dead toplevel, so we return early.
+        if !self.toplevel.alive() | self.toplevel.get_surface().is_none() {
+            return;
```
vberger, line 173: I think we'd need to unset the the grab as well as doing an early return. The client exiting should void any pending grab.

PolyMeilex: Indeed, my bad.

vberger:
I'm not quite sure this is a good long-term solution, probably the grab mechanism in Smithay should be adjusted to automatically detect and drop the grab when the client exits.

I'll open an issue to track this, but in the meantime your change will do, thanks!

PR #296, opened 2021-06-18
Rework DrmRenderSurface into GbmBufferedSurface

Drakulix:
Removes the renderer from the `DrmRenderSurface` allowing anvil to use just one renderer per backend.
Will allow the `ImportEGL`-trait to hold the `EGLBufferReader` (to be done in a future PR).

Since the old `DrmRenderSurface` was dependant on gbm anyway to import buffers, the new `GbmBufferedSurface` does now only supports gbm as an allocator.

Thus it was renamed. This gets rid of tons of generics, which were not very useful to begin with. There are no useful other allocators that can be imported into gbm (only DumbBuffers might work here, but this still can be implemented, just not with this surface). This also allows us to skip our weird gbm->dmabuf->gbm export-import cycle. The buffer is now only exported once. (cc @cmeissl, if I remember correctly, this helps your compositor.)

The only short-coming is, that this exposes the weirdness of the buffer-scanout coordinate system directly to the user, who has to set Transform::Flipped180 to get a correct image. There is nothing we can do about this to "fix" this on the drm side and I was not able to come up with a satisfying work-around that hides this detail.

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -12,19 +12,15 @@ use std::{
 use image::{ImageBuffer, Rgba};
 use slog::Logger;
 
-#[cfg(feature = "egl")]
-use smithay::{
-    backend::{drm::DevPath, egl::display::EGLBufferReader, renderer::ImportDma, udev::primary_gpu},
-    wayland::dmabuf::init_dmabuf_global,
-};
 use smithay::{
     backend::{
-        drm::{DrmDevice, DrmError, DrmEvent, DrmRenderSurface},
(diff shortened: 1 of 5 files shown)


cmeissl:
Yes, this should definitely make it easier for my use-case, thanks 😁. In fact with all the last PRs merged and this one anvil should now run out of the box on aarch64 based i.MX8M. Will check on monday!

cmeissl:
So, I forgot about two other issues that prevent anvil from launching, but the buffer handling now works out of the box :grinning:

vberger (approved):

PR #298, opened 2021-06-19
track suppressed keys

### LINKED ISSUE #242: [Anvil] Locking / Repeating input when keys pressed
I tested this on weston terminal and alacritty, with same result.

## Issue

Press modkey + '1' or '1' + modkey and it starts to keep repeating '1'. It behaves like '1' was never released.

## Environment

- tty has issue
- winit not tested yet

cmeissl:
This PR adds tracking for suppressed keys in anvil. 
A key will be suppressed when a key press triggers a action.
On release the key will be forwarded to the client depending on if
the key has been suppressed or not.

fixes #242


vberger:
Simple and clean, thanks!

PR #304, opened 2021-06-23
compositor: Introduce transaction state tracking

vberger:
This changes the state handling logic of wl_surface to automatically
track subsurface pending in the form of transactions. The role enum
(and the associated type parameter) is no more, and replaced by a
general-purpose typemap-like container on surfaces.

The new logic is introduced in the files:

- `src/wayland/compositor/cache.rs`
- `src/wayland/compositor/transaction.rs`

The rest of the PR is the fallout of these changes, as well as a few
trivial clippy fixes.


src/backend/allocator/dumb.rs
```diff
@@ -39,8 +39,7 @@ impl<A: AsRawFd + 'static> Allocator<DumbBuffer<A>> for DrmDevice<A> {
         // dumb buffers are always linear
         if modifiers
             .iter()
-            .find(|x| **x == Modifier::Invalid || **x == Modifier::Linear)
-            .is_none()
+            .all(|&x| x != Modifier::Invalid && x != Modifier::Linear)
```
Drakulix, line 42: I do not think this correct. What we want is
```suggestion
            .any(|&x| x == Modifier::Invalid || x == Modifier::Linear)
```

Drakulix (commented):
Alright, I tried to give this a full review, but I can hardly say, that I understand anything that is going on.

I have no found any obvious errors, the docs make sense, and I think I have identified one mistake in your clippy fixes. But other then that, I am not sure, if I can be of much help. :sweat_smile:

src/backend/allocator/dumb.rs
vberger: I think it is correct? The original `if` triggers if all entries fail the `find` check (causing the return value to be `None`), and an entry will fail the check if it is neither `Invalid` nor `Linear`. You proposal woul cause the `if` to trigger if at least one entry is `Invalid` or `Linear`.

Drakulix: Well maybe my initial implementation was already wrong then.

DumbBuffers are always going to be linear, so the user needs to be able to work with either Linear or Invalid.

vberger: I'm not sure I follow you. The original code raises an error if the modifier list does not contain `Invalid` nor `Linear`. So was it the expected behavior or should it be changed?

Drakulix: > The original code raises an error if the modifier list does not contain Invalid nor Linear.

That is expected behavior. The list needs to container either of them.

But I got confused, because we are checking here for the reverse of that to return an error, I was working with the wrong assumption on that `if`.

src/wayland/compositor/transaction.rs
```diff
@@ -0,0 +1,248 @@
...
+impl Transaction {
+    pub(crate) fn state(&self) -> BlockerState {
+        use BlockerState::*;
+        self.blockers
+            .iter()
+            .fold(Released, |acc, blocker| match (acc, blocker.state()) {
+                (Cancelled, _) | (_, Cancelled) => Cancelled,
+                (Pending, _) | (_, Pending) => Pending,
+                (Released, Released) => Released,
+            })
```
Drakulix, line 163: Couldn't this be a `find` call? Looks cleaner and easier to understand for me.

```suggestion
            .map(|blocker| blocker.state())
            .find(|state| state == Cancelled || state == Pending)
            .unwrap_or(Released)
```

src/wayland/compositor/transaction.rs
```diff
@@ -0,0 +1,248 @@
...
+                    if place.1 < id {
+                        place.1 = id;
+                    }
+                } else {
+                    // the surface is not in the list, insert it
+                    state.surfaces.push((s, id));
+                }
+            }
+            state.blockers.extend(my_state.blockers);
+        });
```
Drakulix, line 130: These 10 lines or so look like a complete copy of `insert_state`.
I think it would be cleaner to just call that and then do state.blockers, no?

Or may this be such a hot path, that we want to avoid the frequent locking, or can this happen in a multi-threaded contexts and needs to happen atomically?

```suggestion
        for (s, id) in my_state.surfaces {
            self.insert_state(s, id);
        }
        self.with_inner_state(|state| {
            state.blockers.extend(my_state.blockers);
        });
```

Drakulix (approved):
Alright, two comments on stuff, I found to be difficult to understand *and* possibly easier to express. But again I have not really found anything substantial, this seems to be really solid. :+1:

src/wayland/compositor/transaction.rs
vberger: That'd make the logic slightly different. Your proposal would early return on the first `Cancelled` or `Pending`, while if the list contains both a `Pending` and a `Cancelled`, my code will always return `Cancelled`.

src/wayland/compositor/transaction.rs
vberger: Mostly I found it wasteful to lock and unlock the state in a loop.

src/wayland/compositor/transaction.rs
Drakulix: Ah right, but then this needs a comment. I would not have noticed this detail without you pointing it out.

src/wayland/compositor/transaction.rs
vberger: Okay, I re-organized this with an `insert()` method on `TransactionState`. Is that better?

vberger:
So, I added on the top of `cache.rs` and `transaction.rs` a general overview explanation of what the logic of these files do. How is it?

src/wayland/compositor/transaction.rs
Drakulix: Nice solution! :+1:

Drakulix:
> So, I added on the top of `cache.rs` and `transaction.rs` a general overview explanation of what the logic of these files do. How is it?

Clear and relatively easy understandable given the complex problem/logic at hand. I think this is good enough for merge now :)

PR #305, opened 2021-06-23
Large cleanup of the codebase

vberger:
Based on top of #304

This PR does a general cleanup of the Smithay codebase, for stuff that clippy & rustfmt don't handle by themselves

- [x] Ensure that all combinations of cargo features compile without warning, to unclutter the CI output from all these annoying annotations
- [ ] ~~Reorganize the imports in each file to follow a coherence scheme, in the order : 1) imports from std, 2) imports from external crates, 3) `crate::` imports, 4) `super::` imports, 5) `self::` imports~~ There is actually a (currently unstable) rustfmt option that will handle it, so lets not do this by hand and suffer. Postponing that for later.
- [x] Fix the last remaining clippy lints

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -28,6 +28,8 @@ use winit::{
     window::{Window as WinitWindow, WindowBuilder},
 };
 
+use slog::{debug, error, info, o, trace, warn};
+
 /// Errors thrown by the `winit` backends
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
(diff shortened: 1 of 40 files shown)


vberger:
The two last clippy fixes will be solved once we implement popup support in anvil, so I won't consider them blocking this PR.

src/backend/winit.rs
```diff
@@ -28,6 +28,13 @@ use winit::{
     window::{Window as WinitWindow, WindowBuilder},
 };
 
+#[cfg(feature = "use_system_lib")]
+use crate::backend::egl::display::EGLBufferReader;
+#[cfg(feature = "use_system_lib")]
+use wayland_server::Display;
```
Drakulix, line 34: Both these imports are actually not necessary anymore after #309

Drakulix (approved):
:+1:

PR #306, opened 2021-06-24
feature/output-enter-leave

cmeissl:
This PR adds an output map, similar to the window map, that keeps track of the known outputs and
their location.

The output map additionally keeps track of the surfaces shown on a specific output and sends `wl_surface.enter`
and `wl_surface.leave` accordingly. This is neseccary in prepartion for handling HiDPI as the client will
use this information for deciding which scale it should use. 

The shell implementation now uses the output geometry for initial window placement and
maximize/fullscreen handling.

The name of the output is expected to be unique and again this can be used for setting scale factors on specific
outputs.

The winit backend creates a virtual output that gets updated when the window is resized.

Additionally the output map will re-configure maximized/fullscreen windows when the mode changes.
When a output is removed the output map will try to move the now off-screen windows to a different
active output.

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -265,11 +265,18 @@ impl AnvilState<UdevData> {
         }
 
         let (pos_x, pos_y) = pos;
-        let (max_x, max_y) = self.output_map.borrow().size();
+        let output_map = self.output_map.borrow();
+        let max_x = output_map.width();
         let clamped_x = pos_x.max(0.0).min(max_x as f64);
-        let clamped_y = pos_y.max(0.0).min(max_y as f64);
+        let max_y = output_map.height(clamped_x as i32);
 
-        (clamped_x, clamped_y)
+        if let Some(max_y) = max_y {
+            let clamped_y = pos_y.max(0.0).min(max_y as f64);
+
+            (clamped_x, clamped_y)
+        } else {
+            (clamped_x, pos_y)
+        }
     }
 }
 
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -13,6 +13,7 @@ use smithay::{
     },
     reexports::wayland_server::protocol::wl_pointer,
     wayland::{
+        output::Mode,
         seat::{keysyms as xkb, AxisFrame, Keysym, ModifiersState},
         SERIAL_COUNTER as SCOUNTER,
     },
@@ -137,7 +138,12 @@ impl<Backend> AnvilState<Backend> {
 
 #[cfg(feature = "winit")]
 impl AnvilState<WinitData> {
-    pub fn process_input_event<B: InputBackend>(&mut self, event: InputEvent<B>) {
+    pub fn process_input_event<B>(&mut self, event: InputEvent<B>)
+    where
+        B: InputBackend<SpecialEvent = smithay::backend::winit::WinitEvent>,
+    {
+        use smithay::backend::winit::WinitEvent;
+
         match event {
             InputEvent::Keyboard { event, .. } => match self.keyboard_key_to_action::<B>(event) {
                 KeyAction::None | KeyAction::Forward => {}
@@ -162,6 +168,16 @@ impl AnvilState<WinitData> {
             InputEvent::PointerMotionAbsolute { event, .. } => self.on_pointer_move_absolute::<B>(event),
             InputEvent::PointerButton { event, .. } => self.on_pointer_button::<B>(event),
             InputEvent::PointerAxis { event, .. } => self.on_pointer_axis::<B>(event),
+            InputEvent::Special(WinitEvent::Resized { size, .. }) => {
+                self.output_map.borrow_mut().update_mode(
+                    crate::winit::OUTPUT_NAME,
+                    Mode {
+                        width: size.0 as i32,
+                        height: size.1 as i32,
+                        refresh: 60_000,
+                    },
+                );
+            }
             _ => {
                 // other events are not handled in anvil (yet)
             }
@@ -206,17 +222,16 @@ impl AnvilState<UdevData> {
                     }
                 }
                 KeyAction::Screen(num) => {
-                    if let Some(output) = self.backend_data.output_map.get(num) {
-                        let x = self
-                            .backend_data
-                            .output_map
-                            .iter()
-                            .take(num)
-                            .fold(0, |acc, output| acc + output.size.0)
-                            as f64
-                            + (output.size.0 as f64 / 2.0);
-                        let y = output.size.1 as f64 / 2.0;
-                        self.pointer_location = (x as f64, y as f64)
+                    let geometry = self
+                        .output_map
+                        .borrow()
+                        .find_by_index(num, |_, geometry| geometry)
+                        .ok();
+
+                    if let Some(geometry) = geometry {
+                        let x = geometry.x as f64 + geometry.width as f64 / 2.0;
+                        let y = geometry.height as f64 / 2.0;
+                        self.pointer_location = (x, y)
                     }
                 }
             },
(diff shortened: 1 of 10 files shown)


src/wayland/output/mod.rs
```diff
@@ -312,4 +318,34 @@ impl Output {
             .iter()
             .any(|o| o.as_ref().equals(output.as_ref()))
     }
+
+    /// This function allows to run a `Fn` for the matching
+    /// client outputs for a specific `Resource`.
+    pub fn with_output<F, R, II>(&self, resource: &R, mut f: F)
```
vberger, line 324: If the goal is that this closure is invoked for all `wl_output` instances of a given client, why not directly take the `Client` as argument? The user can access the resource in the closure by capturing it. That would make the behavior clearer in my opinion.

vberger (approved):
One small nite but overall this looks like a pretty solid basis on which build upon, thanks a lot! :+1:

I'll leave to @Drakulix to check the backend / udev part of the PR, which I'm not confident in judging. The rest seems good.

anvil/src/input_handler.rs
```diff
@@ -242,47 +257,16 @@ impl AnvilState<UdevData> {
...
-            // find the first one with a greater x
-            .find(|(_idx, x_pos)| *x_pos as f64 > x)
-            // the previous output is the one we are on
-            .map(|(idx, _)| idx - 1)
-            .unwrap_or(self.backend_data.output_map.len() - 1)
-    }
+        let (pos_x, pos_y) = pos;
+        let (max_x, max_y) = self.output_map.borrow().size();
+        let clamped_x = pos_x.max(0.0).min(max_x as f64);
+        let clamped_y = pos_y.max(0.0).min(max_y as f64);
```
Drakulix, line 267: Isn't this a little too simplistic? Like this the cursor can get stuck in an invisible area between two very different monitors (e.g. a horizontal and a vertical one), no?

anvil/src/shell.rs
```diff
@@ -526,24 +629,101 @@ pub fn init_shell<BackendData: 'static>(display: &mut Display, log: ::slog::Logg
...
+                                .borrow()
+                                .with_primary(|output, geometry| {
+                                    let mut primary_output: Option<wl_output::WlOutput> = None;
+                                    output.with_output(wl_surface, |_, output| {
+                                        primary_output = Some(output.to_owned());
+                                    });
+                                    (primary_output, geometry)
+                                })
+                                .ok()
+                        });
```
Drakulix, line 719: Can we de-duplicate at least this interesting construct and extract that into a helper function please? :smile:

anvil/src/udev.rs
```diff
@@ -76,8 +75,14 @@ impl AsRawFd for SessionFd {
     }
 }
 
+struct UdevOutputMap {
+    pub device_id: dev_t,
+    pub crtc: crtc::Handle,
+    pub output_name: String,
+}
```
Drakulix, line 82: I get the feeling, that we might want to add userdata to the output-map at some point. But that exercise can be left to be done later, this is sufficient.

Drakulix (commented):
Some smaller nitpicks, but overall I agree with @vberger here. Nice work! :grin:

src/wayland/output/mod.rs
cmeissl: What do you think about having two functions, one that takes a `Client` and a `FnMut` which will invoke the callback for each output. And a second one to get the first matching output for a resource with the same client.
The second function is more for convenience, like to get a output for a surface when the fullscreen request did not specify an output explicit. While it is possible to use the first function for this use-case i find it really ugly. 

Like:

```
.with_primary(|output, geometry| {
    let primary_output = wl_surface.as_ref().client().and_then(|client| {
        let mut primary_output: Option<wl_output::WlOutput> = None;
        output.with_client_outputs(client, |output| {
            primary_output = Some(output.to_owned());
        });
        primary_output
    });

    (primary_output, geometry)
})
```

vs

```
.with_primary(|output, geometry| {
    let primary_output = output.find(wl_surface);
    (primary_output, geometry)
})
```

anvil/src/shell.rs
cmeissl: Yes, just forgot to move it into a function, good catch.

anvil/src/udev.rs
cmeissl: Yes, i considered this, but wanted a simple solution for now cause i expect that the output will change a bit when implementing HiDPI. Maybe i will add a function to get the (currently) private `Output` struct in `window_map`, then
it would be easy to access a `UserDataMap` from the `Output`.

anvil/src/input_handler.rs
cmeissl: Should not be the case, at the moment the overall size calculated by the output_map is the sum of all widths and the maximum height of the outputs. This is similiar to what was implemented before. I agree that is will get more complicated when the outputs can be placed differently. I left a note in the `output_map` that the current implementation is really simple.

Drakulix: So if two outputs have very different heights, what happens if I move the cursor one the smaller screen below its height? Should still be inside the max-height, right?

cmeissl: Ah, now i know what you mean, sorry for the misunderstanding. I thought you mean it will get stuck like unable to move. Yes, with this simple implementation the cursor could be outside of a screen output if multiple monitors with different resolutions are used. This was intentional, but if it is an issue i can implement output specific clamping. Personally i hate it when i am not able to move the cursor from one screen to another cause their heights are different and I have to move the cursor up before. An alternative could be  to set the cursor to the max height of the next output when moving on the x-axis. Any preference?

src/wayland/output/mod.rs
vberger: I think there is a misunderstanding here, because retrieving a single `WlOutput` from smithay's `Output` abstraction is never the correct thing to do.

If there are several `WlOutput` for a single client, that means that client has bound the global sereval times, and in that case all relevant events should be send to all of them. For example if a client has bound a `WlOutput` two times, it should receive the `WlSurface.enter()` event twice, once for each of its `WlOutput` objects that correspond to the same global.

That's what I was interpreting this function to be for: do some processing (like sending an event) to all duplicated relevant output instances. So I don't get what your `primary_output` example is about, a primary output is a notion that exists at the logical output level (and arguably does not really exists in Wayland),but it has very little to do with the `WlOutput` objects...

cmeissl: Yeah, definitely a misunderstanding, it is clear to me that an event has to be sent to all bound outputs and I do not argue against it. I see that adding a function to get a single output is wrong in most cases and such things are better prevented at the API level :+1: 

The reason for getting a `WlOutput` is that a client is free to not specify a `WlOutput` when requesting `Fullscreen`. In that
case the compositor should decide if and when yes on which output the window will be rendered fullscreen. As the output map is currently in anvil the only information I found that could be set in the xdg state to remember is a `WlOutput`.
But of course there are other solutions, one thing the comes to my mind is to just leave `None` in the xdg state for the choosen output and remember the output name within anvil and only use the geometry for configure.

For now I will just remove this part of the logic as anvil does not handle fullscreen rendering, so maybe it is better to think about it when implementing it.

anvil/src/input_handler.rs
cmeissl: Ok, I think now the logic should be how it was before I changed it and the height is clamped with the current output height of the cursor.

anvil/src/shell.rs
cmeissl: I moved the code to a separate function within the shell and left a few comments what it does.

src/wayland/output/mod.rs
cmeissl: I renamed the function and changed it to take a client, the provided function is called for each client output.
As written in the previous comment is also removed the strange output behavior. I left the re-configure for fullscreen
by using the output geometry. Everything else for fullscreen handling is a topic for a seperate PR :)

Drakulix (approved):

anvil/src/input_handler.rs
cmeissl: Fixed the wrong clamping, but only tested on a two monitor setup. But with two different resolutions.

vberger (approved):

PR #308, opened 2021-06-27
[Anvil] XDG activate support

### LINKED ISSUE #159: Anvil: implement activated state for xdg toplevels
So that applications (notably GTK ones) look like they have focus when they have focus.

PolyMeilex:
Closes #159

### DIFF
--- a/anvil/src/window_map.rs
+++ b/anvil/src/window_map.rs
@@ -1,7 +1,7 @@
 use std::cell::RefCell;
 
 use smithay::{
-    reexports::wayland_server::protocol::wl_surface,
+    reexports::{wayland_protocols::xdg_shell::server::xdg_toplevel, wayland_server::protocol::wl_surface},
     utils::Rectangle,
     wayland::{
         compositor::{with_states, with_surface_tree_downward, SubsurfaceCachedState, TraversalAction},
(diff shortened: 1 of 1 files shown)


anvil/src/window_map.rs
```diff
@@ -252,6 +271,15 @@ impl WindowMap {
...
             let winner = self.windows.remove(i);
+
+            // Take activation away from all the windows
+            for window in self.windows.iter() {
+                window.toplevel.set_activated(false);
+            }
+
+            // Give activation to our winner
+            winner.toplevel.set_activated(true);
+
```
vberger, line 282: Is it really necessary to remove the window (and cause a reorganization of the `Vec`) and insert it back? I think we can just check in the for loop that the current window is not the winner before disabling it.

PolyMeilex: It was not written by me, it is part of bringing surface to top. I just added the activation part in between.
https://github.com/Smithay/smithay/blob/da937d498e33c991d1eebf0de2e96f30fbe90b73/anvil/src/window_map.rs#L255

vberger: Oh, right, bringing it to the top while preserving the ordering of the rest. I forgot about that part, my bad!

vberger (approved):

PR #310, opened 2021-06-27
[WIP] new anvil backend: "raw"

Drakulix:
This adds a new backend named "raw" (for now) to anvil, which is useful for debugging purposes.
As such it is not added to anvil's list of backends (although I am open to change that).

Currently this is just a copy of `anvil/src/udev.rs` without the session, udev or input code. Instead this initializes on a single drm device given via command line. Its main usage is to enable easier debugging on multi-gpu systems, where one gpu can be used to drive the development environment, while anvil is running on the second gpu. This can be far easier to work with for example, when debugging via gdb, instead of having to deal with multiple sessions or debugging via gdbserver from another system.

In the future I could see myself implementing a window on the "client side" to forward inputs easily and possibly even extract the image via dmabufs into the main compositor, but so far this only renders (clients can still be spawned by setting `WAYLAND_DISPLAY` accordingly).

While this looks like a niche use-case, this little hack has proven to be insanely useful on my setup, as I can use the nvidia gpu of my laptop with nouveau for debugging and displaying anvil running through my monitors picture-in-picture mode inside my dev environment. This PR is primarily intended for discussion, if this has any place inside anvil (given anvil is mostly a show-off of smithay). If not, I will likely continue to maintain this on a separate branch.

Note that this is not ready for merge. While parts of the code are duplicated between the udev and this new backend, I do not think everything can easily be shared. But I would like to share at least the `scan_connectors` and `render/render_surface/initial_render` functions, which are (mostly) identical.

Also note, due to a mistake on my end an old version of this is already in the master branch... I got there due to a wildcard add on a formatting commit... sorry

### DIFF
--- a/anvil/src/main.rs
+++ b/anvil/src/main.rs
@@ -14,6 +14,8 @@ mod shell;
 mod state;
 #[cfg(feature = "udev")]
 mod udev;
+#[cfg(feature = "udev")]
+mod raw;
 mod window_map;
 #[cfg(feature = "winit")]
 mod winit;
@@ -27,6 +29,7 @@ static POSSIBLE_BACKENDS: &[&str] = &[
     "--winit : Run anvil as a X11 or Wayland client using winit.",
     #[cfg(feature = "udev")]
     "--tty-udev : Run anvil as a tty udev client (requires root if without logind).",
+    // dont advertise the "raw" backend
 ];
 
 fn main() {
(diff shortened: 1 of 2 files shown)


vberger:
My general opinion on this is that this is worth maintaining it in anvil directly, but we should gate it behind a `debugging` cargo feature, or something with a clear name that this is not intended for general use. This should however be built on the CI, to ensure it remains in sync with the rest of smithay and anvil.

PR #311, opened 2021-06-28
allow the direct session tty to be overriden by using a environment variable

### LINKED ISSUE #302: Allow overriding the tty in a direct session for tty-udev
At the moment when using a `DirectSession` the current `tty` will be used.
When running `anvil --tty-udev` over a ssh connection this will fail.
Consider to allow overriding the `tty` with either a command line argument, an environment variable or both.

cmeissl:
This PR adds an environment variable to the direct session backend that allows the tty path to be overriden.

closes #302

### DIFF
--- a/src/backend/session/direct.rs
+++ b/src/backend/session/direct.rs
@@ -11,7 +11,8 @@
 //! ### Initialization
 //!
 //! To initialize the session you may pass the path to any tty device, that shall be used.
-//! If no path is given the tty used to start this compositor (if any) will be used.
+//! If no path is given the tty used to start this compositor (if any) will be used unless
+//! the [SMITHAY_DIRECT_TTY_ENV] environment variable is set.
 //! A new session and its notifier will be returned.
 //!
 //! ```rust,no_run
@@ -59,7 +60,8 @@ use nix::{
 use std::{
     fmt,
     os::unix::io::RawFd,
-    path::Path,
+    path::{Path, PathBuf},
+    str::FromStr,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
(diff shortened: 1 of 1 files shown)


vberger:
I think we should stick to the common pattern use by several other tools (such as libinput) that environment variables are use as a fallback, rather than an override. In that precise case, that would mean that the environment variable would only be read if the `new()` function is passed `None` as a path (note that this would not change the behavior when using `AutoSession`, which I guess is the main point).

cmeissl:
> I think we should stick to the common pattern use by several other tools (such as libinput) that environment variables are use as a fallback, rather than an override. In that precise case, that would mean that the environment variable would only be read if the `new()` function is passed `None` as a path (note that this would not change the behavior when using `AutoSession`, which I guess is the main point).

I implemented it at first with your proposed behavior, but changed it to an override cause this would allow overriding the tty when 
the compositor uses a "wrong" tty which i thought could be handy in some situations. But yeah, most libs use the environment variables only as a fallback. 

Yes, the main reason for the envrionment variable is using anvil for testing and development with `AutoSession`. In a custom compositor this is of course not an issue as i can just specify the path directly. I just like to be able to run anvil for reference and development on a ssh session too :)

Soo, i will update the PR to use the env as an fallback.

cmeissl:
Ok, I changed the logic and updated the doc.

vberger:
> I implemented it at first with your proposed behavior, but changed it to an override cause this would allow overriding the tty when
> the compositor uses a "wrong" tty which i thought could be handy in some situations.

Well, in that case I'd argue that if the compositor actively chooses a TTY (rather than use the default of `AutoSession`) and chooses a wrong one, then it is a bug to be fixed in that compositor.

That said, thanks, LGTM. :+1:

vberger (approved):

Issue #293, opened 2021-06-14
Consider having direct Wayland/X11 backends instead of winit


vberger:
I'm more and more starting to think this would be desirable, for mostly two reasons:

- winit hides a significant amount of low-level details that we would want for proper handling, meaning at this point I'm not 100% sure using winit does actually simply things that much
- winit forces significant dependencies duplications on us

I don't have a strong opinion about that though, given the winit backend is mostly intended for dev & debug, so release builds using smithay might want to have it disabled anyway, but I'm opening this issue for discussion around it.

(It is definitely *not* a goal for the 0.3 release anyway)

Drakulix:
Yeah, I have come to this conclusion as well.

Especially for the wayland-backend we are not only loosing a lot of information, but we need to abstract over winit, which abstracts over wayland-client, although we could pass through some protocols with minimal changes or atleast work with the protocol-native types instead of the winit ones.

For the new rendering api this would also very much be desirable, as we could just attach dmabufs via wayland-client without relying on wayland-egl or the eglsurface api and much more closely mimic the drm-backend.

> I don't have a strong opinion about that though, given the winit backend is mostly intended for dev & debug, so release builds using smithay might want to have it disabled anyway, but I'm opening this issue for discussion around it.

Which is why I would still keep it. For quick testing the backend is totally fine and might even be useful for some experiments like running anvil on windows/macos. But it is not a proper nested compositor experience and should not be advertised as such.

PR #314, opened 2021-06-28
Graphics tablet support

PolyMeilex:
It took me a few rewrites, but I think the API is finally in kinda usable state now, and I think it's better to just open a PR and hear your opinion, instead of rewriting it once again :smile: 

- winit support is not there, because winit does not expose any tablet specific events (it you ever decide to go #293 route, I would be happy to add tablet support to that).
- This PR does not include `tablet_pad_v2` (tablet buttons, etc) support, it is only for tablet tools (like pens).
- Tablet tool cursors are independent of mouse pointers, but anvil treats them as one for now, as I don't think it's important for anvil.
- I'm not sure how to handle `NewDevice`, `RemovedDevice` events in anvil, as those are libinput specific events, for now I hard-coded `InputBackend<SpecialEvent = LibinputEvent>>` into input handler, but that's a little bit weird, and CI is already complaining about that

### DIFF
--- a/src/backend/input/tablet.rs
+++ b/src/backend/input/tablet.rs
@@ -1,4 +1,5 @@
 use super::{ButtonState, Event, InputBackend, UnusedEvent};
+use bitflags::bitflags;
 
 /// Description of physical tablet tool
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
@@ -35,23 +36,24 @@ pub enum TabletToolType {
     Totem,
 }
 
-/// Describes extra capabilities on a tablet.
-///
-/// Any tool must provide x and y values, extra axes are device-specific.
-#[derive(Debug, Clone, Hash, Eq, PartialEq)]
-pub struct TabletToolCapabilitys {
-    /// Tilt axes
-    pub tilt: bool,
-    /// Pressure axis
-    pub pressure: bool,
-    /// Distance axis
-    pub distance: bool,
-    /// Z-rotation axis
-    pub rotation: bool,
-    /// Slider axis
-    pub slider: bool,
-    /// Wheel axis
-    pub wheel: bool,
+bitflags! {
+    /// Describes extra capabilities on a tablet.
+    ///
+    /// Any tool must provide x and y values, extra axes are device-specific.
+    pub struct TabletToolCapabilitys: u32 {
+        /// Tilt axes
+        const TILT = 1;
+        /// Pressure axis
+        const PRESSURE = 2;
+        /// Distance axis
+        const DISTANCE = 4;
+        /// Z-rotation axis
+        const ROTATION = 16;
+        /// Slider axis
+        const SLIDER = 32;
+        /// Wheel axis
+        const WHEEL = 64;
+    }
 }
 
 /// Tablet tool event
@@ -110,7 +112,7 @@ pub trait TabletToolEvent<B: InputBackend> {
     /// Return the current absolute Y coordinate of the tablet tool event, transformed to screen coordinates.
     fn y_transformed(&self, height: u32) -> f64;
 
-    /// Returns the current pressure being applied on the tool in use, normalized to the range [0, 1].
+    /// Returns the current distance from the tablet's sensor, normalized to the range [0, 1]
     ///
     /// If this axis does not exist on the current tool, this function returns 0.
     fn distance(&self) -> f64;
@@ -118,7 +120,7 @@ pub trait TabletToolEvent<B: InputBackend> {
     /// Check if the distance axis was updated in this event.
     fn distance_has_changed(&self) -> bool;
 
-    /// Returns the current distance from the tablet's sensor, normalized to the range [0, 1]
+    /// Returns the current pressure being applied on the tool in use, normalized to the range [0, 1].
     ///
     /// If this axis does not exist on the current tool, this function returns 0.
     fn pressure(&self) -> f64;
@@ -168,7 +170,8 @@ pub trait TabletToolEvent<B: InputBackend> {
 
     /// Returns the current z rotation of the tool in degrees, clockwise from the tool's logical neutral position.
     ///
-    /// For tools of type Mouse and Lens the logical neutral position is pointing to the current logical north of the tablet. For tools of type Brush, the logical neutral position is with the buttons pointing up.
+    /// For tools of type Mouse and Lens the logical neutral position is pointing to the current logical north of the tablet.
+    /// For tools of type Brush, the logical neutral position is with the buttons pointing up.
     ///
     /// If this axis does not exist on the current tool, this function returns 0.
     fn rotation(&self) -> f64;
@@ -264,7 +267,8 @@ impl<B: InputBackend> TabletToolAxisEvent<B> for UnusedEvent {}
 /// detectable distance of the tablet device. A tool that is out of proximity cannot
 /// generate events.
 ///
-/// On some hardware a tool goes out of proximity when it ceases to touch the surface. On /// other hardware, the tool is still detectable within a short distance (a few cm) off
+/// On some hardware a tool goes out of proximity when it ceases to touch the surface. On
+/// other hardware, the tool is still detectable within a short distance (a few cm) off
 /// the surface.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum ProximityState {
@@ -339,7 +343,8 @@ pub trait TabletToolButtonEvent<B: InputBackend>: TabletToolEvent<B> + Event<B> 
     /// Return the button that triggered this event.
     fn button(&self) -> u32;
 
-    /// For the button of a TabletToolButtonEvent, return the total number of buttons pressed on all devices on the associated seat after the the event was triggered.
+    /// For the button of a TabletToolButtonEvent,
+    /// return the total number of buttons pressed on all devices on the associated seat after the the event was triggered.
     fn seat_button_count(&self) -> u32;
 
     /// Return the button state of the event.
--- a/src/backend/input/tablet.rs
+++ b/src/backend/input/tablet.rs
@@ -1,0 +1,361 @@
+use super::{ButtonState, Event, InputBackend, UnusedEvent};
+
+/// Description of physical tablet tool
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub struct TabletToolDescriptor {
+    /// The tool type is the high-level type of the tool and usually decides the interaction expected from this tool.
+    pub tool_type: TabletToolType,
+    /// Unique hardware serial number of the tool
+    pub hardware_serial: u64,
+    /// Hardware id in Wacom’s format
+    pub hardware_id_wacom: u64,
+    /// Tool capability
+    /// Notifies the client of any capabilities of this tool, beyond the main set of x/y axes and tip up/down detection
+    pub capabilitys: TabletToolCapabilitys,
+}
+
+/// Describes the physical type of a tool. The physical type of a tool generally defines its base usage.
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq)]
+pub enum TabletToolType {
+    /// A generic pen.
+    Pen,
+    /// Eraser
+    Eraser,
+    /// A paintbrush-like tool.
+    Brush,
+    /// Physical drawing tool, e.g. Wacom Inking Pen
+    Pencil,
+    /// An airbrush-like tool.
+    Airbrush,
+    /// A mouse bound to the tablet.
+    Mouse,
+    /// A mouse tool with a lens.
+    Lens,
+    /// A rotary device with positional and rotation data.  
+    Totem,
+}
+
+/// Describes extra capabilities on a tablet.
+///
+/// Any tool must provide x and y values, extra axes are device-specific.
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub struct TabletToolCapabilitys {
+    /// Tilt axes
+    pub tilt: bool,
+    /// Pressure axis
+    pub pressure: bool,
+    /// Distance axis
+    pub distance: bool,
+    /// Z-rotation axis
+    pub rotation: bool,
+    /// Slider axis
+    pub slider: bool,
+    /// Wheel axis
+    pub wheel: bool,
+}
+
+/// Tablet tool event
+pub trait TabletToolEvent<B: InputBackend> {
+    /// Get tablet tool that caused this event
+    fn tool(&self) -> TabletToolDescriptor;
+
+    /// Delta between the last and new pointer device position interpreted as pixel movement
+    fn delta(&self) -> (f64, f64) {
+        (self.delta_x(), self.delta_y())
+    }
+
+    /// Tool position in the device's native coordinate space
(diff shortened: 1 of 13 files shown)


src/backend/libinput/tablet.rs
```diff
@@ -0,0 +1,171 @@
...
+
+impl<E> backend::TabletToolEvent for E
+where
+    E: IsTabetEvent + event::EventTrait,
+{
+    fn tablet(&self) -> TabletDescriptor {
+        let dev = event::EventTrait::device(self);
+
+        // Safety: udev Context has be the same as the one the libinput Context was created from.
+        // As far as I know, currently there is now way to brake this rule in smithway.
```
PolyMeilex, line 55: I'm not 100% sure about that.

vberger:
As a first remark, it looks like your code needs to appropriate `#[cfg(...)]` attributes in anvil, as it currently fails to compile on all feature sets without udev.

src/wayland/tablet/tablet_tool.rs
```diff
@@ -0,0 +1,398 @@
...
+
+    /// Tablet tool is making contact
+    pub fn tip_up(&self, time: u32) {
+        self.inner.borrow_mut().tip_up(time);
+    }
+
+    /// Tablet tool is no longer making contact
+    pub fn tip_down(&self, time: u32) {
+        self.inner.borrow_mut().tip_down(time);
+    }
```
vberger, line 382: It looks like the documentations of these two methods are swapped.

src/wayland/tablet/tablet_tool.rs
```diff
@@ -0,0 +1,398 @@
...
+
+    /// Queue tool slider update
+    pub fn slider_position(&self, slider: f64) {
+        self.inner.borrow_mut().slider_position(slider);
+    }
+
+    /// Queue tool rotation update
+    pub fn rotation(&self, rotation: f64) {
+        self.inner.borrow_mut().rotation(rotation);
+    }
```
vberger, line 372: These methods are documented to queue an update, they should also document when the update will be actually sent to the client.

src/wayland/tablet/tablet_tool.rs
```diff
@@ -0,0 +1,398 @@
...
+        time: u32,
+    ) {
+        let wl_tool = self
+            .instances
+            .iter()
+            .find(|i| i.as_ref().same_client_as(focus.as_ref()));
+
+        if let Some(wl_tool) = wl_tool {
+            tablet.with_focused_tablet(&focus, |wl_tablet| {
+                let serial = SERIAL_COUNTER.next_serial();
```
vberger, line 49: Such methods should take the serial as argument rather than increment it themselves. The serial can then be used by the client to request some action in response to, so the compositor needs to be aware of them, meaning it should provide them.

PolyMeilex:
I'm aware of that, as mentioned:
> I'm not sure how to handle NewDevice, RemovedDevice events in anvil, as those are libinput specific events, for now I hard-coded InputBackend<SpecialEvent = LibinputEvent>> into input handler, but that's a little bit weird, and CI is already complaining about that

I'm already working on fixing that

src/wayland/tablet/tablet_tool.rs
PolyMeilex: The main reason for that, was because `proximity_in()` can be called by `motion()` in some cases, but in most cases `motion` does not use serial for anything, so... would it be ok to simply do nothing with supplied serials in those cases?

vberger: Yes that would be ok. This is more or less what we are already doing with pointer events.

src/backend/input/tablet.rs
```diff
@@ -0,0 +1,361 @@
...
+    pub pressure: bool,
+    /// Distance axis
+    pub distance: bool,
+    /// Z-rotation axis
+    pub rotation: bool,
+    /// Slider axis
+    pub slider: bool,
+    /// Wheel axis
+    pub wheel: bool,
+}
```
vberger, line 55: Would it be relevant to make this struct into a bitfield, to make it more compact?

src/backend/input/tablet.rs
```diff
@@ -0,0 +1,361 @@
...
+    /// If this axis does not exist on the current tool, this function returns 0.
+    fn distance(&self) -> f64;
+
+    /// Check if the distance axis was updated in this event.
+    fn distance_has_changed(&self) -> bool;
+
+    /// Returns the current distance from the tablet's sensor, normalized to the range [0, 1]
+    ///
+    /// If this axis does not exist on the current tool, this function returns 0.
+    fn pressure(&self) -> f64;
```
vberger, line 124: Looks like the docs of `distance()` and `pressure()` are swapped.

src/backend/input/tablet.rs
```diff
@@ -0,0 +1,361 @@
...
+    ///
+    /// If this axis does not exist on the current tool, this function returns 0.
+    fn tilt_y(&self) -> f64;
+
+    /// Check if the tilt y axis was updated in this event.
+    fn tilt_y_has_changed(&self) -> bool;
+
+    /// Returns the current z rotation of the tool in degrees, clockwise from the tool's logical neutral position.
+    ///
+    /// For tools of type Mouse and Lens the logical neutral position is pointing to the current logical north of the tablet. For tools of type Brush, the logical neutral position is with the buttons pointing up.
```
vberger, line 171: This line is too long and needs to be split (rustfmt didn't catch that... ?)

src/backend/input/tablet.rs
```diff
@@ -0,0 +1,361 @@
...
+
+impl<B: InputBackend> TabletToolAxisEvent<B> for UnusedEvent {}
+
+/// The state of proximity for a tool on a device.
+///
+/// The proximity of a tool is a binary state signalling whether the tool is within a
+/// detectable distance of the tablet device. A tool that is out of proximity cannot
+/// generate events.
+///
+/// On some hardware a tool goes out of proximity when it ceases to touch the surface. On /// other hardware, the tool is still detectable within a short distance (a few cm) off
```
vberger, line 267:
```suggestion
/// On some hardware a tool goes out of proximity when it ceases to touch the surface. On
/// other hardware, the tool is still detectable within a short distance (a few cm) off
```

src/backend/input/tablet.rs
```diff
@@ -0,0 +1,361 @@
...
+        match *self {}
+    }
+}
+
+/// Signals that a tool has changed a logical button state on a device with the DeviceCapability::TabletTool capability.
+pub trait TabletToolButtonEvent<B: InputBackend>: TabletToolEvent<B> + Event<B> {
+    /// Return the button that triggered this event.
+    fn button(&self) -> u32;
+
+    /// For the button of a TabletToolButtonEvent, return the total number of buttons pressed on all devices on the associated seat after the the event was triggered.
```
vberger, line 342: This documentation line is too long as well

src/backend/libinput/tablet.rs
```diff
@@ -0,0 +1,188 @@
...
+};
+
+use input as libinput;
+use input::event;
+use input::event::{tablet_tool, EventTrait};
+
+use super::LibinputInputBackend;
+
+/// Marker for tablet tool events
+pub trait IsTabetEvent: tablet_tool::TabletToolEventTrait + EventTrait {}
```
vberger, line 12:
```suggestion
pub trait IsTabletEvent: tablet_tool::TabletToolEventTrait + EventTrait {}
```

src/wayland/tablet_manager/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+
+                        let user_data = seat.user_data();
+                        user_data.insert_if_missing(TabletSeatHandle::default);
+
+                        let instance = tablet_seat;
+                        let tablet_seat = user_data.get::<TabletSeatHandle>();
+
+                        if let Some(tablet_seat) = tablet_seat {
+                            tablet_seat.add_instance(instance);
+                        }
```
vberger, line 82: I'm in favor of using `.unwrap()` when the option cannot possibly be `None` unless something has gone very wrong, which is the case here are we ensure the value is inserted 2 lines above.

src/wayland/tablet_manager/tablet.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub name: String,
+    /// Tablet device USB product id
+    pub id_product: Option<u32>,
+    /// Tablet device USB vendor id
+    pub id_vendor: Option<u32>,
+    /// Path to the device
+    pub syspath: Option<PathBuf>,
+}
+
+impl<D: Device> From<D> for TabletDescriptor {
```
vberger, line 25: Semantically, wouldn't it be more precise for it to be `impl<D: Device> From<&D> for ...` ? That impl does not need to consume the device.

src/wayland/tablet_manager/tablet.rs
```diff
@@ -0,0 +1,98 @@
...
+                    .retain(|i| !i.as_ref().equals(&instance.as_ref()));
+            }));
+
+            seat.tablet_added(&wl_tablet);
+
+            wl_tablet.name(tablet.name.clone());
+
+            if let (Some(id_product), Some(id_vendor)) = (tablet.id_product, tablet.id_vendor) {
+                wl_tablet.id(id_product, id_vendor);
+            }
```
vberger, line 72: What should happen if only one of `id_product` and `id_vendor` is `Some()`? Is this something that can happen?

src/wayland/tablet_manager/tablet_seat.rs
```diff
@@ -0,0 +1,189 @@
...
+            )
+            .finish()
+    }
+}
+
+/// Handle to a tablet seat
+///
+/// TabletSeat extends `Seat` with graphic tablet specyfic functionality
+///
+/// TabletSeatHandle can be used to advertise avalible graphics tablets and tools to wayland clients
```
vberger, line 46:
```suggestion
/// TabletSeatHandle can be used to advertise available graphics tablets and tools to wayland clients
```

src/wayland/tablet_manager/tablet_seat.rs
```diff
@@ -0,0 +1,189 @@
...
+#[derive(Default, Debug, Clone)]
+pub struct TabletSeatHandle {
+    inner: Rc<RefCell<TabletSeat>>,
+}
+
+impl TabletSeatHandle {
+    pub(super) fn add_instance(&self, seat: Main<ZwpTabletSeatV2>) {
+        let mut inner = self.inner.borrow_mut();
+
+        // Notify new instance about avaluble tablets
```
vberger, line 56:
```suggestion
        // Notify new instance about available tablets
```

src/wayland/tablet_manager/tablet_seat.rs
```diff
@@ -0,0 +1,189 @@
...
+        for (desc, tablet) in inner.tablets.iter_mut() {
+            tablet.new_instance(seat.deref(), desc);
+        }
+
+        // Notify new instance about avalible tools
```
vberger, line 61:
```suggestion
        // Notify new instance about available tools
```

src/wayland/tablet_manager/tablet_tool.rs
```diff
@@ -0,0 +1,477 @@
...
+    ) {
+        self.inner
+            .borrow_mut()
+            .proximity_in(pos, focus, tablet, serial, time)
+    }
+
+    /// Notifify that this tool has left proximity.
+    pub fn proximity_out(&self, time: u32) {
+        self.inner.borrow_mut().proximity_out(time);
+    }
```
vberger, line 373: If `motion()` already handles the `proximity_in/proximity_out` logic, is it necessary to expose these methods as well?

vberger (commented):
Thanks a lot, this is a very clean implementation!

I've lest a few comments/questions to nitpick on the details. :)

src/wayland/tablet_manager/tablet.rs
PolyMeilex: I'm not sure to be hones, when it comes to `libinput` both `id_product` and `id_vendor` are always returned (maybe those default to `0`, but there is no mention of that in the doc).

Should we perhaps send `0` if one of ID's was not supplied?

src/wayland/tablet_manager/tablet_tool.rs
PolyMeilex: It is exposed because tablet tool pointers in contrary to mouse pointers can appear out of tin air, for example if the user lifts a pen above the tablet and moves it to another window (without tablet's knowledge) and then lowers down a pen (triggering tool's proximity_in event) then the pointer will just teleport to the new location. Or even better example would be if a user brings a completely new tool into proximity, then it's not really a `motion` per se, it just appeared out of nowhere.

Although motion event can easily emulate this behavior, I just think that the event handler code is a lot easier to understand if it uses those function directly.
[code that I'm talking about](https://github.com/Smithay/smithay/blob/921ef4339769a7456a7f8bf1e902bdc1b2d3cefb/anvil/src/input_handler.rs#L314)

Now, handling of proximity event in anvil looks like this:
```rs
 match evt.state() {
    ProximityState::In => tool.proximity_in(
        self.pointer_location,
        under,
        &tablet,
        SCOUNTER.next_serial(),
        evt.time(),
    ),
    ProximityState::Out => tool.proximity_out(evt.time()),
}
```
It's really easy to understand at first glance.

Here is how it would look like if `proximity_in` and `proximity_out` were not exposed:
```rs
match evt.state() {
    ProximityState::In => tool.motion(
        self.pointer_location,
        under,
        &tablet,
        SCOUNTER.next_serial(),
        evt.time(),
    ),
    ProximityState::Out => tool.motion(
        (0.0,0.0), // Accually tool outside of proximiti does not have pos, but we have to send something
        None, // Tool out of proximity wil never have anything under it
        &tablet,
        SCOUNTER.next_serial(),
        evt.time(),
    ),
}
```

Or maybe I'm just overthink it, and it's not a such big deal, I'll do whatever you prefer.

vberger: I see, thanks for the explanation, that makes total sense. :+1:

src/wayland/tablet_manager/tablet.rs
vberger: If we always get both, I think we could store that as an `Option((u32, u32))`, rather than a pair of options, no?

PolyMeilex: Good point, I'll do that.

vberger (approved):
Alright, I'm good with this PR as it is now, thanks again!

To think smithay is going to have tablet support before touch support. :sweat_smile:

Anyway, @Drakulix do you have something to add?

PolyMeilex:
Sure, it was fun to work on it, I put my tablet hardware to good use :ok_hand: 
I think it would be nice to add some more tablet specific support to anvil (like an ability to move windows with it, or support for multiple pointers for example), but that's not related to Smithay in any way (at least I hope so) so that can be added sometime in the future.

Touch is also definitely something that I'm interested in, but I don't have any touch capable device ATM :cry:

Drakulix:
> Anyway, @Drakulix do you have something to add?

Nope. Looks fine to me!

vberger:
Alright, here we go then!

PR #315, opened 2021-06-28
Fix XWayland panic on exit

cmeissl:
This PR tries to fix the XWayland panic on exit.

The panic is caused by an already borrowed RefCell during XWayland shutdown.
When anvil is exiting the drop implemetation of the XWayland instance will borrow
the associated data and call shutdown.
During a shutdown the wayland client associated with XWayland will be killed explicit which
will invoke the registered destructor callback. The callback again borrows the data to call shutdown which
results in a panic.


vberger (approved):
Thanks, I think this is the appropriate fix. :+1:

PR #317, opened 2021-06-29
Disable pointer grab with dead starting surface

### LINKED ISSUE #295: Investigate the behavior of Grabs when the client exits
Following-up on #294, the pointer grab mechanism has apparently not been design with the possibility that the client could exit while a grab is active. More generally, it should take into account the possibility that the surface which has the grab could be destroyed while the grab is active, and unset the grab when that happens.

vberger:
Fixes #295


PolyMeilex:
I can confirm that it does fix the issue when testing with the same client I used in #294

vberger:
@PolyMeilex I did not undo your over changes, because they actually have a large scope that is specific to xdg-shell resizing interactions so we're going to keep it. :sweat_smile:

Drakulix (approved):
:+1:

PR #318, opened 2021-06-29
Rework `InputBackend` abstractions to track individual devices

Drakulix:
Removes the notion of Seats from `InputBackend` and replaces it with
`DeviceAdded` and `DeviceRemoved` events allowing compositors to
descriminate between single devices.

A new abstraction to model multiple seats on a single input backend
may be be added separately at a later time.


vberger (approved):
Looks like this simplifies the logic quite a lot, neat.

PR #322, opened 2021-06-30
Fix resize state is never finished

### LINKED ISSUE #321: XDG shell move stops working after resize
I'm not 100% sure, but it seems that it is impossible (or maybe just hard) to reproduce this when checking out commits before #286 

Steps to reproduce:
1. Start a client (nautilus in my case, it seems that it's a lot harder to reproduce with some clients, like terminals, no idea why)
2. Resize it
3. Try to move it
4. This will produce this kind of behavior: 
gif

cmeissl:
use the current xdg_toplevel state to check if we are resizing. The configure message
is the result from telling the client that resizing
has stopped and therefore not include the resizing
state. The current state is safe to use because
AckConfigure will move the state to last_acked
and the current state will be set on the next surface commit.

fixes #321

### DIFF
--- a/anvil/src/shell.rs
+++ b/anvil/src/shell.rs
@@ -452,15 +452,30 @@ pub fn init_shell<BackendData: 'static>(display: &mut Display, log: ::slog::Logg
                 .unwrap();
 
                 if let Some(serial) = waiting_for_serial {
-                    if configure.serial > serial {
-                        // TODO: huh, we have missed the serial somehow.
-                        // this should not happen, but it may be better to handle
-                        // this case anyway
-                    }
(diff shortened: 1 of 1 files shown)


vberger (changes requested):
I'm ok with this change, but I'd prefer if it included a comment explaining why the resizing state is checked from the current state rather than from the configure we just received.

cmeissl:
> I'm ok with this change, but I'd prefer if it included a comment explaining why the resizing state is checked from the current state rather than from the configure we just received.

I left a comment why we need to check the current state instead of the configure state

vberger (approved):

PR #323, opened 2021-07-01
Fix libseat build

PolyMeilex:
Since #305 libseat is no longer passing build because of missing slog macros, this PR adds import for those.


vberger:
Okay, given the CI test with `backend_session_libseat` already passes before this PR, I suspect we may have a problem with the CI or the cargo features :thinking:

vberger:
Okay, the problem is that the cargo feature `backend_session_libseat` does not automatically enable the `backend_session` feature, like `backend_session_logind` does. Could you add that in that PR, so that we right the whole thing at once?

PR #327, opened 2021-07-04
Anvil HiDPI

### LINKED ISSUE #104: Make anvil HiDPI-aware
currently anvil is not dpi-aware, and following #102 this leads to graphical bugs when started using the winit backend on an HiDPI monitor.

Parts that need to be made dpi-aware:

- advertising DPI factor in monitors to clients
- rescaling clients buffers depending on the output dpi and their drawing dpi
- processing input events and converting them to logical coordinates before sending them to clients
- probably other things that I don't remember right now

cmeissl:
This PR adds support for HiDPI to anvil.

The scaling supports fractional scaling and will announce the rounded value to the wayland output.
The initial scale of an output can be set with the envorinment variable `ANVIL_SCALE_{short-name of output}`.
The scale can be changed with the keyboard shortcut `LOGO + SHIFT + P` to increase and  `LOGO + SHIFT + M` to decrease in 0.25f32 steps

Toplevels and the pointer (on udev) will be automatically moved after changing the scale so that they appear stable.

closes #104

### DIFF
--- a/anvil/src/drawing.rs
+++ b/anvil/src/drawing.rs
@@ -39,6 +39,7 @@ pub fn draw_cursor<R, E, F, T>(
     frame: &mut F,
     surface: &wl_surface::WlSurface,
     location: Point<i32, Logical>,
+    output_scale: f32,
     log: &Logger,
 ) -> Result<(), SwapBuffersError>
 where
@@ -69,7 +70,7 @@ where
             (0, 0).into()
         }
     };
-    draw_surface_tree(renderer, frame, surface, location - delta, log)
+    draw_surface_tree(renderer, frame, surface, location - delta, output_scale, log)
 }
 
 fn draw_surface_tree<R, E, F, T>(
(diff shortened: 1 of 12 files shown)


vberger:
Overall this looks like a solid start, however it seems you are kind of mixing logical and physical coordinates here, especially when handling pointer motion events.

The general rule I want to follow is that only the drawing code should deal in physical size, everything else should work in logical coordinates.

Notably, that means that input handling, window positioning and output positioning should all be done in logical coordinates.

cmeissl:
> Overall this looks like a solid start, however it seems you are kind of mixing logical and physical coordinates here, especially when handling pointer motion events.
> 
> The general rule I want to follow is that only the drawing code should deal in physical size, everything else should work in logical coordinates.
> 
> Notably, that means that input handling, window positioning and output positioning should all be done in logical coordinates.

I hope you did not look at the winit code :)
I thought about the different coordinates that are used and think that there are 3 different ones.
The global coordinates which are used for output geometry, window positioning and the input events after tranforming them from the physical to global coordinates.
The surface coordinates for sending pointer events to the client relative to the surface origin.
Physical, the renderer will then have to transform the global coordinates to physical ones.

I think this is what i implemented as a PoC in the PR, only winit is somewhat broken and the input handling is wrong.
I concentracted on udev for building a first solution so that i do not have to deal with double dpi handling in winit (my desktop already uses HiDPI and winit runs with scale 2)

vberger (approved):

PR #328, opened 2021-07-04
Winit: Share the resized size instead of using egl display field

i509VCB:
The simple explanation for this pull request is the input handling for winit should be entirely graphics independent. Currently the input has to have an egl surface to resize. This simply adds a Cell<Option<(u32, u32)>> to tell the graphics backend when a resize occurs and resize before rendering. This shifts the responsibility of resizing the egl surface to the graphics backend.


vberger (approved):

vberger:
There are a few clippy warnings, but other than that :+1:

i509VCB:
clippy has been silenced

PR #329, opened 2021-07-05
Introduce Logical/Physical Point/Rectangle

vberger:
The new types are in `src/utils/rectangle.rs`, the rest is mostly the fallout.


Drakulix (approved):
I find it a tad confusing, that a struct named `Point` is also used for something I would rather describe as `Size`. But I get it and I do not think, we would gain much by adding two distinct types. But maybe we could add a type-alias `type Size = Point`? And use size in some places for better code readability?

But probably I will get used to this and this is really not a big problem. Functionality-wise this all looks fine, so I approve of this PR.

src/backend/winit.rs
```diff
@@ -71,11 +75,21 @@ pub enum Error {
 #[derive(Debug, Clone)]
 pub struct WindowSize {
     /// Pixel side of the window
-    pub physical_size: PhysicalSize<u32>,
+    pub physical_size: Point<u32, Physical>,
```
Drakulix, line 78: Should this be a `Size` (judging by the name).

src/backend/winit.rs
```diff
@@ -71,11 +75,21 @@ pub enum Error {
...
     /// Scaling factor of the window
     pub scale_factor: f64,
 }
 
+impl WindowSize {
+    fn logical_size(&self) -> Point<f64, Logical> {
```
Drakulix, line 84: Just like this should be `Size`

src/backend/winit.rs
```diff
@@ -234,7 +248,7 @@ pub enum WinitEvent {
     /// The window has been resized
     Resized {
         /// The new physical size (in pixels)
-        size: (f64, f64),
+        size: Point<f64, Physical>,
```
Drakulix, line 251: Also this should be `Size`, I think?

src/wayland/shell/xdg/mod.rs
```diff
@@ -220,11 +220,11 @@ xdg_role!(
         /// Minimum size requested for this surface
         ///
         /// A value of 0 on an axis means this axis is not constrained
-        pub min_size: (i32, i32),
+        pub min_size: Point<i32, Logical>,
```
Drakulix, line 223: Again, size?

src/wayland/shell/xdg/mod.rs
```diff
@@ -220,11 +220,11 @@ xdg_role!(
...
         /// Maximum size requested for this surface
         ///
         /// A value of 0 on an axis means this axis is not constrained
-        pub max_size: (i32, i32),
+        pub max_size: Point<i32, Logical>,
```
Drakulix, line 227: Again, size?

src/wayland/shell/xdg/mod.rs
```diff
@@ -603,29 +585,29 @@ impl From<ToplevelStateSet> for Vec<xdg_toplevel::State> {
...
-    pub geometry: Option<Rectangle>,
+    pub geometry: Option<Rectangle<i32, Logical>>,
     /// Minimum size requested for this surface
     ///
     /// A value of 0 on an axis means this axis is not constrained
     ///
     /// This is only relevant for xdg_toplevel, and will always be
     /// `(0, 0)` for xdg_popup.
-    pub min_size: (i32, i32),
+    pub min_size: Point<i32, Logical>,
```
Drakulix, line 595: Again, size?

src/wayland/shell/xdg/mod.rs
```diff
@@ -603,29 +585,29 @@ impl From<ToplevelStateSet> for Vec<xdg_toplevel::State> {
...
     /// Maximum size requested for this surface
     ///
     /// A value of 0 on an axis means this axis is not constrained
     ///
     /// This is only relevant for xdg_toplevel, and will always be
     /// `(0, 0)` for xdg_popup.
-    pub max_size: (i32, i32),
+    pub max_size: Point<i32, Logical>,
```
Drakulix, line 602: Again, size?

Drakulix (approved):
I found some things being named `*size*`, but not of the matching type `Size`, but other then that: Lets merge this thing! :)

PR #334, opened 2021-07-11
gles2: add and use cropping to the renderer

### LINKED ISSUE #199: Add a FPS counter to anvil
For general performance monitoring & debugging, might be usefull to have anvil compute its FPS and occasionally dump it in its log, or even display it on the screen.

Drakulix:
Alright, so this is actually kind-of done before #329, but I can rebase it easily, I think.
(This is just using `Rectangle` in a few places, where the new `new`-constructor of #329 would make some code smaller.)

This PR does a bunch of things and is again best review commit-by-commit:
1. Implement the necessary changes to add GL-code inside anvil or more generally out-of-tree.
2. Move the bitmap code into anvil, as the renderer shall only help with buffers.
3. Fixup anvils cursor code, while we are add it and properly load XCursors (that was a fun distraction).
4. Actually add the cropping and scaling features, by modifying the drawn vertices of a texture instead of using GL_SCISSORS. This has the benefit of working, if the user uses a 3d-space through `render_texture` for a graphically more advanced compositor, while GL_SCISSORS can only operate on a 2d-plane.
5. Use GL_SCISSORS to optimize draw-calls that extend beyond our surface/buffer.
6. Make use of our new texture rendering, by adding a spritesheet for numbers (to skip adding a full-blown text-renderer) and rendering FPS. This introduces a new `anvil/debug` feature (like #310, which will be rebased on top of this) and finally closes #199.

I have not yet tested the FPS counter on the udev-backend, so I am marking the PR as a "draft" for now until that is done. But since this already is working nicely on winit, I expect the udev-code to be mostly bug-free. The whole thing should be ready for a review imo.


src/backend/renderer/mod.rs
```diff
@@ -168,26 +168,62 @@ pub trait Frame {
         &mut self,
         texture: &Self::TextureId,
         matrix: Matrix3<f32>,
+        verticies: [Vector2<f32>; 4],
```
vberger, line 171: The documentation of this method needs to be updated for its new argument

src/backend/renderer/mod.rs
```diff
@@ -168,26 +168,62 @@ pub trait Frame {
...
         alpha: f32,
     ) -> Result<(), Self::Error>;
+
     /// Render a texture to the current target as a flat 2d-plane at a given
-    /// position, applying the given transformation with the given alpha value.
+    /// position, cropping at the given offsets and applying the given transformation with the given alpha value.
```
vberger, line 176: It doesn't look like this method does any cropping, it does not take any argument indicating it.

src/backend/renderer/mod.rs
```diff
@@ -168,26 +168,62 @@ pub trait Frame {
...
     ///
     /// This operation is only valid in between a `begin` and `finish`-call.
     /// If called outside this operation may error-out, do nothing or modify future rendering results in any way.
     fn render_texture_at(
         &mut self,
         texture: &Self::TextureId,
         pos: (i32, i32),
+        scale: u32,
```
vberger, line 184: What happens if the compositor wishes to draw a texture whose buffer scale is 2 on an output whose scaling factor is 3?  Should it use `render_texture_from_to` instead? If yes, that might be worth mentioning in the docs.

src/backend/renderer/mod.rs
```diff
@@ -168,26 +168,62 @@ pub trait Frame {
...
+            },
+            transform,
+            alpha,
+        )
+    }
+
+    /// Render part of a texture as given by src to the current target into the rectangle described by dest
+    /// as a flat 2d-plane after applying the given transformations.
+    ///
+    /// This operation is only valid in between a `begin` and `finish`-call.
```
vberger, line 211: Are the comments about `begin`/`finish` still relevant with the render closure-based API?

vberger (commented):
So, it looks like a few `#[cfg(...)]` need to be fixed for the CI to pass.

How confident are you wrt the computations of `render_from_to`? Notably the order in which the transforms & nonuniform scaling are done?

Drakulix:
> How confident are you wrt the computations of `render_from_to`? Notably the order in which the transforms & nonuniform scaling are done?

Oh not at all. :D

The code used to be similar to what sway/wlroots does, but the cropping changed this up a bit.
If my mathematical intuition is correct, the operations are applied to the inverse order they are added to the projection matrix.

wp_viewport also has documented the expected order:

      The coordinate transformations from buffer pixel coordinates up to
      the surface-local coordinates happen in the following order:
        1. buffer_transform (wl_surface.set_buffer_transform)
        2. buffer_scale (wl_surface.set_buffer_scale)
        3. crop and scale (wp_viewport.set*)
      This means, that the source rectangle coordinates of crop and scale
      are given in the coordinates after the buffer transform and scale,
      i.e. in the coordinates that would be the surface-local coordinates
      if the crop and scale was not applied.

I am hoping this code adheres to these rules, but given that we do not implement wp_viewport yet, I was not sure how to best test this. Also given, that two scales are involved here (buffer_scale + viewport_scale), I am not sure, if this *can* be done correctly with this function, because I am not sure, which parts of anvil and smithay are handling the buffer scale right now and how.

src/backend/renderer/mod.rs
Drakulix: Oh right, so we do not some kind of fractional scaling for mixing and matching different output and buffer scales, that are not multiples of each other, I was not really aware of that. I wonder, if the better solution would be to allow a floating point scale here.

vberger: That, or provide the buffer scale as argument as well as the desired output scale, I'd say.

vberger:
Well, once your PR is merged, adding support for `wp_viewporter` will be really easy, so we'll be able to test that then.

src/backend/renderer/gles2/mod.rs
```diff
@@ -67,6 +67,35 @@ struct Gles2Program {
...
+    /// It is also expected to not be external or y_inverted.
+    ///
+    /// Ownership over the texture is taken by the renderer, you should not free the texture yourself.
+    ///
+    /// *Safety*: The renderer cannot make sure `tex` is a valid texture id.
+    pub unsafe fn from_raw(
+        renderer: &Gles2Renderer,
+        tex: ffi::types::GLuint,
+        width: u32,
+        height: u32,
```
vberger, line 86: What about using `Size<i32, Buffer>` here?

vberger (approved):
Small API nitpick, but other than that the looks quite good! :+1:

src/backend/renderer/gles2/mod.rs
Drakulix: Nice catch!

Issue #228, opened 2020-07-16
Documentation update & refresh


vberger:
Following the numerous changes and rework that occurred to Smithay, and in preparation to v0.3, it is necessary to give a general pass on the documentation of the whole crate, as several parts of it have gone outdated.

This means mostly rewriting documentation items that no longer accurately describe the element they are attached to, fixing typos and grammar, as well as finding and fixing dead links in the documentation.

This may also be the occasion of reworking the public-facing module structure (for example flattening unnecessary deep paths by making the submodules private and reexporting their contents).

Already reviewed modules:

- [ ] backend
  - [ ] allocator
    - [ ] dmabuf
    - [ ] dumb
    - [ ] gbm
  - [ ] drm
  - [ ] egl
    - [ ] context
    - [ ] display
    - [ ] ffi
    - [ ] native
    - [ ] surface
  - [ ] input
  - [ ] libinput
  - [ ] renderer
    - [ ] gles2
      - [ ] ffi
  - [x] session
    - [x] auto
    - [x] direct
    - [x] logind
    - [x] libseat
  - [x] udev
  - [x] winit
- [x] utils
  - [x] signaling
- [x] wayland
  - [x] compositor
  - [x] data_device
  - [x] dmabuf
  - [x] explicit_synchronization
  - [x] output
  - [x] seat
  - [x] shell
    - [x] legacy
    - [x] xdg
  - [x] shm
- [x] xwayland

PR #336, opened 2021-07-13
First pass at backend docs for #228

Drakulix:
Probably contains a few spelling mistakes, but I think I caught everything that was out-dated at least.
I will give this another go tomorrow. :)

### DIFF
--- a/src/backend/egl/mod.rs
+++ b/src/backend/egl/mod.rs
@@ -1,30 +1,30 @@
 //! Common traits and types for egl rendering
 //!
-//! Large parts of this module are taken from
-//! [glutin src/api/egl](https://github.com/tomaka/glutin/tree/044e651edf67a2029eecc650dd42546af1501414/src/api/egl/)
+//! This module has multiple responsibilities related to functionality provided by libEGL:
+//! Initializing EGL objects to:
+//! - initialize usage of EGL based [`WlBuffer`](wayland_server::protocol::wl_buffer::WlBuffer)s via `wl_drm`.
+//! - initialize OpenGL contexts from.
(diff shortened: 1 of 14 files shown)


src/backend/egl/mod.rs
```diff
@@ -1,30 +1,30 @@
 //! Common traits and types for egl rendering
 //!
-//! Large parts of this module are taken from
-//! [glutin src/api/egl](https://github.com/tomaka/glutin/tree/044e651edf67a2029eecc650dd42546af1501414/src/api/egl/)
+//! This module has multiple responsibilities related to functionality provided by libEGL:
+//! - Initializing EGL objects to
+//!   - initialize usage of EGL based [`WlBuffer`](wayland_server::protocol::wl_buffer::WlBuffer)s via `wl_drm`.
+//!   - initialize OpenGL contexts from.
+//!   - Import/Export external resources to/from OpenGL
```
vberger, line 7: It seems there is one too many layers of lists here? The topmost list has only one item.

vberger (approved):
Just a tiny nitpick, but otherwise LGTM. :+1: 

Which modules does that cover, so that we can keep track of where we are at on #228

Drakulix:
> Just a tiny nitpick, but otherwise LGTM. +1
> 
> Which modules does that cover, so that we can keep track of where we are at on #228

Should be everything that is missing, although I found most out of date docs in the egl-module.
But I was a little tired yesterday, so I will look through all of it a second time.

Drakulix:
Alright I am done with this one. Fixes #228

vberger (approved):

PR #340, opened 2021-07-20
Implement XdgOutput protocol

PolyMeilex:


src/wayland/output/xdg.rs
```diff
@@ -0,0 +1,177 @@
...
+
+/// Initialize a xdg output manager global.
+pub fn init_xdg_output_manager<L>(display: &mut Display, logger: L) -> Global<ZxdgOutputManagerV1>
+where
+    L: Into<Option<::slog::Logger>>,
+{
+    let log = crate::slog_or_fallback(logger).new(o!("smithay_module" => "xdg_output_handler"));
+
+    display.create_global(
+        3,
```
i509VCB, line 153: magic value? might be worth using the `VERSION` field defined in `ZxdgOutputManagerV1`

PolyMeilex: Sorry, but I kinda don't follow. What do you mean by `VERSION field defined in ZxdgOutputManagerV1`? Do you mean version constants in [zxdg_output_manager_v1](https://docs.rs/wayland-protocols/0.28.6/wayland_protocols/unstable/xdg_output/v1/server/zxdg_output_manager_v1/index.html#constants) module? None of those are equal to the protocol version that I'm using.
(I probably just misunderstood or missed something)

vberger: Yeah, it's not a magic value, it is the maximum version of the protocol currently supported by the code of this module, while the `VERSION` associated constant of `ZxdgOutputManagerV1` is the maximum version supported by the bindings generated by `wayland-server`. We should not use these `VERSION` constants directly in `create_global()` as this would automatically bump the version advertised to clients to the latest of the XML files, even if smithay has not been updated to support them yet.

vberger (approved):

PR #342, opened 2021-07-22
Implement xdg-foreign protocol

i509VCB:
Implementation of the xdg-foreign protocol. This allows a client to export a surface and then allow another import the surface using a handle. 

So what's needed otherwise:
- [x] New release of wayland-rs so the error types from xdg-foreign are accessible. Rebased on #343 
- [x] Implement the parent relationship logic. Untested atm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Additions
+
+#### Clients & Protocols
+
+- Added public api constants for the roles of `wl_shell_surface`, `zxdg_toplevel` and `xdg_toplevel`. See the
+  `shell::legacy` and `shell::xdg` modules for these constants.
+- Whether a surface is toplevel equivalent can be determined with the new function `shell::is_toplevel_equivalent`. 
+- Setting the parent of a toplevel surface is now possible with the `xdg::ToplevelSurface::set_parent` function.
+- Add support for the zxdg-foreign-v2 protocol.
+
 ### Bugfixes
 
 - EGLBufferReader now checks if buffers are alive before using them.
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -65,6 +65,7 @@ use crate::utils::DeadResource;
 use crate::utils::{Logical, Point, Rectangle, Size};
 use crate::wayland::compositor;
 use crate::wayland::compositor::Cacheable;
+use crate::wayland::shell::is_toplevel_equivalent;
 use crate::wayland::{Serial, SERIAL_COUNTER};
 use std::fmt::Debug;
 use std::{
@@ -72,6 +73,7 @@ use std::{
     rc::Rc,
     sync::{Arc, Mutex},
 };
+
 use wayland_protocols::unstable::xdg_shell::v6::server::zxdg_surface_v6;
 use wayland_protocols::xdg_shell::server::xdg_surface;
 use wayland_protocols::{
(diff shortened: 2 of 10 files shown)


src/wayland/shell/xdg/mod.rs
```diff
@@ -1115,6 +1123,30 @@ impl ToplevelSurface {
...
         )
     }
+
+    /// Returns true if the provided surface is the parent of this toplevel surface.
+    pub fn is_parent(&self, parent: &WlSurface) -> bool {
+        match &self.shell_surface {
+            ToplevelKind::Xdg(toplevel) => xdg_handlers::is_parent(toplevel, parent),
+            ToplevelKind::ZxdgV6(toplevel) => zxdgv6_handlers::is_parent(toplevel, parent),
+        }
+    }
```
vberger, line 1133: It'd be better to have a method that returns the parent surface directly, that would then be checked for equality.

src/wayland/shell/xdg/mod.rs
```diff
@@ -1115,6 +1123,30 @@ impl ToplevelSurface {
...
+    pub fn set_parent(&self, parent: Option<ToplevelSurface>) {
+        match &self.shell_surface {
+            ToplevelKind::Xdg(toplevel) => {
+                xdg_handlers::set_parent(toplevel, parent.and_then(|parent| parent.get_surface().cloned()))
+            }
+            ToplevelKind::ZxdgV6(toplevel) => {
+                zxdgv6_handlers::set_parent(toplevel, parent.and_then(|parent| parent.get_surface().cloned()))
+            }
+        }
+    }
```
vberger, line 1149: In general the parent as set by the compositor is not necessarily always a `{z}xdg_toplevel`, but could potentially be other things (like a layer-shell surface for example), so this method should take a `WlSurface` as argument directly.

src/wayland/xdg_foreign/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+    match request {
+        zxdg_exporter_v2::Request::Destroy => {
+            // all is handled by destructor.
+        }
+
+        zxdg_exporter_v2::Request::ExportToplevel { id, surface } => {
+            // Check the role of the surface, this must be xdg_toplevel.
+            let role = compositor::get_role(&surface);
+
+            if role != Some("xdg_toplevel") {
```
vberger, line 191: Same here, the spec says (emphasis mine)

> Only xdg_toplevel *equivalent* surfaces may be exported

so we should allow `zxdg_toplevel` and `wl_shell_surface` as well, and some may be added in the future.

vberger (changes requested):
Thanks for this PR, the general structure looks good, I just have a few remarks. This will need to be rebased on master as well.

i509VCB:
@vberger I think I've resolved your requests here.

vberger:
Hmm, so, seen how you changed your PR, it looks like you missed the fact that there are two different roles `XDG_TOPLEVEL_ROLE` and `ZXDG_TOPLEVEL_ROLE` for the `xdg-shell` and `zxdg_shell_v6` protocols, rather than a single unified role for both.

Taking that into account, I don't think it's necessarily a good idea to move one of them in `xdg/mod.rs`, and then use only that in the checks of `toplevel_like()`.

i509VCB:
Hmm I probably need to make sure I did fully implement the wl_shell_surface parent logic

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+### General
+
+- Added public api constants for the roles of `wl_shell_surface`, `zxdg_toplevel` and `xdg_toplevel`. See the
+  `shell::legacy` and `shell::xdg` modules for these constants.
+- Whether a surface is toplevel equivalent can be determined with the new function `shell::toplevel_like`.
```
vberger, line 9: Looks like this line needs to be updated to the new name you set in your last commit.

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+### General
```
vberger, line 5: Regarding the changelog structure, I'd like to keep the main categories as "Breaking changes" / "Additions" / "Deprecations" / "Bugfixes", so that users can quickly guess their upgrade path from the changelog.

Thematic sub-categories within that are a nice idea, though imo all the entries you added would fit in the "Clients & Protocols" sub-category.

CHANGELOG.md
```diff
@@ -6,6 +6,16 @@
...
+### Additions
+
+#### Clients & Protocols
+
+- Added public api constants for the roles of `wl_shell_surface`, `zxdg_toplevel` and `xdg_toplevel`. See the
+  `shell::legacy` and `shell::xdg` modules for these constants.
+- Whether a surface is toplevel equivalent can be determined with the new function `shell::is_toplevel_equivalent`. 
+- Setting the parent of a toplevel surface is now possible with the `xdg::ToplevelSurface::set_parent` function.
+- Add support for the zxdg-foreign-v2 protocol.
+
```
vberger, line 18: That's very minor, but could you put the "Additions" section before the "Bugfixes" ones? Putting api changes on top of the changelog seems more user-friendly.

src/wayland/xdg_foreign/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+#[derive(Debug)]
+pub struct XdgForeignState {
+    log: ::slog::Logger,
+    shell: Arc<Mutex<ShellState>>,
+    exports: Vec<Export>,
+}
+
+impl XdgForeignState {
+    /// Returns a reference to an export using the handle of the export.
+    pub fn is_export_alive(&self, handle: String) -> bool {
```
vberger, line 71: Couldn't this method just take a `&str`, rather than require a `String` (and thus a cloning of the contents)?

src/wayland/xdg_foreign/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+                self.exports.push(Export {
+                    surface: surface.clone(),
+                    handle: handle.clone(),
+                    inner: exported.deref().clone(),
+                    imports: vec![],
+                    compositor_created: true,
+                });
+
+                return Ok(handle);
+            }
```
vberger, line 100: No, that cannot do. You cannot create an object for a client without sending it to it via a protocol event specifically defined to create it. That will just completely mess up the protocol state because the client does not know about this object, and quickly cause a protocol error. We cannot export a surface of a client without it knowing about it.

From our discussions on the chatroom I though you were talking about exporting a pseudo-surface completely controlled by the compositor, so that clients could set it as parent *as if* it was a surface of an other client.

src/wayland/xdg_foreign/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+        exporter_implementation(request, state.clone(), shell.clone());
+    });
+
+    exporter.assign_destructor(Filter::new(
+        move |exporter: zxdg_exporter_v2::ZxdgExporterV2, _, _| {
+            let state = &mut *destructor_state.lock().unwrap();
+            let exports = &mut state.exports;
+            // Since remove shifts the contents of the vector, we track the amount of removals so
+            // the next element is removed at it's proper index.
+            let mut removals = 0;
```
vberger, line 262: An alternative option, arguably simpler, would be to iterate the `Vec` in reverse order.

vberger (changes requested):

vberger (approved):

PR #354, opened 2021-07-30
WLR Layer Shell Support

PolyMeilex:
Design goal of this implementation was to make its API as similar to `xdg-shell` as possible, the idea is that if someone already knows how to use `xdg-shell` they should be able to start using `layer-shell` by using nearly the same API. So `CachedState`, `with_pending_state`, `current_state`, etc. are all there, and should act just like xdg equivalents.

Anvil implementation of layer shell is lacking at the moment, I wanted to implement everything but the more I added, the hackier the code felt, so I decided to stop. The main missing piece for anvil is exclusive_zone support, but I think that I implemented enough to make it a useful test ground for future changes.
 

https://user-images.githubusercontent.com/20758186/127673456-3296c042-69a6-478f-8a71-dccf66dbef8b.mp4

### DIFF
--- a/src/wayland/shell/wlr_layer/mod.rs
+++ b/src/wayland/shell/wlr_layer/mod.rs
@@ -1,0 +1,454 @@
+//! Utilities for handling shell surfaces with the `wlr_layer_shell` protocol
+//!
+//! This interface should be suitable for the implementation of many desktop shell components,
+//! and a broad number of other applications that interact with the desktop.
+//!
+//! ### Initialization
+//!
+//! To initialize this handler, simple use the [`wlr_layer_shell_init`] function provided in this module.
+//! You need to provide a closure that will be invoked whenever some action is required from you,
+//! are represented by the [`LayerShellRequest`] enum.
+//!
+//! ```no_run
+//! # extern crate wayland_server;
+//! #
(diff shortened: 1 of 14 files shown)


vberger (changes requested):

vberger:
So, a few minor points, mostly regarding documentation. Other than that this looks pretty solid, thank! :+1:

PolyMeilex:
All long lines got eradicated.
Changelog is now fixed and flows the convention.
I rebased on top of master.

While at it, I also exposed the surface role `&str` just like #342 does for other shells

src/wayland/shell/wlr_layer/mod.rs
```diff
@@ -0,0 +1,454 @@
...
+    },
+};
+
+mod handlers;
+mod types;
+
+pub use types::{Anchor, ExclusiveZone, KeyboardInteractivity, Layer, Margins};
+
+/// The role of a wlr_layer_shell_surface
+pub static LAYER_SURFACE_ROLE: &str = "zwlr_layer_surface_v1";
```
i509VCB, line 54: Probably a little better to make this `pub const` here. Static associates a memory address with the value and using `const` allows the values to be inlined a little more easily by rustc.

PolyMeilex: Good point, it's better to use compile time constants for this

vberger (approved):

PR #358, opened 2021-07-31
XDG activation support

PolyMeilex:


anvil/src/window_map.rs
```diff
@@ -297,6 +300,29 @@ impl WindowMap {
...
+            // Take activation away from all the windows
+            for window in self.windows.iter() {
+                window.toplevel.set_activated(false);
+            }
+
+            // Give activation to our winner
+            winner.toplevel.set_activated(true);
+            self.windows.insert(0, winner);
+        }
+    }
```
vberger, line 324: I'm not thrilled by the code duplication with `get_surface_and_bring_to_top`, but I'm not quite sure how to avoid it.

src/wayland/xdg_activation/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// discard it and don't track it any futher.
+    pub fn remove_request(
+        &mut self,
+        token: &XdgActivationToken,
+    ) -> Option<(XdgActivationTokenData, WlSurface)> {
+        self.activation_requests.remove(token)
+    }
+
+    /// Retain activation requests
+    pub fn retain_reqests<F>(&mut self, mut f: F)
```
vberger, line 164:
```suggestion
    pub fn retain_requests<F>(&mut self, mut f: F)
```

anvil/src/window_map.rs
PolyMeilex: I moved the "move the nth window to top" part to a separate method, that helps a little

vberger (approved):

Issue #359, opened 2021-08-03
XDG PositionerState is not updated properly


PolyMeilex:
It took me an hour to figure out why my popups always report None or default values for their positioners. It turns out that [positioner state of XdgPopupSurfaceRoleAttributes](https://github.com/Smithay/smithay/blob/dbd03567ff18e62236a6c288008f15f2c39cbff6/src/wayland/shell/xdg/mod.rs#L323) is never updated.
In fact, `XdgPopupSurfaceRoleAttributes::positioner` is actually never used in the codebase anywhere.

 `xdg_positioner::XdgPositioner` has a `PositionerState` inside of its user data, but it is not exposed to the user:
https://github.com/Smithay/smithay/blob/dbd03567ff18e62236a6c288008f15f2c39cbff6/src/wayland/shell/xdg/xdg_handlers.rs#L164-L167

cmeissl:
Oh, i forgot to set the positioner state in the popup, sorry 😩

For updating the geometry after the initial geometry has been sent (initial configure) we need to add support for reactive popups from the protocol. The default/inital geometry can of course be overridden without this.

PolyMeilex:
I just simply want to flip popup when it gets outside of output, so I just need some way to receive initial gravity. I don't think I need reactive popup for that.

cmeissl:
If the geometry of the popup does not change probably not. Not sure if flipping can be achieved without changing the geometry. If i read the spec correctly the geometry is only allowed to change when the positioner is reactive.

cmeissl:
I opened a pr for v3 support where the inital positioner state has been added to the newpopup event, should make initial placement easier.

PR #360, opened 2021-08-03
Potential fix for #359

PolyMeilex:
I wrapped the `PositionerState` in a `Arc<Mutex<<>>`, not sure if this is the best solution.

While at it, I also made `PositionerState::get_geometry` public, as it is used to get new geometry, when we want to update popup position.

fixes: #359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Breaking changes
+
+- `XdgPopupSurfaceRoleAttributes::positioner` changed its type to Arc<Mutex<PositionerState>>
+
 ### Additions
 
 #### Clients & Protocols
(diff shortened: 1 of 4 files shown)


cmeissl:
I am not sure if the positioner state needs to be mutable. In case the positioner allows re-positioning indicated by set_reactive, wouldn't it be enough to update the geometry in the popup state and send the configure?

vberger:
Ok, so, making the positioner state mutable and automatically forwarded to popup states is almost certainly not the right thing to do, quoting the protocol specification of xdg-shell:

> At the time of the request, the compositor makes a copy of the rules specified by the xdg_positioner. Thus, after the request is complete the xdg_positioner object can be destroyed or reused; further changes to the object will have no effect on previous usages.

However, Smithay currently only implements xdg-shell version 1, and version 3 introduced a [`xdg_popup.reposition` request](https://wayland.app/protocols/xdg-shell#xdg_popup:request:reposition) that seems to be exactly the thing to use to update the positioner state beyond after it is initially set at creation by `get_popup`.

cmeissl:
I have just opened #361 which adds support for v3. In the PR the positioner state has been moved to the PopupState so that it should be handled correctly according to the configure sequence.

PolyMeilex:
Closing in favor of #361, as it implements everything that I need

PR #361, opened 2021-08-03
support for xdg-shell v3

cmeissl:
This PR adds support for `xdg_wm_base` version 3.

## Positioner

The positioner gained support for the new fields and handles the requests for:
- set_reactive
- set_parent_size
- set_parent_configure

**[Breaking]** The positioner state has been moved to the `PopupState` so that it can be updated during handling the `reposition` request. `XdgRequest::NewPopup` has been extended with the position state to make the initial geometry calculation easier.

## Re-Position

The `XdgRequest` has been extended with `RePosition` and the function `send_repositioned` has been
added to `PopupSurface`.
A simple implementation was added to anvil.

## Toplevel-States

The protocol adds new toplevel states for tiling. `send_configure` will filter them if the client protocol version
is lower that the required `SINCE`

## Popup Configure

The `send_configure` now checks for the client protocol version and returns a result indicating an error like
a re-configure is not supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+#### Clients & Protocols
+
+- `XdgPositionerState` moved to `XdgPopupState` and added to `XdgRequest::NewPopup`
+- `PopupSurface::send_configure` now checks the protocol version and returns an `Result`
+
 ### Additions
 
 #### Clients & Protocols
@@ -13,6 +20,7 @@
 - Whether a surface is toplevel equivalent can be determined with the new function `shell::is_toplevel_equivalent`.
 - Setting the parent of a toplevel surface is now possible with the `xdg::ToplevelSurface::set_parent` function.
 - Add support for the zxdg-foreign-v2 protocol.
+- Support for `xdg_wm_base` protocol version 3
 
 ### Bugfixes
 
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -125,6 +125,9 @@ pub const ZXDG_TOPLEVEL_ROLE: &str = "zxdg_toplevel";
 /// [xdg_popup]: self::XDG_POPUP_ROLE
 pub const ZXDG_POPUP_ROLE: &str = "zxdg_popup";
 
+/// Constant for toplevel state version checking
+const XDG_TOPLEVEL_STATE_TILED_SINCE: u32 = 2;
+
 macro_rules! xdg_role {
     ($configure:ty,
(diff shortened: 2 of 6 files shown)


src/wayland/shell/xdg/mod.rs
```diff
@@ -615,6 +618,28 @@ impl ToplevelStateSet {
...
+            )
+        };
+
+        let contains_tiled = self.states.iter().any(|state| is_tiled(state));
+
+        if version >= XDG_TOPLEVEL_STATE_TILED_SINCE || !contains_tiled {
+            return self.states;
+        }
+
+        self.states.into_iter().filter(|state| !is_tiled(state)).collect()
```
vberger, line 641: This is the kind of situation where we'd love `drain_filter` to be stable...

However, I think we could optimize for the happy path of the client using a recent xdg-shell version, in that case we can directly return `self.states` without even iterating the states once.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 - Whether a surface is toplevel equivalent can be determined with the new function `shell::is_toplevel_equivalent`.
 - Setting the parent of a toplevel surface is now possible with the `xdg::ToplevelSurface::set_parent` function.
 - Add support for the zxdg-foreign-v2 protocol.
+- Support for `xdg_wm_base` protocol version 3
+- **[Breaking]** `XdgPositionerState` moved to `XdgPopupState` and added to `XdgRequest::NewPopup`
+- **[Breaking]** `PopupSurface::send_configure` now checks the protocol version and returns an `Result`
```
vberger, line 18: Please put those in their own `### Breaking Changes` section, before the `### Additions` one, and maybe the two latter as sub-items of the first one, some something like that:

```
### Breaking Changes

#### Clients & Protocols

- Support for `xdg_wm_base` protocol version 3
  - `XdgPositionerState` moved to `XdgPopupState` and added to `XdgRequest::NewPopup`
  - `PopupSurface::send_configure` now checks the protocol version and returns an `Result`
```

src/wayland/shell/xdg/mod.rs
cmeissl: Yeah, not the first time drain_filter would come handy 😩
I will split the check into two and move the happy path to the top 👍

cmeissl: done

CHANGELOG.md
cmeissl: I moved the two breaking changes to a `### Breaking Changes` section. I left the support for protocol version 3 in the `Additions`. Is this ok for you?

cmeissl:
@vberger Is there anything left i should address?

vberger:
Nope, this is good, sorry for the delay. Thanks!

PR #365, opened 2021-08-10
Add Foward/Back mouse buttons

i509VCB:
This adds handling for the Forward and Back buttons on mice in the winit and libinput backends.


src/backend/winit.rs
```diff
@@ -472,7 +485,35 @@ impl BackendEvent<WinitInputBackend> for WinitMouseInputEvent {
...
+                        _ => unreachable!(),
+                    }
+                }
+
+                b => b.into(),
+            }
+        } else {
+            match self.button {
+                WinitMouseButton::Other(raw) if raw == 0x115 || raw == 0x116 => {
+                    // Wayland uses the libinput values, 0x115 for forward, 0x116 for back
```
i509VCB, line 506: Do note I need to test this assumption I made still

src/backend/winit.rs
```diff
@@ -472,7 +485,35 @@ impl BackendEvent<WinitInputBackend> for WinitMouseInputEvent {
...
+                    // X11 is 8 for back and 9 for forward.
+                    match raw {
+                        8 => MouseButton::Back,
+                        9 => MouseButton::Forward,
+                        _ => unreachable!(),
+                    }
+                }
+
+                b => b.into(),
+            }
```
vberger, line 502: We can flatten this match with patterns like `WinitMouseButton::Other(8) => MouseButton::Back`, same for the second one with Wayland.

src/backend/winit.rs
vberger: That is likely indeed, but if you can test it for confirmation that'd be great.

src/backend/winit.rs
```diff
@@ -671,6 +712,7 @@ impl InputBackend for WinitInputBackend {
             let resize_notification = &self.resize_notification;
             let logger = &self.logger;
             let window_size = &self.size;
+            let is_x11 = &self.is_x11;
```
vberger, line 715: This is a `bool`, not real point in taking a reference, just copy it.

vberger (changes requested):
The implementation itself looks pretty clean, however I'd like a clarification, because I realize there are two subjects here:

- The conversion of X11 button code to libinput/wayland ones, so that all buttons can be correctly decoded by clients when running the winit/X11 backend
- The addition of other buttons to the Smithay enum so that the compositors can easily react to them as well

Regarding those two questions, I think it'd be great if we could figure out what is the proper way to generically translate X11 button codes into libinput/wayland ones, and settle once an for all the list of relevant buttons that Smithay may reasonnably have in its enum.

psychon:
> is there some known formula for converting x11 button codes to libinput ones?

If I understand your correctly: Seems like you need the inverse of this: https://sources.debian.org/src/xserver-xorg-input-libinput/1.1.0-1/src/xf86libinput.c/?hl=1508#L216-L234
```c

static inline unsigned int
btn_linux2xorg(unsigned int b)
{
	unsigned int button;

	switch(b) {
	case 0: button = 0; break;
	case BTN_LEFT: button = 1; break;
	case BTN_MIDDLE: button = 2; break;
	case BTN_RIGHT: button = 3; break;
	/* tablet button range */
	case BTN_STYLUS: button = 2; break;
	case BTN_STYLUS2: button = 3; break;
	default:
		button = 8 + b - BTN_SIDE;
		break;
	}

	return button;
}
```
Also relevant for the above:
```
$ grep -A 3 BTN_LEFT /usr/include/linux/input-event-codes.h ; echo ;  grep -B 1 BTN_STYLUS2 /usr/include/linux/input-event-codes.h                                                                                       
#define BTN_LEFT		0x110
#define BTN_RIGHT		0x111
#define BTN_MIDDLE		0x112
#define BTN_SIDE		0x113

#define BTN_STYLUS		0x14b
#define BTN_STYLUS2		0x14c
```

PR #367, opened 2021-08-18
backend.libseat: Fix dispatch double borrow

PolyMeilex:
Queue libseat events and dispatch them after libseat dispatch is done to avoid double borrows

### DIFF
--- a/src/backend/session/libseat.rs
+++ b/src/backend/session/libseat.rs
@@ -18,7 +18,10 @@ use std::{
 
 use nix::{errno::Errno, fcntl::OFlag, unistd::close};
 
-use calloop::{EventSource, Poll, PostAction, Readiness, Token, TokenFactory};
+use calloop::{
+    channel::{self, Channel},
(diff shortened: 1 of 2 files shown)


src/backend/session/libseat.rs
```diff
@@ -193,6 +193,24 @@ impl Session for LibSeatSession {
...
+                SeatEvent::Disable => {
+                    self.internal.active.store(false, Ordering::SeqCst);
+                    self.signaler.signal(SessionSignal::PauseSession);
+                    self.internal.seat.borrow_mut().disable().unwrap();
+                }
+            }
+        }
+    }
+
     /// Creates a new session object belonging to this notifier.
```
vberger, line 214: I think this should be using a `calloop` channel, and have it registered as a sub-event source to calloop, so that it can also wake up the event source if necessary, given the whole point is that sometimes libseat can invoke the callback outside of its `dispatch()` method.

vberger (approved):
All is good, thanks!

PR #371, opened 2021-08-25
Direct X11 backend

i509VCB:
Nearing the end of this, so a draft issue.

This pull request is part of series in https://github.com/Smithay/smithay/issues/293.

The X11 backend is renderer agnostic and imports a dmabuf for presentation to a window. The X11 backend also supplies input events, and events indicating window exposure, close requests and resizing.

The TODO before this is truly ready for review:
- [ ] Hook up the x11 backend to the renderer
- [ ] Check for DRI3 extensions
- [ ] Test input implementation
- [ ] Remove stupid hardcoding
- [ ] Remove the `.expect("TODO")`s
- [ ] Clean up the feature mess in Cargo.toml
- [ ] Documentation


src/backend/x11/mod.rs
```diff
@@ -162,8 +163,30 @@ impl X11Backend {
...
+        // Does the X server support dri3?
+        let (dri3_major, dri3_minor) = {
+            let extension = xcb.query_extension(dri3::X11_EXTENSION_NAME.as_bytes())?.reply()?;
+
+            if !extension.present {
+                todo!("Missing extension")
+            }
+
+            // TODO: Figure out what on earth xcb does with these 2 params?
+            let version = xcb.dri3_query_version(1, 2)?.reply()?;
```
psychon, line 183: Since you asked on irc... well:

Here are some docs: https://github.com/freedesktop/xorg-dri3proto/blob/master/dri3proto.txt#L64

And here is the implementation: https://gitlab.freedesktop.org/xorg/xserver/-/blob/18d3131f9a332096825c09106a931c427246ddb7/dri3/dri3_request.c#L45

I don't really know xorg code, but it seems like this checks which protocol version the server (I guess this depends on the graphics driver?) can support. It then sends you this version. It does not do anything else.

You are checking for `major_version < 1`. As far as I can tell, there was never a 0.x version of dri3, so this should never trigger.

src/backend/x11/mod.rs
```diff
@@ -162,8 +163,30 @@ impl X11Backend {
         let (connection, screen_number) = XConnection::new(&logger)?;
         let connection = Arc::new(connection);
         info!(logger, "Connected to screen {}", screen_number);
-
         let xcb = connection.xcb_connection();
+
+        // Does the X server support dri3?
+        let (dri3_major, dri3_minor) = {
+            let extension = xcb.query_extension(dri3::X11_EXTENSION_NAME.as_bytes())?.reply()?;
```
psychon, line 176: I'd recommend `xcb.extension_information(dri3::X11_EXTENSION_NAME)?.is_some()`. x11rb needs the QueryExtension reply as well to send the following `DRI3` `QueryVersion` request (the reply contains the major opcode which is a byte that must be set in the request). If you use `extension_information()`, x11rb caches this reply and saves one round-trip later when it does not have to request the same information again.

(And no, this does not really make any difference. At all. One round-trip should be quite fast. So, I guess this is mostly FYI.)

src/backend/x11/buffer.rs
```diff
@@ -0,0 +1,138 @@
...
+            connection,
+            inner: xid,
+            width: dmabuf.width() as u16,
+            height: dmabuf.height() as u16,
+        })
+    }
+
+    #[allow(dead_code)]
+    pub fn present(&self, window: &Window) -> Result<(), X11Error> {
+        self.connection.copy_area(
```
psychon, line 118: No idea if it's really worth it, but eventually it might make sense to use the present extension. Although I never personally worked with it and I am not 100% sure why that would be better.

src/backend/x11/buffer.rs
```diff
@@ -0,0 +1,138 @@
...
+            self.height,
+        )?;
+
+        Ok(())
+    }
+}
+
+impl Drop for Pixmap {
+    fn drop(&mut self) {
+        let _ = self.connection.free_pixmap(self.inner);
```
psychon, line 136: In case you want to add a TODO for that (and you think that it is worth it): Next version of x11rb will bring `x11rb::protocol::xproto::PixmapWrapper` that wraps a pixmap and calls `free_pixmap` in `Drop`. So, exactly what you are doing here. It even ignores any errors, just like this code. :-)

It's funny how often I saw code like this already since these wrappers were added. I guess they really are useful...

src/backend/x11/connection.rs
```diff
@@ -0,0 +1,135 @@
...
+                (xlib.XCloseDisplay)(display);
+                return Err(ConnectToXError::NoXlibXcb);
+            }
+
+            // Do not drop the connection upon closure since Xlib created the xcb_connection_t.
+            // The Drop impl of `XConnection` will shutdown the Xlib Display
+            XCBConnection::from_raw_xcb_connection(xcb_connection_t, false)
+        };
+
+        if xcb_connection.is_err() {
```
psychon, line 88: Off topic: I was briefly thinking "how can this be an error?!?". Then I looked at the code and saw that x11rb has to parse the `.setup()` that the X11 server sent here and that can go wrong (e.g. the server sent a too short setup claiming to contain 10000 screens). libxcb just does not handle this possibility at all...

Now I am happy about all the safety guarantees that Rust provides.

No one will get this, but since you already wrote on this issue: CC @DemiMarie x11rb is the Rust library I was talking about elsewhere (in case you even care about this - if not, that's totally fine as well).

src/backend/x11/connection.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// Returns a pointer to the xlib display.
+    pub fn xlib_display(&self) -> *mut x11_dl::xlib::Display {
+        self.xlib_display
+    }
+}
+
+// Xlib and libxcb are both thread safe.
+unsafe impl Send for XConnection {}
+unsafe impl Sync for XConnection {}
```
psychon, line 116: As @DemiMarie already wrote: No, you need to call `xlib::XInitThreads()` once first. I'd recommend a global variable containing something like a once_cell for this so that this call even only happens once when another `XConnection` is constructed later on.

src/backend/x11/window.rs
```diff
@@ -0,0 +1,259 @@
...
+        xcb.change_property32(
+            PropMode::REPLACE,
+            window.inner,
+            atoms.WM_PROTOCOLS,
+            AtomEnum::ATOM,
+            &[atoms.WM_DELETE_WINDOW],
+        )?;
+
+        // Block until window creation is complete.
+        cookie.check()?;
```
psychon, line 169: ...Why? I know that everyone does this, but I never understood wait. Sure, this is not really expensive, but it is still one more round-trip to the X11 server.

And if this is about the error checking: Hm. How likely is it that creating the window fails? And the error is not silently lost anyway, it would then show up on the event loop instead...

src/backend/x11/window.rs
```diff
@@ -0,0 +1,259 @@
...
+            self.atoms.UTF8_STRING,
+            title.as_bytes(),
+        )?;
+
+        // Set WM_CLASS
+        let mut raw = Vec::new();
+        raw.extend_from_slice(title.as_bytes());
+        raw.extend_from_slice(b"\n");
+        raw.extend_from_slice(title.as_bytes());
+        raw.push(b'\n');
```
psychon, line 241: - This is half-dangerous. You are not allowed to set `WM_CLASS` while the window is mapped.
- This is wrong: `WM_CLASS` is separated by `\0` and not `\n`.
- This is wrong: I don't think that the window title is a good class and instance name. But I do not really have good suggestions either. `b"smithay\0Smithay\0"`? (If this becomes static: It can be moved to `new()` right after the window was created.)

src/backend/x11/window.rs
```diff
@@ -0,0 +1,259 @@
...
+        )?;
+
+        Ok(())
+    }
+}
+
+impl Drop for WindowInner {
+    fn drop(&mut self) {
+        let _ = self.connection.xcb_connection().destroy_window(self.inner);
+    }
```
psychon, line 258: Yay, another case where the new `WindowWrapper` might be slightly useful (I already mentioned this above for `PixmapWrapper`).

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+    connection: Arc<XConnection>,
+    source: X11Source,
+    window: Arc<WindowInner>,
+    key_counter: Arc<AtomicU32>,
+    depth: Depth,
+    visual_id: u32,
+    gbm_device: Device<RawFd>,
+}
+
+unsafe impl Send for X11Backend {}
```
psychon, line 146: Uhm... this might need a comment. Is this about `gbm_device`? Why is it not `Send` and why is your usage fine?

Why do you need to send the X11Backend between threads anyway?

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+                                        y,
+                                        size: window.size(),
+                                    },
+                                }),
+                                &mut event_window,
+                            )
+                        }
+                    }
+
+                    x11::Event::ResizeRequest(resized) => {
```
psychon, line 505: Uhm. Does this code ever trigger? You are not a window manager, you should not get ResizeRequest events. Instead, you should get ConfigureNotify events.

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+                            && client_message.window == window.inner // Same window
+                        {
+                            (callback)(X11Event::CloseRequested, &mut event_window);
+                        }
+                    }
+
+                    x11::Event::Expose(expose) => {
+                        // TODO: We would ideally use this to determine damage and render more efficiently that way.
+                        //
+                        // Although there is an Expose with damage event somewhere?
```
psychon, line 531: I don't understand the question, but I'll try to answer it anyway:

A expose event contains a rectangle that was exposed (x, y, width, height). If a non-rectangular area is exposed (e.g. Two other windows partially cover your window and now your window is brought to front. Thus, these two rectangles are exposed), then two Expose events are sent. The first one contains `.count == 1` to indicate that another event will follow.

(And if there were four rectangles, then the first one would have `count = 3`, the next one `count = 2` etc...)

src/backend/x11/buffer.rs
```diff
@@ -0,0 +1,138 @@
...
+        let xid = connection.generate_id()?;
+        let mut strides = dmabuf.strides();
+        let mut offsets = dmabuf.offsets();
+        let mut fds = Vec::new();
+
+        for handle in dmabuf.handles() {
+            // XCB closes the file descriptor after sending, so duplicate the file descriptor.
+            let fd: RawFd = fcntl::fcntl(
+                handle,
+                fcntl::FcntlArg::F_DUPFD_CLOEXEC(0), // Why is this 0?
```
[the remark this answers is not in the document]
DemiMarie: I recommend 3; this prevents the duped FD from becoming stdin, stdout, or stderr.

src/backend/x11/mod.rs
i509VCB: Seems to fire under KDE, not sure if KDE has bungled their WM

src/backend/x11/mod.rs
i509VCB: I think this was a relic of the tiny attempt I made with the egl context, but I realized it's not the best approach. Probably should remove it

src/backend/x11/mod.rs
psychon: https://www.x.org/releases/X11R7.6/doc/xproto/x11protocol.html#events:ResizeRequest 

> This event is reported to the client selecting ResizeRedirect on the window [...]

Oh, hey, you *are* a window manager. You are including `RESIZE_REDIRECT` in your event mask in WIndowInner. I missed that earlier. You should instead use `ConfigureNotify` events which are sent from `STRUCTURE_NOTIFY` which you already include.

What `ResizeRedirect` does: It basically makes your window "un-resizable". Only you can resize your own window.

src/backend/x11/buffer.rs
```diff
@@ -52,6 +52,7 @@ impl From<ConnectionError> for CreatePixmapError {
 
 #[derive(Debug)]
 pub struct Pixmap {
+    // TODO: Consider future x11rb WindowWrapper
```
psychon, line 55: The release came faster than I expected: https://docs.rs/x11rb/0.9.0/x11rb/protocol/xproto/struct.WindowWrapper.html

src/backend/x11/connection.rs
DemiMarie: I do care!  For one, this means that Smithay has a chance to have XWayland support that is robust against malicious X servers, assuming that the X server someday gets sandboxed.  That would allow for per-Flatpak X servers that run in the Flatpak sandbox, for example.

Does x11rb use libxcb internally and then validate its output, or does it do everything itself?

i509VCB: x11rb I believe decodes the events and encodes requests itself, libxcb only being used to read and write across the socket.

src/backend/x11/buffer.rs
i509VCB: The fact I somewhat rely on shared references means I would need to look over the implementation to use the wrappers since those have lifetimes.

I'll consider it once I have polished everything up on api side and some parts of implementation.

src/backend/x11/window.rs
i509VCB: I was thinking `b"Smithay\0Wayland_Compositor\0"` as a possibility?

src/backend/x11/buffer.rs
i509VCB: Pixmap was moved to wrapper, window is the contentious part.

src/backend/x11/window.rs
psychon: Sure, fine with me

src/backend/x11/connection.rs
psychon: > Does x11rb use libxcb internally and then validate its output, or does it do everything itself?

There are two options in x11rb. There is `RustConnection` that is pure rust. If you disable the `allow-unsafe-code` feature, then you lose support for FD-passing, but get all of Rust's safety guarantees always. 

With the `allow-unsafe-code` feature, there is also a `XCBConnection`. Just as @i509VCB, this only uses libxcb to send & receive raw bytes (This uses C APIs like `xcb_send_request` and `xcb_wait_for_event` / `xcb_wait_for_reply` from libxcb). Since libxcb does not provide an extra length return, the code has to look at something like three bytes of an X11 packet and figure out the length of the packet, then call the unsafe function `std::slice::from_raw_parts`. Afterwards, the same safe code is used as with `RustConnection`. I am pretty sure that these length calculations are exactly the same math as libxcb does internally, but obviously this is a bit less safe than `RustConnection`.

`XCBConnection` allows sharing the connection with C code that expects a libxcb `xcb_connection_t*` (e.g. cairo or mesa). `RustConnection` is a Rust-only thing that cannot be used in this way.

I bet no one cares, but x11rb even works on Windows. Obviously, there is no FD passing there.

i509VCB:
> libX11 is not thread-safe unless `XInitThreads` is called.

XCB FFI is no longer necessary and is now no longer included in the backend.

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,921 @@
...
+        backend: &X11Backend,
+        format: DrmFourcc,
+        resize: Receiver<Size<u16, Logical>>,
+    ) -> Result<X11Surface, X11Error> {
+        let connection = &backend.connection;
+        let window = backend.window();
+
+        // Determine which drm-device the Display is using.
+        let screen = &connection.setup().roots[backend.screen()];
+        let dri3 = connection.dri3_open(screen.root, 0)?.reply()?;
```
psychon, line 315: Could you add a comment explaining the `0`? https://github.com/freedesktop/xorg-dri3proto/blob/master/dri3proto.txt says this must be a "RandR provider". I don't think I ever heard that concept before.

Seems like you just pick the first one here and servers usually just have one:
```
$ xrandr --listproviders
Providers: number : 1
Provider 0: id: 0x47 cap: 0xf, Source Output, Sink Output, Source Offload, Sink Offload crtcs: 3 outputs: 5 associated providers: 0 name:modesetting
```
Hm. But here, the provider actually has `id: 0x47`...

Ohh, you are actually using a special case in the X11 server that is not documented in the spec:

https://gitlab.freedesktop.org/xorg/xserver/-/blob/6c1e6429bcc4447601d9327c84513421226c45b8/dri3/dri3_request.c#L138

Could you please use `x11rb::NONE` instead of the `0` then? Numerically, that is the same thing, but it makes it a tiny bit clearer that this requests expects an XID.

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,921 @@
...
+
+        let drm_device_fd = dri3.device_fd;
+        // Duplicate the drm_device_fd
```
psychon, line 318: Is it necessary to duplicate the fd? :-(

I would hope that since you own the struct, you can just move members out of it. In fact, doesn't `let drm_device_fd = dri3.device_fd;` already move the FD out? Can't you replace the `fcntl(DUPFD)` call with `dri3.device_fd.into_raw_fd()`

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,921 @@
...
+            ),
+        )
+        .map_err(AllocateBuffersError::from)?;
+
+        if get_drm_node_type_from_fd(drm_device_fd.as_raw_fd())? != DRM_NODE_RENDER {
+            todo!("Attempt to get the render device by name for the DRM node that isn't a render node")
+        }
+
+        // Finally create a GBMDevice to manage the buffers.
+        let device = gbm::Device::new(drm_device_fd.as_raw_fd()).expect("Failed to create gbm device");
```
psychon, line 343: I think this `as_raw_fd()` call doesn't do anything. At least this calls `RawFd::as_raw_fd()`.

...which makes me wonder: Who closes the FD? Looking at the code in `gbm::Device::new`, I think perhaps you want to keep x11rb's `RawFdContainer`. You can still set the `CLOEXEC` frag via `as_raw_fd()` on it. Since it implements `AsRawFd`, you should be able to just pass it as-is to `gbm::Device::new`. Only downside is that this would change the type of `device` from `Device<RawFd>` to `Device<RawFdContainer>`. But at least that would make sure the FD is actually closed when needed.

(Sorry if I misunderstood this gbm code and it actually closes the FD for you. It just really does not look like it...)

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,921 @@
...
+            height: size.h,
+            current,
+            next,
+            resize,
+        })
+    }
+
+    /// Returns a handle to the GBM device used to allocate buffers.
+    pub fn device(&self) -> gbm::Device<RawFd> {
+        self.device.clone()
```
psychon, line 374: Why the `clone()`? I would expect this to return a borrow instead of a clone.

(Also, since x11rb's `RawFdContainer` does not implement `Clone`, this would fail to compile with my suggestion above. Which I think is a good thing because it enforces proper ownership of the FD.)

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,921 @@
...
+
+    /// Returns the format of the buffers the surface accepts.
+    pub fn format(&self) -> DrmFourcc {
+        self.format
+    }
+
+    /// Returns an RAII scoped object which provides the next buffer.
+    ///
+    /// When the object is dropped, the contents of the buffer are swapped and then presented.
+    // TODO: Error type
```
psychon, line 385: Sorry, but I do not understand this todo. Same for the next function.

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,921 @@
...
+        if let Some(new_size) = self.resize.try_iter().last() {
+            self.resize(new_size)?;
+        }
+
+        Ok(Present { surface: self })
+    }
+
+    // TODO: Error type.
+    fn resize(&mut self, size: Size<u16, Logical>) -> Result<(), AllocateBuffersError> {
+        self.width = size.w;
```
psychon, line 396: Shouldn't this be better done at the end of the function where `self.current` and `self.next` are also set? Otherwise these values become wrong when creating new buffers fails and the function does an early return.

src/backend/x11/mod.rs
```diff
@@ -0,0 +1,921 @@
...
+                minor: XFIXES_MINOR_VERSION,
+            }
+            .into());
+        }
+
+        let version = connection
+            .xfixes_query_version(XFIXES_MAJOR_VERSION, XFIXES_MINOR_VERSION)?
+            .reply()?;
+
+        if version.major_version < XFIXES_MAJOR_VERSION {
```
psychon, line 861: This ignores `XFIXES_MINOR_VERSION`. Sure, that constant is currently 0, but... yeah.

I would recommend a `compare_versions` helper function:
```rust
fn compare_versions(actual: (u32, u32), expected: (u32, u32)) -> bool {
    actual.0 > expected.0 || (actual.0 == expected.0 && actual.1 >= expected.1)
}
```
Or perhaps even:
````rust
fn compare_versions(name: &'static str, actual: (u32, u32), expected: (u32, u32)) -> Result<(), MissingExtensionError> {
    if actual.0 > expected.0 || (actual.0 == expected.0 && actual.1 >= expected.1) {
        Ok(())
    } else {
        Err(MissingExtensionError::WrongVersion { ..... })
    }
}
````

File: src/backend/x11/window_inner.rs [hunk 22]
```diff
@@ -0,0 +1,238 @@
...
+
+impl From<Arc<WindowInner>> for Window {
+    fn from(inner: Arc<WindowInner>) -> Self {
+        Window(Arc::downgrade(&inner))
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct WindowInner {
+    // TODO: Consider future x11rb WindowWrapper
```
psychon, line 40: This todo is outdated and can just be removed. `WindowWrapper` borrows the connection and thus cannot be used here... :-/

src/backend/x11/window_inner.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub connection: Weak<RustConnection>,
+    pub id: x11::Window,
+    root: x11::Window,
+    pub atoms: Atoms,
+    pub size: Mutex<Size<u16, Logical>>,
+    /// Maintains the current cursor visibility state.
+    ///
+    /// Required since we cannot enable an already enabled cursor in xfixes.
+    cursor_visible: Arc<AtomicBool>,
```
psychon, line 49: I wondered what this means and found https://www.x.org/releases/X11R7.7/doc/fixesproto/fixesproto.txt saying:

> 	If the client has made no outstanding HideCursor requests a BadMatch error is generated.

Uhm. Just ignore that error?
```rust
// This generates a Match error if we did not call HideCursor before. Ignore that error.
conn.xfixes_show_cursor().ignore_error();
```
That way this whole `cursor_visible` variable can go away, I think.

src/backend/x11/window_inner.rs
```diff
@@ -0,0 +1,238 @@
...
+            | EventMask::STRUCTURE_NOTIFY
+            | EventMask::KEY_PRESS // Key press and release
+            | EventMask::KEY_RELEASE
+            | EventMask::BUTTON_PRESS // Mouse button press and release
+            | EventMask::BUTTON_RELEASE
+            | EventMask::POINTER_MOTION // Mouse movement
+            | EventMask::NO_EVENT,
+            )
+            // Border pixel and color map need to be set if our depth may differ from the root depth.
+            .border_pixel(0)
```
psychon, line 96: Instead of hardcoding a zero, I would suggest `connection.setup().black_pixel`. That will most likely still be the value 0, but... yeah.

src/backend/x11/window_inner.rs
```diff
@@ -0,0 +1,238 @@
...
+            atoms,
+            cursor_visible: Arc::new(AtomicBool::new(true)),
+            size: Mutex::new((properties.size.w, properties.size.h).into()),
+            depth,
+            gc: 0,
+        };
+
+        let gc = connection.generate_id()?;
+        connection.create_gc(gc, window.id, &CreateGCAux::new())?;
+        window.gc = gc;
```
psychon, line 127: Instead of writing `gc: 0` above, you could just do `gc: connection.generate_id()?`. That would save like two lines of code.

src/backend/x11/buffer.rs
```diff
@@ -0,0 +1,152 @@
...
+impl<'c, C> PixmapWrapperExt<'c, C> for PixmapWrapper<'c, C>
+where
+    C: Connection,
+{
+    fn with_buffer(
+        _connection: &'c C,
+        _window: &Window,
+        _buffer: &WlBuffer,
+    ) -> Result<PixmapWrapper<'c, C>, CreatePixmapError> {
+        todo!("Get the buffer data and create the appropriate buffer")
```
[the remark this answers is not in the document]
i509VCB: Yeah I intended to add this for future extension, but it is not necessary at the moment.

anvil/src/x11.rs
```diff
@@ -0,0 +1,320 @@
...
+
+                                    draw_fps(renderer, frame, fps_texture, output_scale as f64, fps)?;
+                                }
+
+                                Ok(())
+                            },
+                        )
+                        .map_err(Into::<SwapBuffersError>::into)
+                        .and_then(|x| x)
+                        .map_err(Into::<SwapBuffersError>::into)
```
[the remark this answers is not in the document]
i509VCB: I recall this unwrapping the layers of a `Result<(), Result<(), Gles2Error>>`

src/backend/x11/mod.rs
i509VCB: Forgot to remove these

src/backend/x11/mod.rs
i509VCB: Changing to `RawFdContainer` seems to cause a segfault when creating the device for some reason.

anvil/src/x11.rs
```diff
@@ -0,0 +1,320 @@
...
+                        Ok(()) => {
+                            // Unbind the buffer and now let the scope end to present.
+                            if let Err(err) = renderer.unbind() {
+                                error!(log, "Error while unbinding buffer: {}", err);
+                            }
+                        }
+
+                        Err(err) => {
+                            if let SwapBuffersError::ContextLost(err) = err {
+                                error!(log, "Critical Rendering Error: {}", err);
```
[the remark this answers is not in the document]
i509VCB: Copied from winit is the excuse here, probably not a great thing to copy either.

src/backend/x11/mod.rs
i509VCB: See comment regarding creation of gbm::Device

src/backend/x11/buffer.rs
i509VCB: Present extension is now used to allow compositors to synchronize with the display's refresh rate.

anvil/src/x11.rs
```diff
@@ -0,0 +1,342 @@
...
+            state.window_map.borrow_mut().refresh();
+            state.output_map.borrow_mut().refresh();
+        }
+    }
+
+    // Cleanup stuff
+    state.window_map.borrow_mut().clear();
+
+    // TODO: Figure out why the renderer is dropped later than everything else and therefore segfaults?
+    drop(renderer);
```
i509VCB, line 341: This is dumb

psychon: Doesn't crash for me if I remove the `drop()`, but I needed the following to make things work (I get a primary node here, not a render node):
```diff
diff --git a/src/backend/x11/mod.rs b/src/backend/x11/mod.rs
index 7b6c67c7d..198c7e4ab 100644
--- a/src/backend/x11/mod.rs
+++ b/src/backend/x11/mod.rs
@@ -309,9 +309,11 @@ impl X11Surface {
         )
         .map_err(AllocateBuffersError::from)?;
 
+        /*
         if get_drm_node_type_from_fd(drm_device_fd.as_raw_fd())? != DRM_NODE_RENDER {
             todo!("Attempt to get the render device by name for the DRM node that isn't a render node")
         }
+        */
 
         // Finally create a GBMDevice to manage the buffers.
         let device = gbm::Device::new(drm_device_fd.as_raw_fd()).expect("Failed to create gbm device");
```

i509VCB (commented):
Reviewing my own code 🤔

src/backend/x11/drm.rs
```diff
@@ -1,21 +1,20 @@
 //! Utilities for checking properties of a drm device.
+//!
+//! Nearly everything in this module is copied from xf86drm.h/c. Ideally we will
```
psychon: Seems like you didn't get to finish your sentence here :-)

PR #377, opened 2021-09-02
anvil: Pass window and output maps using dispatch data AnvilState

i509VCB:
I noticed the xdg_shell and wl_shell cloned the window and output maps into the closures, when it is possible to get these two things from the `AnvilState` via `DispatchData`

99.9% of the file change is just `cargo fmt`


vberger:
Following this, do the window map and output map still need to be in a `Rc<RefCell<_>>` ?

PolyMeilex:
Move and resize grabs need it to be in Rc

vberger (approved):

PR #380, opened 2021-09-17
wayland/seat: Add `KeysymHandle` to allow for keycode conversions.

Drakulix:
Currently keycodes are always converted using xkbcommons
`State::key_get_one_sym` function. This may be not what the compositor
wants, e.g. if it represents keybindings with explicit modifiers.
Applying Shift in this case changes the sym, making it necessary for
the compositor to *undo* this transformation, which is hard or even
impossible and very unnecessary, when we have all the necessary
information in smithay.

Therefor this commit replaces the `Keysym` argument of the filter
closure with a `KeysymHandle`, which allows for different variants of
keysyms to be received. Modified (as previously), unmodified or even
as a raw keycode.

### DIFF
--- a/src/wayland/seat/keyboard.rs
+++ b/src/wayland/seat/keyboard.rs
@@ -282,6 +282,47 @@ struct KbdRc {
     logger: ::slog::Logger,
 }
 
+/// Handle to the underlying keycode to allow for different conversions
+pub struct KeysymHandle<'a> {
+    keycode: u32,
+    keymap: &'a xkb::Keymap,
+    state: &'a xkb::State,
+}
+
+impl<'a> fmt::Debug for KeysymHandle<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.keycode)
+    }
+}
+
+impl<'a> KeysymHandle<'a> {
+    /// Returns the sym for the underlying keycode with all modifications by the current keymap state applied.
+    ///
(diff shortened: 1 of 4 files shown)


src/wayland/seat/keyboard.rs
```diff
@@ -313,22 +354,24 @@ impl KeyboardHandle {
     /// to be compared against. This includes non-character keysyms, such as XF86 special keys.
     pub fn input<F>(&self, keycode: u32, state: KeyState, serial: Serial, time: u32, filter: F)
     where
-        F: FnOnce(&ModifiersState, Keysym) -> bool,
+        F: FnOnce(&ModifiersState, KeysymHandle<'_>) -> bool,
```
vberger, line 357: This is only slightly related, but looking back at that that API, I wonder if it would make sense to have this closure instead return an enum `{ Forward, Intercept(T) }`, and have the `T` value be returned by the `input()` method as an `Option<T>` ?

This would allow compositors to just return the chosen action and process it afterwards, rather than either processing it from withing the callback or storing it in a captured variable (like we do in anvil).

That's a small QoL improvement I suppose, so maybe it's not worth it. What do you think?

vberger (approved):

src/wayland/seat/keyboard.rs
Drakulix: I have added it. It removed less lines from anvil then I anticipated, but it is still a better api imo. I am just not 100% sure about the name of the enum.

PR #394, opened 2021-10-08
Implement XDG-Decoration Protocol

PolyMeilex:
I'm not sure how to check this:
```
Creating an xdg_toplevel_decoration from an xdg_toplevel which has a buffer attached or committed is a client error, 
and any attempts by a client to attach or manipulate a buffer prior 
to the first xdg_toplevel_decoration.configure event must also be treated as errors.
```


PolyMeilex:
> Would it be possible to also implement `server-decoration`?

For sure, that's my plan, but for another PR, as xdg decoration is based on xdg toplevels, and kde's server decoration is based on wl-surfaces.

PR #403, opened 2021-10-23
Refactor Renderer API

Drakulix:
First set of changes discussed on matrix.

For motivation see https://github.com/Smithay/smithay/discussions/363#discussioncomment-1505061 .
TL;DR: A software renderer (and potentially others) cannot support the matrix style `render_texture` api, so lets move that into the `Gles2Renderer`.

Feedback welcome.


vberger:
@Drakulix do you have some time to fix that last doc comment and rebase, so that we can merge this?

Drakulix:
@vberger *Should* be good now!

PR #409, opened 2021-11-04
gbm: use Xrgb8888, if alpha variant is not available

### LINKED ISSUE #408: Accept non-alpha variants of surface formats for primary plane
Quoted from Matrix room:

> So it seems your drm-device (older intel gpu?) supports no alpha color formats.

> Which is not really a problem, because it supports 4-bit formats, where it ignores the alpha channel and because a primary plane cannot be transparent anyway, we should need to accept rendering an alpha-format to a primary-plane supporting the matching non-alpha variant

> So it fix it, we need to adjust the code here, so that it does not do a simple intersection: https://github.com/Smithay/smithay/blob/master/src/backend/drm/surface/gbm.rs#L112 But instead converts alpha formats to non-alpha variants.

> Or add add this here: https://github.com/Smithay/smithay/blob/master/src/backend/drm/surface/mod.rs#L365
If the requested plane is a PlaneType::Primary also add all non-alpha-variants, so the gbm code can later select one for rendering.

Drakulix:
Hopefully fixes #408 

@vurpo Can you please test this branch on your system?

### DIFF
--- a/src/backend/drm/surface/gbm.rs
+++ b/src/backend/drm/surface/gbm.rs
@@ -62,16 +62,21 @@ where
         // Once we have proper color management and possibly HDR support,
         // we need to have a more sophisticated picker.
         // (Or maybe just pick ARGB2101010, if available, we will see.)
-        let code = Fourcc::Argb8888;
+        let mut code = Fourcc::Argb8888;
         let logger = crate::slog_or_fallback(log).new(o!("backend" => "drm_render"));
 
         // select a format
-        let plane_formats = drm
+        let mut plane_formats = drm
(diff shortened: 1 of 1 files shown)


Drakulix:
@vberger Although no one could confirm (yet), that this fixes anything, I would still merge this.
It is probably a good idea and at least does not regress any of my systems. Any objections?

vurpo:
Yes, this fixes the linked issue. Anvil starts correctly in tty-udev mode on my system.

Drakulix:
Thanks for reporting and confirming!

PR #410, opened 2021-11-08
Add KDE server-decoration support

chrisduerr:
This should add a very rough implementation of the server decoration protocol. I've tested it and it seems to work as far as I can tell. My priority was mostly to implement this for myself to force server-side decorations for GTK, this PR is more of an accident.

It's probably somewhat rough around the edges, especially since I'm not particularly familiar with all the wrappers like `Global`/`Main`/`Resource` etc, so please let me know if you have any suggestions for improvements.

### DIFF
--- a/src/wayland/shell/kde/decoration.rs
+++ b/src/wayland/shell/kde/decoration.rs
@@ -72,7 +72,7 @@ pub trait KdeDecorationHandler {
     }
 
     /// Handle decoration object removal for a surface.
-    fn release(&mut self, _surface: &WlSurface) {}
+    fn release(&mut self, _decoration: &OrgKdeKwinServerDecoration, _surface: &WlSurface) {}
 }
 
 /// KDE server decoration state.
--- a/src/wayland/shell/kde/decoration.rs
+++ b/src/wayland/shell/kde/decoration.rs
@@ -62,14 +62,17 @@ pub trait KdeDecorationHandler {
     /// decoration request.
     ///
     /// **It is up to the compositor to prevent feedback loops**, a client is free to ignore modes
-    /// suggested by [`OrgKdeKwinServerDecoration::mode`] and instead keep requesting their
-    /// preferred mode instead.
+    /// suggested by [`OrgKdeKwinServerDecoration::mode`] and instead request their preferred mode
+    /// instead.
     fn request_mode(
         &mut self,
         _surface: &WlSurface,
-        _decoration: &OrgKdeKwinServerDecoration,
-        _mode: WEnum<Mode>,
+        decoration: &OrgKdeKwinServerDecoration,
+        mode: WEnum<Mode>,
     ) {
+        if let WEnum::Value(mode) = mode {
+            decoration.mode(mode);
+        }
     }
 
     /// Handle decoration object removal for a surface.
--- a/src/wayland/shell/kde/decoration.rs
+++ b/src/wayland/shell/kde/decoration.rs
@@ -1,0 +1,129 @@
+//! KDE Window decoration manager
+//!
+//! This interface allows a compositor to announce support for KDE's legacy server-side decorations.
+//!
+//! A client can use this protocol to request being decorated by a supporting compositor.
+//!
+//! ```
+//! extern crate wayland_server;
+//! extern crate smithay;
+//!
+//! use smithay::delegate_kde_decoration;
+//! use smithay::wayland::shell::kde::decoration::{KdeDecorationHandler, KdeDecorationState};
+//!
+//! # struct State { kde_decoration_state: KdeDecorationState };
+//! # let mut display = wayland_server::Display::<State>::new().unwrap();
+//!
+//! // Create the new KdeDecorationState.
+//! let state = KdeDecorationState::new::<State, _>(&display.handle(), None);
+//!
+//! // Insert KdeDecorationState into your compositor state.
+//! // …
+//!
+//! // Implement KDE server decoration handlers.
+//! impl KdeDecorationHandler for State {
+//!     fn kde_decoration_state(&self) -> &KdeDecorationState {
+//!         &self.kde_decoration_state
+//!     }
+//! }
+//!
+//! delegate_kde_decoration!(State);
+//! ```
+
+use slog::Logger;
+use wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decoration::{
+    Mode, OrgKdeKwinServerDecoration,
+};
+use wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decoration_manager::OrgKdeKwinServerDecorationManager;
+use wayland_server::backend::GlobalId;
+use wayland_server::protocol::wl_surface::WlSurface;
+use wayland_server::{Dispatch, DisplayHandle, GlobalDispatch, WEnum};
+
+/// KDE server decoration handler.
+pub trait KdeDecorationHandler {
+    /// Return the KDE server decoration state.
+    fn kde_decoration_state(&self) -> &KdeDecorationState;
+
+    /// Handle decoration manager creation.
+    ///
+    /// Allows setting up the decoration manager, like setting the default decoration mode using
+    /// [`OrgKdeKwinServerDecorationManager::default_mode`].
+    fn setup(&mut self, _kde_decoration_manager: &OrgKdeKwinServerDecorationManager) {}
+
+    /// Handle new decoration object creation.
+    ///
+    /// Called whenever a new decoration object is created, usually this happens when a new window
+    /// is opened.
+    fn new_decoration(&mut self, _surface: &WlSurface, _decoration: &OrgKdeKwinServerDecoration) {}
+
+    /// Handle surface decoration mode requests.
+    ///
+    /// Called when a surface requests a specific decoration mode or acknowledged the compositor's
+    /// decoration request.
+    ///
+    /// **It is up to the compositor to prevent feedback loops**, a client is free to ignore modes
+    /// suggested by [`OrgKdeKwinServerDecoration::mode`] and instead keep requesting their
+    /// preferred mode instead.
+    fn request_mode(
+        &mut self,
+        _surface: &WlSurface,
+        _decoration: &OrgKdeKwinServerDecoration,
+        _mode: WEnum<Mode>,
+    ) {
+    }
+
+    /// Handle decoration object removal for a surface.
+    fn release(&mut self, _surface: &WlSurface) {}
+}
+
+/// KDE server decoration state.
+#[derive(Debug)]
+pub struct KdeDecorationState {
+    pub(crate) logger: Logger,
+    kde_decoration_manager: GlobalId,
+}
+
+impl KdeDecorationState {
+    /// Create a new KDE server decoration global.
+    pub fn new<D, L>(display: &DisplayHandle, logger: L) -> Self
+    where
+        D: GlobalDispatch<OrgKdeKwinServerDecorationManager, ()>
+            + Dispatch<OrgKdeKwinServerDecorationManager, ()>
+            + Dispatch<OrgKdeKwinServerDecoration, WlSurface>
+            + KdeDecorationHandler
+            + 'static,
+        L: Into<Option<Logger>>,
+    {
+        let kde_decoration_manager = display.create_global::<D, OrgKdeKwinServerDecorationManager, _>(1, ());
+        let logger =
+            crate::slog_or_fallback(logger).new(slog::o!("smithay_module" => "kde_decoration_handler"));
+
+        Self {
+            kde_decoration_manager,
+            logger,
+        }
+    }
+
+    /// Returns the id of the [`OrgKdeKwinServerDecorationManager`] global.
+    pub fn global(&self) -> GlobalId {
+        self.kde_decoration_manager.clone()
+    }
+}
+
+#[allow(missing_docs)] // TODO
+#[macro_export]
+macro_rules! delegate_kde_decoration {
+    ($(@<$( $lt:tt $( : $clt:tt $(+ $dlt:tt )* )? ),+>)? $ty: ty) => {
+        $crate::reexports::wayland_server::delegate_global_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
+            $crate::reexports::wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decoration_manager::OrgKdeKwinServerDecorationManager: ()
+        ] => $crate::wayland::shell::kde::decoration::KdeDecorationState);
+
+        $crate::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
+            $crate::reexports::wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decoration_manager::OrgKdeKwinServerDecorationManager: ()
+        ] => $crate::wayland::shell::kde::decoration::KdeDecorationState);
+
+        $crate::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
+            $crate::reexports::wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decoration::OrgKdeKwinServerDecoration: $crate::reexports::wayland_server::protocol::wl_surface::WlSurface
+        ] => $crate::wayland::shell::kde::decoration::KdeDecorationState);
+    };
+}
--- a/src/wayland/shell/kde/decoration.rs
+++ b/src/wayland/shell/kde/decoration.rs
@@ -34,7 +34,9 @@ use slog::Logger;
 use wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decoration::{
     Mode, OrgKdeKwinServerDecoration,
 };
-use wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decoration_manager::OrgKdeKwinServerDecorationManager;
+use wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decoration_manager::{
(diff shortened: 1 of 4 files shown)


chrisduerr:
Another thing I forgot: I'd like to hold off on the Anvil implementation until the API is signed-off on, just to prevent doing it twice.

src/wayland/shell/legacy/mod.rs
```diff
@@ -64,6 +64,7 @@ use super::PingError;
 /// The role of a wl_shell_surface.
 pub const WL_SHELL_SURFACE_ROLE: &str = "wl_shell_surface";
 
+pub mod decoration;
```
i509VCB, line 67: I think this should be `kde_decoration` to make it clear that it is kde's decoration protocol.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+use wayland_server::protocol::wl_surface::WlSurface;
+use wayland_server::{Display, Filter, Global, Main};
+
+/// Events generated by the kde decoration manager.
+#[derive(Debug)]
+pub enum KdeDecorationRequest {
+    /// Decoration manager setup completion.
+    Setup {
+        /// KDE decoration manager.
+        manager: OrgKdeKwinServerDecorationManager,
```
i509VCB, line 43: off topic: KDE seems to have won the competition for most redundant interface names.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+where
+    L: Into<Option<slog::Logger>>,
+    Impl: FnMut(KdeDecorationRequest) + 'static,
+{
+    let callback = Rc::new(RefCell::new(implementation));
+    display.create_global(
+        1,
+        Filter::new(
+            move |(decoration_manager, _): (Main<OrgKdeKwinServerDecorationManager>, _), _, _| {
+                (&mut *callback.borrow_mut())(KdeDecorationRequest::Setup {
```
i509VCB, line 91: I've noticed that when the interface is bound, the `org_kde_kwin_server_decoration_manager::default_mode` event is never sent to clients. The specificiation for the kde decorations says the following:

> This event is emitted directly after binding the interface. It contains the default mode for the decoration. When a new server decoration object is created this new object will be in the default mode until the first request_mode is requested.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+/// Create a new KDE Decoration Manager global
+///
+/// A compositor can set the default mode for new clients at any time using
+/// [`OrgKdeKwinServerDecorationManager::default_mode`]. For individual windows the decorations can
+/// be set using [`OrgKdeKwinServerDecoration::mode`]. Both of these requests will cause the client
+/// to respond with its desired mode using [`KdeDecorationRequest::RequestMode`], which then should
+/// be acknowledged by the compositor using another [`OrgKdeKwinServerDecoration::mode`] event.
+/// **It is up to the compositor to prevent feedback loops**, a client is free to ignore the
+/// suggested decoration mode.
+pub fn init_kde_decoration_manager<L, Impl>(
```
i509VCB, line 77: I think it would make more sense to have init function take the default mode as a parameter. The return value can be adjusted to also return some object that may be used to change this default mode at any time.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+    display: &mut Display,
+    implementation: Impl,
+    _logger: L,
```
i509VCB, line 80: I think this logger deserves a little use, it's trying so hard to be there!

I guess some trace or debug level messages would be fine for some debugging if needed.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+            move |(decoration_manager, _): (Main<OrgKdeKwinServerDecorationManager>, _), _, _| {
+                (&mut *callback.borrow_mut())(KdeDecorationRequest::Setup {
+                    manager: decoration_manager.deref().clone(),
+                });
+
+                let callback = callback.clone();
+                decoration_manager.quick_assign(move |_, request, _| {
+                    let (id, surface) = match request {
+                        ServerDecorationManagerRequest::Create { id, surface } => (id, surface),
+                        _ => return,
```
i509VCB, line 99: Wow the KDE protocol has no destructor for the global, that's going to be fun.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+    display.create_global(
+        1,
+        Filter::new(
+            move |(decoration_manager, _): (Main<OrgKdeKwinServerDecorationManager>, _), _, _| {
+                (&mut *callback.borrow_mut())(KdeDecorationRequest::Setup {
+                    manager: decoration_manager.deref().clone(),
+                });
+
+                let callback = callback.clone();
+                decoration_manager.quick_assign(move |_, request, _| {
```
i509VCB, line 96: We should probably assign the filter to the decoration manager object before we send the Setup event to users of smithay.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+                    let callback = callback.clone();
+                    id.quick_assign(move |_, request, _| match request {
+                        ServerDecorationRequest::RequestMode { mode } => {
+                            (&mut *callback.borrow_mut())(KdeDecorationRequest::RequestMode {
+                                decoration: decoration.clone(),
+                                surface: surface.clone(),
+                                mode,
+                            });
+                        }
+                        ServerDecorationRequest::Release => {
```
i509VCB, line 117: Because the `release` request is a destructor, we can empty this match statement and instead place it in a `id.assign_destructor` or whatever it is called. See xdg-decoration, xdg-foreign and probably a couple other protocols for an example of that.

src/wayland/shell/legacy/decoration.rs
i509VCB, line 99: There is no other request, we can place an `unreachable!()` there.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+/// **It is up to the compositor to prevent feedback loops**, a client is free to ignore the
+/// suggested decoration mode.
+pub fn init_kde_decoration_manager<L, Impl>(
+    display: &mut Display,
+    implementation: Impl,
+    _logger: L,
+) -> Global<OrgKdeKwinServerDecorationManager>
+where
+    L: Into<Option<slog::Logger>>,
+    Impl: FnMut(KdeDecorationRequest) + 'static,
```
i509VCB, line 84: Make sure you add `DispatchData<'_>` to the callback so that compositors can access their global state.

src/wayland/shell/legacy/decoration.rs
```diff
@@ -0,0 +1,128 @@
...
+        Filter::new(
+            move |(decoration_manager, _): (Main<OrgKdeKwinServerDecorationManager>, _), _, _| {
+                (&mut *callback.borrow_mut())(KdeDecorationRequest::Setup {
+                    manager: decoration_manager.deref().clone(),
+                });
+
+                let callback = callback.clone();
+                decoration_manager.quick_assign(move |_, request, _| {
+                    let (id, surface) = match request {
+                        ServerDecorationManagerRequest::Create { id, surface } => (id, surface),
```
i509VCB, line 98: Definitely a little annoying KDE's protocol does not restrict clients from binding it more than once. Also since the protocol marks it as `new_id`, we get forced to have a new object.

i509VCB, line 98: Because KDE's decoration protocol works in terms of `wl_surface`, I think it would be worth it to add a sort of user data storing information about a surface's current mode and updating that value as needed from the server requesting mode changes for a client asking for a different mode.

Uhh I can't recall the specifics of how we would get a user data map from the `WlSurface` at the moment, I imagine someone on matrix can figure that would or explain it further.

src/wayland/shell/legacy/mod.rs
chrisduerr: Oh yeah, that's definitely one thing I wasn't sure about: Where to put it.

It's legacy, but it's not part of the legacy wl_shell. Could create a `decoration` module in the root but then I'd expect the XDG protocol to be found there too.

src/wayland/shell/legacy/decoration.rs
chrisduerr: That's a good point, I've never considered downstream just not setting this in response to the setup event. I've played around with just passing `default_mode` too but I wanted to be able to override it in the future. Both could be done, but maybe it's just easier to always set the default mode at startup and scap the Setup part, since this is the old protocol anyway?

src/wayland/shell/legacy/decoration.rs
chrisduerr: Just copied this from the XDG implementation really and thought the same. I'd assume there it at least was there to future-proof the protocol but assuming this gets merged without bugs there likely won't be any future changes that would require a logger.

Just adding some debug output might make sense, otherwise there's not much reason to have it.

src/wayland/shell/legacy/decoration.rs
chrisduerr: Why?

src/wayland/shell/legacy/decoration.rs
chrisduerr: I saw that assign_destructor call but I'm not sure what it does beyond offering a place for implementing custom destructor logic. Is it also required to free the object itself?

src/wayland/shell/legacy/decoration.rs
chrisduerr: Seems a bit counter-intuitive to defining the enums as non_exhaustive in the first place, but I guess it makes it clear to the reader that there shouldn't be any unimplemented branches.

src/wayland/shell/legacy/decoration.rs
chrisduerr: What would this be for? As a way to debounce duplicate events, potentially solve the feedback loop problem in smithay so the compositor doesn't have to worry?

src/wayland/shell/legacy/decoration.rs
PolyMeilex: > Seems a bit counter-intuitive to defining the enums as non_exhaustive in the first place, but I guess it makes it clear to the reader that there shouldn't be any unimplemented branches. Feels like you'd want a compile-time check for that stuff.

It's non_exhaustive because the protocol is unstable (and also unstandardized as a bonus)

chrisduerr: Yeah, I get the concept just that in this instance I'd assume things shouldn't change because XDG's protocol is the go-to for changes.

src/wayland/shell/legacy/mod.rs
Drakulix: I would put this under `wayland/shell/kde_decoration`.

As long as there are client, that are actively using it, it is (sadly) not legacy. And it belongs to shells, but is not directly related to either `wl_shell` or `xdg_shell`.

src/wayland/shell/legacy/decoration.rs
Drakulix: So the "best" approach would be the most versatile:
- Provide a default-mode at `init`
- But return a handle that allows it too be changed later.

I am not sure, if this is worth for such an old protocol, but regarding @i509VCB 's comment, it should always be as hard as possible for downstream to violate/miss-use a protocol as sensibly possible. So sending the default_mode automatically at setup should be a thing.

src/wayland/shell/legacy/decoration.rs
Drakulix: Maybe log default_mode changes (if implemented)?

src/wayland/shell/legacy/decoration.rs
Drakulix: Pinging @vberger for this question.

src/wayland/shell/legacy/decoration.rs
Drakulix: Yeah basically. Also this is some kind of book-keeping, that might be useful for many compositors. So if we have the opportunity to do this for the user (and rather cheap), we would want to provide that, instead of letting downstream track the surface state manually.

src/wayland/shell/legacy/mod.rs
chrisduerr: How would you feel about `shell/kde/decoration` to match the `shell/xdg/decoration`, which could also be used for other KDE-created protocols? I'm not sure if there are any other KDE protocols or if there will be in the future, but it's definitely an option.

src/wayland/shell/legacy/decoration.rs
chrisduerr: Would depend on *how* it is implemented. If downstream has direct access to the `OrgKdeKwinServerDecorationManager` then default mode can be changed without smithay ever getting any chance to log that. Though the default mode set at startup could still be logged.

src/wayland/shell/legacy/mod.rs
Drakulix: Well according to [wayland.app](https://wayland.app/protocols/) there are many. I am not sure, if we want to support a lot of them, but a `kde` module will certainly make this easier later.

src/wayland/shell/legacy/decoration.rs
vberger: The main difference is that the closure given to `assign_destructor` is also invoked if the client disconnects without explicitly destroying the objects (such as when forcefully disconnected by a protocol error for example).

chrisduerr:
I've already stated it on IRC, but I wanted to make it public here too: I currently don't plan on putting more effort into this PR, because this is basically only needed for GTK so working on the GTK patch would likely be more time-efficient.

That said, if someone else has more interest in this, feel free to clean it up. It should work as intended for anyone just trying to apply a temporary band-aid solution.

I'll leave it up to @Drakulix / @vberger if you want to keep this PR open or not. Probably going to be easier for people to find if it's still open.

PolyMeilex:
Speaking of GTK patch, if someone is dedicated enough to work on it, here is an MR: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/2191 that can be used as a starting point (it works as is, but it needs some work to be merged). This solution would work for both GTK3 and GTK4.

I personally will not work on neither Smithay nor GTK solution, as GTK4 + libadwaita combo that will soonish become the main way of developing GTK apps, will not support SSD title bar  anyway (as libadwaita window does not have title bar by default, it is added in app source code, and therefore we don't have control over it), maybe there is some way to design nice API for libadwaita to accommodate this, but I don't feel like working on it, especially because it reached 1.0 recently. So in libadwaita app enabling SSD would basically hide shadows and that's it, I'd rather respect it's CSD preference instead.

chrisduerr:
> I personally will not work on neither Smithay nor GTK solution, as GTK4 + libadwaita combo that will soonish become the main way of developing GTK apps, will not support SSD title bar anyway

Do you know if it still supports hiding the controls for tiled/maximized windows somehow?

chrisduerr:
Rewrote everything from scratch to port my fork to 0.30.0. Since this forces a more standardized API, this PR might actually be mergeable now.

I don't plan to implement Anvil support, though it should be pretty trivial for someone else to do it if they're interested. But since this is basically just for legacy-support anyway I'm not sure if Anvil would want this.

src/wayland/shell/kde/decoration.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    /// Handle surface decoration mode requests.
+    ///
+    /// Called when a surface requests a specific decoration mode or acknowledged the compositor's
+    /// decoration request.
+    ///
+    /// **It is up to the compositor to prevent feedback loops**, a client is free to ignore modes
+    /// suggested by [`OrgKdeKwinServerDecoration::mode`] and instead keep requesting their
+    /// preferred mode instead.
+    fn request_mode(
```
i509VCB, line 67: no-op is incorrect per the semantics of the protocol:

> The client can request a different mode through the decoration request. The server will acknowledge this by another event with the same mode. So even if a server prefers server-side decoration it's possible to force a client-side decoration.

src/wayland/shell/kde/decoration.rs
```diff
@@ -0,0 +1,129 @@
...
+/// KDE server decoration handler.
+pub trait KdeDecorationHandler {
+    /// Return the KDE server decoration state.
+    fn kde_decoration_state(&self) -> &KdeDecorationState;
+
+    /// Handle decoration manager creation.
+    ///
+    /// Allows setting up the decoration manager, like setting the default decoration mode using
+    /// [`OrgKdeKwinServerDecorationManager::default_mode`].
+    fn setup(&mut self, _kde_decoration_manager: &OrgKdeKwinServerDecorationManager) {}
```
i509VCB, line 51: I'd prefer this to be something set in the `KdeDecorationState`. Unless it would be desirable to have per client decoration defaults.

src/wayland/shell/kde/decoration.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn request_mode(
+        &mut self,
+        _surface: &WlSurface,
+        _decoration: &OrgKdeKwinServerDecoration,
+        _mode: WEnum<Mode>,
+    ) {
+    }
+
+    /// Handle decoration object removal for a surface.
+    fn release(&mut self, _surface: &WlSurface) {}
```
i509VCB, line 76: Is it possible to provide the handle here?

i509VCB (changes requested):

src/wayland/shell/kde/decoration.rs
chrisduerr: What handle?

i509VCB: The kde decoration protocol object

PolyMeilex (approved):
LGTM

vberger (approved):

PR #411, opened 2021-11-09
Add version method to ToplevelSurface

chrisduerr:
Currently the `shell_surface` is not accessible to the outside, meaning
there is no way to query for the XDG shell protocol version supported by
the client.

Adding a `version` method fixes that problem, which allows compositors
to act differently based on the supported protocol version. As an
example this allows setting the state to `State::Maximized` if the
tiling states are not available.


chrisduerr:
Shortly discussed this with @vberger on IRC, who indicated there currently is no way to do this and I didn't just miss something.

An alternative solution would be to expose the shell_surface in some way, though this should be a relatively minimal patch.

vberger (approved):

PR #414, opened 2021-11-10
Add SubAssign for Size

chrisduerr:


vberger (approved):

src/utils/geometry.rs
```diff
@@ -637,6 +637,14 @@ impl<N: AddAssign, Kind> AddAssign for Size<N, Kind> {
     }
 }
 
+impl<N: SubAssign, Kind> SubAssign for Size<N, Kind> {
+    #[inline]
+    fn sub_assign(&mut self, rhs: Self) {
+        self.w -= rhs.w;
+        self.h -= rhs.h
+    }
+}
```
vberger, line 646: Maybe add a `debug_assert!()` to ensure the invariant that a `Size<..>` always has positive coordinates, like is done on the `from()` method for example.

vberger (changes requested):

src/utils/geometry.rs
chrisduerr: I've added a little debug output, however this introduces additional trait constraints. These should be fine for most number types though, let me know what you think.

vberger (approved):

PR #415, opened 2021-11-11
use the latest pending server state if available instead of the current state

cmeissl:
This PR fixes an Issue where consecutively calling the following code
could loose some configured state if the state is not acked by the client
between the calls.

```
ToplevelSurface::with_pending_state(|state| {
                    ...
                });
ToplevelSurface::send_configure();
```

The fix uses the latest non acked configure when available to accumulate
the pending state changes.

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -981,6 +981,27 @@ impl ToplevelSurface {
                     .lock()
                     .unwrap();
                 if let Some(pending) = self.get_pending_state(&mut *attributes) {
+                    // Retrieve the last configured decoration mode
+                    // by checking the last non acked configure,
+                    // if no pending is available the last acked
+                    // and finally fall back to the current state.
+                    // This is necessary as send_configure could be
+                    // called before a client ack's or commits the
+                    // last state. Using the current state could lead
+                    // to unnecessary decoration configures sent to clients.
+                    //
+                    // We have to do this check before adding the pending state
+                    // to the pending configures.
+                    let current_decoration_mode = attributes
+                        .pending_configures
+                        .last()
+                        .map(|c| &c.state)
+                        .or_else(|| attributes.last_acked.as_ref())
+                        .unwrap_or(&attributes.current)
+                        .decoration_mode;
+
+                    let decoration_mode_changed = current_decoration_mode != pending.decoration_mode;
+
                     let configure = ToplevelConfigure {
                         serial: SERIAL_COUNTER.next_serial(),
                         state: pending,
@@ -989,25 +1010,14 @@ impl ToplevelSurface {
                     attributes.pending_configures.push(configure.clone());
                     attributes.initial_configure_sent = true;
 
-                    Some(configure)
+                    Some((configure, decoration_mode_changed))
                 } else {
                     None
                 }
             })
             .unwrap_or(None);
-            if let Some(configure) = configure {
-                let decoration_mode = compositor::with_states(surface, |states| {
-                    let attributes = states
-                        .data_map
-                        .get::<Mutex<XdgToplevelSurfaceRoleAttributes>>()
-                        .unwrap()
-                        .lock()
-                        .unwrap();
-                    attributes.current.decoration_mode
-                })
-                .unwrap_or(None);
-
-                if configure.state.decoration_mode != decoration_mode {
+            if let Some((configure, decoration_mode_changed)) = configure {
+                if decoration_mode_changed {
                     if let Some(data) = self
                         .shell_surface
                         .as_ref()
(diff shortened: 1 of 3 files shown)


src/wayland/shell/xdg/mod.rs
```diff
@@ -209,7 +271,11 @@ macro_rules! xdg_role {
 }
 
 xdg_role!(
-    ToplevelConfigure,
+    ToplevelState,
+    /// A configure message for toplevel surfaces
+    #[derive(Debug, Clone)]
+    ToplevelConfigure {
+    },
```
i509VCB, line 278: nit: empty braces?

cmeissl: Yeah, the macro i defined expects them cause `XdgPopup` requires additional fields. I thought about changing the macro, but decided against it to make the macro not more complicated. But i agree that does not look so nice. I changed the macro to allow zero or more repetitions. Now it is possible to write some funny looking invocations, but i am not aware of an alternative without duplicating macro code. And as this macro is not exported it should not matter.

src/wayland/shell/xdg/mod.rs
```diff
@@ -189,18 +210,59 @@ macro_rules! xdg_role {
...
+
+            /// Gets the latest state that has been configured
+            /// on the server. The state can include changes
+            /// that have been made on the server but not yet
+            /// acked or committed by the client. It does not
+            /// include the current pending state.
+            ///
+            /// This can be used for example to check if the
+            /// pending state is different from the last configured.
+            fn current_server_state(&self) -> &$state {
```
vberger, line 222: Given the documentation on this method, I guess it is intended to be `pub` ?

If so, please split the first sentence on its own line with an empty line, to ensure rustdoc renders it nicely.

vberger (commented):
LGTM apart for a small nit. :+1:

src/wayland/shell/xdg/mod.rs
cmeissl: I was not sure if the two functions should be `pub` but added the doc anyway. If they provide value i am happy to make the pub and fix the small doc issue.

vberger: It seems that this could be useful to have this method public, yes.

cmeissl: The two methods are now public and i changed the docs a bit. Additionally i found two doc issues which i fixed.
The link to `send_configure` was wrong and the link to the `xdg` module in `shell` was broken.

src/wayland/shell/xdg/mod.rs
```diff
@@ -240,12 +243,13 @@ macro_rules! xdg_role {
...
+            /// not been sent to the client.
             ///
-            /// This differs to just checking if the server pending
-            /// state is some in that it also check if a pending
-            /// state is different from the current server state.
-            fn has_pending_changes(&self) -> bool {
+            /// This differs to just checking if the [`server_pending`](#structfield.server_pending)
+            /// state is [`Some`] in that it also checks if a current pending
+            /// state is different from the [`current_server_state`](#method.current_server_state).
+            pub fn has_pending_changes(&self) -> bool {
```
vberger, line 252: I'm realizing that this may be confusing to keep this name for the method, given it does not only checks the pending state. what about `has_unacked_changes` maybe ?

PolyMeilex: There is a case where state can be accked but not yet committed, that would probably cause this to still return `true`.
I have to test out the newest version of this PR to make sure that it's still the case, tho.

cmeissl: @PolyMeilex 

> There is a case where state can be accked but not yet committed, that would probably cause this to still return `true`. I have to test out the newest version of this PR to make sure that it's still the case, tho.

This should be covered [here](https://github.com/Smithay/smithay/pull/415/files#diff-e627a7774fb6d28266da99a8dd7f4b5d37af2e920cc7b84d94a169dff3ba7d94R242), or at least i thought that i covered that case by using the `last_acked` state if it is available in `current_server_state`. 

@vberger 

> I'm realizing that this may be confusing to keep this name for the method, given it does not only checks the pending state. what about `has_unacked_changes` maybe ?

I think `has_unacked_changes` could be missleading. The function checks if something has been configured using 'with_pending_state' on the server, but not sent to the client using 'send_configure'. So i think the nameing is appropriate, but on the other hand i am not the best at naming things XD

PolyMeilex: > This should be covered here, or at least i thought that i covered that case by using the last_acked state if it is available in current_server_state.

Yeah, I know that, I was just mentioning that this should be taken into account as well when naming this method, but now that I look at the code it actually does not matter, as it only cares about ack.

vberger (approved):

PR #417, opened 2021-11-20
Add DispatchData to pointer grabs

PolyMeilex:
At the moment, the only way to pass data to pointer grab is to use Rc, it's quite limiting at times, so I'm trying to find a solution, so here is my attempt.
Feel free to propose any other/better ways to do this, and I'll try to implement it.

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -102,7 +102,7 @@ impl<Backend> AnvilState<Backend> {
             }
             input::ButtonState::Released => wl_pointer::ButtonState::Released,
         };
-        self.pointer.button(button, state, serial, evt.time());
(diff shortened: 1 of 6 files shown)


i509VCB:
Just a note, you'll need to change some stuff in anvil's wlcs integration.

vberger:
I'm uneasy about this change, as the intention with `DispatchData` is that it should be the same underlying type in the whole program, which this code kinda breaks, so it imposes some additional "use this correctly" pressure on the user.

How urgent is this need? The update to wayland-rs 0.30 (when it happens) will significantly change the whole thing anyway.

PolyMeilex:
Not urgent at all, it's a small change that I can easily keep as a patch on my side. 
I'm not super happy with it either, I would rather have it somehow statically typed.
I'll open an issue instead

PR #420, opened 2021-11-23
Rework X11 backend to use EGL for finding gbm device and associated commits

i509VCB:


i509VCB:
I have no earthly idea why thiserror freaks out on github actions but compiles locally all fine.

build.rs
```diff
@@ -17,13 +17,20 @@ fn gl_generate() {
                 "EGL_EXT_create_context_robustness",
                 "EGL_KHR_create_context_no_error",
                 "EGL_KHR_no_config_context",
+                "EGL_EXT_device_base",
+                "EGL_EXT_device_enumeration",
+                "EGL_EXT_device_query",
+                "EGL_EXT_device_drm",
+                "EGL_KHR_stream",
+                "EGL_KHR_stream_producer_eglsurface",
```
Drakulix, line 25: I do not think we need the last two for this feature?

Drakulix (approved):
Looks good to me!

build.rs
Drakulix: Oh right, that was me. Probably an oversight when copying parts of the `EGLDevice` code from my eglstream implementation.

PR #421, opened 2021-11-24
x11: use swapchain and expose age

Drakulix:
Builds upon #419 and #420 to bring buffer age support to the x11 backend as well.

### DIFF
--- a/anvil/src/x11.rs
+++ b/anvil/src/x11.rs
@@ -204,122 +204,116 @@ pub fn run_x11(log: Logger) {
         if state.backend_data.render {
             state.backend_data.render = false;
             let backend_data = &mut state.backend_data;
+            let mut renderer = renderer.borrow_mut();
 
-            match backend_data.surface.present() {
-                Ok(present) => {
-                    let mut renderer = renderer.borrow_mut();
+            // We need to borrow everything we want to refer to inside the renderer callback otherwise rustc is unhappy.
+            let window_map = state.window_map.borrow();
+            let (x, y) = state.pointer_location.into();
+            let dnd_icon = &state.dnd_icon;
+            let cursor_status = &state.cursor_status;
+            #[cfg(feature = "debug")]
+            let fps = backend_data.fps.avg().round() as u32;
+            #[cfg(feature = "debug")]
+            let fps_texture = &backend_data.fps_texture;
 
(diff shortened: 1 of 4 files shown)


i509VCB:
This is great, but a couple of questions:

1. Given the swapchain change, we may want to consider just flattening `Present` away into x11 surface.
2. Maybe mark the function to acquire the buffer as `#[must_use] `?

i509VCB:
I have tested this, everything appears to work as intended.

Drakulix:
> This is great, but a couple of questions:
> 
>     1. Given the swapchain change, we may want to consider just flattening `Present` away into x11 surface.
> 
>     2. Maybe mark the function to acquire the buffer as `#[must_use]`?

Yes, those are probably some good suggestions.

anvil/src/x11.rs
```diff
@@ -328,10 +322,9 @@ pub fn run_x11(log: Logger) {
             window.set_cursor_visible(cursor_visible);
 
             // Send frame events so that client start drawing their next frame
-            state
-                .window_map
-                .borrow()
-                .send_frames(start_time.elapsed().as_millis() as u32);
+            window_map.send_frames(start_time.elapsed().as_millis() as u32);
+            std::mem::drop(window_map);
```
i509VCB, line 326: This seems a bit odd to put in the busy loop?

Drakulix: It is necessary for the next line, because otherwise state is borrowed.

i509VCB: Oh it's probably because window map is in a ref cell

i509VCB: Oh it's probably because window map is in a ref cell

anvil/src/x11.rs
```diff
@@ -204,122 +204,116 @@ pub fn run_x11(log: Logger) {
...
+
+                        Ok(())
+                    },
+                )
+                .map_err(Into::<SwapBuffersError>::into)
+                .and_then(|x| x)
+                .map_err(Into::<SwapBuffersError>::into)
+            {
+                Ok(()) => {
+                    // Unbind the buffer and now let the scope end to present.
```
i509VCB, line 303: There is no scope to drop anymore as we submit on the next line.

PR #422, opened 2021-11-26
[Anvil] Fix ScaleDown scaling up

LoipesMas:
ScaleDown was doing the same thing as ScaleUp (probably someone copypasted and forgot to change the sign).
Now it should properly decrease scale.

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -197,7 +197,7 @@ impl<Backend> AnvilState<Backend> {
 
                     self.output_map
                         .borrow_mut()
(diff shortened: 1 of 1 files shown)


anvil/src/input_handler.rs
```diff
@@ -197,7 +197,7 @@ impl<Backend> AnvilState<Backend> {
 
                     self.output_map
                         .borrow_mut()
-                        .update_scale_by_name(f32::max(1.0f32, current_scale + 0.25f32), output_name);
+                        .update_scale_by_name(f32::max(1.0f32, current_scale - 0.25f32), output_name);
```
i509VCB, line 200: Whoops that was probably me

Drakulix:
Thanks!

PR #423, opened 2021-11-28
Desktop abstractions

Drakulix:
First draft of #363.

## Notable changes:
- Makes existing `Output` more versitile
- Adds `Window` abstraction
- Adds `Space` abstraction
- Adds drawing helpers (`Space::render_output` and `draw_window`)
- Very rough anvil implementation

### space
- [x] xdg-popup handling
- [x] layer-shell support
- [x] track/call output enter/leave for windows
- [x] damage tracking is currently not used for damage of surfaces themselves (instead always the whole window is damaged). This obviously leaves a lot of performance/efficiency gains on the table. 
  - [x] ~~this needs some kind of boolean for each window (was_moved/dirty) to track, which windows should be rendered in full and then this needs to be passed to draw_window to render only with surface damage.~~
  - [x] we actually need to add the accumulated damage of all window surface, because surfaces can be transcluent and we therefor need to render everything that overlaps with the surface damage. Was done in f80b6ba.
  - [x] ~~solution is incomplete, we need to properly track this for multiple outputs/spaces.~~ done
- ~~has no max size and it is awkward to add because it enforces a strict order of:~~
  - updating the size of windows
  - and remapping outputs if necessary
  - and updating the size of the space
  - e.g. If we change the space size and want to remap the outputs and then re-arrange the windows. Should the first two steps error, because we do now have non-accessible windows? Does the operation itself still succeed? Or do we need an api to do all of this atomically?
- ~~we can properly add `MoveGrab` and `ResizeGrab` to smithay directly and add some helpers to use them with spaces~~ tbd in a later pr

### window
- ~~has no max-size (which would also be nice for rendering)~~ will be part of a follow-up PR
- ~~`draw_window` does not play well with multi-gpu setups~~ (will be implemented in a later PR)
- [x] cache bbox
- ~~xwayland support~~ will be part of a larger effort to better support xwayland in smithay

### anvil
- [x] udev backend is not updated at all
- [x] winit makes now use of damage tracking ~~winit backend does not make use of damage-tracking. I think we should not bother. Lets deprecate winit for 0.4 with less features and also add a wayland-backend before 0.4 as an alternative.~~
- [x] render layers again
- [x] render cursors again
- [x] render fps again
- [x] handle fullscreen surfaces
- ~~implement little workspace effect or something similar to show how the desktop abstraction can be temporarily avoided for fancy stuff before resuming efficient rendering.~~ (implemented later)

## Open questions:
- [x] Figure out how to draw external stuff without breaking damage-tracking. Let compositors `map` arbitrary elements with textures? Maybe we can have a trait that can also work for windows and layers? (probably not layers though, since they map to outputs.)
- ~~Figure out a way to atomically update the size and the position of a window (wait for ack on size and next buffer before updating the position). I want to get this rid on first try, because otherwise this feels really bad for tiling. I have no good idea yet how to do that, so suggestions are **very** welcome~~ tbd after wayland-rs 0.30
- ~~Figure out an API to support the size-constraints envisioned by @vberger s first proposal (see #363)~~

My focus obviously was implementing damage-tracking and seeing if that works. I also took some inspiration from the code I wrote for fireplace. So please give feedback for ... well *everything*. I think the api was easy enough to implement for anvil, but it could likely be better.

### DIFF
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -1,6 +1,6 @@
 use crate::{
     backend::renderer::{utils::draw_surface_tree, Frame, ImportAll, Renderer, Texture},
-    desktop::{utils::*, PopupManager, Space},
+    desktop::{space::RenderElement, utils::*, PopupManager, Space},
     utils::{user_data::UserDataMap, Logical, Point, Rectangle},
     wayland::{
         compositor::with_states,
@@ -267,12 +267,12 @@ impl LayerMap {
 }
 
 #[derive(Debug, Default)]
-pub(super) struct LayerState {
-    location: Point<i32, Logical>,
+pub struct LayerState {
+    pub location: Point<i32, Logical>,
 }
 
 type LayerUserdata = RefCell<Option<LayerState>>;
-fn layer_state(layer: &LayerSurface) -> RefMut<'_, LayerState> {
+pub fn layer_state(layer: &LayerSurface) -> RefMut<'_, LayerState> {
     let userdata = layer.user_data();
     userdata.insert_if_missing(LayerUserdata::default);
     RefMut::map(userdata.get::<LayerUserdata>().unwrap().borrow_mut(), |opt| {
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -504,7 +504,7 @@ impl LayerSurface {
 /// Note: This function will render nothing, if you are not using
 /// [`crate::backend::renderer::utils::on_commit_buffer_handler`]
 /// to let smithay handle buffer management.
-pub fn draw_layer<R, E, F, T, P>(
+pub fn draw_layer_surface<R, E, F, T, P>(
     renderer: &mut R,
     frame: &mut F,
     layer: &LayerSurface,
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -438,15 +438,20 @@ impl LayerSurface {
     ) -> Vec<Rectangle<i32, Logical>> {
         let mut damage = Vec::new();
         if let Some(surface) = self.get_surface() {
-            damage.extend(damage_from_surface_tree(surface, (0, 0), for_values));
+            damage.extend(
+                damage_from_surface_tree(surface, (0, 0), for_values)
+                    .into_iter()
+                    .flat_map(|rect| rect.intersection(self.bbox())),
+            );
             for (popup, location) in PopupManager::popups_for_surface(surface)
                 .ok()
                 .into_iter()
                 .flatten()
             {
                 if let Some(surface) = popup.get_surface() {
+                    let bbox = bbox_from_surface_tree(surface, location);
                     let popup_damage = damage_from_surface_tree(surface, location, for_values);
-                    damage.extend(popup_damage);
+                    damage.extend(popup_damage.into_iter().flat_map(|rect| rect.intersection(bbox)));
                 }
             }
         }
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -307,7 +307,7 @@ impl Hash for LayerSurface {
 }
 
 #[derive(Debug)]
-pub struct LayerSurfaceInner {
+pub(crate) struct LayerSurfaceInner {
     pub(crate) id: usize,
     surface: WlrLayerSurface,
     namespace: String,
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -1,6 +1,6 @@
 use crate::{
     backend::renderer::{utils::draw_surface_tree, Frame, ImportAll, Renderer, Texture},
-    desktop::{utils::*, PopupManager},
+    desktop::{utils::*, PopupManager, Space},
     utils::{user_data::UserDataMap, Logical, Point, Rectangle},
     wayland::{
         compositor::with_states,
@@ -423,17 +423,20 @@ impl LayerSurface {
     }
 
     /// Damage of all the surfaces of this layer
-    pub(super) fn accumulated_damage(&self) -> Vec<Rectangle<i32, Logical>> {
+    pub(super) fn accumulated_damage(
+        &self,
+        for_values: Option<(&Space, &Output)>,
+    ) -> Vec<Rectangle<i32, Logical>> {
         let mut damage = Vec::new();
         if let Some(surface) = self.get_surface() {
-            damage.extend(damage_from_surface_tree(surface, (0, 0)));
+            damage.extend(damage_from_surface_tree(surface, (0, 0), for_values));
             for (popup, location) in PopupManager::popups_for_surface(surface)
                 .ok()
                 .into_iter()
                 .flatten()
             {
                 if let Some(surface) = popup.get_surface() {
-                    let popup_damage = damage_from_surface_tree(surface, location);
+                    let popup_damage = damage_from_surface_tree(surface, location, for_values);
                     damage.extend(popup_damage);
                 }
             }
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -84,11 +84,14 @@ impl LayerMap {
         self.zone
     }
 
-    pub fn layer_geometry(&self, layer: &LayerSurface) -> Rectangle<i32, Logical> {
+    pub fn layer_geometry(&self, layer: &LayerSurface) -> Option<Rectangle<i32, Logical>> {
+        if !self.layers.contains(layer) {
+            return None;
+        }
         let mut bbox = layer.bbox_with_popups();
         let state = layer_state(layer);
         bbox.loc += state.location;
-        bbox
+        Some(bbox)
     }
 
     pub fn layer_under<P: Into<Point<f64, Logical>>>(
@@ -98,7 +101,7 @@ impl LayerMap {
     ) -> Option<&LayerSurface> {
         let point = point.into();
         self.layers_on(layer).rev().find(|l| {
-            let bbox = self.layer_geometry(l);
+            let bbox = self.layer_geometry(l).unwrap();
             bbox.to_f64().contains(point)
         })
     }
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -1,6 +1,6 @@
 use crate::{
     backend::renderer::{utils::draw_surface_tree, Frame, ImportAll, Renderer, Texture},
-    desktop::{space::RenderElement, utils::*, PopupManager, Space},
+    desktop::{utils::*, PopupManager, Space},
     utils::{user_data::UserDataMap, Logical, Point, Rectangle},
     wayland::{
         compositor::with_states,
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -16,38 +16,12 @@ use wayland_server::protocol::wl_surface::WlSurface;
 
 use std::{
     cell::{RefCell, RefMut},
-    collections::HashSet,
     hash::{Hash, Hasher},
     rc::Rc,
-    sync::{
-        atomic::{AtomicUsize, Ordering},
-        Arc, Mutex, Weak,
-    },
+    sync::{Arc, Mutex, Weak},
 };
 
-// TODO: Should this be a macro?
-static LAYER_ID: AtomicUsize = AtomicUsize::new(0);
-lazy_static::lazy_static! {
-    static ref LAYER_IDS: Mutex<HashSet<usize>> = Mutex::new(HashSet::new());
-}
-
-fn next_layer_id() -> usize {
-    let mut ids = LAYER_IDS.lock().unwrap();
-    if ids.len() == usize::MAX {
-        // Theoretically the code below wraps around correctly,
-        // but that is hard to detect and might deadlock.
-        // Maybe make this a debug_assert instead?
-        panic!("Out of window ids");
-    }
-
-    let mut id = LAYER_ID.fetch_add(1, Ordering::SeqCst);
-    while ids.iter().any(|k| *k == id) {
-        id = LAYER_ID.fetch_add(1, Ordering::SeqCst);
-    }
-
-    ids.insert(id);
-    id
-}
+crate::utils::ids::id_gen!(next_layer_id, LAYER_ID, LAYER_IDS);
 
 #[derive(Debug)]
 pub struct LayerMap {
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -117,7 +117,12 @@ impl LayerMap {
         bbox
     }
 
-    pub fn layer_under(&self, layer: WlrLayer, point: Point<f64, Logical>) -> Option<&LayerSurface> {
+    pub fn layer_under<P: Into<Point<f64, Logical>>>(
+        &self,
+        layer: WlrLayer,
+        point: P,
+    ) -> Option<&LayerSurface> {
+        let point = point.into();
         self.layers_on(layer).rev().find(|l| {
             let bbox = self.layer_geometry(l);
             bbox.to_f64().contains(point)
@@ -401,7 +406,11 @@ impl LayerSurface {
 
     /// Finds the topmost surface under this point if any and returns it together with the location of this
     /// surface.
-    pub fn surface_under(&self, point: Point<f64, Logical>) -> Option<(WlSurface, Point<i32, Logical>)> {
+    pub fn surface_under<P: Into<Point<f64, Logical>>>(
+        &self,
+        point: P,
+    ) -> Option<(WlSurface, Point<i32, Logical>)> {
+        let point = point.into();
         if let Some(surface) = self.get_surface() {
             for (popup, location) in PopupManager::popups_for_surface(surface)
                 .ok()
@@ -457,12 +466,12 @@ impl LayerSurface {
     }
 }
 
-pub fn draw_layer<R, E, F, T>(
+pub fn draw_layer<R, E, F, T, P>(
     renderer: &mut R,
     frame: &mut F,
     layer: &LayerSurface,
     scale: f64,
-    location: Point<i32, Logical>,
+    location: P,
     damage: &[Rectangle<i32, Logical>],
     log: &slog::Logger,
 ) -> Result<(), R::Error>
@@ -471,7 +480,9 @@ where
     F: Frame<Error = E, TextureId = T>,
     E: std::error::Error,
     T: Texture + 'static,
+    P: Into<Point<i32, Logical>>,
 {
+    let location = location.into();
     if let Some(surface) = layer.get_surface() {
         draw_surface_tree(renderer, frame, surface, scale, location, damage, log)?;
         for (popup, p_location) in PopupManager::popups_for_surface(surface)
--- a/src/desktop/layer.rs
+++ b/src/desktop/layer.rs
@@ -1,0 +1,501 @@
+use crate::{
+    backend::renderer::{utils::draw_surface_tree, Frame, ImportAll, Renderer, Texture},
+    desktop::{utils::*, PopupManager},
+    utils::{user_data::UserDataMap, Logical, Point, Rectangle},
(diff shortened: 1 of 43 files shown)


src/backend/x11/mod.rs
```diff
@@ -349,8 +349,7 @@ pub struct X11Surface {
     format: DrmFourcc,
     width: u16,
     height: u16,
-    current: Option<Slot<BufferObject<()>>>,
-    next: Option<Slot<BufferObject<()>>>,
+    buffer: Option<Slot<BufferObject<()>>>,
```
i509VCB, line 352: Would be great to merge these changes in the x11 backend as their own PR so I can implement multi-window handling in peace.

Drakulix: Good call, I will cherry-pick them.

Drakulix: https://github.com/Smithay/smithay/pull/425

src/desktop/space.rs
```diff
@@ -0,0 +1,419 @@
...
+        let loc = window_state(self.id, window).location;
+        let _ = self.map_window(window, loc);
+
+        // TODO: should this be handled by us?
+        window.set_activated(true);
+        for w in self.windows.iter() {
+            if w != window {
+                w.set_activated(false);
+            }
+        }
```
vberger, line 137: I'm slightly leaning towards "yes"

src/desktop/space.rs
```diff
@@ -0,0 +1,419 @@
...
+        for o in &self.outputs {
+            let o_geo = self.output_geometry(o).unwrap();
+            if w_geo.overlaps(o_geo) {
+                return Some(o.clone());
+            }
+        }
+
+        // TODO primary output
+        self.outputs.get(0).cloned()
+    }
```
vberger, line 242: Shouldn't this really return a (possibly empty) set/vec of outputs, rather than just one output? Also, why does it return an output even if no output overlaps with the window?

src/desktop/space.rs
```diff
@@ -0,0 +1,419 @@
...
+            .last_state
+            .iter()
+            .filter_map(|(id, w)| {
+                if !self.windows.iter().any(|w| w.0.id == *id) {
+                    Some(*w)
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<Rectangle<i32, Logical>>>()
```
vberger, line 398: What about moving the `debug!()` into the `filter_map()` closure, and collecting the result directly into `damage`, rather than manually iterating & pushing every element?

src/desktop/space.rs
```diff
@@ -0,0 +1,419 @@
...
+        damage.retain(|rect| rect.overlaps(output_geo));
+        damage.retain(|rect| rect.size.h > 0 && rect.size.w > 0);
+        for rect in damage.clone().iter() {
+            // if this rect was already removed, because it was smaller as another one,
+            // there is no reason to evaluate this.
+            if damage.contains(rect) {
+                // remove every rectangle that is contained in this rectangle
+                damage.retain(|other| !rect.contains_rect(*other));
+            }
+        }
```
vberger, line 446: I think we can do this in a more efficient way (making this comment to not forget about it, but I don't have the replacement code in my head right now).

src/desktop/space.rs
```diff
@@ -0,0 +1,419 @@
...
+            output_transform
+                .transform_size(output_size)
+                .to_f64()
+                .to_physical(state.render_scale)
+                .to_i32_round(),
+            output_transform,
+            |renderer, frame| {
+                // First clear all damaged regions
+                for geo in &damage {
+                    slog::debug!(self.logger, "Clearing at {:?}", geo);
```
vberger, line 459: Maybe put this as `trace!` instead? That may be extremely spammy

src/desktop/space.rs
```diff
@@ -0,0 +1,419 @@
...
+                for window in self.windows.iter() {
+                    let wgeo = window_rect(window, &self.id);
+                    let mut loc = window_loc(window, &self.id);
+                    if damage.iter().any(|geo| wgeo.overlaps(*geo)) {
+                        loc -= output_geo.loc;
+                        slog::debug!(self.logger, "Rendering window at {:?}", wgeo);
+                        draw_window(renderer, frame, window, state.render_scale, loc, &self.logger)?;
+                        window_state(self.id, window).drawn = true;
+                    }
+                }
```
vberger, line 476: I suppose the longer-term idea would be to only redraw the parts of the window that overlap with the damage spec, right?

src/desktop/space.rs
```diff
@@ -0,0 +1,419 @@
...
+            // if this rect was already removed, because it was smaller as another one,
+            // there is no reason to evaluate this.
+            if damage.contains(rect) {
+                // remove every rectangle that is contained in this rectangle
+                damage.retain(|other| !rect.contains_rect(*other));
+            }
+        }
+
+        let output_transform: Transform = output.current_transform().into();
+        if let Err(err) = renderer.render(
```
vberger, line 449: This nested if is pretty difficult to read, maybe instead do something like:

```rust
let ret = renderer.render(....);

if let Err(err) = ret {
    ....
}
```

src/desktop/space.rs
```diff
@@ -0,0 +1,419 @@
...
+            .collect();
+        state.old_damage.push_front(new_damage);
+
+        // Return if we actually rendered something
+        Ok(!damage.is_empty())
+    }
+
+    pub fn send_frames(&self, all: bool, time: u32) {
+        for window in self.windows.iter().filter(|w| {
+            all || {
```
vberger, line 505: What's that?

src/desktop/window.rs
```diff
@@ -374,6 +374,57 @@ impl Window {
...
+                            }
+
+                            damage.extend(attributes.damage.iter().map(|dmg| {
+                                let mut rect = match dmg {
+                                    Damage::Buffer(rect) => rect.to_logical(attributes.buffer_scale),
+                                    Damage::Surface(rect) => *rect,
+                                };
+                                rect.loc += location;
+                                rect
+                            }));
```
vberger, line 418: On the longer-term, this could take into account the `opaque_region` specified by the client to discard damage that is completely behind an opaque surface.

To transpose that on the space level, the Window would need to have a method returning a `Vec<Region>` describing which parts of it are opaque I guess.

src/desktop/space.rs
```diff
@@ -234,8 +241,124 @@ impl Space {
...
+                                }
+                            },
+                            |_, _, _| true,
+                        )
+                    }
+                    continue;
+                }
+
+                if let Some(surface) = kind.get_surface() {
+                    with_surface_tree_downward(
```
vberger, line 291: I'm wondering if it makes sense to give enter/leave information at the subsurface granularity detail.

Wouldn't it be simpler to just always send the same enter/leave to the whole subsurface tree, based only on intersection with the bounding box? I seems to me that it's unnecessary detail computed and sent to the client.

src/utils/geometry.rs
```diff
@@ -725,6 +725,35 @@ impl<N: Coordinate, Kind> Rectangle<N, Kind> {
...
+    /// Convert to i32 for integer-space manipulations by ceiling float values
+    #[inline]
+    pub fn to_i32_ceil<N: Coordinate>(self) -> Rectangle<N, Kind> {
+        Rectangle {
+            loc: self.loc.to_i32_ceil(),
+            size: self.size.to_i32_ceil(),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
Drakulix: That is a good catch actually, because I think I am already mindlessly using them in their "spirit" instead of what they are actually doing.

But wouldn't it just suffice for `to_i32_floor`  to change the location to `self.loc.to_i32_ceil()` and vise-versa?

src/desktop/window.rs
```diff
@@ -0,0 +1,510 @@
...
+}
+
+fn next_window_id() -> usize {
+    let mut ids = WINDOW_IDS.lock().unwrap();
+    if ids.len() == usize::MAX {
+        // Theoretically the code below wraps around correctly,
+        // but that is hard to detect and might deadlock.
+        // Maybe make this a debug_assert instead?
+        panic!("Out of window ids");
+    }
```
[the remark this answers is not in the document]
Drakulix: yeah me too. I think wrapping-around is fine and "deadlocking because of usize::MAX-windows are created" is even more unlikely?

src/desktop/window.rs
```diff
@@ -0,0 +1,510 @@
...
+    ids.insert(id);
+    id
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum Kind {
+    Xdg(ToplevelSurface),
+    #[cfg(feature = "xwayland")]
+    X11(X11Surface),
+}
```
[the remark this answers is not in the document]
Drakulix: For rendering custom components in a space (without breaking damage tracking) like our FPS counter in anvil, we need a way to attach arbitrary elements to a space anyway. I want to try to design a trait for these elements. Downstream could implement custom shells themselves and attach them to a space by implementing that trait.

Alternatively we can have a custom variant of a `Window`, but I am not really sure, if that is helpful given that smithay will implement all major desktop-shells anyway. And anything custom does likely not cleanly map to our concept of a `Window` anyway. (And if it does, why not just use xdg-shell...)

src/desktop/window.rs
```diff
@@ -0,0 +1,510 @@
...
+        self.0.id.hash(state);
+    }
+}
+
+impl Window {
+    pub fn new(toplevel: Kind) -> Window {
+        let id = next_window_id();
+
+        // TODO: Do we want this? For new lets add Window::commit
+        //add_commit_hook(toplevel.get_surface().unwrap(), surface_commit);
```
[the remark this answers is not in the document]
Drakulix: yes

src/desktop/window.rs
```diff
@@ -0,0 +1,510 @@
...
+                .with_pending_state(|state| {
+                    if active {
+                        state.states.set(xdg_toplevel::State::Activated)
+                    } else {
+                        state.states.unset(xdg_toplevel::State::Activated)
+                    }
+                })
+                .unwrap_or(false),
+            #[cfg(feature = "xwayland")]
+            Kind::X11(ref t) => unimplemented!(),
```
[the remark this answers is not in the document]
Drakulix: Right. I am not sure, if I want to implement an X11 WM in smithay in this PR though. But that is what needs to be done to fully support xwayland with the desktop module.

src/desktop/space.rs
Drakulix: I was wondering, if we should add a `activate`-boolean to `map_window` (and possible `raise_window`). Downstream might not always want to shift focus, just because a new window is spawned. Even though this does not actually change the focus, it looks wrong.

src/desktop/space.rs
Drakulix: Yeah, that is a good idea. I think it always returns an `Output` if possible, because this method is used for fullscreen-ing and maximizing windows. But we could easily return a set/vec of outputs and just sort them by *most overlapping*? And also let downstream/anvil handle the fallback, if no `Output` overlaps.

src/desktop/space.rs
Drakulix: good catch!

src/desktop/space.rs
Drakulix: I actually expect the whole algorithm to be a little on the sub-optimal side. I wanted it to be easily understandable and correct first before optimizing it a whole lot.

src/desktop/space.rs
Drakulix: Locally I have already changed that to trace, it was just super helpful debugging this.

src/desktop/space.rs
Drakulix: I though there are reasons to send frame events periodically to all clients, even those not mapped (the surface-suspension protocol wants to fix this, I think). So I did not want to force the user to do it either way.

src/desktop/window.rs
Drakulix: Yeah definitely. There is definitely room for improvement, but this is getting us probably already 80% of the way in terms of performance. (I still need to actually benchmark/profile it.)

src/desktop/space.rs
Drakulix: Hmm. Not sure actually. While that would probably be fine for most clients, I am not sure if more complex clients like firefox would agree?? But potentially worth a try? This is at least what anvil has been doing.

src/desktop/space.rs
vberger: Ah no, I meant that `all || { ... }` syntax, is that some new rust feature? I've never seen it.

Drakulix: that is just `boolean <OR> { expr returning a boolean }`. Not a closure, just a simple `||` you know like `&&`.

src/utils/geometry.rs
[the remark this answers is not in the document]
vberger: A counter-example (looking only at one coordinate): take a rectangle spanning from 1.9 to 5.1 that we want to "ceil in spirit". 

Its location is 1.9, and its size is 3.2. Just doing floor/ceil would produce an integer rectangle of location 1 and size 4, meaning a rectangle that goes from 1.0 to 5.0, while proper "ceil in spirit" should produce a rectangle going from 1.0 to 6.0.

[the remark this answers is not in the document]
Drakulix: Ah right, the size needs to account for the shift of the location. Makes sense

src/desktop/space.rs
vberger: Ah right, my brain is too hard-wired to read `| | {` as the start of a closure it appears. 😭

src/desktop/space.rs
Drakulix: The code does that now. opaque_region is an optimization that I will likely not tackle inside this PR.

src/utils/geometry.rs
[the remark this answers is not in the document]
Drakulix: Should be fixed in #434. If not the discussion should be moved there.

src/desktop/space.rs
```diff
@@ -399,6 +399,33 @@ impl Space {
...
+                for window in self.windows() {
+                    if let Some(w) = window.toplevel().get_surface() {
+                        if w == surface {
+                            *found.borrow_mut() = Some(window);
+                            return;
+                        }
+                    }
+                },
+            |_, _, _| found.borrow().is_some()
+        );
```
vberger, line 423: I don't think this logic does what you want, at all. There is no point in using `with_surface_tree` here.

It should be something like:

```rust
let mut root = surface.clone();
while let Some(parent) = get_parent(&root) {
    root = parent;
}

let if let Some(window) = self.windows().iter().find(|w| w.toplevel().get_surface() == Some(&root)) {
    window.refresh();
}

```

Drakulix: yeah, I have overcomplicated this a bit...

vberger: It's more than overcomplicated, the `with_surface_tree_*` only go through the subtree whose given surface is the root, so thay don't allow finding to actual root of the subsurface tree, which is what you need here.

src/desktop/window.rs
```diff
@@ -0,0 +1,312 @@
...
+    if let Some(surface) = window.toplevel().get_surface() {
+        draw_surface_tree(renderer, frame, surface, scale, location, damage, log)?;
+        for (popup, p_location) in PopupManager::popups_for_surface(surface)
+            .ok()
+            .into_iter()
+            .flatten()
+        {
+            if let Some(surface) = popup.get_surface() {
+                let damage = damage.iter().cloned().map(|mut geo| { geo.loc -= p_location; geo }).collect::<Vec<_>>();
+                draw_surface_tree(renderer, frame, surface, scale, location + p_location, &damage, log)?;
```
PolyMeilex, line 307: I'm not sure if location of popups here is correct.
example)

In Anodium I do: `location + window.geometry().loc + p_location`, but I'm not sure if that's correct either, as you can see in the screenshot it is also a little bit off in comparison to gnome.

src/desktop/window.rs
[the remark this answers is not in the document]
Drakulix: Custom elements are a thing now.

src/desktop/space.rs
Drakulix: @vberger Wanna give this a try? The corresponding lines are now 469-491 for optimizing the damage values.

src/desktop/window.rs
Drakulix: Lets make another ticket for this one.

src/desktop/space.rs
Drakulix: Does the wayland core protocol say something about this?

vberger: Not much, tbh. The information of enter/leave is mostly used for deciding the scaling factor when drawing, the client cannot assume that its surface is actually visible on all the wl_outputs it is "entered" (as it might be hidden behind an other for example).

So I don't think it's actually possible for a client to rely on precise tracking of the wl_output its surface are entered on.

src/desktop/space.rs
Drakulix: Lines are removed. The overlap optimization already handles rectangles completely occluded by larger ones anyway and the fold operation should be faster.

src/desktop/space/mod.rs
```diff
@@ -0,0 +1,706 @@
...
+                    None
+                }
+            })
+            .collect::<Vec<Rectangle<i32, Logical>>>()
+        {
+            slog::trace!(self.logger, "Removing toplevel at: {:?}", old_toplevel);
+            damage.push(old_toplevel);
+        }
+
+        // lets iterate front to back and figure out, what new windows or unmoved windows we have
```
vberger, line 508: Do I understand correctly that this damage tracking does not take into account Z-order, for example a changed window will cause damage to the output even if it is completely occulted by an other, undamaged window?

If so, maybe worth letting a TODO (and maybe a tracking issue?) to not forget that this is something that has room for improvement in the future?

src/desktop/space/mod.rs
```diff
@@ -0,0 +1,706 @@
...
+
+    /// Maps an [`Output`] inside the space.
+    ///
+    /// Can be safely called on an already mapped
+    /// [`Output`] to update its scale or location.
+    ///
+    /// The scale is the what is rendered for the given output
+    /// and may be fractional. It is independent from the integer scale
+    /// reported to clients by the output.
+    pub fn map_output<P: Into<Point<i32, Logical>>>(&mut self, output: &Output, scale: f64, location: P) {
```
vberger, line 196: I'm not clear on how this interacts with damage tracking for that output. My intuition says that this should mark the whole output as damaged, but I'm not sure if that is what actually happens.

src/desktop/space/mod.rs
Drakulix: Yes correct, because windows can be transparent. We need to expose `opaque_region` similar to surface damage and then iterate twice through the elements.

Once front to back to figure out, what we need to render. Then back to front to render the filtered surfaces.

But that will be another PR 😅

src/desktop/layer.rs
```diff
@@ -97,9 +97,9 @@ impl LayerMap {
         self.zone
     }
 
-    /// Returns the geometry of a given mapped layer.
+    /// Returns the geometry of a given mapped [`LayerSurface.
```
vberger, line 100: Missing the closing ``` `] ```

src/desktop/space/mod.rs
```diff
@@ -193,11 +193,17 @@ impl Space {
     /// The scale is the what is rendered for the given output
     /// and may be fractional. It is independent from the integer scale
     /// reported to clients by the output.
+    ///
+    /// *Note:* Remapping an output does reset it's damage memory.
```
vberger, line 197: So, does that mean the output will be considered as 100% damaged, or 0% damaged?

Drakulix: 100% damaged.
without a memory the output assumes everything is new.

PR #424, opened 2021-11-28
Add more Add/Sub(Assign) methods to geometry

chrisduerr:
The primary motivation behind this patch was the implementation of the
`Add` and `Sub` traits for `Size` and `Point`. This is very useful when
having to reduce a size based on a location (to limit boundaries for
example).

The `AddAssign` and `SubAssign` implementations for `Size` seem to go
hand-in-hand with the normal `Add` and `Sub` implementations and since
they were missing for `Point` too I decided to just implement all
permutations.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -652,6 +652,60 @@ impl<N: SubAssign + fmt::Debug + PartialOrd, Kind> SubAssign for Size<N, Kind> {
     }
 }
 
+impl<N: Add<Output = N>, Kind> Add<Point<N, Kind>> for Size<N, Kind> {
+    type Output = Size<N, Kind>;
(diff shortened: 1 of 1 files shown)


vberger:
> This is very useful when having to reduce a size based on a location (to limit boundaries for example).

Would it be possible to design an explicit api for that instead? It seems to me that adding those cross-type implementations is essentially removing the type-level robustness against misuse that was the original goal...

chrisduerr:
Honestly I feel like sizes and points should be units that can directly operate with each other. Anything else just seems like overhead on the consumer's side.

What problems would there be allowing these operations across type boundaries? It seems to me like there isn't much option for abuse given both are just two dimensional vectors basically.

PR #426, opened 2021-11-29
x11: rework api to be multi-window capable

i509VCB:
This should allow the X11 backend to create multiple windows.

I have not implemented the handling of secondary windows in anvil yet but the single primary window is tested as working.

### DIFF
--- a/src/backend/x11/mod.rs
+++ b/src/backend/x11/mod.rs
@@ -100,7 +100,7 @@ use std::{
     sync::{
         atomic::{AtomicU32, Ordering},
         mpsc::{self, Receiver, Sender},
-        Arc, Weak,
+        Arc, Mutex, MutexGuard, Weak,
     },
 };
 use x11rb::{
@@ -345,7 +345,7 @@ pub struct X11Surface {
     connection: Weak<RustConnection>,
     window: Window,
     resize: Receiver<Size<u16, Logical>>,
-    swapchain: Swapchain<gbm::Device<DrmNode>, BufferObject<()>>,
+    swapchain: Swapchain<Arc<Mutex<gbm::Device<DrmNode>>>, BufferObject<()>>,
     format: DrmFourcc,
     width: u16,
     height: u16,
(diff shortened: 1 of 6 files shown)


src/backend/x11/mod.rs
```diff
@@ -441,7 +441,7 @@ impl X11Surface {
     /// This will fail if the backend has already been used to create a surface.
     pub fn new(
         backend: &mut X11Backend,
-        device: gbm::Device<DrmNode>,
+        device: Arc<Mutex<gbm::Device<DrmNode>>>,
```
Drakulix, line 420: I wonder, if this is something we should rather handle internally.
Although you can now share allocators between different backends with this PR, users should probably not be encouraged to do that and rather create multiple `gbm::Device`s for different backends.

src/backend/x11/mod.rs
```diff
@@ -472,7 +472,7 @@ impl X11Surface {
     }
 
     /// Returns a handle to the GBM device used to allocate buffers.
-    pub fn device(&self) -> &gbm::Device<DrmNode> {
+    pub fn device(&self) -> &Arc<Mutex<gbm::Device<DrmNode>>> {
```
Drakulix, line 469: If we decide to hide the `Arc<Mutex<_>>` (see comment above), we could return a `MutexGuard<'_, gbm::Device<DrmNode>>` here instead.

Drakulix (commented):
Looks good on first glance.

Do you plan to implement multi-window support for anvil in this PR or follow up with that later?

src/backend/x11/mod.rs
i509VCB: Well the constructor for `X11Surface` needs to get the gbm device somehow. We don't make the X11Backend or X11Handle create the device at the moment.

To propagate the device to new X11Surfaces, we would need to either bootstrap off one surface or make an intermediate object for the device that internally holds the Mutex?

i509VCB:
@Drakulix I will probably deal with multi-window in anvil at a later date because some of the input abstraction and windowed backend event handling is not yet configured for multiple outputs.

src/backend/x11/mod.rs
Drakulix: I though about giving the `X11Backend` a `create_surface` function and making this constructor private.

i509VCB: Hmm that seems like a fine way around this issue

Drakulix:
I think this PR is good as soon as CI agrees.

i509VCB:
All the errors here should have been clippy update for 1.57, so rerun actions I guess?

Drakulix:
> All the errors here should have been clippy update for 1.57, so rerun actions I guess? 

Or rebase onto master and push?

i509VCB:
this appears to build now

PR #429, opened 2021-12-06
x11: Do not hold lock when dispatching events

i509VCB:
This should prevent compositors from deadlocking when creating a window while x11 is dispatching events.

One case where a deadlock occurs is when creating a new window when `Super + O` is pressed and creating said window when the `X11Event::Input` is received.

---

Days since i5 forgot to drop a MutexGuard: 0


Drakulix (approved):
Looks good!

PR #434, opened 2021-12-16
Geometry additions

Drakulix:
So this is the result of a bunch of changes, that have piled up in #423 related to our geometry helpers.
I have extracted them to make review easier and quicker.
This also should address previous commits from @vberger on #423 related to `Rectangle::to_i32_{ceil, floor, round}` (4284ae471df6a5ddc966fe48c27daa2d933035e6).

Best reviewed commit-by-commit, if anything is controversial (like maybe 100f7ad70e0d8fd333048287e25f987cb2b42b76) I can always drop them from this particular PR.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -17,14 +17,21 @@ pub struct Buffer;
 #[derive(Debug)]
 pub struct Raw;
 
+/// Trait for types serving as a coordinate for other geometry utils
 pub trait Coordinate:
     Sized + Add<Self, Output = Self> + Sub<Self, Output = Self> + PartialOrd + Default + Copy + fmt::Debug
 {
+    /// Downscale the coordinate
     fn downscale(self, scale: Self) -> Self;
+    /// Upscale the coordinate
     fn upscale(self, scale: Self) -> Self;
+    /// Convert the coordinate to a f64
     fn to_f64(self) -> f64;
+    /// Convert to this coordinate from a f64
     fn from_f64(v: f64) -> Self;
+    /// Test if the coordinate is not negative
     fn non_negative(self) -> bool;
+    /// Returns the absolute value of this coordinate
     fn abs(self) -> Self;
 }
 
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -833,6 +833,19 @@ impl<N: Coordinate, Kind> Rectangle<N, Kind> {
         )
     }
 
+    /// Clamp rectangle to min and max corners resulting in the overlapping area of two rectangles
+    #[inline]
+    pub fn intersection(self, other: impl Into<Rectangle<N, Kind>>) -> Self {
+        let other = other.into();
+        Rectangle::from_extemities(
+            (self.loc.x.max(other.loc.x), self.loc.y.max(other.loc.y)),
+            (
+                (self.loc.x + self.size.w).min(other.loc.x + other.size.w),
+                (self.loc.y + self.size.h).min(other.loc.y + other.size.h),
+            ),
+        )
+    }
+
     /// Compute the bounding box of a given set of points
     pub fn bounding_box(points: impl IntoIterator<Item = Point<N, Kind>>) -> Self {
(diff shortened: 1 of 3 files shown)


src/utils/geometry.rs
```diff
@@ -725,6 +725,29 @@ impl<N: Coordinate, Kind> Rectangle<N, Kind> {
...
+    #[inline]
+    pub fn to_i32_floor<N: Coordinate>(self) -> Rectangle<N, Kind> {
+        Rectangle::from_extemities(self.loc.to_i32_floor(), (self.loc + self.size).to_i32_ceil())
+    }
+
+    /// Convert to i32 for integer-space manipulations by ceiling float values
+    #[inline]
+    pub fn to_i32_ceil<N: Coordinate>(self) -> Rectangle<N, Kind> {
+        Rectangle::from_extemities(self.loc.to_i32_ceil(), (self.loc + self.size).to_i32_floor())
+    }
```
vberger, line 838: I think those method should have more explicit names, or at least more explicit documentation. Their current behavior is the opposite from my I would have intuited given their names: it seemed to me natural to expect that ceil-ing a rectangle would yield a bigger rectangle than floor-ing it.

But overall, I'd say we'd better avoid using the words floor/ceil. What about `to_i32_grow` and `to_i32_shrink` for example?

src/utils/geometry.rs
```diff
@@ -208,6 +208,7 @@ floating_point_coordinate_impl! {
  */
 
 /// A point as defined by its x and y coordinates
+#[repr(C)]
```
vberger, line 271: Why do we need that?

src/utils/geometry.rs
```diff
@@ -833,6 +833,19 @@ impl<N: Coordinate, Kind> Rectangle<N, Kind> {
         )
     }
 
+    /// Clamp rectangle to min and max corners resulting in the overlapping area of two rectangles
+    #[inline]
+    pub fn overlapping_area(self, other: impl Into<Rectangle<N, Kind>>) -> Self {
```
vberger, line 902: Nitpick, but I tend to think naming this "intersection" would be more succinct and slightly clearer.

src/utils/geometry.rs
```diff
@@ -194,6 +227,31 @@ macro_rules! floating_point_coordinate_impl {
...
+                }
+            }
+            #[inline]
+            fn saturating_mul(self, other: Self) -> Self {
+                match self * other {
+                    x if x == Self::INFINITY => Self::MAX,
+                    x if x == Self::NEG_INFINITY => Self::MIN,
+                    x => x,
+                }
+            }
```
vberger, line 254: I'm not quite sure about this one, feels wrong to clamp infinite values on float. :thinking:

src/utils/geometry.rs
Drakulix: Oh right, I totally swapped them around. But that just shows, that they should be better docs and probably better names.

src/utils/geometry.rs
Drakulix: I currently don't, but that allows e.g. to read in a `Vec<Rectangle<i32, Logical>>` as a `&[i32]` in the OpenGL renderer.

src/utils/geometry.rs
Drakulix: Yeah, I wasn't sure either, but Infinity is not a very useful value inside a compositor for coordinates. I can easily drop that though and just do the normal operations.

vberger: Well, `1e306` is not a useful value either, but at least infinity (and its tendency to produce NaN) can make potential bugs more obvious I'd say

Drakulix: Granted. :+1:

src/utils/geometry.rs
Drakulix: I hope its cleaner now?

vberger: Yep, that looks good now :+1:

PR #436, opened 2021-12-17
x11: internal window to send close backend event to

i509VCB:
Dumb mistake I made. Per the docs of X11Source:

> The returned instance will use `SendRequest` to cause a `ClientMessageEvent` to be sent to
> the given window with the given type. The expectation is that this is a window that was
> created by us. Thus, the event reading thread will wake up and check an internal exit flag,
> then exit.

@Drakulix you probably want to test this on cosmic before merging.


Drakulix (approved):
Works! Just waiting for CI.

PR #441, opened 2021-12-22
x11: Don't use modifiers on old dri3 versions

Drakulix:
Older dri3 versions (< 1.2) do only support dmabufs with one plane.
Lets limit the modifiers early, if we detect that, so we do not end
up allocating buffers the interface does not accept.

### DIFF
--- a/src/backend/x11/mod.rs
+++ b/src/backend/x11/mod.rs
@@ -367,7 +367,12 @@ impl X11Handle {
             return Err(X11Error::InvalidWindow);
         }
 
-        let modifiers = modifiers.collect::<Vec<_>>();
+        let mut modifiers = modifiers.collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


src/backend/x11/mod.rs
```diff
@@ -367,7 +367,10 @@ impl X11Handle {
             return Err(X11Error::InvalidWindow);
         }
 
-        let modifiers = modifiers.collect::<Vec<_>>();
+        let mut modifiers = modifiers.collect::<Vec<_>>();
+        if !(window.0.extensions.dri3 >= Some((1, 2))) {
```
i509VCB, line 371: Make sure you put a comment here indicating why we limit the modifiers here, the content of the commit message should be fine.

i509VCB (approved):

PR #442, opened 2021-12-23
Fix use_mode when trying to apply higher resolution then the current one 

dragonnn:
Current use_mode implementation fails when the user tries to set a mode with higher resolution the the initial output mode.
The issue is cased by two places:
- the test_buffer uses the previous mode size, so when a new mode tries to commit on it and has higher resolution it fails with ENOSPC. This is fixed by the "make use_mode use the new mode size on test_buffer 9a7cd23" commit
-  after the test is successful the real buffer will reject the new pending mode on next commit unless resize on swapchain is called. This is fixed by the "call resize after applying new pending model 5ce5a8a" commit.

### DIFF
--- a/src/backend/drm/surface/atomic.rs
+++ b/src/backend/drm/surface/atomic.rs
@@ -354,7 +354,7 @@ impl<A: AsRawFd + 'static> AtomicDrmSurface<A> {
                 source,
             })?;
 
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
Thanks!

PR #443, opened 2021-12-25
Wayland/input method and text input

rano-oss:
The modules should provide initial support and cover most of the functionality. ~~Surface handling of the popup surface and modifier handling/compositor shortcuts have not been added so far. I wanted to avoid duplicate effort even if it is not 100% finished.~~ Also I plan to add fixes and the rest of the functionality as i write my own input method and test with other input methods.

Please let me know if you have any suggestions for improvements.

~~ps: I also just realized that my tooling have been broken lately, hence the last commit with fmt, clippy and check~~

Edit: Should now support everything defined in the protocol as far as I can tell


anvil/src/input_handler.rs
```diff
@@ -121,6 +126,7 @@ impl<Backend> AnvilState<Backend> {
                         .get_surface_and_bring_to_top(self.pointer_location);
                     self.keyboard
                         .set_focus(under.as_ref().map(|&(ref s, _)| s), serial);
+                    self.text_input.set_focus(under.as_ref().map(|&(ref s, _)| s));
```
PolyMeilex, line 129: What is the reason behind `|&(ref s, _)|`, from my quick testing `|(s,_)|` gives the same result and looks cleaner.

anvil/src/input_handler.rs
```diff
@@ -460,6 +466,7 @@ impl AnvilState<UdevData> {
                         let serial = SCOUNTER.next_serial();
                         self.keyboard
                             .set_focus(under.as_ref().map(|&(ref s, _)| s), serial);
+                        self.text_input.set_focus(under.as_ref().map(|&(ref s, _)| s));
```
PolyMeilex, line 469: Same here

src/wayland/input_method/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+    .unwrap();
+    let state = xkb::State::new(&keymap);
+    (keymap.get_as_string(xkb::KEYMAP_FORMAT_TEXT_V1), state)
+}
+
+impl fmt::Debug for InputMethod {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("InputMethod").finish()
+    }
+}
```
PolyMeilex, line 187: Could we move this manual impl to `KeyboardState` (as it is the reason for manual impl in the first place), and this way we will also preserver `rate`, `delay`, `keymap` in the debug log.

anvil/src/input_handler.rs
rano-oss: Alright, I'll replace that, it was basically a quick copy past

src/wayland/input_method/mod.rs
rano-oss: I'll fix it when I get around too it, thanks

anvil/src/input_handler.rs
rano-oss: Done

anvil/src/input_handler.rs
rano-oss: done

src/wayland/input_method/mod.rs
rano-oss: Done

rano-oss:
Now supports serial handling and popup surface handling

src/wayland/input_method/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+pub trait InputMethodSeatTrait {
+    /// Get input method seat associated with this seat
+    /// this is also used to set xkb config parameters that will be sent to the input method
+    /// Input methods need different keyboard languages for different input methods
+    /// E.g Norwegian pinyin user will want to use a nordic keyboard layout
+    /// but a Taiwanese person will write their input method with a us layout
+    fn add_input_method(
+        &self,
+        repeat_rate: i32,
+        repeat_delay: i32,
```
chrisduerr, line 266: These two parameters must be swapped. Right now they're the exact opposite of `add_keyboard` and that will definitely screw people up.

rano-oss: Alright, done.

src/wayland/input_method_manager/input_method_popup_surface.rs
```diff
@@ -0,0 +1,79 @@
...
+impl<D> Dispatch<ZwpInputPopupSurfaceV2, InputMethodPopupSurfaceUserData, D> for InputMethodManagerState {
+    fn request(
+        _state: &mut D,
+        _client: &wayland_server::Client,
+        _resource: &ZwpInputPopupSurfaceV2,
+        _request: <ZwpInputPopupSurfaceV2 as wayland_server::Resource>::Request,
+        _data: &InputMethodPopupSurfaceUserData,
+        _dhandle: &wayland_server::DisplayHandle,
+        _data_init: &mut wayland_server::DataInit<'_, D>,
+    ) {
```
i509VCB, line 73: I think we should match on the request here, but so nothing so that we know if new requests are added in future.

rano-oss: Fixed :)

anvil/src/drawing.rs
```diff
@@ -66,6 +66,17 @@ pub fn draw_dnd_icon(
...
 }
 
+pub fn draw_input_popup_surface(
+    surface: wl_surface::WlSurface,
+    location: impl Into<Point<i32, Logical>>,
+) -> SurfaceTree {
+    SurfaceTree {
+        surface,
+        position: location.into(),
+        z_index: 99, /* Cursor should always be on-top */
```
Drakulix, line 76: This is not a cursor, isn't it? :thinking: 
Please update the comment and rather use 98 (so the cursor is the only thing on top of the input-popup).

anvil/src/winit.rs
```diff
@@ -218,6 +219,15 @@ pub fn run_winit(log: Logger) {
                 }
             }
 
+            // draw input method square if any
+            let input_method = state.seat.input_method();
+            let (x, y, _, height) = input_method.coordinates();
+            input_method.with_surface(|surface| {
+                if surface.alive() {
```
Drakulix, line 226: Can we move the alive-check into `input_method.with_surface`? So that downstream does not have to deal with it.

src/wayland/input_method_manager/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+use super::{seat::XkbConfig, text_input_manager::TextInputHandle};
+
+const MANAGER_VERSION: u32 = 1;
+
+mod input_method;
+mod input_method_keyboard_grab;
+mod input_method_popup_surface;
+
+/// Extends [Seat] with input method functionality
+pub trait InputMethodSeatTrait {
```
Drakulix, line 90: Maybe just call it `InputMethodSeat`, the "Trait" is a little redundant.

src/wayland/text_input_manager/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+pub use text_input::TextInputUserData;
+
+use super::input_method_manager::InputMethodHandle;
+
+const MANAGER_VERSION: u32 = 1;
+
+mod text_input;
+
+/// Extends [Seat] with text input functionality
+pub trait TextInputSeatTrait {
```
Drakulix, line 78: Same here, `TextInputSeat` is enough imo.

wlcs_anvil/src/main_loop.rs
```diff
@@ -116,6 +117,15 @@ pub fn run(channel: Channel<WlcsEvent>) {
                 }
             }
 
+            // draw input method square if any
+            let input_method = state.seat.input_method();
+            let (x, y, _, height) = input_method.coordinates();
+            input_method.with_surface(|surface| {
+                if surface.alive() {
```
Drakulix, line 124: Same as with anvil.

Drakulix (commented):
Some nitpicks, but generally this looks very good! Thanks!

src/wayland/mod.rs
```diff
@@ -51,13 +51,15 @@ pub mod buffer;
 pub mod compositor;
 pub mod data_device;
 pub mod dmabuf;
+pub mod input_method_manager;
```
Drakulix, line 54: Also a little nitpicky, but generally the modules are named after the protocols.
And these protocols are named `input_method` and `text_input` not `*_manager`, that is just the internal type.
So if you don't mind please rename these.

rano-oss: Done :)

wlcs_anvil/src/main_loop.rs
rano-oss: fixed

src/wayland/text_input_manager/mod.rs
rano-oss: fixed

src/wayland/input_method_manager/mod.rs
rano-oss: Fixed

anvil/src/winit.rs
rano-oss: Fixed

anvil/src/drawing.rs
rano-oss: fixed(?)

anvil/src/state.rs
```diff
@@ -287,6 +296,9 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
             *cursor_status3.lock().unwrap() = new_status;
         });
 
+        seat.text_input();
```
Drakulix, line 299: Why is this call here? I looks a little out of place.

rano-oss: It adds text input to the seat, should I rename it to add_text_input()?
I previously also used it to access the a handle, but I removed the need for that

Drakulix: Or maybe `init_text_input`. Also please make sure, that other functions that need the userdata added by the function don't unwrap and crash, if the user has not called `init_text_input`. Also maybe add hints to their docs, that this call is required.

rano-oss: I have this in the docs:

_This module provides you with utilities to handle input methods, it must be used in conjunction with the text input module to work._

`
seat.init_text_input();
// Add text input capabilities to a seat, needed for the input method to work
seat.add_input_method(XkbConfig::default(), 200, 25);
// Add input method capabilities to a seat
`

_This module provides you with utilities to handle text input surfaces, it must be used in conjunction with the input method module to work._

`
seat.init_text_input();
// Add text input capabilities to a seat
seat.add_input_method(XkbConfig::default(), 200, 25);
// Add input method capabilities to a seat, , needed for text input to work
`

Is that enough? :)

PolyMeilex:
@rano-oss friendly ping, looks like wlcs build is failing after latest changes.

rano-oss:
@PolyMeilex thanks for the ping, I hope it is alright that I request a review

anvil/src/input_handler.rs
```diff
@@ -172,7 +174,7 @@ impl<Backend> AnvilState<Backend> {
         // subsurface menus (for example firefox-wayland).
         // see here for a discussion about that issue:
         // https://gitlab.freedesktop.org/wayland/wayland/-/issues/294
-        if !pointer.is_grabbed() && !keyboard.is_grabbed() {
+        if !pointer.is_grabbed() && (!keyboard.is_grabbed() || input_method.keyboard_grabbed()) {
```
PolyMeilex, line 177: Could you explain this `if`? Existing comment is already kinda big, but it looks like input_method keyboard grab is some sort of special case, and I find it hard to understand. Perhaps update the comment to include explanation of this.

EDIT: Or perhaps this is just a typo and `!` should be before bracket?

anvil/src/winit.rs
```diff
@@ -218,6 +219,13 @@ pub fn run_winit(log: Logger) {
                 }
             }
 
+            // draw input method square if any
```
PolyMeilex, line 222: Kinda nitpicky, but I think that `square` is not the right term here, square always has aspect ratio of 1:1, while input method popup is just any rectangle. Perhaps just call it `input method surface`?

src/wayland/input_method/input_method_popup_surface.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Used to access the relative location of an input popup surface
+    pub fn coordinates(&self) -> (i32, i32, i32, i32) {
+        let inner = self.inner.lock().unwrap();
+        (
+            inner.x + inner.l_x,
+            inner.y + inner.l_y,
+            inner.width,
+            inner.height,
+        )
+    }
```
PolyMeilex, line 50: Have you considered using Smithay's Rectangle type, instead of tupe? It is usually used to store position + size.

src/wayland/input_method/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+        let user_data = self.user_data();
+        user_data.insert_if_missing(InputMethodHandle::default);
+        let input_method = user_data.get::<InputMethodHandle>().unwrap().clone();
+        input_method.configure_keyboard(xkb_config, repeat_delay, repeat_rate);
+    }
+
+    fn input_method(&self) -> Option<&InputMethodHandle> {
+        let user_data = self.user_data();
+        user_data.get::<InputMethodHandle>()
+    }
```
PolyMeilex, line 108: Just making sure that I understand correctly, we return Option here instead of inserting `InputMethodHandle::default` because input method needs to have the keyboard configured first using `add_input_method`?

If the keyboard configuration requirement is not there, I imagine that it could look like this:
```rs
trait InputMethodSeat {
   fn input_method(&self) -> &InputMethodHandle;
}

impl InputMethodSeat for Seat {
   fn input_method(&self) -> &InputMethodHandle {
        let user_data = self.user_data();
        user_data.insert_if_missing(InputMethodHandle::default);
        user_data.get::<InputMethodHandle>().unwrap()
   }
}

let seat: Seat = ...;
seat.input_method().configure_keyboard(...);
seat.input_method().watever_else(...);
```

src/wayland/seat/mod.rs
```diff
@@ -49,7 +49,8 @@
 //! This module further defines the `"cursor_image"` role, that is assigned to surfaces used by clients
 //! to change the cursor icon.
 
-mod keyboard;
+#[allow(missing_docs)]
+pub mod keyboard;
```
PolyMeilex, line 53: Is this right? This exposes `KeymapFile` to the consumers of the API. Do they need it somewhere? Perhaps you meant `pub(crate) mod keyboard`?  Using `pub(crate)` also removes the need for all the  `#[allow(missing_docs)]` notations that this PR adds.

If this is indeed meant to be used by the public, we should document it.

PolyMeilex (changes requested):
In general, this looks good to me, just some minor questions and nitpicks.

src/wayland/seat/mod.rs
i509VCB: This whole module heirarchy is just a mess and needs to be reordered

PolyMeilex: Good point, I would move `KeymapFile` struct to `utils` module, if it is used by two different protocols. (pub(crate) question still applies even if it is in utils)

anvil/src/input_handler.rs
rano-oss: Ah, no you are correct. Input methods are special cases, they are a replacement for the regular keyboards but they just redirect keyboard input to a client(the input method client). Focus is updated for text input, then redirected to the original keyboard surface handler. Input methods do not change surface handling in any way.

Maybe I misunderstood some detail about how this grab is supposed to work?

anvil/src/winit.rs
rano-oss: Will do

src/wayland/input_method/input_method_popup_surface.rs
rano-oss: Good suggestion, I will update

src/wayland/input_method/mod.rs
rano-oss: Yes, that is correct, but I will try your suggestion and see how that turns out.

src/wayland/seat/mod.rs
rano-oss: Ah pub(crate) is correct, I wrote this before I knew how that worked. I will fix

anvil/src/winit.rs
rano-oss: fixed

src/wayland/input_method/input_method_popup_surface.rs
rano-oss: Updated

src/wayland/seat/mod.rs
rano-oss: fixed

src/wayland/input_method/mod.rs
rano-oss: I figured it was best to leave it as is

anvil/src/input_handler.rs
rano-oss: Updated comment

src/wayland/seat/keyboard/keymap_file.rs
```diff
@@ -12,12 +12,14 @@ use nix::{
...
 use slog::error;
 
+#[allow(missing_docs)]
 #[derive(Debug)]
 pub struct KeymapFile {
     sealed: Option<SealedFile>,
     keymap: CString,
 }
 
+#[allow(missing_docs)]
```
PolyMeilex, line 22: Are those allow missing docs lines still needed? Those types are now internal types not crate level public. If rustc still complains that those need docs then it's fine I guess.

rano-oss: Apparently not :)

anvil/src/input_handler.rs
```diff
@@ -172,7 +174,9 @@ impl<Backend> AnvilState<Backend> {
         // subsurface menus (for example firefox-wayland).
         // see here for a discussion about that issue:
         // https://gitlab.freedesktop.org/wayland/wayland/-/issues/294
-        if !pointer.is_grabbed() && !keyboard.is_grabbed() {
+        // Input methods is a special case in that they only use the keyboard input,
```
PolyMeilex, line 177: Small grammar nitpick:
```suggestion
        // Input methods are a special case in that they only use the keyboard input,
```
`methods` is plural, therefore `are` is more appropriate
Or perhaps it should be:
```suggestion
        // Input method is a special case in that they only use the keyboard input,
```

PolyMeilex (approved):

PR #445, opened 2021-12-31
seat: Export KeysymHandle

Drakulix:
An oversight, interesting Rust allows this, given it is part of the public api through the closure in `KeyboardHandle::input`.


vberger (approved):
LGTM, but rustfmt disagrees. :sweat_smile:

PR #450, opened 2022-01-06
0.30: Update wayland-rs delegate traits

PolyMeilex:
This is PR updates wayland-rs in order to: 
- Use new delegate traits that don't take &mut self
- Removes all generic parameters that were previously used for DisplayHandle 

PR: If you see some unused D in some places, consider this a mistake, I might miss some of them.


Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [package]
+wayland-sys = { git  = "https://github.com/Smithay/wayland-rs.git", optional = true }
```
vberger, line 2: Looks like this line has nothing to do here, is it a failpaste?

examples/compositor.rs
```diff
@@ -63,192 +57,21 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            global_data,
-            data_init,
-        );
-    }
-}
+delegate_global_dispatch!(App: [WlCompositor] => CompositorState);
+delegate_dispatch!(App: [WlCompositor] => CompositorState);
+delegate_dispatch!(App: [WlSurface] => CompositorState);
+delegate_dispatch!(App: [WlRegion] => CompositorState);
+delegate_dispatch!(App: [WlCallback] => CompositorState);
```
vberger, line 73: FYI you should be able to combine the various delegates into a single macro call:
```suggestion
delegate_dispatch!(App: [WlCompositor, WlSurface, WlRegion, WlCallback] => CompositorState);
```

vberger (commented):
I looks like the PR also contains a bunch of reformatted unrelated code?

examples/compositor.rs
PolyMeilex: Wow, that's cool!

Cargo.toml
PolyMeilex: yep, indeed

PolyMeilex:
> I looks like the PR also contains a bunch of reformatted unrelated code?

Yep, backend code got reformatted by accident, I removed that from the PR.

PR #455, opened 2022-01-07
Implement WlOutput and XdgOutput

PolyMeilex:


src/wayland/output/handlers.rs
```diff
@@ -0,0 +1,211 @@
...
+        let output = data_init.init(
+            resource,
+            OutputUserData {
+                global_data: global_data.clone(),
+            },
+        );
+
+        let mut inner = global_data.inner.0.lock().unwrap();
+
+        trace!(inner.log, "New global instantiated.");
```
vberger, line 45: If we're going to have these trace messages, maybe they could be a little more informative, maybe at least tell that it is an output that is instantiated.

src/wayland/output/mod.rs
```diff
@@ -339,82 +355,95 @@ impl Output {
...
     /// with the matching client output
-    pub fn enter(&self, surface: &wl_surface::WlSurface) {
-        if let Some(client) = surface.as_ref().client() {
-            self.with_client_outputs(client, |output| surface.enter(output))
+    pub fn enter(&self, dh: &mut DisplayHandle<'_>, surface: &wl_surface::WlSurface) {
+        let client = dh.get_client(surface.id()).ok();
+
+        if let Some(client) = client {
+            self.with_client_outputs(dh, &client, |dh, output| surface.enter(dh, output))
         }
```
vberger, line 425: If we're going to just match it using `if let`, what's the point of using `.ok()` here?

src/wayland/output/mod.rs
```diff
@@ -339,82 +355,95 @@ impl Output {
...
     /// with the matching client output
-    pub fn leave(&self, surface: &wl_surface::WlSurface) {
-        if let Some(client) = surface.as_ref().client() {
-            self.with_client_outputs(client, |output| surface.leave(output))
+    pub fn leave(&self, dh: &mut DisplayHandle<'_>, surface: &wl_surface::WlSurface) {
+        let client = dh.get_client(surface.id()).ok();
+
+        if let Some(client) = client {
+            self.with_client_outputs(dh, &client, |dh, output| surface.leave(dh, output))
         }
```
vberger, line 435: Same here

src/wayland/output/handlers.rs
PolyMeilex: Indeed, not it does say output global, and prints name of the output as well

PR #457, opened 2022-01-08
Allow checking if KeyboardHandle is focused on a surface

dragonnn:
This PR add a simple function to KeyboardHandle with allows checking if they is currently a focused surface with grabs the keyboard.
Motivation behind it is when writing a WM with internal build in shell for example by using imgui keyboard events routed to it needs to be omitted if the Keyboard is focused on a surface.


vberger:
Thanks!

PR #459, opened 2022-01-09
0.30: Update Seat to allow data passing

PolyMeilex:
- Allows users to pass arbitrary data into grabs
- Implements `is_focused` method
- Implements `from_resource`  (Useful for potential multiseat support)
- Implements `user_data` method (I believe some yet unported modules will make use of it)

### DIFF
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -90,8 +90,8 @@ impl ShmHandler for App {
     }
 }
 
-impl SeatHandler for App {
-    fn seat_state(&mut self) -> &mut SeatState {
+impl SeatHandler<Self> for App {
+    fn seat_state(&mut self) -> &mut SeatState<Self> {
         &mut self.seat_state
     }
 }
@@ -100,7 +100,7 @@ struct App {
     compositor_state: CompositorState,
     xdg_shell_state: XdgShellState,
     shm_state: ShmState,
-    seat_state: SeatState,
+    seat_state: SeatState<Self>,
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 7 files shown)


PolyMeilex:
Converting to draft for now, as I have to test it in smallvil after the huge rebase that happened in the meantime.

vberger:
General question, as I'm not sure I've followed this through. What is the rationale for introducing a new type parameter `T`, rather than just using the global `D` state (especially given that's what the examples do)?

PolyMeilex:
The idea was that you can keep grabs encapsulated as they used to be, and pass grab specific state or something along those lines, but personally I don't have any plans to use it, I will just pass D as examples do.

Also. passing D requires us to trick borrow checker by doing something like this:
```rs
// Clone is here only because we can't borrow self two times. calling anything on self.pointer inside of grab will result in a deadlock
self.pointer.clone().motion(self, cx, pos, under, serial, event.time());
``` 
So I can imagie a case when someone does no want to do this.

vberger:
Alright, I see. :+1:

One point though: if there is a deadlock risk, I think it should be clearly stated in the docs.

PolyMeilex:
Not sure where to document this, should every seat method state this? Previously the compositor would just panic as it was using refcell so the issue was obvious, but now that mutexes are mandatory it just deadlocks.

vberger:
I think the documentation of the `PointerGrab` trait would be a good place, as it is an indication for implements of that trait: any interaction with the pointer from within the grab callbacks should be done through the provided inner handle, and not an instance of the base `Pointer`, otherwise it'll deadlock.

PR #460, opened 2022-01-10
EGL and Gles2 cleanups and additions

Drakulix:
Best reviewed on a commit-by-commit basis.

Addresses some short-comings I found implementing smithay-egui.

6e1f6ab1f3e4d3e705c55c51d6c17d09b6f39068 makes it possible to attach user_data to an `EGLContext`. Given that GL-resources like textures or buffers are local to a context, this is the obvious place to store them.

e0ad187411cbcb4e7d7586b9676c47a07069686f makes it possible to retrieve the `EGLContext` of a `Gles2Renderer` to actually access the user_data from downstream.

7ac8872a5ebac9cf8a6570da3ccaa4199dbe1751 and 07bf1e563955c00b2ec50bb51e48ce1c1562f577 expose rendering specifics of the `Gles2Frame`, that make it easier to correctly orientate and size externally drawn content.

04f6434b76c9ac0d2d1d7ed10873c9673005cbf3 allows downstream to retrieve the raw-egl handles, if they want to. We already supported that for the `EGLDevice` and `EGLDisplay` anyway, and this is useful for example to hook up RenderDoc.

08df10c958bc73181e5ce5ee7dfa63669c5174cf fixes up the existing texture filtering code of the `Gles2Renderer`, which was previously not applied correctly.

29b0b2b1bdad43bdc1e8a017880a565f4ca34aa2 fixes resizing the window of the winit-backend in anvil, which broke with #423.

Finally 864b379 adds a utility function to transform `Rectangle`s and some tests.

### DIFF
--- a/src/backend/renderer/gles2/mod.rs
+++ b/src/backend/renderer/gles2/mod.rs
@@ -1428,4 +1428,9 @@ impl Gles2Frame {
 
         Ok(())
     }
+
+    /// Projection matrix for this frame
+    pub fn projection(&self) -> &[f32; 9] {
+        self.current_projection.as_ref()
+    }
 }
--- a/src/backend/renderer/gles2/mod.rs
+++ b/src/backend/renderer/gles2/mod.rs
@@ -1017,12 +1017,25 @@ impl Drop for Gles2Renderer {
 }
 
 impl Gles2Renderer {
+    /// Get access to the underlying [`EGLContext`].
+    ///
+    /// *Note*: Modifying the context state, might result in rendering issues.
+    /// The context state is considerd an implementation detail
+    /// and no guarantee is made about what can or cannot be changed.
+    /// To make sure a certain modification does not interfere with
+    /// the renderer's behaviour, check the source.
+    pub fn egl_context(&self) -> &EGLContext {
+        &self.egl
+    }
+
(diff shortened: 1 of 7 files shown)


src/backend/renderer/gles2/mod.rs
```diff
@@ -1017,12 +1017,21 @@ impl Drop for Gles2Renderer {
 }
 
 impl Gles2Renderer {
+    /// Get access to the underlying [`EGLContext`].
+    ///
+    /// *Note*: Modifying the context state, might result in rendering issue.
+    /// To make sure a certain modification is safe, check the source.
```
vberger, line 1031: "check the source" in a documentation is pretty unusual :laughing: maybe this could be expanded a little?

Drakulix: The thing is OpenGL context state is huge and complicated and we *might* adjust the renderer in the future.

If we document every state the renderer needs, we effectively turn this into an api, rather then being an implementation detail. And I dont want to break the api for every small renderer change.

So the idea for downstream *should* be more along the lines of: reset everything you've changed in the end *or* look at the source. 😅

Drakulix: done

vberger (approved):

PR #448, opened 2022-01-05
Update anvil to show off desktop abstractions

Drakulix:
Currently WIP.

- [ ] Depends on #423. (needs to be merged.)
- [ ] History needs cleanup.
- [ ] RenderDoc support needs to be moved into another PR or flashed out.
- [ ] Output scale changes need fixup.
- [ ] Desktop abstractions need to support XWayland to not downgrade anvils feature set.

PR #462, opened 2022-01-12
Popup Grab

cmeissl:
This PR adds support for popup grabs.

It introduces a new `PopupGrab` which is returned from `PopupManager::grab_popup`.
The `PopupGrab` implements `KeyboardGrab` and `PointerGrab` and can be set by the compositor
on the `Seat` handles.
Additional the main logic of the grab is accessible from the `PopupGrab` to allow the user to implement
a custom grab strategy.


src/desktop/popup/manager.rs
```diff
@@ -56,24 +60,62 @@ impl PopupManager {
...
+                }
+
+                toplevel_popups
+            }
+        })?;
+
+        // The parent of a grabbing popup must either be another xdg_popup with an active explicit grab, or an xdg_popup or xdg_toplevel, if there are no explicit grabs already taken.
+        let parent_role = get_role(&parent);
+
+        if parent_role != Some(XDG_TOPLEVEL_ROLE) && parent_role != Some(XDG_POPUP_ROLE) {
```
Drakulix, line 87: This should take layer_shell into account, shouldn't it? Maybe we need some helper-function somewhere "is_toplevel_like" or something like that.

src/desktop/popup/manager.rs
```diff
@@ -56,24 +60,62 @@ impl PopupManager {
...
+            root.as_ref()
+                .post_error(
+                    wayland_protocols::xdg_shell::server::xdg_popup::Error::InvalidGrab as u32,
+                    "The parent of a popup grab must be either another xdg_popup with an active grab or and xdg_toplevel".to_string());
```
Drakulix, line 91: "or and"?

src/desktop/popup/manager.rs
```diff
@@ -56,24 +60,62 @@ impl PopupManager {
...
+            if toplevel_popups.is_dismissed(&parent) {
+                popup.send_done();
+                return Err(PopupGrabError::ParentDismissed);
+            }
+
+            if !toplevel_popups.has_grab(&parent) {
+                root.as_ref()
+                .post_error(
+                    wayland_protocols::xdg_shell::server::xdg_popup::Error::InvalidGrab as u32,
+                    "The parent of a popup grab must be either another xdg_popup with an active grab or and xdg_toplevel".to_string());
```
Drakulix, line 105: see above

src/desktop/popup/manager.rs
```diff
@@ -56,24 +60,62 @@ impl PopupManager {
...
+                return Err(PopupGrabError::InvalidGrab);
```
Drakulix, line 106: The way I read the spec, it can be a popup without an explicit-grab, as long as there are no other grabs happening. (If there are, *then* the parent needs to be specifically the one-and-only popup with that grab. This does not seem to account for that?

anvil/src/shell.rs
```diff
@@ -725,7 +725,20 @@ pub fn init_shell<BackendData: Backend + 'static>(
                     surface,
                     seat,
                 } => {
-                    warn!(log_ref, "Unhandled grab");
+                    debug!(log_ref, "Popup grab");
```
Drakulix, line 728: Should we not check for already active grabs? Or should a popup-grab override a resize or move grab?

Drakulix (commented):
Overall I really like the approach, I think the interface/api is pretty solid and fits nicely into the module. But I am not sure, every detail is correct yet, but maybe I have misinterpreted some code.

src/desktop/popup/manager.rs
cmeissl: I am really not sure, the spec states:

> The parent of a grabbing popup must either be an xdg_toplevel surface or another xdg_popup with an explicit grab. If the parent is another xdg_popup it means that the popups are nested, with this popup now being the topmost popup.

But yeah, that would limit the usage a bit.

src/desktop/popup/manager.rs
cmeissl: To be honest i am quiet confused by the spec.

At first it states

> The parent of a grabbing popup must either be an xdg_toplevel surface or another xdg_popup with an explicit grab. If the parent is another xdg_popup it means that the popups are nested, with this popup now being the topmost popup.

then shorty after

> The parent of a grabbing popup must either be another xdg_popup with an active explicit grab, or an xdg_popup or xdg_toplevel, if there are no explicit grabs already taken.

and near the end

> If the parent is a grabbing popup which has already been dismissed, this popup will be immediately dismissed. If the parent is a popup that did not take an explicit grab, an error will be raised.

src/desktop/popup/manager.rs
cmeissl: Oops, thanks for pointing that out.

src/desktop/popup/manager.rs
Drakulix: So other shells can have popups, but not explicit grabs, if we would follow the spec?

src/desktop/popup/manager.rs
Drakulix: Very much understandable

> [...] or another xdg_popup with an explicit grab.  [...]


+1 for only xdg-popups with an explicit grab being valid targets

> [...]  either be another xdg_popup with an active explicit grab, or an xdg_popup or xdg_toplevel, if there are no explicit grabs already taken.


Clear point for either xdg_toplevel or xdg_popup, if no grabs exist, **or** an xdg_popup with a grab.

> [...] If the parent is a popup that did not take an explicit grab, an error will be raised.


So now we get an error, if the parent popup has no grab.. So +1 for only grabbed popups being valid parents.

It seems to contradict itself, so I would argue, that we should open an issue https://gitlab.freedesktop.org/wayland/wayland-protocols.

For the implementation we should probably follow what other compositors are doing right now?

anvil/src/shell.rs
cmeissl: Initially i had a check for that, but there was a race. The grab is called as a result of an input event, so it is possible that the DefaultGrab in Pointer has still set a ClickGrab which would make a simple check for an active grab fail. Also a grab on a nested popup could already have an active popup grab. But maybe it is possible to use a serial as a condition. Something like if there is a active grab with either the provided serial or the serial stored in the popup grab from the last grab it is ok to override the grab.

src/desktop/popup/manager.rs
cmeissl: Yeah, but i have looked at what weston and wlroots have implemented. Weston has a similar check, but this is no suprise as weston has no support for `wlr-layer-shell`. wlroots seems to only check if the grabbed popup is the topmost by inspecting if the grabbed popup has children or not. The layer layer shell spec specifies the following:

> zwlr_layer_surface_v1::set_keyboard_interactivity
>
> Set how keyboard events are delivered to this surface. By default, layer shell surfaces do not receive keyboard events; this request can be used to change this.
> This setting is inherited by child surfaces set by the get_popup request.

So it seems that it allows the grab. I already found some other issues in my implementation and will try to account for layer shell in the fixes. For example returning the keyboard focus is implemented wrong.

cmeissl:
The grab handling has been implemented in #467. #448 adds support in anvil.

PR #463, opened 2022-01-12
desktop: Fix damage for removed popups

Drakulix:
Fixes missing damage for removed popups inside the window bounds by tracking them like other toplevels. Drawing operations are still grouped, because popups without a parent are not useful. The only api that changes for downstream is `Window::accumulated_damage`, which now only returns the windows damage (as the docs suggested anyway) instead of including popup damage.


vberger (approved):
LGTM. I'll see about these clippy errors.

PR #466, opened 2022-01-14
Apply transformations correctly

Drakulix:
Damage values for rendered textures on transformed outputs are currently off, because we calculate the positions in the cpu before uploading them to the gpu in opengl coordinate space.

The first commit simply applies the transformation beforehand.

The second fixes a slight mistake in the Transform::matrices and then finally applies buffer-transformations correctly instead.
To do that it:
- Moves `Transform` into utils::geometry
- Changes conversion from and into buffer-coordinates to take
  `Transform` arguments.
- `Renderer` take `Buffer`-space instead of `Physical` (so render-space) damage now
- buffer_transform is taken into account everywhere

### DIFF
--- a/src/backend/renderer/gles2/mod.rs
+++ b/src/backend/renderer/gles2/mod.rs
@@ -1295,13 +1295,19 @@ impl Frame for Gles2Frame {
                 let rect_constrained_loc = rect
                     .loc
                     .constrain(Rectangle::from_extemities((0f64, 0f64), dest.size.to_point()));
-                let rect_clamped_size = rect.size.clamp((0f64, 0f64), dest.size);
+                let rect_clamped_size = rect.size.clamp(
+                    (0f64, 0f64),
+                    (dest.size.to_point() - rect_constrained_loc).to_size(),
+                );
+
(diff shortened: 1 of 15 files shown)


src/backend/renderer/gles2/mod.rs
```diff
@@ -1290,24 +1290,24 @@ impl Frame for Gles2Frame {
...
                 let rect = rect.to_f64();
 
                 let rect_constrained_loc = rect
                     .loc
-                    .constrain(Rectangle::from_extemities((0f64, 0f64), dest.size.to_point()));
-                let rect_clamped_size = rect.size.clamp(
-                    (0f64, 0f64),
-                    (dest.size.to_point() - rect_constrained_loc).to_size(),
-                );
+                    .constrain(Rectangle::from_extemities((0f64, 0f64), src.to_point()));
```
vberger, line 1298: Do I understand correctly that the change from using `dest.size` to `src.size` is the mistake that is being fixed?

Drakulix: Correct. The damage is relative to the buffer, so in cases where dest != src, the clamping may be wrong.

PR #467, opened 2022-01-15
Add support for popup grabs to desktop abstraction

cmeissl:
The PR adds support for popup grabs to the desktop abstraction.

A popup grab is only allowed before the initial commit, to check for that protocol error the field `committed` 
has been added to the `XdgPopupSurfaceRoleAttributes`.

To support exclusive keyboard focus a `KeyboardGrab` similiar to the existing `PointerGrab` has been added.

During testing i found an issue with the current popup handling that prevents menu popups from working correctly.
If a popup has no grab it should not receive the keyboard focus, for that a `WindowSurfaceType` has been added
that can be used to filter the surfaces in `surfaces_under`.


Drakulix (approved):
Looks good to me now and seems to follow the spec. Pinging @vberger because this touches a lot of wayland code as well and I am not 100% sure I didn't miss something.

vberger (approved):
Sorry for the late review. That looks very neat, thanks a lot!

PR #469, opened 2022-01-15
fix setting the parent of a ToplevelSurface

cmeissl:
`set_parent` of `ToplevelSurface` always removed the current parent, this PR fixes that.

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -1169,15 +1169,14 @@ impl ToplevelSurface {
     /// The parent must be another toplevel equivalent surface.
     ///
     /// If the parent is `None`, the parent-child relationship is removed.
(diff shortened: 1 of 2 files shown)


src/wayland/shell/xdg/mod.rs
```diff
@@ -1170,14 +1170,13 @@ impl ToplevelSurface {
     ///
     /// If the parent is `None`, the parent-child relationship is removed.
     pub fn set_parent(&self, parent: Option<wl_surface::WlSurface>) -> bool {
```
i509VCB, line 1172: WlSurface -> &WlSurface would probably make this nicer to use.

cmeissl: yeah, done.

vberger (approved):

PR #470, opened 2022-01-16
desktop.space: Fix multioutput rendering

PolyMeilex:


src/desktop/space/mod.rs
```diff
@@ -600,6 +600,9 @@ impl Space {
                     clear_color,
                     &damage
                         .iter()
+                        // Map from global space to output space
+                        .map(|geo| Rectangle::from_loc_and_size(geo.loc - output_geo.loc, geo.size))
+                        // Map from logical to phisical
```
vberger, line 605:
```suggestion
                        // Map from logical to physical
```

src/desktop/space/mod.rs
```diff
@@ -626,6 +629,8 @@ impl Space {
                             .iter()
                             .flat_map(|d| d.intersection(geo))
                             .map(|geo| Rectangle::from_loc_and_size(geo.loc - loc, geo.size))
+                            // Map from global compositor space to output one
+                            .map(|geo| Rectangle::from_loc_and_size(geo.loc - output_geo.loc, geo.size))
```
Drakulix, line 633:
```suggestion
                            // Map from global compositor space to output one
                            .map(|geo| Rectangle::from_loc_and_size(geo.loc - output_geo.loc, geo.size))
                            // Map from output space to surface relative coordinates
                            .map(|geo| Rectangle::from_loc_and_size(geo.loc - loc, geo.size))
```

Drakulix (approved):

src/desktop/space/mod.rs
Drakulix: Wait, why is this second one necessary, "loc" already is subtracted by `output_geo.loc` in [line 627](https://github.com/Smithay/smithay/pull/470/files/de0c2e0016498361f40e42b2816c36c6f24f3155#diff-55c984b4634eecc8afa1b426c76e53c642a67f51f7cc73647927edbead721be7R627)?

@PolyMeilex Can you try changing `loc` to only contain `element.location` and nothing else? That should work as well..

PolyMeilex: Yep, that was my initial question on matrix, I'm not sure why it is double substracted.

PolyMeilex: Nope that did not work, it only shows up on the screen when it is subtracted in two places 🤷

Drakulix: Well in that case, I am going to debug that at first, before we merge this. It is just too weird.

Drakulix (changes requested):
See above. Needs more investigation.

Drakulix (approved):

vberger (approved):

PR #473, opened 2022-01-18
Add layer definition to RenderElement

dragonnn:
This PR add support for definition of layer a RenderElement passed to render_output should be draw on.
This is done via layer function in RenderElement trait, it should return a layer an element should be draw on.
A default implementation is provided which makes the element render on top of all other elements to preserve the current behavior and not introduce any braking changes.

Marking PR as draft because I hope some comments/suggestion for improvements, mostly probably for for the whole render loop chain got pretty scary.
Maybe add some simple functions is_top, is_above_background etc. to at least skip the `c.layer() == RenderLayer::AboveBackground`? 
Another option would be to wrap the `&[DynamicRenderElements<R>]` in a struct and add something similar to `layers_on`.
Thanks for any feedback :).

### DIFF
--- a/src/desktop/space/element.rs
+++ b/src/desktop/space/element.rs
@@ -35,6 +35,7 @@ pub type DynamicRenderElements<R> =
 pub(super) type SpaceElem<R> =
     dyn SpaceElement<R, <R as Renderer>::Frame, <R as Renderer>::Error, <R as Renderer>::TextureId>;
 
+/// Helper struct for iterating over diffrent layers of `DynamicRenderElements`
 pub(super) struct DynamicRenderElementMap<'a, R: Renderer>(pub(super) &'a [DynamicRenderElements<R>]);
 
 impl<'a, R> DynamicRenderElementMap<'a, R>
@@ -44,30 +45,37 @@ where
     R::Error: 'static,
     R::Frame: 'static,
 {
+    /// Iterate over `DynamicRenderElements` with layer `RenderLayer::Bottom`
     pub fn iter_bottom(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
         self.iter_layer(RenderLayer::Bottom)
     }
 
+    /// Iterate over `DynamicRenderElements with layer `RenderLayer::AboveBackground`
     pub fn iter_above_background(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
         self.iter_layer(RenderLayer::AboveBackground)
     }
 
+    /// Iterate over `DynamicRenderElements` with layer `RenderLayer::BeforeWindows`
     pub fn iter_before_windows(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
         self.iter_layer(RenderLayer::BeforeWindows)
     }
 
+    /// Iterate over `DynamicRenderElements` with layer `RenderLayer::AfterWindows`
     pub fn iter_after_windows(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
         self.iter_layer(RenderLayer::AfterWindows)
     }
 
+    /// Iterate over `DynamicRenderElements` with layer `RenderLayer::BeforeOverlay`
     pub fn iter_before_overlay(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
         self.iter_layer(RenderLayer::BeforeOverlay)
     }
 
+    /// Iterate over `DynamicRenderElements` with layer `RenderLayer::Top`
     pub fn iter_top(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
         self.iter_layer(RenderLayer::Top)
     }
 
+    /// Iterate over `DynamicRenderElements` with provided `layer`
     pub fn iter_layer(&'a self, layer: RenderLayer) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
         Box::new(
             self.0
--- a/src/desktop/space/element.rs
+++ b/src/desktop/space/element.rs
@@ -12,6 +12,7 @@ use wayland_server::protocol::wl_surface::WlSurface;
 
 /// Enum for indicating on with layer a render element schould be draw
 #[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum RenderLayer {
     /// Bellow every other elements
     Bottom,
@@ -25,6 +26,56 @@ pub enum RenderLayer {
     BeforeOverlay,
     /// Above anything else
     Top,
+}
+
+/// Elements rendered by [`Space::render_output`] in addition to windows, layers and popups.
+pub type DynamicRenderElements<R> =
+    Box<dyn RenderElement<R, <R as Renderer>::Frame, <R as Renderer>::Error, <R as Renderer>::TextureId>>;
+
+pub(super) type SpaceElem<R> =
+    dyn SpaceElement<R, <R as Renderer>::Frame, <R as Renderer>::Error, <R as Renderer>::TextureId>;
+
+pub(super) struct DynamicRenderElementMap<'a, R: Renderer>(pub(super) &'a [DynamicRenderElements<R>]);
+
+impl<'a, R> DynamicRenderElementMap<'a, R>
+where
+    R: Renderer + ImportAll + 'static,
+    R::TextureId: 'static,
+    R::Error: 'static,
+    R::Frame: 'static,
+{
+    pub fn iter_bottom(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
+        self.iter_layer(RenderLayer::Bottom)
+    }
+
+    pub fn iter_above_background(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
+        self.iter_layer(RenderLayer::AboveBackground)
+    }
+
+    pub fn iter_before_windows(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
+        self.iter_layer(RenderLayer::BeforeWindows)
+    }
+
+    pub fn iter_after_windows(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
+        self.iter_layer(RenderLayer::AfterWindows)
+    }
+
+    pub fn iter_before_overlay(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
+        self.iter_layer(RenderLayer::BeforeOverlay)
+    }
+
+    pub fn iter_top(&'a self) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
+        self.iter_layer(RenderLayer::Top)
+    }
+
+    pub fn iter_layer(&'a self, layer: RenderLayer) -> Box<dyn Iterator<Item = &SpaceElem<R>> + 'a> {
+        Box::new(
+            self.0
+                .iter()
+                .filter(move |c| c.layer() == layer)
+                .map(|c| c as &SpaceElem<R>),
+        )
+    }
 }
 
 /// Trait for custom elements to be rendered during [`Space::render_output`].
--- a/src/desktop/space/element.rs
+++ b/src/desktop/space/element.rs
@@ -20,14 +20,14 @@ pub enum RenderZindex {
     Bottom = 20,
     /// Default zindex for Windows
     Shell = 30,
-    /// Default zindex for Windows PopUps
-    PopUpsShell = 40,
     /// WlrLayer::Top default zindex
-    Top = 50,
+    Top = 40,
+    /// Default zindex for Windows PopUps
+    Popups = 50,
     /// Default Layer for RenderElements
-    Overlay = 80,
+    Overlay = 60,
     /// Default Layer for Overlay PopUp
-    PopUpsOverlay = 100,
+    PopupsOverlay = 70,
 }
 
 /// Elements rendered by [`Space::render_output`] in addition to windows, layers and popups.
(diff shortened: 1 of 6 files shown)


src/desktop/space/mod.rs
```diff
@@ -613,18 +614,53 @@ impl Space {
...
+                            .layers_on(WlrLayer::Background)
+                            .map(|l| l as &SpaceElem<R>)
+                            .chain(
+                                custom_elements
+                                    .iter()
+                                    .filter(|c| c.layer() == RenderLayer::AboveBackground)
+                                    .map(|c| c as &SpaceElem<R>),
+                            )
+                            .chain(layer_map.layers_on(WlrLayer::Bottom).map(|l| l as &SpaceElem<R>)),
+                    )
```
Drakulix, line 632:
```suggestion
                    )
                    .chain(
                        custom_elements
                            .iter()
                            .filter(|c| c.layer() == RenderLayer::AboveBackground)
                            .map(|c| c as &SpaceElem<R>),
                    )
                    .chain(layer_map.layers_on(WlrLayer::Bottom).map(|l| l as &SpaceElem<R>))
```

These two chains can be moved into the upper chain for consistency, no?

src/desktop/space/mod.rs
```diff
@@ -613,18 +614,53 @@ impl Space {
...
-                            .map(|l| l as &SpaceElem<R>),
+                            .map(|l| l as &SpaceElem<R>)
+                            .chain(
+                                custom_elements
+                                    .iter()
+                                    .filter(|c| c.layer() == RenderLayer::BeforeOverlay)
+                                    .map(|c| c as &SpaceElem<R>),
+                            )
+                            .chain(layer_map.layers_on(WlrLayer::Overlay).map(|l| l as &SpaceElem<R>)),
+                    )
```
Drakulix, line 657:
```suggestion
                    )
                    .chain(
                        custom_elements
                            .iter()
                            .filter(|c| c.layer() == RenderLayer::BeforeOverlay)
                            .map(|c| c as &SpaceElem<R>),
                    )
                    .chain(layer_map.layers_on(WlrLayer::Overlay).map(|l| l as &SpaceElem<R>))
```

Same here

src/desktop/space/element.rs
```diff
@@ -55,6 +72,11 @@ where
         damage: &[Rectangle<i32, Logical>],
         log: &slog::Logger,
     ) -> Result<(), R::Error>;
+
+    /// Returns they layer the elements schould be draw on, defaults to Top
+    fn layer(&self) -> RenderLayer {
```
Drakulix, line 77: I am not sure about the name `layer`, given that it might be possible to confuse this with layer-shell, but I don't really have a better idea.

Drakulix (commented):
In general this looks pretty good.
Yes the chain is getting a little insane, but I don't really see, how we could break it apart. Without causing a bunch of duplicated code. (Which I would want to avoid, given the complexity of damage tracking in general.)

Drakulix:
> Maybe add some simple functions is_top, is_above_background etc. to at least skip the c.layer() == RenderLayer::AboveBackground?

This is likely a good idea, but we could just name the function `is_render_layer` and then make each filter do:
`.filter(|x| is_render_layer(x, RenderLayer::BeforeOverlay))` ?

Those that is still pretty long. Maybe add `is_top` to `RenderLayer` so we can call `filter(|c| c.layer().is_top())` ?

The goal should probably be to make rustfmt treat all our chains a one-liners, so this keeps being easily readable.
For that goal we might also want to add an internal function inside `Space::render_output` for casting, something like:

```rust
fn as_space_elem<R>(elem: &impl SpaceElement<R, <R as Renderer>::Frame, <R as Renderer>::Error, <R as Renderer>::TextureId>) -> &SpaceElem<R> {
 elem as &SpaceElem<R>
}
```

So we can just call `.map(as_space_elem)`? But I don't know how much shorter that would really be.

dragonnn:
> The goal should probably be to make rustfmt treat all our chains a one-liners, so this keeps being easily readable. For that goal we might also want to add an internal function inside `Space::render_output` for casting, something like:

Yeah exactly what I was thinking that too, getting rustfmt to treat all chains as one line, I will test different option and try to come up with a solution that works

dragonnn:
What about that solution?
Needs a extra struct but I think it is worth it. Render loop looks now much better

Drakulix:
> What about that solution (will add some docs comments when it is done)?

I like that! Though I am not a big fan of having this definition inside the `space/mod.rs`, because that is already a monster to begin with. Maybe we can move it (and the type definitions) in the `element.rs`-module.

> Only thinking if it is good to skip the .map(|w| w as &SpaceElem<R>) in the struct, this way it could be public

This struct is an implementation detail to improve readability, it does not need to be part of the public api at all. (Like `SpaceElement`, which is deliberately not part of the public api. Only `RenderElement` should be.)

Also could you make `RenderLayer` non_exhaustive? If we add support for other shells in the future, we might have to add new render layers.

dragonnn:
> I like that! Though I am not a big fan of having this definition inside the `space/mod.rs`, because that is already a monster to begin with. Maybe we can move it (and the type definitions) in the `element.rs`-module.

If you are fine with moving `SpaceElem<R>` to elements.rs I am perfectly fine with that, my first idea was to put it there but wasn't sure about moving `SpaceElem<R>`

> Also could you make `RenderLayer` non_exhaustive? If we add support for other shells in the future, we might have to add new render layers.

Sure :)

Drakulix:
Sure, move it. no problem

dragonnn:
Ok I think everything now should be fine. If you have any more thoughts/needed changes let me know

src/desktop/space/element.rs
dragonnn: Yeah I didn't come with a better name too, since it needs to line up with RenderElement. 
RenderElementLayer sounds to verbose to me, but could be one option?

src/desktop/space/element.rs
```diff
@@ -10,6 +10,82 @@ use std::{
...
+    /// Right before programs windows are draw
+    BeforeWindows,
+    /// Right after programs windows are draw
+    AfterWindows,
+    /// Above WlrLayer::Top but bellow WlrLayer::Overlay
+    BeforeOverlay,
+    /// Above anything else
+    Top,
+}
+
```
vberger, line 30: Between that and the layer shell layers, this is getting pretty heavy. At this point I'm wondering if using a numeric z-index might make more sense?

Basically, assign a numeric value to each defined protocol layer (so maybe `background = 10, bottom = 20, shell = 30, top = 40, overlay = 50` for example), and when drawing elements, we just sort them according to that z-index using a *stable* sort (to preserve the ordering within a layer) and draw that?

This would give a lot of flexibility to the downstream users of smithay when drawing their content, hopefully with a reasonably small runtime cost (collect the render elements in a Vec and sort it). What do you think @Drakulix?

Drakulix: Oh right. I mean, that does not cover any use case you cannot already do with the enum.

But I guess downstream would not have to carefully manage the order of each and every custom element, if they could just use a z-index.

dragonnn: That is an interesting and flexible idea, but would definitely need more changes in the whole render_output

Drakulix: Not necessarily. You could just sort the array, when we create the DynamicList (and we would need to pass a Vec instead of a slice to do that without copying the references) and then filter for indices between a given range instead of layers.

dragonnn:
I am pretty sure I did screw up some default values, most I am not sure how popus should be handled. 
Also, can we rework the damage loops to the whole array? I think not, because they must be a reason why window_popups and layer_popups are handled in this order?
Any comments/suggestion are welcome

Drakulix (commented):
Looks nice! :+1:
I think this is the right approach now.

Drakulix:
> I am pretty sure I did screw up some default values, mostly I am not sure how popus should be handled in this.

See also the comments in the review. Popups should indeed be on top of most things, except for Overlays. Overlays need to overlay everything with the exception, that they also can have Popups. :sweat_smile:
So we need to copy the layer of the parent into the `RenderPopup` of parent (if is a layer-surface) to determine, which z_index it should have.

> Also, can we rework the damage loops to the whole array? I think not, because they must be a reason why window_popups and layer_popups are handled in this order? Any comments/suggestion are welcome

Actually the old order is broken. E.g. if you do a context menu on a `Background`-layer via right-click, it should not be hidden by Windows (but it would be right now). Also again for damage the order does not matter, but now with the z-index we can finally optimize for `opaque_region` (in a layer PR), which is super nice. So you should just collect the damage from all elements of the sorted array in order, so we can later expand on that.

src/desktop/space/element.rs
```diff
@@ -10,22 +10,22 @@ use std::{
...
+    /// WlrLayer::Bottom default zindex
+    Bottom = 20,
+    /// Not used yet?
+    Shell = 30,
+    /// Default zindex for Windows
+    Top = 40,
+    /// Default Layer for RenderElements
+    Overlay = 50,
+    /// Default Layer for PopUps?
+    PopUp = 60,
```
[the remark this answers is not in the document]
Drakulix: yes, but swap `PopupsShell` and `Top`

src/desktop/space/element.rs
```diff
@@ -163,6 +112,9 @@ where
         damage: &[Rectangle<i32, Logical>],
         log: &slog::Logger,
     ) -> Result<(), R::Error>;
+    fn z_index(&self) -> u8; //{
```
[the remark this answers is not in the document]
dragonnn: RenderElements have a default, look above and for the implementation of SpaceElement for them, the default to RenderZindex::Overlay.
That is the default for SpaceElement overall, I did have a default for it but comment it out to find where I need implement it, but yes it will be better without one

dragonnn:
I think I covered everything, check if the popups are now done right, not 100% sure about it

src/desktop/space/element.rs
```diff
@@ -10,6 +10,33 @@ use std::{
...
+#[repr(u8)]
+pub enum RenderZindex {
+    /// WlrLayer::Background default zindex
+    Background = 10,
+    /// WlrLayer::Bottom default zindex
+    Bottom = 20,
+    /// Default zindex for Windows
+    Shell = 30,
+    /// Default zindex for Windows PopUps
+    PopUpsShell = 40,
```
Drakulix, line 24: Please just rename to `Popups` and move after `Top` (the top-layer can also have popups + popups should also be rendered before the top-layer).

src/desktop/space/popup.rs
```diff
@@ -70,6 +74,7 @@ impl LayerSurface {
                         RenderPopup {
                             location: offset,
                             popup,
+                            parent_layer: RenderZindex::Overlay,
```
Drakulix, line 77: This should only be the case, if this `LayerSurface::layer` actually returns `Overlay`, all other popups should be on the same `Popups` layer.

src/desktop/space/element.rs
```diff
@@ -10,6 +10,33 @@ use std::{
...
+    /// WlrLayer::Top default zindex
+    Top = 50,
+    /// Default Layer for RenderElements
+    Overlay = 80,
+    /// Default Layer for Overlay PopUp
+    PopUpsOverlay = 100,
```
Drakulix, line 30: Also please stylize popups as "Pop**u**p", not "Pop**U**p, to make it consistent to the Pop**u**pManager and other structs already in smithay.

Drakulix (commented):
Almost done! Just a little mixup with popup orders, I think.

src/desktop/space/popup.rs
dragonnn: Changed a lite bit the logic, now z_index is resolved when RenderPopup is created, that made it a lite bit easier to get the logic right for Windows and LayerSurface separately

dragonnn:
Yeah I expect some changes more with Popups 😅 , they are not easy to get right. I think now it should be right

Drakulix (approved):

src/desktop/space/element.rs
```diff
@@ -10,6 +10,33 @@ use std::{
...
+    /// WlrLayer::Background default zindex
+    Background = 10,
+    /// WlrLayer::Bottom default zindex
+    Bottom = 20,
+    /// Default zindex for Windows
+    Shell = 30,
+    /// WlrLayer::Top default zindex
+    Top = 40,
+    /// Default zindex for Windows PopUps
+    Popups = 50,
```
vberger, line 26: So window popups are always drawn above the `wlrlayer::top layer`?

I though popups were considered as if they were part of their parent surface, as far as z-index is concerned. Why do they have their own layer?

src/desktop/space/mod.rs
```diff
@@ -498,23 +490,31 @@ impl Space {
...
+                + self.windows.len()
+                + window_popups.len()
+                + layer_popups.len(),
+        );
+
+        render_elements.extend(&mut custom_elements.iter().map(|l| l as &SpaceElem<R>));
+        render_elements.extend(&mut self.windows.iter().map(|l| l as &SpaceElem<R>));
+        render_elements.extend(&mut window_popups.iter().map(|l| l as &SpaceElem<R>));
+        render_elements.extend(&mut layer_map.layers().map(|l| l as &SpaceElem<R>));
+        render_elements.extend(&mut layer_popups.iter().map(|l| l as &SpaceElem<R>));
```
vberger, line 505: Are those `&mut` really needed?

src/desktop/space/element.rs
Drakulix: Well imagine a background layer having a popup (e.g. a context menu), if you can open it, your mouse cursor is obviously on the background layer. But should the menu extend below an open window? Does that make any sense?

The xdg-shell protocol does not define how they should be stacked (except in relation to other xdg-popups of the same toplevel), but I think this is the expected behavior.

src/desktop/space/mod.rs
dragonnn: Right, I think they where need when I was doing append, works now without them, Fixed, thanks :)

src/desktop/space/element.rs
vberger: Right, good point.

vberger (approved):

PR #474, opened 2022-01-19
desktop.layer: Send frames to popups

Drakulix:
Fixes layer-shell popups no updating correctly.


vberger:
I'm pretty sure the protocol allows to have popups on popups, and it seems that this code only handles the topmost level of popup stacking?

I think this is important, because I suspect this is how are handled submenus in context menus for example.

vberger (approved):

PR #477, opened 2022-01-21
space: Pass a location to `RenderElement::draw`

Drakulix:
Poly noticed, that rendering `RenderElement`s is of, when the output is not mapped to (0, 0), which is correct, because it is difficult to figure out for a custom element right now where it actually should draw. This PR adds a `location` parameter to the `draw` function, because we already do this calculation in smithay anyway.

@PolyMeilex please test. :)


PolyMeilex:
Yep, works like a charm, SurfaceTree based cursors are displayed as expected now, and custom udev one as well.

vberger (approved):

PR #480, opened 2022-01-22
winit: Fix returned buffer age

Drakulix:
Previously if the buffer age of a `WinitGraphicsBackend`s back buffer was queried, it would never return a defined value, because it's surface will never have been bound, because the `Renderer::bind`-call would deferr its `make_current` call until the actual rendering.

Additionally this PR adjusts the `WinitGraphicsBackend::buffer_age` docs to warn the user of this behavior.

### DIFF
--- a/src/backend/egl/surface.rs
+++ b/src/backend/egl/surface.rs
@@ -81,18 +81,28 @@ impl EGLSurface {
     }
 
     /// Returns the buffer age of the underlying back buffer
-    pub fn buffer_age(&self) -> i32 {
+    pub fn buffer_age(&self) -> Option<i32> {
         let surface = self.surface.load(Ordering::SeqCst);
         let mut age = 0;
-        unsafe {
+        let ret = unsafe {
             ffi::egl::QuerySurface(
                 **self.display,
                 surface as *const _,
                 ffi::egl::BUFFER_AGE_EXT as i32,
                 &mut age as *mut _,
+            )
+        };
+        if ret == ffi::egl::FALSE {
+            slog::debug!(
+                self.logger,
+                "Failed to query buffer age value for surface {:?}: {}",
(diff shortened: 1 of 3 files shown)


src/backend/egl/surface.rs
```diff
@@ -81,18 +81,28 @@ impl EGLSurface {
...
         let mut age = 0;
-        unsafe {
+        if unsafe {
             ffi::egl::QuerySurface(
                 **self.display,
                 surface as *const _,
                 ffi::egl::BUFFER_AGE_EXT as i32,
                 &mut age as *mut _,
+            )
+        } == ffi::egl::FALSE
```
vberger, line 94: If think this check should be split, to improve readability:

```rust
let ret = unsafe { .... };
if ret == ffi::egl::FALSE {
    ...
}
```

src/backend/winit/mod.rs
```diff
@@ -302,12 +302,19 @@ impl WinitGraphicsBackend {
...
     }
 
-    /// Retrieve the buffer age of the current backbuffer of the window
-    pub fn buffer_age(&self) -> usize {
+    /// Retrieve the buffer age of the current backbuffer of the window.
+    ///
+    /// This will only return a meaningful value, if this `WinitGraphicsBackend`
+    /// is currently bound (by previously calling [`WinitGraphicsBackend::bind`]).
+    ///
+    /// Otherwise and on error this function returns `None`.
```
vberger, line 310: So there can be errors other than the fact that the backend is not currently bound? Would they be worth reporting?

Drakulix: They are usually flukes, no other errors are defined by the function, but is defined as possible to fail.

vberger (approved):

PR #482, opened 2022-01-25
gbm: Allow usage of other allocators for GbmBufferedSurface

Drakulix:
Right now the `GbmBufferedSurface` abstraction forces the use of a `GbmDevice` as an allocator.

This means users have to clone the underlying file descriptor to initialize multiple surfaces of the the same `GbmDevice`.
smithays internal code explicitly avoids this, because it may contribute to wrong or more difficult cleanup procedures, might prove to be a security hole, if the clone is not CLOEXEC, and even can mess up reference counting of buffer handles inside libgbm.

We already have a much better solution in smithay through implementations of the `Allocator`-trait for `Rc<RefCell<impl Allocator>>` and `Arc<Mutex<impl Allocator>>`. This PR adds a generic type to the `GbmBufferedSurface` to allow these variants to be used as well, as long as they produce a `gbm::BufferObject`.

This also updates anvil to make use of this feature to not encourage downstream to use the *wrong* solution.

### DIFF
--- a/src/backend/drm/surface/gbm.rs
+++ b/src/backend/drm/surface/gbm.rs
@@ -4,12 +4,12 @@ use std::sync::Arc;
 
 use drm::buffer::PlanarBuffer;
 use drm::control::{connector, crtc, framebuffer, plane, Device, Mode};
-use gbm::{BufferObject, Device as GbmDevice};
+use gbm::BufferObject;
 
 use crate::backend::allocator::{
     dmabuf::{AsDmabuf, Dmabuf},
     gbm::GbmConvertError,
-    Format, Fourcc, Modifier, Slot, Swapchain,
+    Allocator, Format, Fourcc, Modifier, Slot, Swapchain,
 };
 use crate::backend::drm::{device::DevPath, surface::DrmSurfaceInternal, DrmError, DrmSurface};
 use crate::backend::SwapBuffersError;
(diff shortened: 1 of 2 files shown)


src/backend/drm/surface/gbm.rs
```diff
@@ -334,9 +340,10 @@ impl<A: AsRawFd + 'static> Drop for FbHandle<A> {
     }
 }
 
-fn attach_framebuffer<A>(drm: &Arc<DrmSurface<A>>, bo: &BufferObject<()>) -> Result<FbHandle<A>, Error>
+fn attach_framebuffer<A, D>(drm: &Arc<DrmSurface<D>>, bo: &BufferObject<()>) -> Result<FbHandle<D>, Error<A>>
 where
-    A: AsRawFd + 'static,
+    A: std::error::Error + Send + Sync,
```
vberger, line 345: I don't think using `A` as a name for an error type is a good idea, especially when `A` represents the allocator in the rest of the file.

Also, this is pretty surprising that the caller can chose the error type when invoking this function, but I guess it's because the `Error::GbmError` is never constructed in it, right?

src/backend/drm/surface/gbm.rs
```diff
@@ -385,7 +392,7 @@ where
 
 /// Errors thrown by a [`GbmBufferedSurface`]
 #[derive(Debug, thiserror::Error)]
-pub enum Error {
+pub enum Error<A: std::error::Error + Send + Sync + 'static> {
```
vberger, line 395: Same here, using `A` as an associated type representing an error feels wrong. We can go multi-letters, like `AllocError` or something like that to still make it clear what this type represents, maybe?

src/backend/drm/surface/gbm.rs
Drakulix: Yes, also this function is private, so that is not really a concern for downstream.

But you are right about the generic name, I will change that.

src/backend/drm/surface/gbm.rs
```diff
@@ -406,14 +413,14 @@ pub enum Error {
...
+    GbmError(#[source] E),
     /// Error exporting as Dmabuf
     #[error("The allocated buffer could not be exported as a dmabuf: {0}")]
     AsDmabufError(#[from] GbmConvertError),
 }
 
-impl From<Error> for SwapBuffersError {
-    fn from(err: Error) -> SwapBuffersError {
+impl<A: std::error::Error + Send + Sync + 'static> From<Error<A>> for SwapBuffersError {
+    fn from(err: Error<A>) -> SwapBuffersError {
```
vberger, line 423: There's this `A` as well that should be changed for consistency I think.

PR #484, opened 2022-01-27
drm: Use EventMetadata to pass on frame timings

Drakulix:
For precise low-latency rendering it is important to now, when a frame event happened. We currently pass non of the timing related event fields from drm-rs to downstream. Lets make use of the calloop-Metadata to parse and change that.

### DIFF
--- a/src/backend/drm/device/mod.rs
+++ b/src/backend/drm/device/mod.rs
@@ -3,10 +3,11 @@ use std::cell::RefCell;
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::path::PathBuf;
 use std::sync::{atomic::AtomicBool, Arc};
+use std::time::{Instant, SystemTime};
 
 use calloop::{EventSource, Interest, Poll, PostAction, Readiness, Token, TokenFactory};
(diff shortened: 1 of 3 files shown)


src/backend/drm/device/mod.rs
```diff
@@ -336,7 +361,16 @@ where
                 for event in events {
                     if let Event::PageFlip(event) = event {
                         trace!(self.logger, "Got a page-flip event for crtc ({:?})", event.crtc);
-                        callback(DrmEvent::VBlank(event.crtc), &mut ());
+                        let metadata = EventMetadata {
+                            time: if self.has_monotonic_timestamps {
+                                Time::Monotonic(unsafe { std::mem::zeroed::<Instant>() } + event.duration)
```
PolyMeilex, line 366: Comment here would be helpful, adding something to zeroed Instant is non-obvious operation.

Drakulix: That is fair, added.

PolyMeilex (approved):
Looks good to me now.

PR #485, opened 2022-01-29
Remove socket initialization from udev.rs

chrisduerr:
Since `AnvilState::init` already initializes the socket, it shouldn't be
necessary to duplicate this in the `udev.rs` backend.


PolyMeilex (approved):
Indeed

vberger (approved):

PR #488, opened 2022-01-31
drm: Replace inner Option on DrmNode and use mem::forget

i509VCB:
Let's clean up the internals of DrmNode a bit.

### DIFF
--- a/src/backend/drm/node/mod.rs
+++ b/src/backend/drm/node/mod.rs
@@ -7,7 +7,7 @@ use libc::dev_t;
 
 use std::{
     fmt::{self, Display, Formatter},
-    fs, io,
+    fs, io, mem,
     os::unix::prelude::{AsRawFd, IntoRawFd, RawFd},
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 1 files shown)


src/backend/drm/node/mod.rs
```diff
@@ -120,22 +115,22 @@ impl Display for DrmNode {
 }
 
 impl IntoRawFd for DrmNode {
-    fn into_raw_fd(mut self) -> RawFd {
-        self.fd.take().unwrap()
+    fn into_raw_fd(self) -> RawFd {
+        let fd = self.fd;
+        mem::forget(self);
```
Drakulix, line 120: But doesn't that leak memory?

i509VCB: Nope:
> Takes ownership and “forgets” about the value without running its destructor.
>
> Any resources the value manages, such as heap memory or a file handle, will linger forever in an unreachable state. > However, it does not guarantee that pointers to this memory will remain valid.

i509VCB: Plus the semantics defined by IntoRawFd indicate the caller takes ownership of the open fd

Drakulix: Yes for the Fd, that is correct, but we leak the memory of "self".

Drakulix: Misunderstood, stack variables are of course still deallocated, just not "dropped"

Drakulix (approved):

PR #489, opened 2022-02-01
seat: Always update pointer handles current location

Drakulix:
While a focus is active `PointerHandle::current_location` will always return the starting position of the grab, which forces compositors to track the pointer location themselves.

This changes this behavior to always update the current position correctly.

### DIFF
--- a/src/wayland/seat/pointer.rs
+++ b/src/wayland/seat/pointer.rs
@@ -213,6 +213,7 @@ impl PointerHandle {
     ) {
         let mut inner = self.inner.borrow_mut();
         inner.pending_focus = focus.clone();
+        inner.location = location;
         inner.with_grab(move |mut handle, grab| {
             grab.motion(&mut handle, location, focus, serial, time);
         });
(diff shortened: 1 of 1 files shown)


vberger:
Hmm, ok, so, re-reading that code I remember the original intent: the location is only updated in the *inner* method because in theory grabs may need to alter the location of the pointer compared to what the input backend reported. Main example of this would be the `pointer_constraints` protocol.

So I think here, the problem is rather that some grab implementation does not invoke the `PointerInnerHandle::motion()` method... ?

Drakulix:
> So I think here, the problem is rather that some grab implementation does not invoke the `PointerInnerHandle::motion()` method... ?

Almost no grabs actually do that, as far as I am aware. Neither the internal `DnDGrab` not the `ResizeSurfaceGrab` or `MoveSurfaceGrab` from anvil. Only the `PopupGrab`.

So would that be a better solution? Couldn't something like pointer_constraints reset the location again? At least we should document this problem, if we want to fix the grabs instead.

vberger:
Oh wow, speak about an oversight. Yeah, the grabs need to be fixed, the API was really designed with this in mind. The grabs not invoking `PointerInnerHandle::motion()` means the clients don't receive location updates for the pointer as long as the grab is active, which will certainly cause issues.

PR #491, opened 2022-02-09
desktop: Take xdg geometry location into acount

PolyMeilex:
What is the definition of `geometry` in desktop abstraction terms? My understanding was that it should work like xdg geometry, but looking at the code it erases location of xdg toplevels in relation to root position.
So the geometry does not align with window contents:
image

So using `+=` seems to fix the problem.


vberger (approved):

PolyMeilex:
Closing in favor of proper API rework, that I will PR in a bit

PR #492, opened 2022-02-09
Refactor anvil udev code

chrisduerr:
This patch was written while implementing the udev backend for my
compositor to make things a little easier to read and understand. I
think the improvements would also help others to grasp grasp the code a
bit faster.

I've tried to keep all the changes minimal and objective, so hopefully
there shouldn't be any major style changes but just improvements. No
functionality has changed.

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -1,7 +1,7 @@
 use std::{
+    borrow::Cow,
     cell::RefCell,
     collections::hash_map::{Entry, HashMap},
-    io::Error as IoError,
     os::unix::io::{AsRawFd, RawFd},
     path::PathBuf,
     rc::Rc,
@@ -73,7 +73,7 @@ use crate::{
     state::{AnvilState, Backend},
 };
 
-#[derive(Clone)]
+#[derive(Copy, Clone)]
 pub struct SessionFd(RawFd);
 impl AsRawFd for SessionFd {
     fn as_raw_fd(&self) -> RawFd {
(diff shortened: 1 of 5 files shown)


anvil/src/udev.rs
```diff
@@ -247,13 +246,6 @@ pub fn run_udev(log: Logger) {
             state.output_map.borrow_mut().refresh();
         }
     }
-
-    // Cleanup stuff
-    state.window_map.borrow_mut().clear();
-
-    event_loop.handle().remove(session_event_source);
-    event_loop.handle().remove(libinput_event_source);
-    event_loop.handle().remove(udev_event_source);
```
chrisduerr, line 256: As far as I understand, removal of these should happen automatically through RAII and order doesn't matter.

Of course this can be done explicitly, but doing so is likely confusing when it is unnecessary. I haven't noticed any issues, should this code actually be **necessary** please let me know.

anvil/src/udev.rs
```diff
@@ -304,95 +296,97 @@ fn scan_connectors(
         let encoder_infos = connector_info
             .encoders()
             .iter()
-            .filter_map(|e| *e)
-            .flat_map(|encoder_handle| device.get_encoder(encoder_handle))
+            .flatten()
+            .flat_map(|encoder_handle| device.get_encoder(*encoder_handle))
             .collect::<Vec<EncoderInfo>>();
-        'outer: for encoder_info in encoder_infos {
```
chrisduerr, line 310: This function looks like a massive delta, but I haven't changed much. It's mostly just less indentation which makes it possible to get rid of ugly things like loop annotations.

Drakulix (approved):
Even after carefully checking everything, I cannot find any edge-case that should be influenced by these changes.

Overall the code quality has definitely improved and I am suggesting we merge this.

PolyMeilex (approved):
Looks good to me as well :ok_hand:

PR #495, opened 2022-02-14
Add support for GPUs without instanced rendering

chrisduerr:
The patch 31b3088 introduced instanced rendering for efficiently
updating only the damaged area of the output. However due to this new
feature, support for older GPUs without instanced rendering was dropped.

This patch adds an alternative implementation which is chosen at runtime
when the target GPU does not support instanced rendering. It uses a
combination of uploading duplicate information and multiple draw calls
to somewhat efficiently mimic the effects of instanced rendering.

I've verified that with this patch it is now possible to run Anvil on my
PinePhone.

### DIFF
--- a/src/backend/renderer/gles2/mod.rs
+++ b/src/backend/renderer/gles2/mod.rs
@@ -1171,7 +1171,7 @@ const fn triangle_verts() -> [ffi::types::GLfloat; 12 * MAX_RECTS_PER_DRAW] {
     let mut i = 0;
     loop {
         // Top Left.
-        verts[i * 12 + 0] = 0.0;
+        verts[i * 12] = 0.0;
         verts[i * 12 + 1] = 0.0;
 
         // Bottom left.
--- a/src/backend/renderer/gles2/mod.rs
+++ b/src/backend/renderer/gles2/mod.rs
@@ -1,5 +1,6 @@
 //! Implementation of the rendering traits using OpenGL ES 2
 
+use std::borrow::Cow;
 use std::convert::TryFrom;
 use std::ffi::CStr;
 use std::fmt;
@@ -180,6 +181,7 @@ pub struct Gles2Renderer {
     destruction_callback_sender: Sender<CleanupResource>,
     min_filter: TextureFilter,
     max_filter: TextureFilter,
+    supports_instancing: bool,
     logger_ptr: Option<*mut ::slog::Logger>,
     logger: ::slog::Logger,
     _not_send: *mut (),
@@ -196,6 +198,7 @@ pub struct Gles2Frame {
     size: Size<i32, Physical>,
     min_filter: TextureFilter,
     max_filter: TextureFilter,
+    supports_instancing: bool,
 }
 
 impl fmt::Debug for Gles2Frame {
@@ -441,7 +444,7 @@ impl Gles2Renderer {
 
         context.make_current()?;
 
-        let (gl, exts, logger_ptr) = {
+        let (gl, exts, logger_ptr, supports_instancing) = {
             let gl = ffi::Gles2::load_with(|s| crate::backend::egl::get_proc_address(s) as *const _);
             let ext_ptr = gl.GetString(ffi::EXTENSIONS) as *const c_char;
             if ext_ptr.is_null() {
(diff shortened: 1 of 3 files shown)


src/backend/renderer/gles2/mod.rs
```diff
@@ -487,16 +490,10 @@ impl Gles2Renderer {
...
-            {
-                return Err(Gles2Error::GLExtensionNotSupported(&[
-                    "GL_EXT_instanced_arrays",
-                    "GL_EXT_draw_instanced",
-                ]));
-            }
+            // Check if GPU supports instanced rendering.
+            let supports_instancing = gl_version >= version::GLES_3_0
+                || (exts.iter().any(|ext| ext == "GL_EXT_instanced_arrays")
+                    && exts.iter().any(|ext| ext == "GL_EXT_draw_instanced"));
```
chrisduerr, line 496: While @Drakulix has recommended on IRC to add an environment variable to override this, I've chosen against doing so because I don't see much practical application for it.

When specifically testing instanced/uninstanced one is likely already compiling from source anyway and in that scenario just setting it to a fixed value is trivial. Even while writing this patch I didn't see any need to make this logic more complicated.

src/desktop/space/mod.rs
```diff
@@ -542,6 +542,37 @@ impl Space {
...
+                    //     Rectangle::from_loc_and_size((300, 450), (100, 100)),
+                    //     Rectangle::from_loc_and_size((450, 450), (100, 100)),
+                    //     Rectangle::from_loc_and_size((600, 450), (100, 100)),
+
+                    //     Rectangle::from_loc_and_size((0, 600), (100, 100)),
+                    //     Rectangle::from_loc_and_size((150, 600), (100, 100)),
+                    //     Rectangle::from_loc_and_size((300, 600), (100, 100)),
+                    //     Rectangle::from_loc_and_size((450, 600), (100, 100)),
+                    //     Rectangle::from_loc_and_size((600, 600), (100, 100)),
+                    // ],
```
chrisduerr, line 575: Not actually part of my patch, but this is very useful for testing because it forces the drawing of 25 rectangles which exceeds the limit of 10 to verify that logic works too.

src/backend/renderer/gles2/mod.rs
```diff
@@ -487,16 +490,10 @@ impl Gles2Renderer {
...
-                && !(exts.iter().any(|ext| ext == "GL_EXT_instanced_arrays")
-                    && exts.iter().any(|ext| ext == "GL_EXT_draw_instanced"))
-            {
-                return Err(Gles2Error::GLExtensionNotSupported(&[
-                    "GL_EXT_instanced_arrays",
-                    "GL_EXT_draw_instanced",
-                ]));
-            }
+            // Check if GPU supports instanced rendering.
+            let supports_instancing = gl_version >= version::GLES_3_0
```
i509VCB, line 494: Could it be useful to have an environment variable for all us with high end cards to test this?

chrisduerr: See https://github.com/Smithay/smithay/pull/495#discussion_r805519150.

I wrote this patch myself on a computer that supports instanced rendering. I just don't see why it would be any easier to set an environment variable compared to just changing the code when you're compiling from source anyway.

src/backend/renderer/gles2/mod.rs
Drakulix: The practical application is:
1. downstream depending on a released version of smithay
2. encounters a bug
3. wants to verify if it happens with both renderer paths

Sure you could also add a "patch"-statement in your Cargo.toml for this case.

But I agree, that both code paths are similar enough (especially given they both use the same shader), that this kind of scenario should be very rare. So this is likely fine the way it is.

Drakulix (approved):
Looks good!

Just a minor nitpick regarding a left TODO comment, but otherwise this is ready for merge.

src/backend/renderer/gles2/mod.rs
chrisduerr: Yeah that's basically the only thing I could come up with too, but if you're on older hardware opting into instanced simply isn't an option. And if you're on newer hardware then it really wouldn't be of much help when the garbage path of non-instanced rendering works, especially because the "backup" plan is much more likely to be broken than what basically everyone uses.

PR #497, opened 2022-02-18
include feature hints in docs

wash2:
- includes feature hints in docs

feature_hint2
feature_hints


PolyMeilex (approved):
This would be a nice addition in my opinion. :+1:

i509VCB:
I guess I will copy this over to sctk since we have some features but not nearly as many

PR #498, opened 2022-02-18
Replace redundant Result/Error typealiases

i509VCB:
Specifically, aliases to io::Error, io::Result and nix::Result


PolyMeilex (approved):
LGTM

PR #499, opened 2022-02-22
Add client touch capability support

chrisduerr:
While Smithay already supports libinput's touch events, it did not
support forwarding touch events to clients yet.

This patch implements a new `TouchHandle`, which can be acquired from
`Seat::get_touch`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 - `MouseButton` is now non-exhaustive.
 - Remove `Other` and add `Forward` and `Back` variants to `MouseButton`. Use the new `PointerButtonEvent::button_code` in place of `Other`.
 - `GrabStartData` has been renamed to `PointerGrabStartData`
+- Added `TouchHandle` for Wayland client touch support (see `Seat::get_touch`)
+- **[Breaking]**: The `slot` method on touch events no longer returns an `Option` and multi-touch capability is thus opaque to the compositor
 
 #### Backends
 
@@ -61,10 +63,10 @@
 #### Backends
 
 - New `x11` backend to run the compositor as an X11 client. Enabled through the `backend_x11` feature.
-- `x11rb` event source integration used in anvil's XWayland implementation is now part of smithay at `utils::x11rb`. Enabled through the `x11rb_event_source` feature. 
+- `x11rb` event source integration used in anvil's XWayland implementation is now part of smithay at `utils::x11rb`. Enabled through the `x11rb_event_source` feature.
 - `KeyState`, `MouseButton`, `ButtonState` and `Axis` in `backend::input` now derive `Hash`.
 - New `DrmNode` type in drm backend. This is primarily for use a backend which needs to run as client inside another session.
-- The button code for a `PointerButtonEvent` may now be obtained using `PointerButtonEvent::button_code`. 
+- The button code for a `PointerButtonEvent` may now be obtained using `PointerButtonEvent::button_code`.
 - `Renderer` now allows texture filtering methods to be set.
 - `backend::renderer` has a new `utils`-module that can take care of client buffer management for you.
 - `EGLSurface::buffer_age` can be used to query the surface buffer age.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,7 @@
 - `MouseButton` is now non-exhaustive.
 - Remove `Other` and add `Forward` and `Back` variants to `MouseButton`. Use the new `PointerButtonEvent::button_code` in place of `Other`.
 - `GrabStartData` has been renamed to `PointerGrabStartData`
-- Added `TouchHandle` for Wayland client touch support (see `Seat::get_touch`)
-- **[Breaking]**: The `slot` method on touch events no longer returns an `Option` and multi-touch capability is thus opaque to the compositor
+- The `slot` method on touch events no longer returns an `Option` and multi-touch capability is thus opaque to the compositor
 
 #### Backends
 
@@ -59,6 +58,7 @@
 - Added a `KeyboardGrab` similar to the existing `PointerGrab`
 - `wayland::output::Output` now has a `current_scale` method to quickly retrieve its set scale.
 - `wayland::shell::wlr_layer::KeyboardInteractivity` now implements `PartialEq` and `Eq`.
+- Added `TouchHandle` for Wayland client touch support (see `Seat::get_touch`)
 
 #### Backends
 
--- a/src/backend/input/mod.rs
+++ b/src/backend/input/mod.rs
@@ -347,9 +347,9 @@ impl From<Option<u32>> for TouchSlot {
     }
 }
 
-impl Into<i32> for TouchSlot {
-    fn into(self) -> i32 {
-        self.id.map(|id| id as i32).unwrap_or(-1)
+impl From<TouchSlot> for i32 {
+    fn from(slot: TouchSlot) -> i32 {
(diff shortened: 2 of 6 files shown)


src/backend/input/mod.rs
```diff
@@ -336,22 +336,27 @@ impl<B: InputBackend> PointerMotionAbsoluteEvent<B> for UnusedEvent {
 /// Touch events are grouped by slots, usually to identify different
 /// fingers on a multi-touch enabled input device. Events should only
 /// be interpreted in the context of other events on the same slot.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct TouchSlot {
-    id: u64,
+    id: Option<u32>,
```
chrisduerr, line 341: I've decided to change the `TouchSlot` here because it already was creating some friction inside my compositor when handling touch events and just created more problems when it came to converting the `Option<TouchSlot>` back to a `i32` to send it to the client.

So since I'd assume nobody is using touch yet (since it doesn't work for clients), changing it now will prevent headaches in the future.

src/wayland/seat/mod.rs
```diff
@@ -313,6 +317,59 @@ impl Seat {
...
+    /// #
+    /// # use smithay::wayland::seat::Seat;
+    /// #
+    /// # let mut display = wayland_server::Display::new();
+    /// # let (mut seat, seat_global) = Seat::new(
+    /// #     &mut display,
+    /// #     "seat-0".into(),
+    /// #     None
+    /// # );
+    /// let touch_handle = seat.add_touch();
```
i509VCB, line 342: This example doesn't really explain anything more than the function docs itself.

If the seat creation are made visible it would be more jusitifed to have.

chrisduerr: Correct, but I'd say the same applies for the existing code examples for keyboard/pointer?

> If the seat creation are made visible it would be more jusitifed to have.

What do you mean by this? Removing the `#` before the seat stuff?

chrisduerr:
I've uploaded some videos for the people unable to test with touchscreens (these will be gone in 24h, unfortunately the videos are too big for GH):

Touch Input

Multi-touch

src/wayland/seat/touch.rs
```diff
@@ -0,0 +1,184 @@
...
+
+        // Select all WlTouch instances associated to the active WlSurface.
+        for handle in &self.known_handles {
+            if handle.as_ref().same_client_as(surface.as_ref()) {
+                focus.handles.push(handle.clone());
+            }
+        }
+
+        let (x, y) = (location - focus.surface_offset).into();
+        let serial = SERIAL_COUNTER.next_serial().into();
```
PolyMeilex, line 117: The serial can be used by the client to request some action, so the compositor needs to be aware of them. Otherwise it can't implement it's own serial checking rules. 
So we should not hide any serials from the user.

@vberger Correct me if I'm wrong

src/wayland/seat/touch.rs
```diff
@@ -0,0 +1,184 @@
...
+        self.with_focused_handles(slot, |handle| {
+            handle.down(serial, time, surface, slot.into(), x, y)
+        });
+    }
+
+    fn up(&self, time: u32, slot: TouchSlot) {
+        let serial = SERIAL_COUNTER.next_serial().into();
```
PolyMeilex, line 124: Same here.

src/wayland/seat/touch.rs
```diff
@@ -0,0 +1,184 @@
...
+            Some(slot) => slot,
+            None => return,
+        };
+
+        let (x, y) = (location - focus.surface_offset).into();
+        self.with_focused_handles(slot, |handle| handle.motion(time, slot.into(), x, y));
+    }
+
+    fn shape(&self, slot: TouchSlot, major: f64, minor: f64) {
+        self.with_focused_handles(slot, |handle| handle.shape(slot.into(), major, minor));
```
PolyMeilex, line 139: This is missing a version check, shape is available since version 6

src/wayland/seat/touch.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    fn orientation(&self, slot: TouchSlot, orientation: f64) {
+        self.with_focused_handles(slot, |handle| handle.orientation(slot.into(), orientation));
```
PolyMeilex, line 143: Version check as well, orientation is available since version 6.

PolyMeilex (changes requested):
I have no way to test it, but overall this looks really good 👍
Just a few minor nitpicks.

src/wayland/seat/touch.rs
chrisduerr: Just for my personal information: What kind of action can clients request here?

src/wayland/seat/touch.rs
chrisduerr: Do you know if there's some minimum level Smithay assumes to be present? Because I noticed release is only part of version 3 but I don't think Smithay ever checks version for that?

Drakulix: > Do you know if there's some minimum level Smithay assumes to be present? Because I noticed release is only part of version 3 but I don't think Smithay ever checks version for that?

That would be a bug then. Compositors are required to support all previous versions for stable protocols.

src/wayland/seat/touch.rs
PolyMeilex: I'm not familiar with touch based clients, so I don't know what they do often, but it can basically be anything, lets say xdg-shell move/resize, or show_window_menu for example.

src/wayland/seat/touch.rs
chrisduerr: I think Smithay gets around this by never actually handling release but just using automatic destructors using `assign_destructor`. So I think the existing implementations are okay? I just noticed that there weren't any version checks for release.

PolyMeilex (approved):
Ok, looks good to me now.

PolyMeilex:
Actually, could we get a short changelog entry for this? Quick note about the addition and maybe mention the braking change of touch slot (although I doubt anyone was using it)

chrisduerr:
> Actually, could we get a short changelog entry for this? Quick note about the addition and maybe mention the braking change of touch slot (although I doubt anyone was using it)

Yep, of course. Was just waiting for feedback on the general design first.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 - `MouseButton` is now non-exhaustive.
 - Remove `Other` and add `Forward` and `Back` variants to `MouseButton`. Use the new `PointerButtonEvent::button_code` in place of `Other`.
 - `GrabStartData` has been renamed to `PointerGrabStartData`
+- Added `TouchHandle` for Wayland client touch support (see `Seat::get_touch`)
```
PolyMeilex, line 17: Technically, this should be under [Additions > Clients & Protocols](https://github.com/Smithay/smithay/blob/032c0de390d17219f53065bb69c99d5ef1762974/CHANGELOG.md#clients--protocols-1) as this is not "Braking Change" per se

Drakulix (approved):
I agree with @PolyMeilex's remarks, but other than that, this is good for merge.

Drakulix:
@chrisduerr Could you rebase on master? I pushed some fixes to make CI on master green again and would like to verify CI works on this branch as well before merging it.

chrisduerr:
Done.

PolyMeilex:
Thank you!

PR #500, opened 2022-02-27
Update render matrix for rotated outputs

chrisduerr:
When the output is rotated by 90°/270° using a transform in the
`Gles2Renderer::render` function, the resulting matrix would cause the
buffer to be stretched along one and compressed along the other axis.

Since width and height are swapped when rotating by 90 degrees
(counter-)clockwise, the resulting output matrix needs to adjust for
this difference.

### DIFF
--- a/src/backend/renderer/gles2/mod.rs
+++ b/src/backend/renderer/gles2/mod.rs
@@ -1732,12 +1732,23 @@ impl Renderer for Gles2Renderer {
         // We account for OpenGLs coordinate system here
         let flip180 = Matrix3::new(1.0, 0.0, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0, 1.0);
 
+        // Matrix handling the width/height swap when output is rotated by 90°/270°.
+        let rotation_stretch = match transform {
+            Transform::_90 | Transform::_270 | Transform::Flipped90 | Transform::Flipped270 => {
+                let factor = size.w as f32 / size.h as f32;
+                Matrix3::new(factor, 0.0, 0.0, 0.0, 1. / factor, 0.0, 0.0, 0.0, 1.0)
+            }
+            Transform::Normal | Transform::_180 | Transform::Flipped | Transform::Flipped180 => {
+                Matrix3::identity()
+            }
+        };
+
         let mut frame = Gles2Frame {
             gl: self.gl.clone(),
             tex_programs: self.tex_programs.clone(),
             solid_program: self.solid_program.clone(),
             // output transformation passed in by the user
-            current_projection: flip180 * transform.matrix() * renderer,
+            current_projection: flip180 * transform.matrix() * renderer * rotation_stretch,
             transform,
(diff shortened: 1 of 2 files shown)


chrisduerr:
I'm not exactly sure this is the correct solution, all I know is that it works for me and is in the right location suggested by @Drakulix. If a more complicated approach is required, please let me know.

Drakulix:
> I'm not exactly sure this is the correct solution, all I know is that it works for me and is in the right location suggested by @Drakulix. If a more complicated approach is required, please let me know.

The solution definitely works, but I creates some questions around the api.

Is the `size` pre- or post-transformation? If it is pre-transformation most stuff should be fine, but I wonder if we need the additional matrix, or if we just could swap the size around (we have a function `transform_size` for that) before line 1098 to fix this instead. If it is post-transformation, the `glViewport`-call is simply wrong.

With which assumption did you test this? We should update the docs of the `Renderer`-trait accordingly. Additionally we might want to rename the argument to something more descriptive like `src_size` or `dst_size`?

chrisduerr:
> Is the size pre- or post-transformation?

The size is the *real* output size, so untransformed.

> If it is pre-transformation most stuff should be fine, but I wonder if we need the additional matrix, or if we just could swap the size around (we have a function transform_size for that) before line 1098 to fix this instead. 

Good question, I've only tried it downstream and that obviously didn't work. It seems like doing it in line 1098 works just as well as the matrix stuff, would you prefer it that way? I just thought that the matrix would be the more appropriate place since that's what is going to be transformed again later when the client turns things around again for example. But I do agree that a `swap(w, h)` is far easier to understand than a matrix transform.

> With which assumption did you test this? We should update the docs of the Renderer-trait accordingly.

Which assumption regarding the pre- or post-transformation? I can certainly document that.

> Additionally we might want to rename the argument to something more descriptive like src_size or dst_size?

Maybe just output_size?

Drakulix:
> Good question, I've only tried it downstream and that obviously didn't work. It seems like doing it in line 1098 works just as well as the matrix stuff, would you prefer it that way? I just thought that the matrix would be the more appropriate place since that's what is going to be transformed again later when the client turns things around again for example. But I do agree that a `swap(w, h)` is far easier to understand than a matrix transform.

Yes, I think I would prefer the readability of a swap here.

> Which assumption regarding the pre- or post-transformation? I can certainly document that.

Yes. please.

> Maybe just output_size?

Sounds good to me! :+1:

Drakulix:
Sorry I left this open for so long. Can you verify that after #526 / #527 this still works as intended and fix the conflict? Then I would be glad to merge this.

chrisduerr:
Cannot confirm this still works atm because smithay's latest version doesn't work at all anymore for me. So I'll have to look into that first tomorrow.

chrisduerr:
After updating my compositor to 0.30.0, I can confirm this still works.

Drakulix (approved):

PR #501, opened 2022-02-28
Implement `From<Transform>` for `wl_output::Transform`

chrisduerr:
I'd assume most Wayland consumers probably mostly want to deal with
Smithay's `Transform`, so it's nice to keep that type consistent in
consumer crates and then have it converted to the protocol's types on
demand.

I've refrained from making changes to the `change_current_state` method
in this patch to keep the patch as uncontroversial as possible, but
changing the parameters to `impl Into<Option<T>>` for at least the
transform of that method similar to how the Size/Point types are
implemented would probably help downstream.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -1,6 +1,8 @@
 use std::fmt;
 use std::ops::{Add, AddAssign, Sub, SubAssign};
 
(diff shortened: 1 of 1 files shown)


Drakulix:
CI is picking up on some unrelated clippy notes.
Given the simplicity of this patch, I am going to merge it anyway.

PR #502, opened 2022-02-28
Move shared input event functions to common traits

chrisduerr:
The current state of input events is somewhat of a tragedy for
consumers. You write your implementation in the match branch for the
individual event, then notice there are a lot of similarities, but after
you try to move it into a function you realize that all the input events
are completely separate events that just have methods named the same
way.

To allow downstream to actually make use of these traits as proper
shared functionality and write their own helper abstractions, new traits
have been introduced that are implemented for all events.

These two new traits, `PositionEvent` and `TouchEvent`, allow accessing
the slot and location of any generic input event which provides this
functionality.


Drakulix:
Overall this is probably a rather good idea, but I am not 100% sold on the `PositionEvent` yet.

I think at minimum it needs some docs, that describe the absolute nature of input-coordinates provided.
Because otherwise it is a bit weird, that the `PositionAbsolute`-events implement it, but the relative `Position`-events do not.

Maybe this should instead be named differently, though that might lead to bikeshedding. Another option would be to make this just a property of all TouchEvents, although a position would likely have no meaning for TouchFramEvents, right?

chrisduerr:
> but the relative Position-events do not.

Ah fair enough, I just didn't know about them because I never used them. I can rename it to `AbsolutePositionEvent` too maybe?

> although a position would likely have no meaning for TouchFramEvents, right?

Correct, there are multiple touch events that do not have any position associated. That includes TouchUp, Frame and Cancel I believe.

Drakulix:
> > but the relative Position-events do not.
> 
> Ah fair enough, I just didn't know about them because I never used them. I can rename it to `AbsolutePositionEvent` too maybe?

But that is so long. I think "MotionAbsolute" is a rather bad name to begin with.
Maybe we should just "Motion" for everything, that deals with deltas, like the `PointerMotionEvent`,
and then use "Position" for everything with absolute values, like the `PointerMotionAbsoluteEvent` and the `TouchMotion` event.

So those would become `PointerPositionEvent` and `TouchPositionEvent`. And then the trait could keep being named `PositionEvent`.

Does that sound reasonable?

chrisduerr:
> Maybe we should just use "Motion" for everything, that deals with deltas, like the PointerMotionEvent,
and then use "Position" for everything with absolute values, like the PointerMotionAbsoluteEvent and the TouchMotion event.

That seems like a bigger change though, affecting all events and traits?

I don't care *that* much about the length of a trait name tbh, since that's rarely in a location where brevity is particularly important. That said, I don't think the naming is particularly great but it's consistent with the existing events and traits to some degree.

Drakulix:
Yeah, we can go with "AbsolutePositionEvent" for now to keep the consistency and then bike shed about better names later. Agreed

chrisduerr:
I think this should be good now @Drakulix?

Drakulix (approved):
Indeed! LGTM

PR #507, opened 2022-03-05
0.30: Remove no longer needed alive checks

PolyMeilex:
Those checks existed because old version of compositor module required alive checks, but that's no longer the case.

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -61,7 +61,6 @@
 //! the subhandler you provided, or via methods on the [`ShellState`]
 //! that you are given (in an `Arc<Mutex<_>>`) as return value of the `init` function.
 
-use crate::utils::DeadResource;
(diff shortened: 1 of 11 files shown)


src/wayland/shell/xdg/mod.rs
```diff
@@ -919,11 +911,7 @@ impl ToplevelSurface {
     /// Retrieve the shell client owning this toplevel surface
     ///
     /// Returns `None` if the surface does actually no longer exist.
-    pub fn client(&self, dh: &mut DisplayHandle<'_>) -> Option<ShellClient> {
-        if !self.alive(dh) {
-            return None;
-        }
-
+    pub fn client(&self) -> Option<ShellClient> {
```
i509VCB, line 914: This will always return `Some`

PolyMeilex: Indeed, I missed that one.

PR #508, opened 2022-03-05
0.30: introduce delegate_ macros

i509VCB:
This should make using a delegate type much easier.


PolyMeilex (approved):
This looks great!
CI is complaining about lack of docs comments, but I guess we can worry about that later.

vberger (approved):

PR #509, opened 2022-03-05
0.30: Introduce IsAlive trait

PolyMeilex:
This introduces new IsAlive trait that we can use to implement `is_alive` check on any wayland object we want.  For now I implemented it for most common objects that users may want to alive check.

### DIFF
--- a/src/utils/alive_tracker.rs
+++ b/src/utils/alive_tracker.rs
@@ -17,11 +17,6 @@ impl Default for AliveTracker {
 }
 
 impl AliveTracker {
-    /// Create new alive tracker
-    pub fn new() -> Self {
-        Self::default()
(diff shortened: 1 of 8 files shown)


i509VCB:
@PolyMeilex you probably need to rebase this

PolyMeilex:
I will just draft it for now, as after a discussion with Drakulic we think that `is alive()` pattern should be discouraged, so it's better to add it as late as possible, or never if we don't end up needing it. So I will just add surface_destroyed, toplevel_destoryed callbacks, and we'll see how it goes from there.

PolyMeilex:
Looks like everything is rebased correctly :+1: 
So it's good to go as far as I'm concerned

PR #511, opened 2022-03-05
shm 0.30: fix some safety issues and cleanup

i509VCB:


src/wayland/shm/pool.rs
```diff
@@ -188,7 +193,6 @@ unsafe fn map(fd: RawFd, size: usize) -> Result<*mut u8, ()> {
     ret.map(|p| p as *mut u8).map_err(|_| ())
 }
 
-// mman::munmap should really be unsafe... why isn't it?
```
i509VCB, line 191: context here is that these functions are now unsafe.

vberger (approved):
Maybe we should merge this once the `DataInit` stuff is fixed in wayland-rs (I have a good idea what needs to be done) before merging this, so that we don't forget the related `FIXME` ?

i509VCB:
@vberger I will write a seperate pull request to update wayland-rs to the new alpha so the fixme is resolved. I have removed the comment for now under the assumption the update is done.

PR #513, opened 2022-03-08
0.30: update to alpha6

i509VCB:
WlEglSurface is no longer Sync, so EGLNativeDisplay is only implemented for `Mutex<WlEglSurface>`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,11 +43,11 @@ slog-stdlog = { version = "4", optional = true }
 tempfile = { version = "3.0", optional = true }
 thiserror = "1.0.25"
 udev = { version = "0.6", optional = true }
-wayland-egl = { version = "0.30.0-alpha5", optional = true }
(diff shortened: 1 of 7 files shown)


PolyMeilex (approved):
LGTM

Drakulix:
Actually this is not the correct fix.

You cannot bind the same EGLSurface across multiple threads, so EGLNativeSurface should just not require `Sync`, only `Send`.

Same goes for EGLContext, which cannot be bound from multiple threads and is thus also not Sync.

Only the EGLDisplay and EGLImage/Buffer/BufferReader are truely Sync.

So we should not need the mutex.

i509VCB:
Hmm okay, so this needs some larger changes to be correct.

i509VCB:
Also added suppression on the delegate macro docs for now since we need to know if the code compiles or not.

Drakulix (approved):
EGL side now also looks good to me

PR #514, opened 2022-03-08
0.30: wl_output v4

i509VCB:


src/wayland/output/mod.rs
```diff
@@ -213,6 +214,7 @@ impl Output {
             inner: Arc::new((
                 Mutex::new(Inner {
                     name,
+                    description: format!("{} - {} - {}", physical.make, physical.model, name),
```
vberger, line 217: Looks like that doesn't compile, `name` is moved on the previous line.

PolyMeilex (approved):

PR #516, opened 2022-03-08
Use the window geometry to find the window under the pointer location

cmeissl:
Currently it is not possible to focus a window underneath another window by clicking through the dropshadow.
Instead of using the bounding box of the window, which includes the drop shadow, use the geometry to
find the window under the current pointer location.


vberger (approved):
I think the correct thing would be to use the input regions of the subsurface tree, but that's still an improvement.

PR #519, opened 2022-03-10
0.30: port small parts of desktop

i509VCB:
Waiting for #518 to be merged first.

I have chosen to port parts that would not conflict with #503 is completed since this pull request changes a lot of the element api. This means future pull requests will come to finish the rest and is also why the module is not enabled.


i509VCB:
I'll do this all at a later time.

PR #520, opened 2022-03-11
Use `Serial` type for touch events instead of u32

chrisduerr:


chrisduerr:
I just noticed this when updating my local smithay version, probably best to fix immediately rather than having people get confused later why one API is different than the others.

i509VCB (approved):

vberger (approved):

PR #524, opened 2022-03-15
x11: use EGLSurface

Drakulix:


src/backend/x11/mod.rs
```diff
@@ -146,72 +126,124 @@ pub enum X11Event {
...
+                ));
+            }
+            xlib_sys::xlib_xcb::XSetEventQueueOwner(
+                display,
+                xlib_sys::xlib_xcb::XEventQueueOwner::XCBOwnsEventQueue,
+            );
+            let screen = xlib_sys::xlib::XDefaultScreen(display);
+            let xcb_raw = xlib_sys::xlib_xcb::XGetXCBConnection(display);
+            (
+                XCBConnection::from_raw_xcb_connection(xcb_raw, true)?,
```
i509VCB, line 172: This should be false https://docs.rs/x11rb/latest/x11rb/xcb_ffi/struct.XCBConnection.html#method.from_raw_xcb_connection

Reason being that Xlib effectively owns the xcb connection, therefore to close this we need to use `XDestroyDisplay`

src/backend/x11/mod.rs
```diff
@@ -146,72 +126,124 @@ pub enum X11Event {
...
+#[derive(Debug)]
+pub struct X11Backend {
+    connection: Arc<XCBConnection>,
+    source: X11Source<XCBConnection>,
+    inner: Arc<Mutex<X11Inner>>,
+    log: Logger,
+}
+
+unsafe impl Send for X11Connection {}
+unsafe impl Sync for X11Connection {}
```
i509VCB, line 201: Safety: Send + Sync because Xlib is initialized with threads.

src/backend/egl/ffi.rs
```diff
@@ -130,6 +130,7 @@ pub mod egl {
     include!(concat!(env!("OUT_DIR"), "/egl_bindings.rs"));
 
     pub const RESOURCE_BUSY_EXT: u32 = 0x3353;
+    pub const PLATFORM_XCB_EXT: u32 = 0x31DC;
```
i509VCB, line 133: Manually adding these is annoying, can't just update upstream since the egl.xml isn't parsed correctly.

Might want to look into phosphorus at some point and finish that egl pull request for that?

src/backend/x11/mod.rs
```diff
@@ -146,72 +126,124 @@ pub enum X11Event {
...
+                return Err(X11Error::ConnectionFailed(
+                    x11rb::errors::ConnectError::UnknownError,
+                ));
+            }
+            xlib_sys::xlib_xcb::XSetEventQueueOwner(
+                display,
+                xlib_sys::xlib_xcb::XEventQueueOwner::XCBOwnsEventQueue,
+            );
+            let screen = xlib_sys::xlib::XDefaultScreen(display);
+            let xcb_raw = xlib_sys::xlib_xcb::XGetXCBConnection(display);
```
i509VCB, line 170: Let's make sure this isn't null.

src/backend/x11/mod.rs
```diff
@@ -146,72 +126,124 @@ pub enum X11Event {
...
+    log: Logger,
+}
+
+unsafe impl Send for X11Connection {}
+unsafe impl Sync for X11Connection {}
 
-        let extensions = Extensions::check_extensions(&*connection, &logger)?;
+impl X11Backend {
+    /// Initializes the X11 backend by connecting to the X server.
+    pub fn new(connection: X11Connection, context: &EGLContext) -> Result<X11Backend, X11Error> {
```
i509VCB, line 205: This is pribably too early to involve egl, I'd prefer egl getting involved at the point of window creation or as a sort of X11EglSurface

src/backend/egl/ffi.rs
Drakulix: Well, at least its not much..

phire:
Testing on same setup as #914, this throws:

`thread 'main' panicked at 'Failed to create EGLDisplay: DisplayCreationError(BadParameter)', anvil/src/x11.rs:63:57`

winit also fails to create an EGLDisplay:

`Failed to initialize Winit backend: EGL error: Display creation failed with error: One or more argument values are invalid.`

Drakulix:
> thread 'main' panicked at 'Failed to create EGLDisplay: DisplayCreationError(BadParameter)', anvil/src/x11.rs:63:57

This *might* simply be the case, because of how outdated the branch is.
We have since added a bunch of nvidia specific workarounds especially for newer drivers.
E.g. https://github.com/Smithay/smithay/pull/788

So I guess, I need to rebase this mess.

phire:
If you rebase, I'll check it again. 

Otherwise, I can debug what's going on.

PR #525, opened 2022-03-16
0.30: dmabuf

i509VCB:


Drakulix (approved):
Looks good to me. Nothing obvious missing from my dmabuf-protocol-expertise. But I cannot really comment on the use of the new 0.30 api, as I am not used to that.

PolyMeilex (approved):
I can't say much about dmabuf details, but in general it looks good.

src/wayland/dmabuf/dispatch.rs
```diff
@@ -0,0 +1,250 @@
...
+            } => {
+                // Client is killed if the if statement is not taken.
+                if let Some(dmabuf) = data.create_dmabuf(dh, params, width, height, format, flags) {
+                    if state.dmabuf_state().globals.get(&data.id).is_some() {
+                        if !state.dmabuf_imported(&DmabufGlobal { id: data.id }, dmabuf.clone()) {
+                            // Buffer import failed.
+                            //
+                            // The reason for failure must be a renderer/implementation specific error.
+                            // This means we create an invalid wl_buffer and tell the client import failed.
+                            params.failed(dh);
```
vberger, line 229: The protocol allows us to either create an invalid buffer (as is done here) or kill the client with a protocol error. Could you explain why you chose the former rather than the latter?

I'm not saying we should do the latter, I just would like to know your reasoning for choosing this.

src/wayland/dmabuf/mod.rs
```diff
@@ -11,471 +11,470 @@
...
+        formats: Vec<Format>,
+        filter: F,
+        logger: L,
+    ) -> DmabufGlobal
+    where
+        D: GlobalDispatch<zwp_linux_dmabuf_v1::ZwpLinuxDmabufV1, GlobalData = DmabufGlobalData>
+            + BufferHandler
+            + DmabufHandler
+            + 'static,
+        F: for<'c> FnMut(&'c Client) -> bool + Send + 'static,
```
vberger, line 170: Do we need `FnMut` here? I'm under the impression that most filter implementations woudl not need to have mutable internal state, so `Fn` would be enought. That'd allow to remove the inner `Mutex<_>`, and the user could still put one in the closure if they really need mutable state in the filter.

src/wayland/dmabuf/mod.rs
```diff
@@ -11,471 +11,470 @@
...
+        display.remove_global(self.globals.remove(&global.id).unwrap());
+        DMABUF_GLOBAL_IDS.lock().unwrap().remove(&global.id);
+    }
+}
+
+/// Data associated with a dmabuf global.
+#[allow(missing_debug_implementations)]
+pub struct DmabufGlobalData {
+    #[allow(clippy::type_complexity)]
+    filter: Arc<Mutex<dyn for<'c> FnMut(&'c Client) -> bool + Send>>,
```
vberger, line 211: Why does this need to be in an `Arc`? It does not seem to be cloned anywhere.

src/wayland/dmabuf/mod.rs
```diff
@@ -11,471 +11,470 @@
...
+    };
+}
+
+const GLOBAL_VERSION: u32 = 3;
+
+impl DmabufParamsData {
+    /// Emits a protocol error if the params have already been used to create a dmabuf.
+    ///
+    /// This returns true if the protocol object has already been used.
+    fn if_used_then_err(
```
vberger, line 306: This does not seem to be a very explicit name for what it does when seeing it a call site.

What about something like `ensure_unused`, which returns `true` if it is indeed not yet used?

src/wayland/dmabuf/dispatch.rs
i509VCB: My rationale is as follows:
- Invalid parameters should result in a protocol error (fulfilling option one in the protocol)
- implementation specific errors should just result in import failure assuming all parameters are valid. 

I decided to read into this further, there are really two errors the renderer could have during import:
1. The number of planes is not valid for the format. This is something the renderer has to know and cannot be communicated easily to the frontend without some changes. Vulkan for example can get this info from https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkDrmFormatModifierPropertiesEXT.html
2. Implementation failure, the device was hotplugged mid import.

Since there are two cases, import should probably return a result so that an invalid plane count for a format can result in the client being killed.

src/wayland/dmabuf/mod.rs
i509VCB: I believe it's for send or sync, I'll look at it again.

src/wayland/dmabuf/dispatch.rs
i509VCB: Per matrix discussion, the protocol heavily hints that create_immed should just cause client failure.

vberger (approved):

PR #527, opened 2022-03-17
Use physical damage in the renderer

cmeissl:
As discussed I changed the renderer to take `Physical` damage instead of `Buffer` damage. This greatly simplifies the calling code as the damage, which is provided in `Logical` coordinates from `Space`, can now be easily converted to `Physical` without using the buffer `Transform` and `Scale`.

In `render_texture` the `tex_coords` have been replaced by a matrix that describes how the vertices are translated to texture coordinates, including `y_inverted`.

~~For supporting fractional scale and crop the render functions are now generic over the `Coordinate`.~~

---

TODO:
- [x] Update docs on render functions

### DIFF
--- a/anvil/src/drawing.rs
+++ b/anvil/src/drawing.rs
@@ -123,7 +123,7 @@ where
             Transform::Normal,
             &*damage
                 .iter()
-                .map(|rect| rect.to_buffer(1, Transform::Normal, &self.size))
+                .map(|rect| rect.to_f64().to_physical(scale).to_i32_round())
                 .collect::<Vec<_>>(),
             1.0,
         )?;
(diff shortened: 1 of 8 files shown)


i509VCB:
Don't forget the changelog!

cmeissl:
> Don't forget the changelog!

I intentionally did not add an entry to the changelog cause there is already one for the deskop abstractions and damage tracking.
Do we want to add changelog entries for code that touches already unreleased changes (which are already included in the changelog)?

Drakulix (approved):
Looks really solid! :+1:

Drakulix:
> > Don't forget the changelog!
> 
> I intentionally did not add an entry to the changelog cause there is already one for the deskop abstractions and damage tracking. Do we want to add changelog entries for code that touches already unreleased changes (which are already included in the changelog)?

No, the changelog should just accurately reflect the changes compared to the previously release version. So if all entries are still holding up, there is no need to add anything.

PR #529, opened 2022-03-19
Fix damage tracking with fractional output scale

cmeissl:
Based on https://github.com/Smithay/smithay/pull/527

This PR changes the renderer damage including `clear` to use `f64` just like the `dst` already did.
The problem with damage tracking and fractional output scale is that rounding the `dst` and or `damage` will
result in mismatches with the damage provided to `clear`.


Drakulix (approved):
Seems really obvious in hindsight. Thanks a ton

PR #533, opened 2022-03-20
clamp the buffer damage to the buffer dimensions

cmeissl:
Some clients use i32::MAX for damaging the buffer so we need to clamp the damage to the actual buffer dimensions.
Can be observed by running foot which will print `ERRO [GL] GL_INVALID_VALUE in glTexSubImage2D(xoffset 0 + width 2147483647 > 24), smithay_module: renderer_gles2` into the anvil log.

### DIFF
--- a/src/backend/renderer/utils.rs
+++ b/src/backend/renderer/utils.rs
@@ -132,16 +132,20 @@ where
                 if let Entry::Vacant(e) = data.textures.entry(texture_id) {
                     if let Some(buffer) = data.buffer.as_ref() {
                         let surface_size = surface_size.unwrap();
+                        let buffer_dimensions = data.buffer_dimensions.unwrap();
                         let buffer_damage = attributes
                             .damage
                             .iter()
-                            .map(|dmg| match dmg {
-                                Damage::Buffer(rect) => *rect,
-                                Damage::Surface(rect) => rect.to_buffer(
-                                    attributes.buffer_scale,
-                                    attributes.buffer_transform.into(),
-                                    &surface_size,
-                                ),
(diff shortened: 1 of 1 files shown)


Drakulix:
I wonder if we should rather fix this in the wayland/compositor module, when we receive the damage, instead of the renderer/utils.
Because I already have similar code in #528 for the multigpu-code (which has to access `attributes.damage` directly in some cases) and that feels kinda redundant.

Basically, if none of our code wants to deal with this behaviour and we are not really loosing any information, why don't do it earlier? Ping @vberger on an opinion, if our wayland-handlers should deal with it or leave it to the renderers.

cmeissl:
> I wonder if we should rather fix this in the wayland/compositor module, when we receive the damage, instead of the renderer/utils. Because I already have similar code in #528 for the multigpu-code (which has to access `attributes.damage` directly in some cases) and that feels kinda redundant.
> 
> Basically, if none of our code wants to deal with this behaviour and we are not really loosing any information, why don't do it earlier? Ping @vberger on an opinion, if our wayland-handlers should deal with it or leave it to the renderers.

Yes, that would probably be easier. I think the correct place would be during a commit as we need the transform and scale for that.

vberger:
> Basically, if none of our code wants to deal with this behaviour and we are not really loosing any information, why don't do it earlier? Ping @vberger on an opinion, if our wayland-handlers should deal with it or leave it to the renderers.

That would imply that the wayland helpers would inspect the contents of the buffer currently attached to the surface (probably at commit time) to figure its dimensions, but if that's not a problem, we can do that.

cmeissl:
superseded by #539

PR #537, opened 2022-03-21
Fix error due to unknown DRM rotation property

### LINKED ISSUE #536: Rotation causes UnknownPropterty error
The recent patch (https://github.com/Smithay/smithay/pull/521) caused a regression with my PinePhone where it would error out due to the following problem:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DrmError(UnknownProperty { handle: 37, name: "rotation" })', src/udev.rs:279:91
```

It seems like the property shouldn't be blindly set for all devices, since doing so will cause a hard error for devices that do not support it.

Also as a small sidenote: I'd really love to know the actual cause for this PR and commit, just as a potential opportunity to actually look into it myself. However the lack of a commit message detailing the exact reason for adding this kinda makes this more difficult than it should be.

chrisduerr:
This resolves a regression introduced in 6260e46, which could lead to
compositors not finding any viable GPU due to their lack of the
`rotation` property.

Since this property is only set to ensure it has a sane default value,
it should not be necessary on devices where it is not present at all.

Fixes #536.

### DIFF
--- a/src/backend/drm/surface/atomic.rs
+++ b/src/backend/drm/surface/atomic.rs
@@ -862,11 +862,9 @@ impl<A: AsRawFd + 'static> AtomicDrmSurface<A> {
                 self.plane_prop_handle(primary, "CRTC_H")?,
                 property::Value::UnsignedRange(mode.size().1 as u64),
             );
-            req.add_property(
-                primary,
-                self.plane_prop_handle(primary, "rotation")?,
-                property::Value::Bitmask(1u64),
-            );
+            if let Ok(prop) = self.plane_prop_handle(primary, "rotation") {
+                req.add_property(primary, prop, property::Value::Bitmask(1u64));
+            }
         }
 
         // and finally the others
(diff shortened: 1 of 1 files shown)


chrisduerr:
This is just a proposed fix, it seems to work for my device at least. Since I'm not fully familiar with the original issue, I'm not sure if this is the best solution. But I feel like it should definitely be superior to the current one.

i509VCB:
I think it might be worth logging if rotation is not an available property.

chrisduerr:
> I think it might be worth logging if rotation is not an available property.

Why? Seems like it's irrelevant when it doesn't exist anyway, no? If the only goal is setting it to a sane value when it exists, why would you log it when it doesn't?

Drakulix (approved):
Yep, that seems like a good solution.

i509VCB:
> 

If the rotation property isn't available and the crtc is upside-down down then you have a problem.

Issue #494, opened 2022-02-14
Crash when double-clicking foot titlebar to fullscreen


chrisduerr:
I've run into this issue while debugging something else and am not interested in looking into this myself right now. But I thought it might still be of interest.

The following backtrace will be produced when opening the foot terminal emulator and double-clicking the titlebar to fullscreen it (reproducible on all backends):

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/desktop/utils.rs:150:57
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2681f253bcdb31a274411d2be456e7c6a1c67d62/library/std/src/panicking.rs:577:5
   1: core::panicking::panic_fmt
             at /rustc/2681f253bcdb31a274411d2be456e7c6a1c67d62/library/core/src/panicking.rs:135:14
   2: core::panicking::panic
             at /rustc/2681f253bcdb31a274411d2be456e7c6a1c67d62/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
             at /rustc/2681f253bcdb31a274411d2be456e7c6a1c67d62/library/core/src/option.rs:752:21
   4: smithay::desktop::utils::damage_from_surface_tree::{{closure}}::{{closure}}
             at /home/undeadleech/programming/rust/smithay/src/desktop/utils.rs:150:34
   5: core::iter::adapters::map::map_fold::{{closure}}
             at /rustc/2681f253bcdb31a274411d2be456e7c6a1c67d62/library/core/src/iter/adapters/map.rs:84:28
   6: core::iter::traits::iterator::Iterator::fold
             at /rustc/2681f253bcdb31a274411d2be456e7c6a1c67d62/library/core/src/iter/traits/iterator.rs:2203:21
   7: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
             at /rustc/2681f253bcdb31a274411d2be456e7c6a1c67d62/library/core/src/iter/adapters/map.rs:124:9
   8: core::iter::traits::iterator::Iterator::for_each
             at /rustc/2681f253bcdb31a274411d2be456e7c6a1c67d62/library/core/src/iter/traits/iterator.rs:772:9
   9: <alloc::vec::Vec<T,A> as alloc::vec::spec_extend::SpecExtend<T,I>>::spec_extend
[...]

PR #539, opened 2022-03-21
Fix surface-damage indefinitely accumulating

Drakulix:
wayland/compositor is designed to expect downstream to drain/clear
the accumulated damage from `SurfaceData` on commit. To handle
multiple renderers the renderer/utils module did not do that, causing
damage to pile up indefinitely. This is obviously wrong, so this commit
attempts to fix this, by accumulating damage into `SurfaceState`
(clearing `SurfaceData` in the process) and keeping at most four commits
of damage for different renderers to consume.

### DIFF
--- a/src/backend/renderer/utils.rs
+++ b/src/backend/renderer/utils.rs
@@ -5,11 +5,10 @@ use crate::{
     utils::{Buffer, Logical, Point, Rectangle, Size, Transform},
     wayland::compositor::{
         is_sync_subsurface, with_surface_tree_upward, BufferAssignment, Damage, SubsurfaceCachedState,
-        SurfaceAttributes, TraversalAction,
+        SurfaceAttributes, SurfaceData, TraversalAction,
     },
 };
-#[cfg(feature = "desktop")]
-use std::collections::HashSet;
+use std::collections::VecDeque;
 use std::{
     any::TypeId,
     cell::RefCell,
@@ -19,14 +18,19 @@ use wayland_server::protocol::{wl_buffer::WlBuffer, wl_surface::WlSurface};
 
 #[derive(Default)]
 pub(crate) struct SurfaceState {
+    pub(crate) commit_count: usize,
     pub(crate) buffer_dimensions: Option<Size<i32, Buffer>>,
     pub(crate) buffer_scale: i32,
     pub(crate) buffer_transform: Transform,
     pub(crate) buffer: Option<WlBuffer>,
+    pub(crate) damage: VecDeque<Vec<Rectangle<i32, Buffer>>>,
+    pub(crate) renderer_seen: HashMap<(TypeId, usize), usize>,
     pub(crate) textures: HashMap<(TypeId, usize), Box<dyn std::any::Any>>,
     #[cfg(feature = "desktop")]
-    pub(crate) damage_seen: HashSet<crate::desktop::space::SpaceOutputHash>,
+    pub(crate) space_seen: HashMap<crate::desktop::space::SpaceOutputHash, usize>,
 }
+
+const MAX_DAMAGE: usize = 4;
 
 impl SurfaceState {
     pub fn update_buffer(&mut self, attrs: &mut SurfaceAttributes) {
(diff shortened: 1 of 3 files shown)


src/backend/renderer/utils.rs
```diff
@@ -19,31 +18,72 @@ use wayland_server::protocol::{wl_buffer::WlBuffer, wl_surface::WlSurface};
 
 #[derive(Default)]
 pub(crate) struct SurfaceState {
+    pub(crate) commit_count: usize,
```
cmeissl, line 21: `commit_count` may be a bit misleading as it is only incremented if there was a `BufferAssignment`

src/backend/renderer/utils.rs
```diff
@@ -19,31 +18,72 @@ use wayland_server::protocol::{wl_buffer::WlBuffer, wl_surface::WlSurface};
...
                 self.textures.clear();
-                #[cfg(feature = "desktop")]
-                self.damage_seen.clear();
+                let (new_commit_count, overflow) = self.commit_count.overflowing_add(1);
+                self.commit_count = new_commit_count;
+                if overflow {
+                    // we need to avoid bugs in `damage_since` (see below)
+                    self.damage.clear();
+                } else {
+                    self.damage.push_front({
```
cmeissl, line 63: Maybe use a variable for the calculated damage to reduce the nesting as it may be hard to read.

src/backend/renderer/utils.rs
```diff
@@ -52,13 +92,40 @@ impl SurfaceState {
...
-                self.damage_seen.clear();
+                self.commit_count = self.commit_count.wrapping_add(1);
+                self.damage.clear();
             }
             None => {}
         }
     }
 
+    pub(crate) fn damage_since(&self, commit: Option<usize>) -> Vec<Rectangle<i32, Buffer>> {
+        if commit
```
cmeissl, line 103: also here it may be worth to introduce a variable for the `map.unwrap_or`

cmeissl (commented):
Looks good to me :+1:

I will test it a bit later but from what I can see this should also fix #494 as the damage is cleared on buffer removal.

src/backend/renderer/utils.rs
Drakulix: Well true, but it essentially is, what we are seeing as a commit, because merged commits are indistinguishable due to the merging. Also `on_commit_buffer_handler` is meant to be called on each commit anywhere, any usage other than that is undefined behavior anyway.

src/backend/renderer/utils.rs
```diff
@@ -19,31 +18,70 @@ use wayland_server::protocol::{wl_buffer::WlBuffer, wl_surface::WlSurface};
...
                     }
                 }
                 self.textures.clear();
-                #[cfg(feature = "desktop")]
-                self.damage_seen.clear();
+                let (new_commit_count, overflow) = self.commit_count.overflowing_add(1);
+                self.commit_count = new_commit_count;
+                if overflow {
+                    // we need to avoid bugs in `damage_since` (see below)
+                    self.damage.clear();
```
vberger, line 61: In this code path, `attrs.damage` is not even read, is that expected?

Also, looking at `damage_since()`, it appears that function will always return and empty `Vec` if `self.damage` is empty, so it seems to me that we are effectively discarding all damage information on overflow, shouldn't there be at least an addition of a full damage over the surface or something like that?

Drakulix: It will only return an empty vector, if the buffer dimensions are not set (and thus we have no buffer). Otherwise it will return full-damage, because damage.len() will be too small, resulting in `recent_enough` evaluating to false.

> In this code path, attrs.damage is not even read, is that expected?

But this is a good catch, this code path should probably clear attrs.damage.

vberger: > Otherwise it will return full-damage, because damage.len() will be too small, resulting in recent_enough evaluating to false.

Oh, right, ok. Maybe let's add a comment on the computation of `recent_enough` explaining that? That in case of overflow, we'll actually have `self.commit_count < commit`

vberger (approved):

PR #540, opened 2022-03-22
Remove winit constraint from EGLNativeSurface impl

chrisduerr:
To make it possible for wayland clients to use Smithay's EGL
abstractions, it's necessary to have an available implementation for
`EGLNativeSurface`.

Currently the implementation for `WlEglSurface` is only available with
the `backend_winit` feature enabled, which pulls in a ton of additional
dependencies. Since it's not actually required, just implementing it
with the `backend_egl` enabled should be more appropriate.


chrisduerr:
I think it would also be worthwhile to implement `EglNativeDisplay` for `wayland_client::Display`, but I decided to not do that for now since `smithay` does not have a direct dependency on `wayland-client` yet.

i509VCB:
I think just WlEglSurface is fine.



> I think it would also be worthwhile to implement `EglNativeDisplay` for `wayland_client::Display`, but I decided to not do that for now since `smithay` does not have a direct dependency on `wayland-client` yet.
> 
> Maybe it is worth it to extract this code to a separate library though, if it provides some features that clients might want. I feel like it would be more at home in the wayland-egl crate maybe?

Won't work, it relies on a WlSurface + some size and extent data provided in the WlEglSurface.

chrisduerr:
> Won't work, it relies on a WlSurface + some size and extent data provided in the WlEglSurface.

That still doesn't seem like something that can only done in a *compositor*-focused library? Even the code is smithay is for the client-side as far as I understand it. So wouldn't it be more appropriate to have it in a library that focuses on clients?

chrisduerr:
People are better off using Glutin these days anyway, so I don't think there's much point in this PR.

PR #541, opened 2022-03-22
0.30 buffer management

i509VCB:
This adds a way for multiple types of WlBuffer to coexist. The interesting part of this is the `ManagedBuffer` type which statically asserts a buffer is managed by Smithay.

This is done in two commits:
- introducing the `buffer` module
- porting shm to use `buffer` module along with renderer changes.

### DIFF
--- a/src/backend/renderer/mod.rs
+++ b/src/backend/renderer/mod.rs
@@ -10,16 +10,13 @@
 use std::collections::HashSet;
 use std::error::Error;
 
-use crate::utils::{Buffer, Physical, Point, Rectangle, Size, Transform};
+use crate::utils::{Buffer as BufferCoord, Physical, Point, Rectangle, Size, Transform};
 
 #[cfg(feature = "wayland_frontend")]
-use crate::wayland::{buffer::ManagedBuffer, compositor::SurfaceData};
+use crate::wayland::{buffer::Buffer, compositor::SurfaceData};
 use cgmath::Matrix3;
 #[cfg(feature = "wayland_frontend")]
-use wayland_server::{
-    protocol::{wl_buffer, wl_shm},
-    DisplayHandle, Resource,
-};
+use wayland_server::protocol::wl_shm;
 
 #[cfg(feature = "renderer_gl")]
 pub mod gles2;
@@ -108,7 +105,7 @@ pub trait Unbind: Renderer {
 /// A two dimensional texture
 pub trait Texture {
     /// Size of the texture plane
-    fn size(&self) -> Size<i32, Buffer> {
+    fn size(&self) -> Size<i32, BufferCoord> {
         Size::from((self.width() as i32, self.height() as i32))
     }
 
@@ -157,7 +154,7 @@ pub trait Frame {
     ) -> Result<(), Self::Error> {
         self.render_texture_from_to(
             texture,
-            Rectangle::from_loc_and_size(Point::<i32, Buffer>::from((0, 0)), texture.size()),
+            Rectangle::from_loc_and_size(Point::<i32, BufferCoord>::from((0, 0)), texture.size()),
             Rectangle::from_loc_and_size(
                 pos,
                 texture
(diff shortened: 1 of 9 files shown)


src/wayland/buffer/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+    ///
+    /// At this point the buffer is no longer usable by Smithay.
+    fn buffer_destroyed(&mut self, buffer: &ManagedBuffer);
+}
+
+/// The buffer manager.
+///
+/// This type allows a [`WlBuffer`] to have data associated and managed by Smithay.
+#[derive(Debug)]
+pub struct BufferManager(());
```
PolyMeilex, line 44: If we want to keep this type, wouldn't this be cleaner?
```suggestion
pub struct BufferManager;
```

src/wayland/buffer/mod.rs
```diff
@@ -0,0 +1,139 @@
+//! Buffer management utilities.
+//!
+//! This module provides the [`Buffer`] type to represent a [`WlBuffer`] managed by Smithay and data
```
vberger, line 3:
```suggestion
//! This module provides the [`ManagedBuffer`] type to represent a [`WlBuffer`] managed by Smithay and data
```

src/wayland/buffer/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+use crate::utils::UnmanagedResource;
+
+/// Handler trait for associating data with a [`WlBuffer`].
+///
+/// This trait primarily allows compositors to be told when a buffer is destroyed.
+///
+/// # For buffer abstractions
+///
+/// Buffer abstractions (such as [`shm`](crate::wayland::shm)) should require this trait bound in dispatch
+/// requirements to ensure access to the [`BufferManager`].
```
vberger, line 26: Is this still relevant?

src/wayland/buffer/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+{
+    fn request(
+        self: Arc<Self>,
+        _: &mut Handle<D>,
+        data: &mut D,
+        _: ClientId,
+        msg: Message<ObjectId>,
+    ) -> Option<Arc<dyn ObjectData<D>>> {
+        // WlBuffer has a single request which is a destructor.
+        assert_eq!(msg.opcode, 0);
```
vberger, line 128: Let's make that a `debug_assert!()`, no need to pollute release code with such checks.

src/wayland/shm/handlers.rs
```diff
@@ -96,22 +100,13 @@ where
  * wl_shm_pool
  */
 
-/// User data of WlShmPool
-#[derive(Debug)]
-pub struct ShmPoolUserData {
-    inner: Arc<Pool>,
-}
```
vberger, line 103: Why did you move this to `mod.rs`?

src/backend/renderer/mod.rs
```diff
@@ -256,6 +256,7 @@ pub trait ImportMemWl: ImportMem {
     /// with an empty list `&[]`, the renderer is allowed to not update the texture at all.
     fn import_shm_buffer(
         &mut self,
+        dh: &mut DisplayHandle<'_>,
         buffer: &wl_buffer::WlBuffer,
```
vberger, line 260: The fact that this function takes a `wl_buffer` instead of a `ManagedBuffer` makes the whole thing pretty awkward. I think we should find a way to allow `ManagedBuffer` to cover EGL buffers as well.

src/wayland/buffer/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+
+        /// Data associated with the managed buffer.
+        data: Arc<BufferData>,
+    },
+
+    /// Buffer not managed by Smithay.
+    ///
+    /// There is no user data that may be taken from the buffer.
+    Unmanaged(WlBuffer),
+}
```
vberger, line 134: I'm wondering if the inner fields of this enum should be public, or if we should wrap it into a unit struct so that `Buffer` effectively is an opaque type from the point of view of the user.

i509VCB: Not sure it's worth exposing the inner fields. Okay the user has an object ID and buffer data. It's not exactly clear what to do with the returned data.

If this were to be public, then I'd need to make more details of BufferData public.

Plus I have added functions to get the data in side these enum variants anyways.

vberger: Ok, right, my bad, for some reason I though this was actually the public type. :shrug: All is good then.

PolyMeilex (approved):

PR #551, opened 2022-04-05
0.30: polling source for Display

i509VCB:
Roughly inspired by the code in anvil here: https://github.com/Smithay/smithay/blob/master/anvil/src/state.rs#L71


vberger:
This is an extremely thin wrapper tbh, at this point I'm wondering if just a standalone function `fn(&Display<_>) -> Generic<RawFd>` wouldn't achieve the same thing in a much simpler way?

PR #553, opened 2022-04-07
WIP: 0.30 Data device

PolyMeilex:


src/wayland/data_device/mod.rs
```diff
@@ -1,313 +1,117 @@
...
-        self.selection = new_selection;
-        self.send_selection();
-    }
+impl<D> DataDeviceState<D> {
+    pub fn new<F, C, L>(display: &mut Display<D>, callback: C, action_choice: F, logger: L) -> Self
+    where
+        L: Into<Option<::slog::Logger>>,
+        D: GlobalDispatch<WlDataDeviceManager, GlobalData = ()> + 'static,
+        D: DataDeviceHandler,
+        F: FnMut(DndAction, DndAction) -> DndAction + Send + 'static,
```
i509VCB, line 58: These two callbacks could be defined in a trait function?

PolyMeilex: I initially tried to do so but failed, they are called from grabs.

PolyMeilex: oh, wait nevermind, we have access to D in grabs now. I guess I will try again.

src/wayland/data_device/device.rs
```diff
@@ -0,0 +1,173 @@
...
+        fd: RawFd,
+    },
+}
+
+#[derive(Debug)]
+pub struct DataDeviceUserData {
+    pub(crate) wl_seat: WlSeat,
+}
+
+impl<D> DelegateDispatchBase<WlDataDevice> for DataDeviceState<D> {
```
i509VCB, line 67: Since we do plan to have server side grabs, it may be best to implement this via ObjectData implementations instead of the dispatch traits (since the high level dispatch machinery only allows one implementation).

src/wayland/data_device/server_dnd_grab.rs
```diff
@@ -182,147 +189,170 @@ where
...
-            _ => unreachable!(),
+            (*callback.lock().unwrap())(ServerDndEvent::Finished);
+            data.active = false;
         }
-    });
+        Request::SetActions {
+            dnd_actions,
+            preferred_action,
+        } => {
+            let dnd_actions = match dnd_actions {
```
i509VCB, line 325: Could be `into_result` and then `unwrap_or`

PolyMeilex: Oh, that method used to not exist before, that's nice!

src/wayland/data_device/mod.rs
PolyMeilex: Yep, enforcing D in grabs allows for those to be moved to trait. :rocket: :rocket: :rocket: :rocket: :rocket:

src/wayland/data_device/device.rs
```diff
@@ -0,0 +1,164 @@
...
+use crate::wayland::{
+    compositor,
+    data_device::seat_data::{SeatData, Selection},
+    seat::Seat,
+    Serial,
+};
+
+use super::{dnd_grab, DataDeviceHandler, DataDeviceState, DataOfferUserData};
+
+static DND_ICON_ROLE: &str = "dnd_icon";
```
i509VCB, line 24: pub const

src/wayland/data_device/device.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// Events that are generated by interactions of the clients with the data device
+#[derive(Debug)]
+pub enum DataDeviceEvent<D> {
```
i509VCB, line 28: Since we use a handler trait, these enum variants could just be turned into trait functions.

src/wayland/data_device/server_dnd_grab.rs
```diff
@@ -182,147 +189,167 @@ where
...
-where
-    C: FnMut(ServerDndEvent) + 'static,
+pub fn handle_server_dnd<D>(
+    hanldler: &mut D,
+    offer: &WlDataOffer,
+    request: wl_data_offer::Request,
+
+    metadata: &super::SourceMetadata,
+    offer_data: &Arc<Mutex<ServerDndOfferData>>,
+    callback: &Arc<Mutex<dyn FnMut(ServerDndEvent) + Send + Sync>>,
```
i509VCB, line 258: I would like this to be a trait based system as well. But it differs too much from client to client grabs, so you would have to use ObjectData to implement this specialized version.

src/wayland/data_device/device.rs
i509VCB: Or at least const

src/wayland/data_device/mod.rs
```diff
@@ -316,215 +209,205 @@ pub fn set_data_device_focus(seat: &Seat, client: Option<Client>) {
...
-            }
-        },
-        _ => unreachable!(),
-    });
-
-    ddm.deref().clone()
+/// Userdata of WlDataOffer
+#[allow(missing_docs)]
+#[derive(Debug)]
+pub enum DataOfferUserData {
```
i509VCB, line 261: Do we want to expose these implementation details to users?

src/wayland/data_device/mod.rs
```diff
@@ -29,285 +27,180 @@
...
-    SendSelection {
-        /// the requested mime type
-        mime_type: String,
-        /// the fd to write into
-        fd: RawFd,
-    },
+    ///
+    /// * `seat` - The seat on which the DnD action was finished.
+    #[allow(unused_variables)]
+    fn dropped(&mut self, seat: Seat<Self>) {}
```
i509VCB, line 109: Warning can be suppressed by naming the variable `_seat`

PolyMeilex: The idea was to not call it with _ so autocompleation works better for those, but yeah maybe it's not worth it.

src/wayland/data_device/mod.rs
PolyMeilex: This will go away when we switch to manual object impl

src/wayland/data_device/mod.rs
Drakulix: In those cases you can also just do `let _ = seat;` inside the implementation.

src/wayland/data_device/mod.rs
i509VCB: Ah I guess I was thinking too far ahead.

src/wayland/data_device/server_dnd_grab.rs
```diff
@@ -1,87 +1,68 @@
...
         }
     }
 }
 
-impl<C> PointerGrab for ServerDnDGrab<C>
+impl<D> PointerGrab<D> for ServerDnDGrab<D>
 where
-    C: FnMut(ServerDndEvent) + 'static,
+    D: DataDeviceHandler,
+    D: ServerDndGrabHandler,
```
i509VCB, line 51: We could require `trait ServerDndGrabHandler: DataDeviceHandler` to remove the redundant trait requirements.

Same with the client.

i509VCB (commented):
I'll look a bit deeper on a proper computer.

src/wayland/data_device/mod.rs
```diff
@@ -29,285 +27,173 @@
...
 
-enum Selection {
-    Empty,
-    Client(wl_data_source::WlDataSource),
-    Compositor(SourceMetadata),
+/// Event generated by the interactions of clients with a server initiated drag'n'drop
+pub trait ServerDndGrabHandler {
+    /// The client chose an action
+    #[allow(unused_variables)]
+    fn action(&mut self, action: DndAction) {}
```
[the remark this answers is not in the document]
PolyMeilex: Server dnd grab is started by the server itself

src/wayland/data_device/mod.rs
```diff
@@ -29,285 +27,173 @@
...
-    NewSelection(Option<wl_data_source::WlDataSource>),
+    #[allow(unused_variables)]
+    fn new_selection(&mut self, source: Option<WlDataSource>) {}
+
+    /// A client requested to read the server-set selection
+    ///
+    /// * `mime_type` - the requested mime type
+    /// * `fd` - the fd to write into
+    #[allow(unused_variables)]
+    fn send_selection(&mut self, mime_type: String, fd: RawFd) {}
```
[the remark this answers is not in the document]
PolyMeilex: Default impl removed

src/wayland/data_device/mod.rs
```diff
@@ -29,285 +27,173 @@
...
+    ///
+    /// The client can no longer interact
+    fn cancelled(&mut self) {}
+
+    /// The client requested for data to be sent
+    ///
+    /// * `mime_type` - The requested mime type
+    /// * `fd` - The FD to write into
+    #[allow(unused_variables)]
+    fn send(&mut self, mime_type: String, fd: RawFd) {}
```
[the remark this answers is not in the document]
PolyMeilex: Default impl removed

src/wayland/data_device/server_dnd_grab.rs
```diff
@@ -182,147 +159,192 @@ where
...
+}
+
+fn handle_server_dnd<D>(
+    handler: &mut D,
+    offer: &WlDataOffer,
+    request: wl_data_offer::Request,
+
+    data: &ServerDndData,
+
+    dh: &mut wayland_server::DisplayHandle<'_>,
```
i509VCB, line 256: Are these spaces intentional?

i509VCB (approved):

src/wayland/data_device/server_dnd_grab.rs
PolyMeilex: Nope, fixed

PR #554, opened 2022-04-08
desktop: allow to retrieve the surface under a point

cmeissl:
This PR changes the input handling to always use the input region(s) to find the surface under a point.
Using the window bounding box blocks clicking through drop shadows and using the window geometry
is wrong as it will block input on decorations (breaking move and resize).

### DIFF
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -410,6 +410,7 @@ impl EGLDisplay {
     }
 
     /// Exports an [`EGLImage`] as a [`Dmabuf`]
+    #[allow(clippy::not_unsafe_ptr_arg_deref)]
     pub fn create_dmabuf_from_image(
         &self,
         image: EGLImage,
@@ -435,6 +436,8 @@ impl EGLDisplay {
         let mut num_planes: nix::libc::c_int = 0;
         let mut modifier: ffi::egl::types::EGLuint64KHR = 0;
         if unsafe {
+            // TODO: clippy warns us here that we might dereference a raw pointer in a non unsafe public function
+            // For now we add a allow rule, but this should be addressed in the future
             ffi::egl::ExportDMABUFImageQueryMESA(
                 **self.display,
                 image,
@@ -450,6 +453,8 @@ impl EGLDisplay {
         let mut strides: Vec<ffi::egl::types::EGLint> = Vec::with_capacity(num_planes as usize);
         let mut offsets: Vec<ffi::egl::types::EGLint> = Vec::with_capacity(num_planes as usize);
         unsafe {
+            // TODO: clippy warns us here that we might dereference a raw pointer in a non unsafe public function
+            // For now we add a allow rule, but this should be addressed in the future
             if ffi::egl::ExportDMABUFImageMESA(
                 **self.display,
                 image,
--- a/src/desktop/space/mod.rs
+++ b/src/desktop/space/mod.rs
@@ -29,6 +29,8 @@ pub use self::element::*;
 use self::output::*;
 use self::window::*;
 
+use super::WindowSurfaceType;
+
 crate::utils::ids::id_gen!(next_space_id, SPACE_ID, SPACE_IDS);
 
 /// Represents two dimensional plane to map windows and outputs upon.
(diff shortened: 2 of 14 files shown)


Drakulix (commented):
I fail to see, how exactly this fixes things.

It appears the most important change is in the `input_handler`, because due to the new `space.surface_under` function no window would get raised anymore, in cases where `window.surface_under` does not return something, but `space.window_under` previously did.

**But** `window.surface_under` does also not use the `input_region` as far as I am aware, because all of those functions end up in `under_surface_tree`, which has not reference to the input region whatsoever.

So while this an ever so slight improvement, this isn't really fixed, right?
Also we probably don't want to force `under_surface_tree` to always just take the input_region into account, do we? Because than we would not have any utility function any more for cases where graphical overlap is important.

So maybe we want to introduce a new enum similar to the `WindowSurfaceType` here. Something like `RegionType`, which could be `BoundingBox`, `InputRegion` and even `OpagueRegion` (which could later be used to optimize our damage-tracking algorithm even more).

(Note that I am not against merging it as is, but as always I would prefer a more versatile solution.)

cmeissl:
> It appears the most important change is in the input_handler, because due to the new space.surface_under function no window would get raised anymore, in cases where window.surface_under does not return something, but space.window_under previously did.

Yes, but depending on the semantics it should be correct now. Orginally the bounding box was used, so clicking outside of the input region, something like a drop shadow, returned the window which got raised. But afterwards `under_from_surface_tree` did not return a surface so it has set `None` as the focus. Both things are probably wrong in this context. If I click on the drop shadow I want the window behind it to be raised. The same applies to the surface that is currently focused (at least this is how it works on other compositors afaik). I changed it to use the geometry instead of the bounding box, which made it worse as decorations are sometimes not within the geomerty, this is what broke move and resize. I also checked what for example gnome does if I explicitely use an empty input region. The window will then be completely click through, event the cursor changes to the cursor defined by the window below.

> But window.surface_under does also not use the input_region as far as I am aware, because all of those functions end up in under_surface_tree, which has not reference to the input region whatsoever.

`under_from_surface_tree` internally uses `SurfaceState::contains_point` which uses the input region.

> So while this an ever so slight improvement, this isn't really fixed, right?
Also we probably don't want to force under_surface_tree to always just take the input_region into account, do we? Because than we would not have any utility function any more for cases where graphical overlap is important.
>
> So maybe we want to introduce a new enum similar to the WindowSurfaceType here. Something like this?

I do think it fixes the things I wrote, but I agree it could be extended to support more use-cases. I named it `surface_under` to align it with what `Window` is doing and also to have the same semantics as `Window::surface_under`.

> Also we probably don't want to force under_surface_tree to always just take the input_region into account, do we?

But that is what has been implemented for `under_from_surface_tree` long before this change. It only uses some optimization internally before using the input region, like if the point is not inside of the bbox there is no need to check the input region as
the protocol specifies that the input region has to be clamped to the surface bbox anyway.

Drakulix:
> `under_from_surface_tree` internally uses `SurfaceState::contains_point` which uses the input region.

Ah! I missed that. In that case we can do a follow up PR to fix/adjust this. But that means this is the correct approach for now.

Drakulix (approved):

src/backend/egl/display.rs
```diff
@@ -410,6 +410,7 @@ impl EGLDisplay {
     }
 
     /// Exports an [`EGLImage`] as a [`Dmabuf`]
+    #[allow(clippy::not_unsafe_ptr_arg_deref)]
```
vberger, line 413: Why did that trigger?

src/desktop/space/mod.rs
```diff
@@ -121,12 +123,37 @@ impl Space {
         self.windows.iter()
     }
 
+    /// Finds the topmost surface under this point if any and returns it
+    /// together with the location of this surface and its window.
+    pub fn surface_under<P: Into<Point<f64, Logical>>>(
```
vberger, line 128: Let's write in the documentation that this takes input regions into account to find the surface.

src/backend/egl/display.rs
i509VCB: Rust 1.60, EGLImage is typedefed to a pointer.

src/desktop/space/mod.rs
cmeissl: Yeah, I will extend the docs. But then we should also change the docs on the existing methods using the input region like window and under_from_surface_tree?

vberger: Sounds right, yeah

PolyMeilex:
Nice, I can confirm that this fixes resize in a lot of apps that broke after #516

src/desktop/space/mod.rs
cmeissl: I updated the docs on `Space::surface_under` and extended the docs on `Window::surface_under` and `under_from_surface_tree`.

src/backend/egl/display.rs
cmeissl: From Matrix:

> well our EGL code is not really managing unsafe properly tbh.
And EGLImage is just a typedef.
>
> I think it needs an overhaul anyway. So for now its fine, I guess.
>
> If you add a #allow-annotation just make sure to add a TODO somewhere as well.

PR #555, opened 2022-04-10
desktop: damage whole element on first appearance

cmeissl:
fixes elements moving from one output to another within a space without damaging there whole surface.
This is necessary as the damage seen is tracked per space, but the element geometry is tracked per output.

Issue reported by @PolyMeilex on matrix


PolyMeilex (approved):
Looks good to me, but clippy disagrees, so this will have to wait for merge of #554

Drakulix:
#554 is merged. After a quick rebase and green CI, I would be happy to merge this.

PR #560, opened 2022-04-18
desktop: place a window relative to its geometry

cmeissl:
the xdg-shell protocol defines that a window should be placed relative to its geometry. the geometry should only include user visible parts, so non funtional parts like drop shadows should be ignored for placement.

from the [spec](https://wayland.app/protocols/xdg-shell#xdg_surface:request:set_window_geometry):
> When maintaining a position, the compositor should treat the (x, y) coordinate of the window geometry as the top left corner of the window.

this also fixes maximize for a lot of clients where before some parts of the decorations (menu bar) where not visible and the size did not align with the output size.

placing a window at (0,0) works now fine for most clients, but it seems that winit (or at least glutin) based clients do not put their decorations inside the geometry when run on anvil. so for clients like `alacritty` placing a window at `(0,0)` would move the top bar out of the visible bounds of the output. the client toolkit examples seem to work fine as they define a negative geometry to include the decorations.

on gnome the window geometry looks correct for alacritty:
```
[ 346182.829]  -> xdg_surface@31.set_window_geometry(-4, -28, 808, 632)
```

but on anvil the geometry does not include the decorations:
```
[ 368479.083]  -> xdg_surface@16.set_window_geometry(0, 0, 800, 600)
```

not sure what causes this behavior.

(looks like  the geometry is correctly updated after a maximize, only the initial geometry seems to be wrong)

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -257,9 +257,8 @@ impl<Backend> AnvilState<Backend> {
                     WindowSurfaceType::ALL,
                 )
                 .map(|(s, loc)| (s, loc + layer_loc));
-        } else if let Some((window, surface, location)) = space.surface_under(pos, WindowSurfaceType::ALL) {
-            let window_loc = space.window_location(&window).unwrap();
-            under = Some((surface, location + window_loc));
+        } else if let Some((_, surface, location)) = space.surface_under(pos, WindowSurfaceType::ALL) {
+            under = Some((surface, location));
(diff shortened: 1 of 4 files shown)


Drakulix (approved):
I think the weird behaviour of alacritty caused this to be implemented in a weird way in the first place.

I agree, that this is overall the better approach and that we should instead figure out, why winit-clients seem to be misbehaving instead. :+1:

PR #562, opened 2022-04-20
XWayland 0.30

i509VCB:
Well it's time to port the XWayland connection utilities to 0.30

# Changes
- Add a `display` field to `XWaylandEvent::Ready`. This is partially in relation to this issue (https://github.com/rust-lang/rust/issues/90308) about `set_env` possibly being unsound and maybe being deprecated. This is made available if someone wanted to set the XWayland environment variable in `Command`. We still call `set_env`, but this may change in the future.
- `XWayland::start` now takes a `LoopHandle<D>`. This is done to remove the generic on the `XWayland` type.
- `XWaylandEvent::Ready` now returns a UnixStream for the XWayland wayland client connection. 0.30 changes to how clients are managed have mandated this change.
- Mark `XWaylandEvent` and `XWaylandSource` as `#[must_use]`. Since you need to connect the client yourself, we do this so you do not forget in the event handler. `XWaylandSource` needs to be registered to the event loop, so it is also marked as `#[must_use]`.

## Semantic changes
- The compositor must shutdown XWayland's client connection now.


i509VCB:
@psychon could you review this?

src/xwayland/xserver.rs
```diff
@@ -80,37 +76,46 @@ pub struct XWayland<Data> {
 /// disconnecting several time in a row, but only a single connection will
 /// be active at any given time.
 #[derive(Debug)]
+#[must_use = "Connection events must be handled to prevent fd leaking"]
```
psychon, line 79: Huh? "Leak" sounds like calling `std::mem::forget` on e.g. a `UnixStream`: A file descriptor ends up being unowned.

Is this really that kind of leak? Shouldn't `UnixStream`'s `Drop` impl clean that up?

Put differently: I (apparently) do not understand the reason text.

src/xwayland/xserver.rs
```diff
@@ -80,37 +76,46 @@ pub struct XWayland<Data> {
...
+        ///
+        /// This should be used to create a Wayland client using [`Display::insert_client`](wayland_server::Display::insert_client).
+        ///
+        /// When the client is disconnected, you should [`shutdown`](XWayland::shutdown) XWayland.
+        client: UnixStream,
+
+        /// The display number the XWayland server is available at.
+        ///
+        /// This may be used if you wish to set the `DISPLAY` variable manually when spawning processes that
+        /// may use XWayland.
```
psychon, line 96: Should this comment mention something like `format!(":{}", event.display)`? That might be helpful in figuring out how to set `$DISPLAY`.

psychon (commented):
> @psychon could you review this?

I'm not sure if I can. I'm a bit "Out of the loop".

> Well it's time to port the XWayland connection utilities to 0.30

From the above, I guess that something somewhere broke API and you are now piece-wise porting to "the new world". If so, all the changes seem sensible.  I left some minor comments.

The only real change seems to be due to:

> XWaylandEvent::Ready now returns a UnixStream for the XWayland wayland client connection. 0.30 changes to how clients are managed have mandated this change.

And, yeah, well, I cannot really say much about that.

TL;DR: Looks fine to me, but I do not know anything about 0.30.

src/xwayland/xserver.rs
i509VCB: Didn't know that UnixStream had a Drop implementation (even if you need to go like 8 structs deep in std)

src/xwayland/xserver.rs
i509VCB: Yes, I should note that.

src/xwayland/xserver.rs
vberger: Well yeah, it's more like "this event needs to be handled for XWayland to work at all" tbh

vberger:
This PR does change the logic on quite a few places, and the fact that the `Client` is no longer in control of the `XWayland` instance means that the `shutdown()` method not longer does its job.

I think we should either remain closer to the previous API, and maybe require the user to provide a callback for client creation, or this module needs to be redesigned in a deeper way.

i509VCB:
Already merged in drakulix's pull request

PR #567, opened 2022-05-02
Fractional (output) scaling fixes

cmeissl:
This PR fixes some fractional scaling issues I have encountered during implementing crop & scale.
Additional it changes the default texture filtering to linear which shows better results in most cases.


Drakulix (approved):
:+1: on all three changes

PR #568, opened 2022-05-02
anvil: change implicit keyboard focus for toplevels only

cmeissl:
It seems only toplevels should receive an implicit keyboard focus.
This fixes an issue in firefox-wayland where the menu would just close on selecting an entry.

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -157,8 +157,16 @@ impl<Backend> AnvilState<Backend> {
     }
 
     fn update_keyboard_focus(&mut self, serial: Serial) {
-        // change the keyboard focus unless the pointer is grabbed
-        if !self.pointer.is_grabbed() {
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Is there a way to document this somehow? So downstream compositors don't end up with the same mistakes?
Maybe add a note on the `WindowSurfaceType` to give downstream an explanation when they should use which variant?

Otherwise :+1:

cmeissl:
Thinking about the Issue again I think the fix here may be actually wrong/incomplete. Maybe it would be a better solution to always use the root surface for the keyboard focus (unless a grab is active). That would also transfer the focus when a subsurface or popup is clicked to the root (or stay at the root). So for a window all surface types would be checked, including the popups, but the window surface would be provided for the keyboard focus.

Drakulix (approved):

PR #569, opened 2022-05-02
seat: do not send an implicit motion event when setting a grab

cmeissl:
this fixes an issue in epiphany not opening the menu on the first click.

@vberger
I have tested `weston-dnd` and `nautilus`, both seem to be happy with this change.
The focus is already removed inside the `motion` handler of the grab, so this change
should not alter the current behavior. I can add an explicit `motion` call with `None` where
the grab gets set, but from what I can tell the focus was also not removed during initialization
before.


vberger:
> The focus is already removed inside the motion handler of the grab, so this change
> should not alter the current behavior. I can add an explicit motion call with None where
> the grab gets set, but from what I can tell the focus was also not removed during initialization
> before.

Normally, as soon as a dnd grab started, the server would have sent a `wl_pointer.leave` event. Can you make sure your PR maintain this behavior? Some clients expect it IIRC.

cmeissl:
> > The focus is already removed inside the motion handler of the grab, so this change
> > should not alter the current behavior. I can add an explicit motion call with None where
> > the grab gets set, but from what I can tell the focus was also not removed during initialization
> > before.
> 
> Normally, as soon as a dnd grab started, the server would have sent a `wl_pointer.leave` event. Can you make sure your PR maintain this behavior? Some clients expect it IIRC.

I can add a None motion call to remove the focus, but I fail to see how the existing code enforces this behavior. The existing code just calls motion with the current focus when setting a grab and I have not found where the focus is actually set to None for DnD. So most likely it just genrates a motion event for the same surface cause at the start of the gab the surface already has the focus?

vberger:
It is enforced here, which is triggered by the initial synthetic motion event:

https://github.com/Smithay/smithay/blob/d5d8b670246d0da26bfd4fd19c4cbad5ca71bf9d/src/wayland/data_device/dnd_grab.rs#L62-L63

cmeissl:
> triggered

Ah, I saw that line, but failed to realize that the grab is already active at that point and will remove the focus, thanks 😅
Will update the PR tomorrow to make sure the focus is actually removed during dnd initialization. IIRC at least the move surface grab also removes the focus on motion, so maybe that should be fixed in the PR too.

cmeissl:
@vberger Anything left to address in the PR?

vberger (approved):

cmeissl:
Is there something blocking this PR or should I try to rebase it on 0.30?

vberger:
I guess I forgot about it, sorry.

Though, given our current trajectory, if it can be adapted directly to the 0.30 branch, it might be better, given the plan is to merge 0.30 into master soon.

cmeissl:
> I guess I forgot about it, sorry.
> 
> Though, given our current trajectory, if it can be adapted directly to the 0.30 branch, it might be better, given the plan is to merge 0.30 into master soon.

Ok, I guess that could be a good start to get familiar with 0.30 in smithay. Will see if I can rebase it or just replicate it on top of 0.30.

cmeissl:
@vberger Updated to 0.30

src/wayland/seat/pointer/mod.rs
```diff
@@ -67,12 +67,11 @@ impl<D> PointerInternal<D> {
         }
     }
 
-    fn set_grab<G: PointerGrab<D> + 'static>(&mut self, serial: Serial, grab: G, time: u32) {
+    fn set_grab<G: PointerGrab<D> + 'static>(&mut self, serial: Serial, grab: G, unfocus: bool) {
```
vberger, line 70: I wonder if, in terms of best-practices, it would not be better to replace this `bool` argument with an enum? Something like:

```rust
pub enum Focus {
    Keep,
    Clear
}
```

This way at call site it would be much easier to see at a glance what that argument means.

cmeissl: Yeah, this is definitely more expressive, changed `set_grab` to take a `Focus` enum.

vberger (approved):

PR #570, opened 2022-05-05
X11 wm in smithay

Drakulix:
WIP PR to track progress of implementing a re-parenting xwayland window manager in smithay and integrating it with the `desktop` module.


src/xwayland/xwm.rs
```diff
@@ -561,6 +561,7 @@ impl X11WM {
                             None
                         },
                     });
+                    // TODO: If the window is not configured as part of this callback, we need to send a synthetic configure event
```
psychon, line 573: Actually, I would suggest to *always* send a synthetic configure event after this. I learnt this the hard way: https://github.com/awesomeWM/awesome/commit/cd63cabaddabd61bbbea44bf4dc8a4f575901e92

(Yes, this commit adds dead code. `send_notice` will always end up being `true`. `client_send_configure()` is a helper function that sends a synthetic configure event.)

Don't ask me about the details, but I think I remember digging into Java's source code and finding some code that detects KWin and then behaves ICCCM conforming. There was a comment next to this claiming that this was a work-around for a bug in KWin. So... yeah, "mechanism not policy". Thanks, X11.

psychon: If only there would be something better that could replace X11... ;-)

Sorry, I don't know such a list. My general rule of thumb is that I stopped debugging bug reports involving Java, because it always ended up being Java's fault and taking a lot of time.

For proper timestamps... yeah, well, X11 is no longer used (much) across high-latency network connections and so the kind of races that proper timestamps are supposed to avoid are no longer that important. And the spec only talks about "normal" clients. My "bad excuse" is that it does not directly apply to WMs. ;-)

I did eventually write https://github.com/awesomeWM/awesome/commit/ec63afc822 and https://github.com/awesomeWM/awesome/commit/ce814b4dbfbda5088955fb977555b4fcc0ff79fc to reduce the number of `XCB_CURRENT_TIME` uses. Dunno if that ever made any differences.

So.... dunno. Feel free to ask for help if some stuff behaves weirdly. And this "Java always wants a synthetic ConfigureNotify since it fails to properly track its position"-bug has become my pet peeve since I wasted sooo much time on debugging it back then.

Drakulix:
Seems to be good enough to handle simple clients like gedit now.

It is still missing window decorations, but I would rather start incrementally making this better than trying to address it all at once.
Steam and fullscreen games are another thing known not to work properly.

src/xwayland/xwm/surface.rs
```diff
@@ -0,0 +1,878 @@
...
+            Some(atom) if atom == AtomEnum::WM_TRANSIENT_FOR.into() => self.update_transient_for(),
+            Some(_) => Ok(()), // unknown
+            None => {
+                self.update_title()?;
+                self.update_class()?;
+                self.update_protocols()?;
+                self.update_hints()?;
+                self.update_normal_hints()?;
+                self.update_transient_for()?;
+                self.update_net_state()?;
```
psychon, line 514: Perhaps a comment saying that `_NET_WM_STATE` may only be modified by the WM and thus does not need to be in the list above?

src/xwayland/xwm/surface.rs
```diff
@@ -0,0 +1,878 @@
...
+        if let Some(surface) = self.state.lock().unwrap().wl_surface.as_ref() {
+            KeyboardTarget::enter(surface, seat, data, keys, serial);
+        }
+    }
+
+    fn leave(&self, seat: &Seat<D>, data: &mut D, serial: Serial) {
+        if self.input_mode() == InputMode::None {
+            return;
+        } else if let Some(conn) = self.conn.upgrade() {
+            if let Err(err) = conn.set_input_focus(InputFocus::NONE, x11rb::NONE, x11rb::CURRENT_TIME) {
```
psychon, line 803: Hm... :thinking: 

Xwayland really is a special case. In a "normal WM" this would be a really bad idea. But here it might actually be fine to send all input to `/dev/null`...

psychon (commented):
I scrolled through this and focused on X11-specific things. Here's the result. Thus, I ignored most of the code since I do not really know enough Wayland / smithay for this. Hopefully the X11 bits are useful.

src/xwayland/xwm/surface.rs
```diff
@@ -0,0 +1,878 @@
...
+                .width(rect.size.w as u32)
+                .height(rect.size.h as u32)
+                .border_width(0);
+            if let Some(frame) = state.mapped_onto {
+                let win_aux = ConfigureWindowAux::default()
+                    .width(rect.size.w as u32)
+                    .height(rect.size.h as u32)
+                    .border_width(0);
+                conn.configure_window(frame, &aux)?;
+                conn.configure_window(self.window, &win_aux)?;
```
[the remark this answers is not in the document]
Drakulix: But wouldn't that configure the reparented window to be offset from it's parent? Or is the location passed here global?

I thought I would want to send the location only to the frame (if it is no override-redirect window) and the size to both of them.

src/xwayland/xwm/surface.rs
```diff
@@ -0,0 +1,878 @@
...
+
+    /// Set the X11 windows as mapped/unmapped affecting its visibility.
+    ///
+    /// It is an error to call this function on override redirect windows
+    pub fn set_mapped(&self, mapped: bool) -> Result<(), X11SurfaceError> {
+        if self.override_redirect {
+            if mapped {
+                return Ok(());
+            } else {
+                return Err(X11SurfaceError::UnsupportedForOverrideRedirect);
```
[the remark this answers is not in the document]
Drakulix: I thought override-redirect windows map themselves and I just get a `map_notify`, so downstream doesn't need to map it? (and shouldn't explicitly?)

src/xwayland/xwm/surface.rs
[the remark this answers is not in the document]
psychon: Oh. Seems like I totally missed the two different window arguments. Sorry for the noise & dunno how i was able to miss this.

src/xwayland/xwm/surface.rs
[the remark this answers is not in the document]
psychon: From the point of view of the WM: Sure.

I am just suggesting one of two changes:

- Change the doc to only say "it is an error to call this on override redirect windows *with argument false*" (making it clear that the function can be called on these function
- Change the implementation to `if self.override_redirect { return Err(X11SurfaceError::UnsupportedForOverrideRedirect); }` so that it does what its doc comment says.

I guess the best option would be to extend the doc comment a bit? Something like "Override redirect windows cannot be unmapped through this function" might be what is intended? I'm not sure.

src/xwayland/xwm/mod.rs
```diff
@@ -657,6 +666,57 @@ impl X11WM {
...
+            pixels,
+        )?;
+        let cursor = self.conn.generate_id()?;
+        self.conn
+            .render_create_cursor(cursor, picture.picture(), hotspot.x, hotspot.y)?;
+        if let Some(old_cursor) = self.cursor.replace(cursor) {
+            let _ = self.conn.free_cursor(old_cursor);
+        };
+        self.conn
+            .change_window_attributes(self.screen.root, &ChangeWindowAttributesAux::new().cursor(cursor))?;
```
psychon, line 717: AFAIR, X11 resources are usually reference counted internally in the X11 server. Thus, you should be able to `free_cursor` directly after `change_window_attributes()`. But I am not a 100% sure about this.

(And I am only guessing that `self.cursor` exists to only free the cursor when another one is needed.)

anvil/src/shell/x11.rs
```diff
@@ -0,0 +1,278 @@
...
+            // Assumes that at least one output exists
+            .expect("No outputs found");
+        let geometry = self.state.space.output_geometry(output).unwrap();
+
+        window.set_maximized(true).unwrap();
+        window.configure(geometry).unwrap();
+        window.user_data().insert_if_missing(OldGeometry::default);
+        window.user_data().get::<OldGeometry>().unwrap().save(old_geo);
+        self.state.space.map_element(elem, geometry.loc, false);
+    }
```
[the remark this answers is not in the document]
Drakulix: Good catch, fixed.

src/xwayland/xwm/surface.rs
```diff
@@ -0,0 +1,878 @@
...
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_TOOLBAR => Some(WmWindowType::Toolbar),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_TOOLTIP => Some(WmWindowType::Tooltip),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_UTILITY => Some(WmWindowType::Utility),
+                _ => None,
+            })
+    }
+
+    fn change_net_state(&self, added: &[Atom], removed: &[Atom]) -> Result<(), ConnectionError> {
+        let conn = self.conn.upgrade().ok_or(ConnectionError::UnknownError)?;
+        conn.grab_server()?;
```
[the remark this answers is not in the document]
Drakulix: Fixed using [scopeguard](https://crates.io/crates/scopeguard). There were quite a few of these cases.

src/xwayland/xwm/surface.rs
```diff
@@ -0,0 +1,878 @@
...
+        match (
+            state.hints.as_ref().and_then(|hints| hints.input).unwrap_or(true),
+            state.protocols.contains(&WMProtocol::TakeFocus),
+        ) {
+            (false, false) => InputMode::None,
+            (true, false) => InputMode::Passive, // the default
+            (true, true) => InputMode::LocallyActive,
+            (false, true) => InputMode::GloballyActive,
+        }
+    }
```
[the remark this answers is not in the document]
Drakulix: Yes, it did save a few lines, but not a whole lot given most `InputMode`-code was in the keyboard handling code anyway, which has to check both of these flags. Plus the checks actually became less descriptive, so I'll just let is bet the way it is now.

src/xwayland/xwm/surface.rs
Drakulix: Yes, a normal X11 Server would likely always have something focused. But with wayland surfaces being in the picture this seems to be the correct solution and so far this hasn't caused any issue. I'll just remove this line and keep the focus on the last X11 window, if any clients actually start misbehaving because of it.

src/xwayland/xwm/mod.rs
Drakulix: Seems like it works! thanks :)

src/xwayland/xwm/surface.rs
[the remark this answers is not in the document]
Drakulix: > Change the implementation to if self.override_redirect { return Err(X11SurfaceError::UnsupportedForOverrideRedirect); } so that it does what its doc comment says.

Done. I think this was from a previous iteration, that merged `MappedNotify` and `MapRequest` events until I decided to make it more explicit.
Didn't notice this implementation detail at first. Good call as well.

anvil/src/shell/element.rs
```diff
@@ -0,0 +1,508 @@
...
+    fn is_in_input_region(&self, point: &Point<f64, Logical>) -> bool {
+        if self.decoration_state().is_ssd {
+            point.y < HEADER_BAR_HEIGHT as f64
+                || match self {
+                    WindowElement::Wayland(w) => {
+                        SpaceElement::is_in_input_region(w, &(*point - Point::from((0.0, 32.0))))
+                    }
+                    #[cfg(feature = "xwayland")]
+                    WindowElement::X11(w) => {
+                        SpaceElement::is_in_input_region(w, &(*point - Point::from((0.0, 32.0))))
```
cmeissl, line 378: This should probably use `HEADER_BAR_HEIGHT` instead of the hardcoded `32.0`

anvil/src/shell/ssd.rs
```diff
@@ -0,0 +1,251 @@
...
+
+impl WindowElement {
+    pub fn decoration_state(&self) -> RefMut<'_, WindowState> {
+        self.user_data().insert_if_missing(|| {
+            RefCell::new(WindowState {
+                is_ssd: false,
+                ptr_entered_window: false,
+                header_bar: HeaderBar {
+                    pointer_loc: None,
+                    width: 128,
```
cmeissl, line 234: What is the reason behing using `128` as the default `width`? Wouldn't `0` also work as long as the width is not known?

anvil/src/shell/x11.rs
```diff
@@ -0,0 +1,290 @@
...
+    fn configure_request(
+        &mut self,
+        _xwm: XwmId,
+        window: X11Surface,
+        _x: Option<i32>,
+        _y: Option<i32>,
+        w: Option<u32>,
+        h: Option<u32>,
+    ) {
+        // Nope
```
cmeissl, line 83: Is this some left-over?

anvil/src/state.rs
```diff
@@ -259,26 +262,60 @@ impl<BackendData> XdgActivationHandler for AnvilState<BackendData> {
...
+                        w.set_ssd(true);
+                        Mode::ServerSide
+                    }
+                    _ => {
+                        w.set_ssd(false);
+                        Mode::ClientSide
+                    }
+                });
+            });
+            toplevel.send_configure();
```
cmeissl, line 294: Is it possible that this will send a configure before the initial commit?

cmeissl (commented):
I tried to review the code itself as I am not experienced with X11 in any way. Left some small nits and questions.

src/xwayland/xwm/surface.rs
```diff
@@ -0,0 +1,943 @@
...
+            conn,
+            atoms,
+            state: Arc::new(Mutex::new(SharedSurfaceState {
+                alive: true,
+                wl_surface: None,
+                mapped_onto: None,
+                geometry,
+                title: String::from(""),
+                class: String::from(""),
+                instance: String::from(""),
```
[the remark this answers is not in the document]
Drakulix: While the protocol might technical allow clients to not set these values (I am not sure though), the ICCCM demands them and I am not aware of any clients, that don't do that.

src/xwayland/xwm/surface.rs
```diff
@@ -0,0 +1,943 @@
...
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_DIALOG => Some(WmWindowType::Dialog),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_MENU => Some(WmWindowType::Menu),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_NOTIFICATION => Some(WmWindowType::Notification),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_NORMAL => Some(WmWindowType::Normal),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_POPUP_MENU => Some(WmWindowType::PopupMenu),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_SPLASH => Some(WmWindowType::Splash),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_TOOLBAR => Some(WmWindowType::Toolbar),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_TOOLTIP => Some(WmWindowType::Tooltip),
+                x if *x == self.atoms._NET_WM_WINDOW_TYPE_UTILITY => Some(WmWindowType::Utility),
+                _ => None,
```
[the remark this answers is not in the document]
Drakulix: These should be all defined by the EWMH and following specifications. Given how old these are, I don't expect any clients to craft up new ones and then expect all WMs to pick them up.

src/desktop/space/wayland/window.rs
```diff
@@ -171,8 +53,9 @@ impl SpaceElement for Window {
         }
 
         let mut surface_list = output_surfaces(output);
+        let surface = self.toplevel().wl_surface();
```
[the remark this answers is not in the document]
Drakulix: I am confused.. that is exactly what I am doing here?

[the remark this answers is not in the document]
cmeissl: Tt was me that was probably looking at the wrong side of the diff, please ignore that...

src/desktop/space/wayland/window.rs
```diff
@@ -207,15 +90,11 @@ impl SpaceElement for Window {
         self.user_data().insert_if_missing(WindowOutputUserData::default);
         let state = self.user_data().get::<WindowOutputUserData>().unwrap().borrow();
 
+        let surface = self.toplevel().wl_surface();
```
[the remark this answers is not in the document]
cmeissl: also this can be ignored, sorry

cmeissl (approved):
LGTM

src/xwayland/xwm/mod.rs
```diff
@@ -657,6 +660,55 @@ impl X11WM {
 
         surface.state.lock().unwrap().wl_surface = Some(wl_surface);
     }
+
+    /// Set the default cursor used by X clients
+    pub fn set_cursor(
+        &mut self,
+        pixels: &[u8],
```
i509VCB, line 670: What format are the pixels supposed to be in?

src/xwayland/xwm.rs
```diff
@@ -0,0 +1,476 @@
...
+        width: Option<u32>,
+        height: Option<u32>,
+        border_width: Option<u32>,
+        stacking: Option<(StackMode, X11Window)>,
+    },
+}
+
+/// The runtime state of the XWayland window manager.
+#[derive(Debug)]
+pub struct X11WM {
```
i509VCB: On naming are we supposed to use `WM` or `Wm`?

src/utils/x11rb.rs
```diff
@@ -36,6 +36,8 @@ use calloop::{
 pub struct X11Source {
     connection: Arc<RustConnection>,
     channel: Option<Channel<Event>>,
+    #[cfg(feature = "xwayland")]
+    pub(crate) sender: SyncSender<Event>,
```
i509VCB, line 40: What is this sender for? Another internal pub(crate) field isn't exactly very nice for untangling the single monolithic crate in the future.

src/xwayland/xwm/mod.rs
```diff
@@ -0,0 +1,1167 @@
...
+    /// To grant the wish you have to call `X11Surface::set_mapped(true)` for the window to become visible.
+    fn map_window_request(&mut self, xwm: XwmId, window: X11Surface);
+    /// Override redirect window was mapped.
+    ///
+    /// This is a notification. The XWM cannot prohibit override redirect windows to become mapped.
+    fn mapped_override_redirect_window(&mut self, xwm: XwmId, window: X11Surface);
+    /// Window was unmapped.
+    fn unmapped_window(&mut self, xwm: XwmId, window: X11Surface);
+    /// Window was destroyed
+    fn destroyed_window(&mut self, xwm: XwmId, window: X11Surface);
```
i509VCB, line 179: Is it garunteed that unmap would always be called before destroy? Or could destroy happen first because of a client disconnecting abruptly.

src/xwayland/xwm/mod.rs
Drakulix: RGBA. You are right, that should be documented.

src/xwayland/xwm.rs
Drakulix: Well technically it is an acronym, so `WM`. But I see the appeal for `Wm`, what does you gut say?

src/utils/x11rb.rs
Drakulix: It is used in the `X11Injector`-type to inject events in our processing loop, which is used to ease the wl_surface-id-dance.
This sender already exists inside the source without xwayland, the feature just adds an otherwise unused way to access that sender. So for separating this is not really a problem, we could just make the sender part of a public api of the `X11Source`, if we would split it apart.

src/xwayland/xwm/mod.rs
Drakulix: I have never seen the XServer deliver these out of order.

Drakulix: Perfect, that this is a non-issue.

src/xwayland/xwm.rs
i509VCB: I'd think `Wm`, as that is in line with how Unix (UnixStream) is used through out the standard library.

Drakulix: Sure. I can change that.

src/xwayland/xwm/mod.rs
```diff
@@ -0,0 +1,1172 @@
+//!
+//! Xwayland Window Manager module
+//!
+//! Provides an [`X11Wm`] type, which will register itself as a window manager for a previously spawned Xwayland instances,
+//! allowing backwards-compatibility by seemlessly integrating X11 windows into a wayland compositor.
+//!
+//! To use this functionality you must first spawn an [`XWayland`](super::XWayland) instance to attach a [`X11Wm`] to.
+//!
+//! ```norun
```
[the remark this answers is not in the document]
Drakulix: Good catch. Should be fixed

PolyMeilex (commented):
I don't see anything obviously wrong codewise. LGTM

src/xwayland/xwm/mod.rs
```diff
@@ -664,13 +664,18 @@ impl X11WM {
         surface.state.lock().unwrap().wl_surface = Some(wl_surface);
     }
 
-    /// Set the default cursor used by X clients
+    /// Set the default cursor used by X clients.
+    ///
+    /// `pixels` is expected to be in `rgba`-format with each channel encoded as an u8.
+    ///
+    /// This function will panic, if `pixels` is not at least `size.w * size.h * 4` long.
```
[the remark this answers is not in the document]
Drakulix: Good catch. Updated the assert.

anvil/src/shell/element.rs
```diff
@@ -0,0 +1,510 @@
...
+    },
+    xwayland::X11Surface,
+};
+
+use super::ssd::HEADER_BAR_HEIGHT;
+use crate::AnvilState;
+
+#[cfg(not(feature = "xwayland"))]
+#[derive(Debug, Clone, PartialEq)]
+pub enum WindowElement {
```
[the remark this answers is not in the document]
Drakulix: Nope, I think this is a left-over from when I was still using the `space_elements!`-macro for generating this enum, which didn't really like the feature bounds. But that auto-generated code didn't work any more once I implemented SSDs.

Anyways, Fixed.

PolyMeilex (approved):

PR #575, opened 2022-05-12
input: Allow focus to be grabbed by generic `*Handler`-trait objects instead of `WlSurface`s.

Drakulix:
This is a first attempt at decoupling a large portion of the `wayland/seat` module from `WlSurface` as the only object to receive any input events.

## Motivation

Currently all custom input-handling code has to be build from ground-up in downstream compositors. smithay either provides raw input events or converts those (almost automatically, except for focus handling) into appropriate wayland-events.

There is no middle ground, where some pre-processing can take place other components could benefit from. Especially implementing window-decorations on the server-side is tedious currently. Custom shells are a giant pain.

The plan with this is to allow custom *elements* to be attached to a `Window` or even a `Space`, that can handle input events or trigger grabs like any wayland client inside smithay.

## Drawbacks

The current implementation is quite rough around some edges, that need discussion:

- The new `Pointer/KeyboardHandler`-traits don't require `Clone`. A lot of the old code assumed, that the focus object is clonable, given `WlSurface`s do have this property.
  - Grabs are sometimes initialized from existing grab's start data. Because the `GrabStartData` contains the focus it is also not clonable anymore. Instead a new `take_grab_start_data` method was introduced.
    - Additionally `grab_start_data` now returns a `Ref<StartData>`, which exposes the underlying `RefCell`. This makes double-borrows and thus runtime errors a little too easy.
  - The `ClickGrab` requires the focus to be clonable. As a result only `WlSurface`s will exhibit the click-grab semantics now. Other focused object will not receive events outside of their surface. (Specialization could help with this...)
  - The `PointerHandler` has no **implicit** `pending_focus` anymore, because this cannot be done without cloning. Working around this using `Rc<RefCell<_>>` and the like proved to be not working, because it breaks down-casting, which is required for some `WlSurface` specific-code (e.g. PopupGrabs).
    - Instead grabs can now call `PointerInnerHandle::set_pending_focus` **explicitly** to pass through a focus object to be restored after a grab ends. This puts more responsibility into grab-implementations..
    - Given the focus is still passed into the handler on every `motion`-call, the code calling into the `PointerHandler` arguably needs a way to clone or create new handles for recieving focus anyway. Maybe requiring `Clone` would actually be more desirable?
- Calling `motion` with `None` as focus requires a type-parameter now. Internally pointer/keyboard do have a `DummyHandler` for this. While we could expose it to the user, I opted to provide a `motion_no_focus` function instead. It feels more ergonomic, but I am not sure it is better. Maybe we should split `motion` and `set_focus`, which would also remove the effective requirement for downstream to be able to provide the focus object for every pointer movement.

## TODOs / Open Questions

- [ ] #569 needs to be ported to this in some form
- [ ] Changing keyboard repeat infos currently does not propagate to wl_keyboards (where would we put this code?)
- [ ] Decide where to put this
  - Right now it resides (like `Output`) inside `src/wayland`, although the module could be used without `wayland_frontend`.
  - Maybe introduce a new toplevel module?
- [ ] Decide what types to use. Currently things like `ButtonState` or `KeyState` are either:
  - Taken from `backend::input` (which also has no feature gate, so maybe this is the easiest solution?)
  - Or defined locally in the module + conversation implementations via `Into`/`From`
  - (or wayland-definitions could be used directly, but I have already removed the code in an attempt to make this usable without)
- [ ] What to do with `src/wayland/seat/touch`? It is missing a lot of the functionality of `pointer`/`keyboard` like grabs. Right now a similar implementation would not make a lot of sense imo.

## Alternatives

- Status Quo - maybe downstream *should* implement its own dispatching? (I disagree.)

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -16,7 +16,9 @@ use smithay::{
     wayland::{
         compositor::with_states,
         output::Scale,
-        seat::{keysyms as xkb, AxisFrame, FilterResult, Keysym, ModifiersState},
+        seat::{
+            keysyms as xkb, Axis, AxisFrame, AxisSource, ButtonState, FilterResult, Keysym, ModifiersState,
+        },
         shell::wlr_layer::{KeyboardInteractivity, Layer as WlrLayer, LayerSurfaceCachedState},
         Serial, SERIAL_COUNTER as SCOUNTER,
     },
@@ -96,7 +98,7 @@ impl<Backend> AnvilState<Backend> {
                     && (data.layer == WlrLayer::Top || data.layer == WlrLayer::Overlay)
                 {
                     self.keyboard
-                        .set_focus(Some(layer.get_surface().unwrap()), serial);
+                        .set_focus(Some(layer.get_surface().unwrap().clone()), serial);
                     self.keyboard
                         .input::<(), _>(keycode, state, serial, time, |_, _| FilterResult::Forward);
                     return KeyAction::None;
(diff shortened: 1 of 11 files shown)


vberger:
> Changing keyboard repeat infos currently does not propagate to wl_keyboards (where would we put this code?)

Why not have a method on the keyboard handler trait?

> Decide where to put this

Indeed, given this is getting pretty generic over non-wayland stuff, it might make sense to put it into some new crate-level module that would regroup such helpers :thinking: 

> Currently things like ButtonState or KeyState are [...] taken from backend::input

That sounds like the best solution to me tbh

> What to do with src/wayland/seat/touch?

First of all, figure out what are the precise semantics of grabs wrt to touch events, if any. And then eventually port the touch module to the same model (but probably not in this PR).

Drakulix:
Now tracked in https://github.com/Smithay/smithay/pull/689

PR #454, opened 2022-01-07
Rework wayland rs 0.30

vberger:
## TODOs

- [x] tablet_manager
- [x] drop xdg_foreign
- [x] drop explicit-synchronization
- [x] drop wl_shell
- [x] Reimplement [keyboard::has_focus](https://github.com/Smithay/smithay/blob/f061ced7aa3366df091ab2b8af03d1730fb967fa/src/wayland/seat/keyboard/mod.rs#L503) - https://github.com/Smithay/smithay/pull/618
- [x] [shm::with_buffer_contents error todo](https://github.com/Smithay/smithay/blob/rework-wayland-rs-0.30/src/wayland/shm/mod.rs#L192)
- [x] Docs
- [x] Tests

### anvil
- [x] https://github.com/Smithay/smithay/pull/584
- [x] xwayland: https://github.com/Smithay/smithay/pull/586
- [x] ~(maybe code cleanup?)~
- [x] Update wlcs_anvil

### smallvil
- [x] Resize Grab

### Open PRs

- [x] https://github.com/Smithay/smithay/pull/613
- [x] https://github.com/Smithay/smithay/pull/602

### Backports

- [x] https://github.com/Smithay/smithay/pull/585
- [x] https://github.com/Smithay/smithay/pull/582
- [x] https://github.com/Smithay/smithay/pull/581
- [x] https://github.com/Smithay/smithay/pull/577
- [x] https://github.com/Smithay/smithay/pull/574
- [x] https://github.com/Smithay/smithay/pull/573
- [x] https://github.com/Smithay/smithay/pull/568
- [x] https://github.com/Smithay/smithay/pull/571 ?
- [x] https://github.com/Smithay/smithay/pull/567
- [x] https://github.com/Smithay/smithay/pull/566
- [x] https://github.com/Smithay/smithay/pull/557 + https://github.com/Smithay/smithay/pull/561 + https://github.com/Smithay/smithay/pull/565
- [x] https://github.com/Smithay/smithay/pull/617
- [x] https://github.com/Smithay/smithay/pull/588

PR #576, opened 2022-05-12
0.30: Port multigpu renderer

Drakulix:
Actually felt waaay to easy, but I am not going to complain.

### DIFF
--- a/src/backend/renderer/multigpu/egl.rs
+++ b/src/backend/renderer/multigpu/egl.rs
@@ -1,7 +1,8 @@
 //! Implementation of the multi-gpu [`GraphicsApi`] using
 //! EGL for device enumeration and OpenGL ES for rendering.
 
-use wayland_server::DisplayHandle;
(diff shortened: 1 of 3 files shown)


PolyMeilex:
Clippy is not happy, but besides that it looks good.

src/backend/renderer/multigpu/egl.rs
```diff
@@ -1,7 +1,7 @@
 //! Implementation of the multi-gpu [`GraphicsApi`] using
 //! EGL for device enumeration and OpenGL ES for rendering.
 
-use wayland_server::DisplayHandle;
+use wayland_server::{protocol::wl_buffer, DisplayHandle};
```
i509VCB, line 4: Missing Wayland frontend cfg guards?

Drakulix:
I think every CI-step, that passes on #454 passes now.

i509VCB (approved):

PR #584, opened 2022-05-18
0.30: Anvil

Drakulix:
## Notes so far on the porting effort:

- ~~Why don't we have a delegate-macro for wl_output/xdg-output?~~ Fixed
- ~~Why don't we have a delegate-macro for wlr_layer_shell?~~ Fixed
- `Seat::add_pointer`'s callback is the place to get the `CursorImage` and still replies on a `Arc<Mutex<_>>`, can we get rid of that?
- The `CompositorHandler::commit` callback still handles a lot of stuff, we probably want to make use of our new `post_commit_hooks` (though maybe after the initial port / 0.30 merge).

## Things missing:

- [x] X11 Backend
- [x] Udev Backend
- ~~[ ] XWayland (depends on #562)~~ TBD in #586
- [x] ~~Clients are crashing on startup~~


i509VCB:
> Why don't we have a delegate-macro for wl_output/xdg-output?
> Why don't we have a delegate-macro for wlr_layer_shell?

probably forgot?

PolyMeilex:
> Why don't we have a delegate-macro for wl_output/xdg-output?
> Why don't we have a delegate-macro for wlr_layer_shell?

Those were one of the first modules to be ported, and we never went back to add macros to them once we started using `delegate_module_name!` pattern. Just an oversight.

src/backend/egl/display.rs
```diff
@@ -617,8 +617,8 @@ impl EGLDisplay {
     /// if called for the same [`Display`] multiple times, as only one egl display may be bound at any given time.
     #[cfg(all(feature = "use_system_lib", feature = "wayland_frontend"))]
     #[allow(unreachable_code, unused_variables)] // Temporary warnings
```
i509VCB, line 619: dead warnings

anvil/src/shell.rs
```diff
@@ -221,47 +211,58 @@ impl PointerGrab for ResizeSurfaceGrab {
...
-                with_states(self.window.toplevel().wl_surface().unwrap(), |states| {
+                with_states(self.window.toplevel().wl_surface(), |states| {
                     let mut data = states
                         .data_map
                         .get::<RefCell<SurfaceData>>()
                         .unwrap()
                         .borrow_mut();
                     if let ResizeState::Resizing(resize_data) = data.resize_state {
-                        data.resize_state = ResizeState::WaitingForFinalAck(resize_data, serial);
+                        data.resize_state = ResizeState::WaitingForFinalAck(resize_data, event.serial);
```
PolyMeilex, line 259: Once #598 is merged, send_configure() serial should be used here, instead of unrelated button serial. Ack serial is the serial that was used to send the configure. As protocol description of `ack_configure` states `serial - is the serial from the configure event`

i509VCB (approved):

anvil/src/shell.rs
PolyMeilex: Let's leave this for another PR, as it's not related, and requires more investigation as to why are we waiting for that ack to begin with.

PolyMeilex (approved):

PR #586, opened 2022-05-20
0.30/anvil xwayland

Drakulix:
Continuation of #562, but rebased on top of #584 to have something to actually test it with.

Simplifies some places and more importantly moves client creation into the xwayland module again, so it we ~~*could*~~ **can** implement proper shutdown ~~if `Client::kill` would not need a `DisplayHandle`~~.


anvil/src/xwayland/mod.rs
```diff
@@ -181,16 +185,22 @@ impl X11State {
...
+                    let surface = self.client.object_from_protocol_id(dh, id);
+
                     match surface {
-                        None => {
+                        Err(_) => {
                             self.unpaired_surfaces.insert(id, (msg.window, location));
                         }
-                        Some(surface) => self.new_window(msg.window, surface, location),
+                        Ok(surface) => {
+                            info!(
```
vberger, line 195: Seems to me it should be `debug!` here

vberger (approved):
LGTM

PR #587, opened 2022-05-23
Add DRM format tables

i509VCB:
Adds a bunch of data that is not implicitly available with DRM formats.

This data is generated with a macro that exposes some functions.

This provides the following data related to formats:
- bits per pixel
- opaque alternatives
- whether a format has an alpha channel, useful for graphics apis like Vulkan where there are no opaque formats (instead swizzling the alpha component).

Also some formats need an update drm-fourcc crate, I tried to update it but last time I tried it seemed to fail to find the headers (even on a system with the headers installed)?

This intentionally does not include other graphics api format conversions. Some formats defined in graphics apis may only work with specific endian values or may require extra data (like Vulkan with component swizzling).

## Testing

Some tests are included to assert the data is mostly correct:
- A format with alpha must not have it's opaque equivalent be the same format.
- A format with alpha must have the same bpp as it's opaque format.
- An opaque format must not have alpha, and must not have an opaque format.
- A format with opaque alternative must have alpha.

## Questions

Is `allocator` the right module to place this in? Or should this table be maintained in a separate crate?

I have no idea how to represent the YUV formats here, so these are not done yet unless this is something that is essentially to this pull request.

### DIFF
--- a/src/backend/allocator/format.rs
+++ b/src/backend/allocator/format.rs
@@ -1,0 +1,391 @@
+//! Format info tables for DRM formats.
+//!
+//! This module provides three functions, [`get_opaque`], [`has_alpha`] and [`get_bpp`].
+//!
+//! [`get_opaque`] returns the opaque alternative of a DRM format with an alpha channel.
+//!
+//! ```
+//! # use smithay::backend::allocator::Fourcc;
+//! # use smithay::backend::allocator::format::get_opaque;
+//! assert_eq!(Some(Fourcc::Xrgb8888), get_opaque(Fourcc::Argb8888));
+//! ```
+//!
+//! [`has_alpha`] returns true if the format has an alpha channel.
+//!
+//! ```
+//! # use smithay::backend::allocator::Fourcc;
+//! # use smithay::backend::allocator::format::has_alpha;
+//! assert!(has_alpha(Fourcc::Argb8888));
+//! assert!(!has_alpha(Fourcc::Xrgb8888));
+//! ```
+//!
+//! [`get_bpp`] returns the number of bits per pixel of a format.
+//!
+//! ```
+//! # use smithay::backend::allocator::Fourcc;
+//! # use smithay::backend::allocator::format::get_bpp;
+//! assert_eq!(get_bpp(Fourcc::Argb8888), Some(32));
+//! ```
+
+/// Macro to generate table lookup functions for formats.
+///
+/// See the module documentation for usage details.
+macro_rules! format_tables {
+    (
+        $($fourcc: ident {
+            $(opaque: $opaque: ident,)?
+            alpha: $alpha: expr,
+            bpp: $bpp: expr $(,)?
+        }),*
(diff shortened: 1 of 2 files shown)


src/backend/allocator/format.rs
```diff
@@ -0,0 +1,399 @@
...
+            ]
+        }
+    };
+}
+
+format_tables! {
+    // 8-bit bpp Red
+    R8 {
+        bpp => 8,
+    },
```
vberger, line 91: In terms of readability and future maintainability, I tend to think it'd be more convenient to have those formatted as single-line entries?

like
```
R8 { bpp => 8 },
```

src/backend/allocator/format.rs
```diff
@@ -0,0 +1,399 @@
...
+
+/// Macro to generate table lookup functions for formats.
+///
+/// See the module documentation for usage details.
+macro_rules! format_tables {
+    (
+        $($fourcc: ident {
+            $(opaque => $opaque: ident,)?
+            bpp => $bpp: expr,
+        }),*
```
vberger, line 37: Nitpick: let's use `:` to match the struct definition format, and optionally allow a trailing `,` rather than require it. IMO that'd be more practical for future reading and maintenance of this table.

```suggestion
        $($fourcc: ident {
            $(opaque : $opaque: ident,)?
            bpp : $bpp: expr $(,)?
        }),*
```

Drakulix:
> Is `allocator` the right module to place this in? Or should this table be maintained in a separate crate?

Given that `Format` is defined in the allocator module, it fits. 


> I have no idea how to represent the YUV formats here, so these are not done yet unless this is something that is essentially to this pull request.

I don't think they have to be done in this PR, but we should make sure, that we don't run into issues later down the line, when we want to support those. So we probably should figure out how to represent them.

src/backend/allocator/format.rs
```diff
@@ -0,0 +1,336 @@
...
+            }
+        }
+
+        /// Returns true if the format has an alpha channel.
+        ///
+        /// This function may be useful to know if the alpha channel may need to be swizzled when rendering.
+        ///
+        /// Unknown formats will always return `false`.
+        pub const fn has_alpha(fourcc: $crate::backend::allocator::Fourcc) -> bool {
+            get_opaque(fourcc).is_some()
```
Drakulix, line 62: Could there be alpha formats, which don't have an opaque counter part?

i509VCB: Yes there is one, `Axbxgxrx106106106106`.

Although we will need to update drm-fourcc for it.

Drakulix: So then this implementation doesn't work, does it?

i509VCB: Yes I'll need to mark whether a format has alpha explicitly and then if it has an opaque equivalent.

Drakulix (approved):
LGTM

PR #593, opened 2022-05-25
0.30 gl: fix compile error without wayland_frontend

i509VCB:
Trivial change, waiting for CI to finish to see how many more cfg configurations need fixing.


i509VCB:
I guess the examples need their features adjusted, well for the other pull request.

PR #594, opened 2022-05-25
Move smallvil to new crate

i509VCB:
This makes smallvil something you can copy and paste more easily.

Smallvil also has a CI step now.

~~(Waiting to see if ci is valid)~~


Drakulix (approved):
LGTM!

PR #596, opened 2022-05-26
manage globals using DisplayHandle

i509VCB:
This does now require turbofish syntax for global creation. This makes global registration much more flexible.


Drakulix (approved):
:+1:

PR #598, opened 2022-05-27
xdg-shell & layer-shell: Don't generate configure serial behind user's back

PolyMeilex:
Without knowing the serial it's impossible to handle ack_configure. 

Because of that, Anvil uses button click serial instead of configure serial.

EDIT:
An alternative solution would be to return the serial, but all Smithay modules accept the serial instead of returning it, so I kept the convention here.


i509VCB:
Does the layer shell follow this convention as well?

i509VCB (approved):
Other than the question around the layer shell this seems fine.

Although I guess drakulix asked Levans to look at this before merging?

vberger:
> Without knowing the serial it's impossible to handle ack_configure.

Just trying to understand the motivation, what is the kind of handling that needs to be done and that is not already done internally by Smithay?

PolyMeilex:
For example anvil is doing ack configure check on final resize event. And instead of checking the actual configure serial it uses button click serial that ended the grab to estimate if next ack is related to the configure that we did. It does `if configure.serial >= button.serial` which seems confusing to me. It works because we know that click happens one serial before configure, but this is non obvious when reading the code.

As to why are we checking the ack, I'm not 100% sure, protocol guarantees that the client will always ack & commit the newest configure, so we know that next commit will be in response to our configure, but I guess that for some reason we want make sure that it's the case before stoping the resize?


As to layer shell, yeah if we decide that there is need for this, layer shell should also reflect those changes.

PolyMeilex:
I guess I have to investigate the anvil implementation, and what are the motivation for those checks.

vberger:
> As to why are we checking the ack, I'm not 100% sure, protocol guarantees that the client will always ack & commit the newest configure, so we know that next commit will be in response to our configure, but I guess that for some reason we want make sure that it's the case before stoping the resize?

Yeah, my point is that, `Topelevel::current_state()` already allows the compositor to access the last state that was acked by the client. Isn't that enough?

Though overall, I'm not opposed to making the serials more explicit at this level anyway, I suppose that can simplify synchronization in complex situations.

PolyMeilex:
I dropped all ack related code in smallvil, and resize seem to work the same. 
I basically changed `end of resize -> wait for ack -> wait to commit` to `end of resize -> wait to commit`
So I'm not sure what was the initial reasoning for this check, 

Sadly, I can't use Weston as a reference because it does not even wait for the commit after end of resize and the window ends up in the wrong place 90% of the time. And sway has a ton of tiling specific code that I don't want to dig through :smile: 

So let's assume that ack check is not needed for resize and this PR just solves a theoretical problem.
I will add the same change to layer shell, and then we can decide if we want this change or not.

PR #600, opened 2022-05-29
use multiple dispatch implementations for buffers

i509VCB:
This simplifies the codebase and removes the need for a Buffer wrapper type that would need conversion all around compositor implementations.

### DIFF
--- a/src/wayland/dmabuf/mod.rs
+++ b/src/wayland/dmabuf/mod.rs
@@ -102,7 +102,9 @@ use std::{
 
 use nix::unistd;
 use wayland_protocols::wp::linux_dmabuf::zv1::server::{zwp_linux_buffer_params_v1, zwp_linux_dmabuf_v1};
-use wayland_server::{backend::GlobalId, Client, DisplayHandle, GlobalDispatch, Resource, WEnum};
+use wayland_server::{
+    backend::GlobalId, protocol::wl_buffer, Client, DisplayHandle, GlobalDispatch, Resource, WEnum,
+};
 
 use crate::{
     backend::allocator::{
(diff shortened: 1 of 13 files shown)


src/wayland/dmabuf/mod.rs
```diff
@@ -288,15 +290,17 @@ pub enum ImportError {
...
 ///
 /// If the buffer is not managed by the dmabuf handler (whether the buffer is a different kind of buffer,
 /// such as an shm buffer or is not managed by smithay), this function will return an [`UnmanagedResource`]
 /// error.
-pub fn get_dmabuf(buffer: &Buffer) -> Result<Dmabuf, UnmanagedResource> {
-    Ok(buffer.buffer_data::<Dmabuf>()?.ok_or(UnmanagedResource)?.clone())
+///
+/// [`WlBuffer`]: wl_buffer::WlBuffer
+pub fn get_dmabuf(buffer: &wl_buffer::WlBuffer) -> Result<Dmabuf, UnmanagedResource> {
+    Ok(buffer.data::<Dmabuf>().ok_or(UnmanagedResource)?.clone())
```
vberger, line 303:
```suggestion
    buffer.data::<Dmabuf>().cloned().ok_or(UnmanagedResource)
```

vberger (approved):
One small nit, otherwise LGTM

PolyMeilex (approved):
This looks a lot cleaner now, nice :ok_hand:

src/wayland/dmabuf/mod.rs
i509VCB: My rationale here is to avoid the clone if the data is not a dmabuf.

PolyMeilex: Calling Cloned on None will just return None:
```rs
        match self {
            Some(t) => Some(t.clone()),
            None => None,
        }
```
So this is basically free, or is this just stylistic choice?

i509VCB: Ah okay, so the clone is effectively free then.

PR #601, opened 2022-05-30
Update WlSeat to version 7

PolyMeilex:
Ver 6 was about touch events that we already have: https://github.com/wayland-project/wayland/commit/6a18a87727c64719c68168568b9ab1e4d7c2d9c1
Ver 7 is about sealing the keymap files:  https://github.com/wayland-project/wayland/commit/905c0a341ddf0a885811d19e2b79c65a3f1d210c

The Sealing could allow for some optimizations, but I'm not sure which OSes support that, so I just left the inefficient version where we create a file per client.

### DIFF
--- a/src/wayland/seat/keyboard/keymap_file.rs
+++ b/src/wayland/seat/keyboard/keymap_file.rs
@@ -32,11 +32,11 @@ impl KeymapFile {
         }
     }
 
-    pub fn with_fd<F>(&self, supports_seeled: bool, cb: F) -> Result<(), std::io::Error>
+    pub fn with_fd<F>(&self, supports_sealed: bool, cb: F) -> Result<(), std::io::Error>
     where
         F: FnOnce(RawFd, usize),
     {
-        if let Some(file) = supports_seeled.then(|| self.sealed.as_ref()).flatten() {
+        if let Some(file) = supports_sealed.then(|| self.sealed.as_ref()).flatten() {
             cb(file.as_raw_fd(), file.size);
             Ok(())
         } else {
--- a/src/wayland/seat/keyboard/keymap_file.rs
+++ b/src/wayland/seat/keyboard/keymap_file.rs
@@ -1,0 +1,101 @@
+use std::{
+    ffi::{CStr, CString},
+    fs::File,
+    io::{Seek, Write},
+    os::unix::prelude::{AsRawFd, FromRawFd, RawFd},
+    path::PathBuf,
+};
+
+use nix::{
+    fcntl::{FcntlArg, SealFlag},
+    sys::memfd::MemFdCreateFlag,
+};
+use slog::error;
+
+#[derive(Debug)]
+pub struct KeymapFile {
+    sealed: Option<SealedFile>,
+    keymap: CString,
+}
+
+impl KeymapFile {
+    pub fn new(keymap: CString, log: slog::Logger) -> Self {
+        let sealed = SealedFile::new(&keymap);
+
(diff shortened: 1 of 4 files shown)


HeavyRain266:
As far as I'm aware, from systems where you can actually run wayland (linux & freebsd), seems like only linux support that kind of feature but on the other hand freebsd uses libinput layer and will have to look into kernel code to see if there is sealing.

src/wayland/seat/keyboard/mod.rs
```diff
@@ -544,6 +545,13 @@ impl KeyboardHandle {
         let ret = tempfile().and_then(|mut f| {
             f.write_all(self.arc.keymap.as_bytes())?;
             f.flush()?;
+
+            if kbd.version() >= 7 {
+                if let Err(err) = nix::fcntl::fcntl(f.as_raw_fd(), FcntlArg::F_GET_SEALS) {
```
vberger, line 550: Is this call actually setting any kind of seals on the file?

PolyMeilex: My bad, the error log is misleading. This is basically a check if selling is supported, so the log should just be "Sealing not supported".

(If selling is supported, we should then return an existing fd of a sealed file)

I guess this log does not make much sense if the whole thing is not implemented.
I guess I'll implement the whole thing, it is supported on Linux and FreeBSD and that's the only targets we care about anyway.

PolyMeilex: Done, replaced with seal support

src/wayland/seat/keyboard/keymap_file.rs
```diff
@@ -0,0 +1,101 @@
...
+        file.write_all(keymap)?;
+        file.flush()?;
+
+        file.seek(std::io::SeekFrom::Start(0))?;
+
+        nix::fcntl::fcntl(
+            file.as_raw_fd(),
+            FcntlArg::F_ADD_SEALS(
+                SealFlag::F_SEAL_SEAL
+                    | SealFlag::F_SEAL_SHRINK
```
i509VCB, line 84: Nix only appears to enable this on android and Linux.

Also we do need to test things compile on more than just Linux at some point lol

PolyMeilex: [FreeBSD manual](https://www.freebsd.org/cgi/man.cgi?query=fcntl&sektion=2) list those, so I guess it's time to report a bug in `nix`

src/wayland/seat/keyboard/keymap_file.rs
```diff
@@ -0,0 +1,101 @@
...
+            error!(log, "Error when creating sealed keymap file: {}", err);
+        }
+
+        Self {
+            sealed: sealed.ok(),
+            keymap,
+        }
+    }
+
+    pub fn with_fd<F>(&self, supports_seeled: bool, cb: F) -> Result<(), std::io::Error>
```
vberger, line 35:
```suggestion
    pub fn with_fd<F>(&self, supports_sealed: bool, cb: F) -> Result<(), std::io::Error>
```

vberger (approved):
LGTM, just one small typo.

src/wayland/seat/keyboard/keymap_file.rs
vberger: Yeah, this needs some followup updates in the file too :sweat_smile:

PolyMeilex: Done :sweat_smile:

PR #602, opened 2022-05-30
Update WlCompositor to version 5

PolyMeilex:
Ver 5: https://github.com/wayland-project/wayland/commit/fceabb7e9e167e1100c244bb4f460c8a346a5478

Offset in `wl_surface::Attach` got replaced by `wl_surface::Offset`

### DIFF
--- a/src/backend/renderer/utils.rs
+++ b/src/backend/renderer/utils.rs
@@ -164,9 +164,7 @@ impl RendererSurfaceState {
     ///
     /// Once delta is taken this method returns `None` to avoid processing it several times.
     pub fn take_accumulated_buffer_delta(&mut self) -> Point<i32, Logical> {
-        let accumulated_buffer_delta = self.accumulated_buffer_delta;
-        self.accumulated_buffer_delta = (0, 0).into();
-        accumulated_buffer_delta
+        std::mem::take(&mut self.accumulated_buffer_delta)
     }
 }
 
--- a/src/backend/renderer/utils.rs
+++ b/src/backend/renderer/utils.rs
@@ -4,7 +4,7 @@ use crate::{
     backend::renderer::{buffer_dimensions, Frame, ImportAll, Renderer},
     utils::{Buffer as BufferCoord, Logical, Point, Rectangle, Size, Transform},
     wayland::compositor::{
-        add_destruction_hook, is_sync_subsurface, with_surface_tree_upward, BufferAssignment, Damage,
+        self, add_destruction_hook, is_sync_subsurface, with_surface_tree_upward, BufferAssignment, Damage,
         SubsurfaceCachedState, SurfaceAttributes, SurfaceData, TraversalAction,
     },
 };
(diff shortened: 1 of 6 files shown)


src/wayland/compositor/handlers.rs
```diff
@@ -166,6 +166,14 @@ where
     ) {
         match request {
             wl_surface::Request::Attach { buffer, x, y } => {
+                if surface.version() >= 5 && (x != 0 || y != 0) {
+                    surface.post_error(
+                        handle,
+                        wl_surface::Error::InvalidOffset,
+                        "Passing non-zero x,y is protocol violation",
```
i509VCB, line 173: probably worth mentioning version 5 here

src/wayland/compositor/handlers.rs
```diff
@@ -258,6 +266,21 @@ where
...
+                PrivateSurfaceData::with_states(surface, |states| {
+                    match &mut states.cached_state.pending::<SurfaceAttributes>().buffer {
+                        Some(BufferAssignment::NewBuffer { delta, .. }) => {
+                            *delta = (x, y).into();
+                        }
+                        _ => {
+                            error!(
+                                state.compositor_state().log,
+                                "Got wl_surface::Offset request, but there is no pending buffer available."
+                            );
```
vberger, line 279: I'd say this kind `error!` logging should be reserved to problem in the compositor (or interaction with the OS), not clients misbehaving. I'd rather have this as a `debug!` loglevel

src/wayland/compositor/handlers.rs
PolyMeilex: Done

src/wayland/compositor/handlers.rs
PolyMeilex: Done

Drakulix: I don't see anything in the protocol, that requires `attach` to be called before `offset`. But because it is double-buffered, we have to handle that. I think setting an offset first, then attaching the buffer and assuming the offset will be used by the compositor after commit is perfectly legal.

PolyMeilex: Indeed,  https://github.com/Smithay/smithay/pull/602/commits/f5752848315067239f63ee2ce749400065b960b2 should address this, it reworks the handling of the offset around the new version, and retrofits pre 5 way.

i509VCB (approved):
I'm not familiar with the specifics in compositor, you'll probably want a second opinion.

src/backend/renderer/utils.rs
```diff
@@ -37,8 +37,9 @@ const MAX_DAMAGE: usize = 4;
 
 impl SurfaceState {
     pub fn update_buffer(&mut self, dh: &DisplayHandle, attrs: &mut SurfaceAttributes) {
+        let _delta = attrs.buffer_delta.take();
```
Drakulix, line 40: While I don't think this PR needs to use the offset everywhere to upgrade to version 5, leaving this unused here feels also not very useful. We should make it at least accessible to downstream.

What about adding a `buffer_delta` field to `SurfaceState` + some shorts docs?

Drakulix (approved):
Minor nitpick, otherwise LGTM!

src/backend/renderer/utils.rs
PolyMeilex: Done in: https://github.com/Smithay/smithay/pull/602/commits/00e45a9e7fb394328add121fe6ddbe48803362b4

vberger (approved):

src/backend/renderer/utils.rs
```diff
@@ -156,27 +157,16 @@ impl RendererSurfaceState {
...
-        if self.buffer_delta_ack {
-            None
-        } else {
-            self.buffer_delta_ack = true;
-            self.buffer_delta
-        }
+    pub fn take_accumulated_buffer_delta(&mut self) -> Point<i32, Logical> {
+        let accumulated_buffer_delta = self.accumulated_buffer_delta;
+        self.accumulated_buffer_delta = (0, 0).into();
+        accumulated_buffer_delta
```
Drakulix, line 169:
```suggestion
        std::mem::replace(&mut self.accumulated_buffer_delta, (0, 0).into())
```

Drakulix (approved):
LGTM!

src/backend/renderer/utils.rs
vberger: Actually, I think even `std::mem::take(&mut self.accumulated_buffer_delta)` would work, thanks to the `Default` impl of `Point`.

PolyMeilex: Done using `std::mem::take`

PR #603, opened 2022-06-03
allocator: Add depth to format table

Drakulix:


src/backend/allocator/format.rs
```diff
@@ -82,6 +83,18 @@ macro_rules! format_tables {
             }
         }
 
+        /// Returns the depth of the specified format.
```
i509VCB, line 86: what makes a depth different from the bpp?

i509VCB: Also let's mention this in the module docs

src/backend/allocator/format.rs
```diff
@@ -94,218 +107,238 @@ macro_rules! format_tables {
 
 format_tables! {
     // 8-bit bpp Red
-    R8 { alpha: false, bpp: 8 },
+    R8 { alpha: false, bpp: 8, depth: 8 },
 
     // TODO: Update drm-fourcc
     // 16-bit bpp Red with padding (x:R)
     // R10 { bpp: 16 },
     // R12 { bpp: 16 },
```
i509VCB, line 115: Let's add the depth to these also, we do need to update drm-fourcc to get these variants though

src/backend/allocator/format.rs
Drakulix: It is the amount of used bits per pixel. So any padding or non-alpha ("X") parts of the format are not counted towards the depth, but the bpp.

Older drm apis require that information to create a framebuffer.

Drakulix: Added to module docs

src/backend/allocator/format.rs
Drakulix: done

src/backend/allocator/format.rs
```diff
@@ -26,6 +26,16 @@
...
 //! assert_eq!(get_bpp(Fourcc::Argb8888), Some(32));
 //! ```
+//!
+//! [`get_depth`] returns the number of used bits per pixel of a format
+//! (excluding padding or non-alpha "X" parts of the format).
+//!
+//! ```
+//! # use smithay::backend::allocator::Fourcc;
+//! # use smithay::backend::allocator::format::get_depth;
+//! assert_eq!(get_depth(Fourcc::Argb8888), Some(32))
```
i509VCB, line 36: doc tests failed

vberger (approved):

PR #604, opened 2022-06-04
Allow `xdg_activation_token_v1::destroy` after token is constructed

ids1024:
Fixes protocol error starting `gtk4-demo`. I'm not sure if this could be handled better or if `destroy` should actually be doing anything.

### DIFF
--- a/src/wayland/xdg_activation/handlers.rs
+++ b/src/wayland/xdg_activation/handlers.rs
@@ -54,7 +54,7 @@ fn get_activation_token(
         let mut token_constructed = false;
 
         move |id, req, _| {
-            if !token_constructed {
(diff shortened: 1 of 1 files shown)


i509VCB (approved):
Fix could look a little nicer, but this will work.

(I've doubled checked and 0.30 already has this fixed)

PolyMeilex (approved):
>  or if destroy should actually be doing anything.

Our `assign_destructor()` block already does the cleanup, so there is no need to do anything beyond that. LGTM

ids1024:
> (I've doubled checked and 0.30 already has this fixed)

The `rework-wayland-rs-0.30` branch? Seems like `anvil` in that branch segfaults when `gtk4-demo` is started, so I can't compare this and the other issue I was looking into.

PR #606, opened 2022-06-04
0.30 General cleanup

PolyMeilex:
This is a:
- `TODO` blocks hunt
- removal of all `object_info().is_err()` based alive checks.
- drop of explicit sync module
- drop of xdg foreign module


Drakulix (approved):
:+1:

PR #609, opened 2022-06-07
gbm: Test formats in case binding fb fails

Drakulix:
This makes `GbmBufferedSurface::new` internally more flexible.

We prefer a format with alpha support, so that the GbmBufferedSurface can also be used for overlays and cursor planes with hardware blending.

However some hardware does not support alpha formats on primary planes (binding the framebuffer fails with *unsupported argument*), **although** it is returned as a supported plane format... We already handle the case, where the plane format is known to be unsupported by falling-back to `XRGB8888`. Possibly this only happens with certain modifier combinations, but no matter if this is buggy driver behaviour or not, we can be more defensive in testing options.

This commit essentially extends that fallback to all cases, where surface creation fails. This should fix newer intel gpus and presumably qemu drivers, as reported by @wash2 and @h1771t respectively. Please test, if this enables you to run anvil on your configurations.

While I personally have not seen any regressions on my hardware, I would welcome others to check, if anvil still launches for them. (Which is why I added a bunch of reviewers.)


i509VCB (approved):
No noticeable regressions on my hardware. All my hardware has worked correctly with the existing GbmBufferedSurface.

(6700xt on desktop)
(Intel 11th gen on laptop)

h1771t:
anvil from github.com/Smithay/smithay modesets for me in qemu even without this patch. Looks like reason is that anvil with --tty-udev uses the dmabuf allocator and not gbm which cosmic-comp uses?

I managed to try the patch against github.com/pop-os/smithay which cosmic-comp uses. But then then cosmic-comp modeset issue still persists for me, until I change the following:
*src/backend/drm/surface/atomic.rs* in 
```
impl<A: AsRawFd + 'static> AtomicDrmSurface<A> {
fn create_test_buffer(
```
Change `.add_framebuffer(&db, 32, 32)` to `.add_framebuffer(&db, 24, 32)`

I used to have to change this in both gbm.rs and atomic.rs, but now with this patch changing only atomic.rs gets the modeset working for me in qemu on cosmic-comp.

Drakulix:
> I managed to try the patch against github.com/pop-os/smithay which cosmic-comp uses. But then cosmic-comp modeset issue still persists for me, until I change the following: _src/backend/drm/surface/atomic.rs_ in
> 
> Change `.add_framebuffer(&db, 32, 32)` to `.add_framebuffer(&db, 24, 32)`
> 

Good catch, I was only updating the gbm code. #fd064ae should now also address this and I tried to make sure every other occurrence of `add_framebuffer` and `create_dumb_buffer` now also does not use any hardcoded values anymore.

Could you re-test @h1771t ?

h1771t:
Confirmed - applying the 3 patches to github.com/pop-os/smithay and launching cosmic-comp with kms backend successfully modesets in qemu for me. Thank you.

PR #610, opened 2022-06-07
allocator: Fix overflow

Drakulix:
Fixes a potential overflow inside the `Swapchain`, if the numbers of cycled slots is not consistent.
Currently if an allocated slot does not get used for a long time, it will overflow causing a panic.
Instead we should reset unused slots.

Fixes https://github.com/pop-os/cosmic-comp/issues/10

As far as I know this is not reproducable in anvil, but given cosmic-comp does a little more complex buffer management and I cannot see any obvious misuse of the api, this seems to be the correct fix. And securing the swapchain against an overflow is good practice anyway.

@PolyMeilex and @h1771t please test and confirm this fixes the problem on your systems.

### DIFF
--- a/src/backend/allocator/swapchain.rs
+++ b/src/backend/allocator/swapchain.rs
@@ -173,25 +173,28 @@ where
     /// Buffers can always just be safely discarded by dropping them, but not
     /// calling this function before may affect performance characteristics
     /// (e.g. by not tracking the buffer age).
-    pub fn submitted(&self, slot: &Slot<B>) {
+    pub fn submitted(&mut self, slot: &Slot<B>) {
         // don't mess up the state, if the user submitted and old buffer, after e.g. a resize
         if !self.slots.iter().any(|other| Arc::ptr_eq(&slot.0, other)) {
             return;
         }
 
         slot.0.age.store(1, Ordering::SeqCst);
-        for other_slot in &self.slots {
+        for other_slot in &mut self.slots {
             if !Arc::ptr_eq(other_slot, &slot.0) && other_slot.buffer.is_some() {
                 let res = other_slot
                     .age
                     .fetch_update(Ordering::SeqCst, Ordering::SeqCst, |age| {
                         if age > 0 {
-                            Some(age + 1)
+                            age.checked_add(1)
                         } else {
                             Some(0)
                         }
                     });
-                assert!(res.is_ok());
+                // If the age overflows the slot was not used for a long time. Lets clear it
+                if res.is_err() {
+                    *other_slot = Default::default();
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


h1771t:
Confirmed - this fixes the allocator overflow panic for me. Previously cosmic-comp would consistently panic for me with kms backend, with this patch cosmic-comp no longer panics.

What is the ideal way to test github patches? Given cosmic-comp depends on smithay from github.com/pop-os/smithay, I could not just edit cosmic-comp Cargo.toml and point it to github.com/Smithay/smithay with fix/allocator_overflow branch. I guess cosmic-comp depends on patches in github.com/pop-os/smithay that are not yet in github.com/Smithay/smithay. So the build would fail.

I ended up doing a standard cosmic-comp build, then getting a mail friendly patch from the pull using https://patch-diff.githubusercontent.com/raw/Smithay/smithay/pull/610.patch, applying it to directory ~/.cargo/git/checkouts/smithay-1fd533335f33e976/6b7db88, deleting files ~/cosmic-comp/target/debug/build/smithay-6*, then cargo build, and run.

Not ideal I know, what is the recommended way to be testing github pulls with cosmic-comp?

Drakulix:
> Confirmed - this fixes the allocator overflow panic for me. Previously cosmic-comp would consistently panic for me with kms backend, with this patch cosmic-comp no longer panics.
> 
> What is the ideal way to test github patches? Given cosmic-comp depends on smithay from github.com/pop-os/smithay, I could not just edit cosmic-comp Cargo.toml and point it to github.com/Smithay/smithay with fix/allocator_overflow branch. I guess cosmic-comp depends on patches in github.com/pop-os/smithay that are not yet in github.com/Smithay/smithay. So the build would fail.
> 
> I ended up doing a standard cosmic-comp build, then getting a mail friendly patch from the pull using https://patch-diff.githubusercontent.com/raw/Smithay/smithay/pull/610.patch, applying it to directory ~/.cargo/git/checkouts/smithay-1fd533335f33e976/6b7db88, deleting files ~/cosmic-comp/target/debug/build/smithay-6*, then cargo build, and run.
> 
> Not ideal I know, what is the recommended way to be testing github pulls with cosmic-comp?

Yeah that is a little annoying, I have to agree.

What I would have done is:
- Clone pop-os/smithay
- Cherry-pick this PR
- Change the patch in cosmic-comp's Cargo.toml to point to your local copy (using `path =` instead of `git =`

In the future I hope to get rid of pop-os/smithay again, but right now it carries primarily [this patch](https://github.com/Smithay/smithay/pull/558), which will have to be reworked into a new community crate (smithay-wlr), but likely only after the big 0.30-rework (https://github.com/Smithay/smithay/pull/454) is done... (As you can see, everything is in flux.)
And eventually (though that might take a lot of time) cosmic-comp might be able to depend on released stable versions of smithay. :sweat_smile:

h1771t:
I understand the flux at present. Thanks for the pointers on cleaner way to test, I will do that going forward.

vberger (approved):

PR #613, opened 2022-06-08
Update to wayland-rs beta 4

PolyMeilex:


i509VCB:
This todo could also be fixed now https://github.com/Smithay/smithay/blob/rework-wayland-rs-0.30/src/wayland/shm/mod.rs#L192

vberger (approved):

PR #614, opened 2022-06-09
The Vulkan allocator

i509VCB:
## Scope

First is a thin lightweight wrapper around ash to get an Instance and PhysicalDevices. This helper is kept intentionally lightweight to prevent essentially replicating a crate like vulkano. Creation of a logical `Device` is intentionally not covered since extensions to `DeviceCreateInfo` and queue creation are hard to model in a public API. The Vulkan allocator and a future renderer are expected to initialize their own logical device(s) and queues.

The Vulkan allocator is modeled around the `Allocator` trait. Implementing memory import/export and dmabuf import could be done, but it involves much more complex lifetimes (since memory import/export requires copy commands, which means command execution and needing to delay destruction of images until command execution finishes). In the future we could expand the allocator api to handle this as an alternative to gbm's memory import/export functions. This may be harder to do but will allow custom Vulkan renderers to be build more easily (since all the external memory logical will be in one type you can just pull in).

There are a few things we do need to do before merging:
- [x] ~~A new ash release (https://github.com/ash-rs/ash/pull/632). 0.37.1 will not eliminate the builders, so if we do not want to wait for 0.38.0 then we will need to readd the builders.~~ Will use 0.37.1 pre-release with a set revision for now.
- [x] Double check we do not have validation errors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,8 @@ members = [
 
 [dependencies]
 appendlist = "1.4"
+# Intentionally pick a commit from "0.37-stable" branch since additions for 0.37.1 are used
+ash = { git = "https://github.com/ash-rs/ash", rev = "62960ad680207e8496ae19ebd11f7c7ae8422e27", optional = true }
 bitflags = "1"
 calloop = "0.10.1"
 cgmath = "0.18.0"
@@ -42,6 +44,7 @@ libloading = { version="0.7.0", optional = true }
 nix = "0.22"
 once_cell = "1.8.0"
 rand = "0.8.4"
+scopeguard = { version = "1.1.0", optional = true }
 slog = "2"
 slog-stdlog = { version = "4", optional = true }
 tempfile = { version = "3.0", optional = true }
@@ -69,7 +72,7 @@ gl_generator = { version = "0.14", optional = true }
 pkg-config = { version = "0.3.17", optional = true }
 
 [features]
-default = ["backend_drm", "backend_gbm", "backend_libinput", "backend_udev", "backend_session_logind", "backend_x11", "backend_winit", "desktop", "renderer_gl", "renderer_multi", "xwayland", "wayland_frontend", "slog-stdlog"]
+default = ["backend_drm", "backend_gbm", "backend_libinput", "backend_udev", "backend_session_logind", "backend_x11", "backend_winit", "desktop", "renderer_gl", "renderer_multi", "xwayland", "wayland_frontend", "slog-stdlog", "backend_vulkan"]
 backend_winit = ["winit", "backend_egl", "wayland-egl", "renderer_gl"]
 backend_x11 = ["x11rb", "x11rb/dri3", "x11rb/xfixes", "x11rb/present", "x11rb_event_source", "backend_gbm", "backend_drm", "backend_egl"]
 backend_drm = ["drm", "drm-ffi"]
@@ -78,6 +81,7 @@ backend_egl = ["gl_generator", "libloading"]
 backend_libinput = ["input"]
 backend_session = []
 backend_udev = ["udev", "input/udev"]
+backend_vulkan = ["ash", "scopeguard"]
 backend_session_logind = ["dbus", "backend_session", "pkg-config"]
 backend_session_elogind = ["backend_session_logind"]
 backend_session_libseat = ["backend_session", "libseat"]
@@ -105,3 +109,7 @@ required-features = ["wayland_frontend"]
 [[example]]
 name = "compositor"
 required-features = ["wayland_frontend"]
+
+[[example]]
+name = "vulkan"
+required-features = ["backend_vulkan"]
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -78,6 +83,7 @@ backend_egl = ["gl_generator", "libloading"]
 backend_libinput = ["input"]
 backend_session = []
 backend_udev = ["udev", "input/udev"]
+backend_vulkan = ["backend_drm", "ash", "scopeguard"]
```
Drakulix, line 86: I assume a `renderer_vulkan` would not need `backend_drm`, no? Does `backend_vulkan` strictly need it?

i509VCB: backend_drm is a result of DrmNode being used. Unless we wanted to split that out into another crate or feature there is no real way around it.

src/backend/allocator/vulkan/mod.rs
```diff
@@ -0,0 +1,770 @@
...
+        modifiers: &[DrmModifier],
+    ) -> Result<VulkanImage, Self::Error> {
+        self.create_buffer_with_usage(
+            width,
+            height,
+            fourcc,
+            modifiers,
+            // More than likely images from the allocator will be sampled or used with color attachments when
+            // exported.
+            ImageUsageFlags::SAMPLED | ImageUsageFlags::COLOR_ATTACHMENT,
```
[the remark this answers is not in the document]
i509VCB: > Do we need to account for what other implementation might do to the buffer after importing them?

Yes.

The relevant section of the Vulkan specification (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceImageFormatProperties.html) for determining if an image format is supported states the following:

> If format is not a supported image format, or if the combination of format, type, tiling, usage, and flags is not supported for images, then vkGetPhysicalDeviceImageFormatProperties returns VK_ERROR_FORMAT_NOT_SUPPORTED.

Now wait aren't dmabufs just memory we assign meaning to? Yes.

In Vulkan we rely on the image handle we create to provide us information about what type of memory allocation is needed and then bind the memory to the image. With the drm tiling extension we use in the allocator, the drm tiling could of course change what type of memory the implementation wants to allocate.

Declaring this ahead of time is important for the dmabuf other graphics apis may import as well. Clients and the compositor if they run Vulkan may use the same function above.

If the compositor needs more restrictive usage flags than clients (since it uses client images as textures), then the following applies:
> The limitations on an image format that are reported by vkGetPhysicalDeviceImageFormatProperties have the following property: if usage1 and usage2 of type [VkImageUsageFlags](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageUsageFlags.html) are such that the bits set in usage1 are a subset of the bits set in usage2, and flags1 and flags2 of type [VkImageCreateFlags](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageCreateFlags.html) are such that the bits set in flags1 are a subset of the bits set in flags2, then the limitations for usage1 and flags1 must be no more strict than the limitations for usage2 and flags2, for all values of format, type, and tiling.

TLDR: Allocating an image with less restrictive usage flags may result in the image not being importable with more restrictive usages.

You do have a point with specifying the default usage flags during construction, which may allow the implementation to request the least restrictive usage flags it needs (which would possibly provide more usable formats).

src/backend/vulkan/mod.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// * The Vulkan implementation is not at least 1.2
+    /// * If the Vulkan implementation is not at least Vulkan 1.2, the `VK_KHR_driver_properties` device
+    ///   extension is not available.
+    pub fn driver(&self) -> Option<&DriverInfo> {
+        self.info.driver.as_ref()
+    }
+
+    /// Returns the major and minor numbers of the primary node which corresponds to this physical device's DRM
+    /// device.
+    pub fn primary_node(&self) -> Option<DrmNode> {
```
Drakulix, line 521: Couldn't we just guard these functions with `#[cfg(feature = "backend_drm")]`?

Cargo.toml
```diff
@@ -58,6 +58,11 @@ x11rb = { version = "0.9.0", optional = true }
 xkbcommon = "0.4.0"
 scan_fmt = { version = "0.2.3", default-features = false }
 
+# Vulkan dependencies
+# TODO: Pending new release
+scopeguard = { version = "1.1.0", optional = true }
+ash = { git = "https://github.com/ash-rs/ash", optional = true }
```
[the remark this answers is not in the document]
i509VCB: The ash release with lifetimes instead of builders will not come for some time.

For now I've chosen a revision that will be part of the planned `0.37.1` release of ash and made the changes to the codebase to reflect that.

Cargo.toml
```diff
@@ -24,6 +24,8 @@ members = [
 
 [dependencies]
 appendlist = "1.4"
+# Intentionally pick one commit back from the "release 0.37.1" branch since the release commit is likely to be rebased
+ash = { git = "https://github.com/ash-rs/ash", rev = "28a4868dc302013e00569dc5393f8a0b43ed4a70", optional = true }
```
MarijnS95, line 28: Why not pick a commit from the `0.37-stable` branch, which intentionally exists as a sort of "master" branch but for "backporting" non-breaking changes?

`0.37.1-release~` is also part of that branch, all you need to do is refine the comment.

PR #620, opened 2022-06-09
0.30: wlcs_anvil

Drakulix:
Based on #619 to check if CI is happy


wlcs_anvil/src/lib.rs
```diff
@@ -27,7 +27,7 @@ static SUPPORTED_EXTENSIONS: &[WlcsExtensionDescriptor] = extension_list!(
     ("wl_compositor", 4),
     ("wl_subcompositor", 1),
     ("wl_data_device_manager", 3),
-    ("wl_seat", 5),
+    ("wl_seat", 7),
     ("wl_output", 3),
```
vberger, line 31: I guess we can bump it to 4 now?

PR #623, opened 2022-06-11
0.30 Port tablet protocol

PolyMeilex:


Drakulix (commented):
Looks good, but clippy is complaining

PR #625, opened 2022-06-13
Fix mouse qemu

h1771t:
pointer absolute input events for anvil in udev mode to allow mouse to work in qemu.


anvil/src/input_handler.rs
```diff
@@ -586,6 +587,40 @@ impl AnvilState<UdevData> {
...
+        let serial = SCOUNTER.next_serial();
+
+        let max_x = self
+            .space
+            .outputs()
+            .fold(0, |acc, o| acc + self.space.output_geometry(o).unwrap().size.w);
+        let max_y = self
+            .space
+            .outputs()
+            .fold(0, |acc, o| acc + self.space.output_geometry(o).unwrap().size.h);
```
Drakulix, line 600: This is not entirely correct. Outputs in smithay only extend to the right, so the maximum-y coordinate is the largest monitor, but not the sum of all.

h1771t:
oops soz this was a mistake.

PR #626, opened 2022-06-13
0.30: Break `Request`-enums into trait methods

Drakulix:


src/wayland/shell/xdg/mod.rs
```diff
@@ -730,12 +731,130 @@ impl Cacheable for SurfaceCachedState {
...
+    /// - `positioner` - The state of the positioner at the time the reposition request was made.
+    ///     The positioner state can be used by the compositor
+    ///     to calculate the best placement for the popup.
+    ///     For example the compositor should prevent that a popup
+    ///     is placed outside the visible rectangle of a output.
+    /// - `token` - The passed token will be sent in the corresponding xdg_popup.repositioned event.
+    ///     The new popup position will not take effect until the corresponding configure event
+    ///     is acknowledged by the client. See xdg_popup.repositioned for details.
+    ///     The token itself is opaque, and has no other special meaning.
+    fn reposition_request(
```
i509VCB, line 850: `_request` suffix is redundant imo.

src/wayland/shell/xdg/mod.rs
```diff
@@ -730,12 +731,130 @@ impl Cacheable for SurfaceCachedState {
...
 
-    /// Xdg requests
-    fn request(&mut self, dh: &DisplayHandle, request: XdgRequest);
+    /// A new shell client was instantiated
+    fn new_client(&mut self, dh: &wayland_server::DisplayHandle, client: ShellClient) {}
+    /// The pong for a pending ping of this shell client was received
+    ///
+    /// The `ShellHandler` already checked for you that the serial matches the one
+    /// from the pending ping.
+    fn client_pong(&mut self, dh: &wayland_server::DisplayHandle, client: ShellClient) {}
```
i509VCB, line 745: Style nit: can we have an space between each function?

src/wayland/shell/xdg/mod.rs
Drakulix: yeah, but it communicates the intent and `move` is not a valid method name.

src/wayland/shell/xdg/mod.rs
i509VCB: On realization this is a bad spot to place the comment...

src/wayland/shell/xdg/mod.rs
i509VCB: move can be done as `r#move` or `move_`.

Under the reasoning you provided, you want to communicate to compositors that listening to those events is optional?

Drakulix: > Under the reasoning you provided, you want to communicate to compositors that listening to those events is optional?

yes

Drakulix: That depends on your perspective on the functions imo.

`request_reposition` makes sense from the viewpoint of the xdg-surface. It **requests a reposition**.
`reposition_request` makes sense from the viewpoint of a downstream compositor. It receives a **reposition request**.

I would argue the latter is more in line of what users of smithay would expect, but I am fine either way.

vberger (approved):

PR #632, opened 2022-06-14
Fix mouse in qemu when running anvil in --tty-udev mode

h1771t:
pointer absolute event in udev mode, clamp height to max height from outputs.

Gets mouse working in qemu, when running anvil in --tty-udev mode


h1771t:
am I doing pull req wrong? why is it showing 121 files changed when I changed 1 file. My change was already based off the branch I'm requesting pull for, so changes should be minimal?

h1771t:
Let me close this and try again. Sorry.

PolyMeilex:
> am I doing pull req wrong? why is it showing 121 files changed when I changed 1 file. My change was already based off the branch I'm requesting pull for, so changes should be minimal?

If you are trying to merge something to rework branch, you have to be based on rework branch as well.
If you are merging to master, you have to be based on master.
The same applies the other way around.

Also, there is no need to reopen the PR every time, you can just push or force push to already open one, it will be easier to track that way. If you don't know how then nevermind, that's fine too.

PR #633, opened 2022-06-14
pointer absolute events for anvil udev mode.

h1771t:
allows mouse to work in qemu when running anvil in --tty-udev mode


h1771t:
interesting that https://github.com/Smithay/smithay/actions/runs/2495605126 failed to compile, the code does compile for me. Need to figure out why or what is different on the ci test to what I do locally then.

PolyMeilex:
There is a feature guard on pointer absolute event trait import/use statement. You have to move it so it is compiled even if x11 and winit features are disabled

h1771t:
thank you for pointing out the issue.

If I change line 28 from:
`#[cfg(any(feature = "winit", feature = "x11"))]`
to
`#[cfg(any(feature = "winit", feature = "x11", feature="udev"))]`

Would that suffice?

vberger:
> How would I use cargo locally to build with only the features that the ci tests are building with? I am new with rust and cargo.

From inside the anvil directory, you can run commands like `cargo build --features "x11 udev"`

PR #642, opened 2022-06-20
[WIP] CI: expand WLCS tests

vberger:


src/wayland/compositor/handlers.rs
```diff
@@ -168,6 +174,10 @@ where
     ) {
         match request {
             wl_surface::Request::Attach { buffer, x, y } => {
+                if !crate::wayland::shell::xdg::ensure_configured_if_xdg(surface) {
```
i509VCB, line 177: Layer shell also has this. I'd prefer something generic since a stablized layer shell will result in three role objects with these semantics.

PR #645, opened 2022-06-22
dmabuf: Make `DmabufGlobal` Clone & Copy

Drakulix:


src/wayland/dmabuf/mod.rs
```diff
@@ -244,10 +245,9 @@ pub struct DmabufParamsData {
 /// A handle to a registered dmabuf global.
 ///
 /// This type may be used in equitability checks to determine which global a dmabuf is being imported to.
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
```
i509VCB, line 248: nit: what ordering do derive traits generally appear in?

i509VCB (approved):
An open question: what happens if we disable globals more than once?

src/wayland/dmabuf/mod.rs
Drakulix: 🤷‍♀️

PR #652, opened 2022-06-27
udev: Handle absolute pointer devices

Drakulix:
So I merged #633 accidentally into the `rework-wayland-rs-0.30` branch instead of master.
So here is a new PR. Sorry @h1771t !


h1771t:
This is my fault not yours, I based the PR on `rework-wayland-rs-0.30` not on `master`. I should be basing any pull requests I do on `master`. Apologies on my part.

Drakulix:
@PolyMeilex  @i509VCB @vberger  I already reviewed this as part of the original PR, can someone quickly glance over it and make sure nothing is obviously wrong?

i509VCB:
Nothing looks wrong, I'm not really the input expert though, so Poly might know better than me.

anvil/src/input_handler.rs
```diff
@@ -586,6 +590,48 @@ impl AnvilState<UdevData> {
...
+            .max_by_key(|o| self.space.output_geometry(o).unwrap().size.h)
+            .unwrap();
+
+        let max_y = self.space.output_geometry(max_h_output).unwrap().size.h;
+
+        self.pointer_location.x = evt.x_transformed(max_x);
+        self.pointer_location.y = evt.y_transformed(max_y);
+
+        // clamp to screen limits
+        // this event is never generated by winit
```
PolyMeilex, line 617: I don't get this comment, absolute move events are produced by winit as far as I remember, so I guess that this refers to the fact that winit does not require clamping?

PolyMeilex (approved):

PolyMeilex:
> This is my fault not yours, I based the PR on `rework-wayland-rs-0.30` not on `master`. I should be basing any pull requests I do on `master`. Apologies on my part.

I don't think your fault is so obvious, rework branch used to be our main development branch at the time, so it was probably a good target, it should just been rebased to master once all of the 0.30 rework shenanigans were done, and master became our target branch again, but yeah we missed it, not a big deal.

anvil/src/input_handler.rs
h1771t: the comment was copied from code that already existed: https://github.com/Smithay/smithay/blob/670e8a3ab19103df0428db91d30f757e4262fa3c/anvil/src/input_handler.rs#L574

Since I'm calling the same function in new code which was called in the old code, I thought the comment remained relevant? I'm not really sure, I could remove this line https://github.com/Smithay/smithay/blob/670e8a3ab19103df0428db91d30f757e4262fa3c/anvil/src/input_handler.rs#L617 and leave the `clamp to screen limits` comment if that fits better?

Drakulix: Yeah I think that would be better.

Given you can't push to this branch, I'll quickly do that.

PR #659, opened 2022-07-06
Feature/opaque regions

cmeissl:
This PR adds support for opaque regions in the renderer and extends the space to skip damage when applicable


src/backend/renderer/mod.rs
```diff
@@ -602,6 +602,217 @@ pub fn buffer_type(buffer: &wl_buffer::WlBuffer) -> Option<BufferType> {
...
+                | drm_fourcc::DrmFourcc::Yvu411
+                | drm_fourcc::DrmFourcc::Yvu420
+                | drm_fourcc::DrmFourcc::Yvu422
+                | drm_fourcc::DrmFourcc::Yvu444
+                | drm_fourcc::DrmFourcc::Yvyu
+        ));
+    }
+
+    if let Ok(format) = crate::wayland::shm::with_buffer_contents(buffer, |_, metadata| metadata.format) {
+        return Some(!matches!(
```
Drakulix, line 702: And this can maybe be just moved into `src/wayland/shm`

src/backend/renderer/mod.rs
```diff
@@ -602,6 +602,217 @@ pub fn buffer_type(buffer: &wl_buffer::WlBuffer) -> Option<BufferType> {
...
+        .map(|x| x.format.clone())
+    {
+        return Some(!matches!(
+            format,
+            super::egl::Format::RGB
+                | super::egl::Format::Y_UV
+                | super::egl::Format::Y_U_V
+                | super::egl::Format::Y_XUXV
+        ));
+    }
```
Drakulix, line 811: And I guess this is short enough, that it could stay. But potentially this could also be a `has_alpha` function of the `EGLBufferReader`.

src/backend/renderer/utils.rs
```diff
@@ -112,6 +116,66 @@ impl RendererSurfaceState {
...
+                            let dest_size = self.surface_view.unwrap().dst;
+
+                            let rect_constrained_loc = rect
+                                .loc
+                                .constrain(Rectangle::from_extemities((0, 0), dest_size.to_point()));
+                            let rect_clamped_size = rect
+                                .size
+                                .clamp((0, 0), (dest_size.to_point() - rect_constrained_loc).to_size());
+
+                            let rect = Rectangle::from_loc_and_size(rect_constrained_loc, rect_clamped_size);
```
Drakulix, line 140: But, if the rendering of a buffer from it's `src` to `dst` is scaled, shouldn't the opaque_region here also be scaled rather then cut of?

Drakulix (commented):
Looks good to me except for the missing docs!

src/backend/renderer/mod.rs
```diff
@@ -602,6 +602,217 @@ pub fn buffer_type(buffer: &wl_buffer::WlBuffer) -> Option<BufferType> {
...
 
+/// Returns if the buffer has an alpha channel
+///
+/// Returns `None` if the type is not known to smithay
+/// or otherwise not supported (e.g. not initialized using one of smithays [`crate::wayland`]-handlers).
+///
+/// Note: This is on a best-effort, but will never return false for a buffer
+/// with a format that supports alpha.
+#[cfg(feature = "wayland_frontend")]
+pub fn buffer_has_alpha(buffer: &wl_buffer::WlBuffer) -> Option<bool> {
```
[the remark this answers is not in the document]
cmeissl: sure, I somehow expected that comment 😅

src/backend/renderer/utils.rs
cmeissl: I am not 100% sure, but I think the client should submit the correct size already.

The opaque region is defined as:
> The opaque region is specified in surface-local coordinates.

So it *should* be the same as the logical surface damage.

src/backend/renderer/mod.rs
```diff
@@ -602,6 +602,217 @@ pub fn buffer_type(buffer: &wl_buffer::WlBuffer) -> Option<BufferType> {
...
+    if let Ok(dmabuf) = crate::wayland::dmabuf::get_dmabuf(buffer) {
+        return Some(!matches!(
```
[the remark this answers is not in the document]
cmeissl: Good point, replaced with a call to `format::has_alpha`

src/backend/renderer/mod.rs
cmeissl: Moved

src/backend/renderer/mod.rs
cmeissl: Added `egl_buffer_has_alpha` to the reader

Drakulix (approved):

PR #661, opened 2022-07-07
layer_shell: Add new_popup callback

Drakulix:
Currently there is no event generated by the [zwlr_layer_surface_v1::get_popup](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:get_popup) request.

This is an issue when trying to implement positioning logic for popup with layershell surfaces as parent.
On the [xdg_surface::get_popup](https://wayland.app/protocols/xdg-shell#xdg_surface:request:get_popup) call they don't have yet a parent, making it impossible to know, which location they should be positioned relative to.

On the first commit, it is already too late, given the xdg-shell protocol mandates that configure-events are only send at very specific times:
> For version 2 or older, the configure event for an xdg_popup is only ever sent once for the initial configuration. Starting with version 3, it may be sent again if the popup is setup with an xdg_positioner with set_reactive requested, or in response to xdg_popup.reposition requests.

Thus every non-xdg-shell, that supports popups, like layer-shell, needs it's own `new_popup`-callback for downstream to be able to implement positioning logic. This PR adds such a callback to the `WlrLayerShellHandler`.

### DIFF
--- a/src/wayland/shell/wlr_layer/handlers.rs
+++ b/src/wayland/shell/wlr_layer/handlers.rs
@@ -268,6 +268,13 @@ where
                         .unwrap()
                         .parent = Some(parent_surface);
                 });
+
+                WlrLayerShellHandler::new_popup(
+                    state,
+                    dh,
+                    make_surface_handle(layer_surface),
+                    crate::wayland::shell::xdg::handlers::make_popup_handle(&popup),
+                );
             }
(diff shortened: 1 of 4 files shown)


PolyMeilex (approved):
Indeed, I did not consider it when implementing layer shell, LGTM

PR #662, opened 2022-07-07
Compositor driven window crop and scale

cmeissl:
This PR adds compositor driven window crop and scale.

---
Todo:
- [ ] popup position has issues
- [X] ~~popup input is not transformed (where can the `InputTransformed` be assigned?)~~
      For now `Window::refresh` also updates the popup transforms and `PopupManager::commit`
      will re-extract the transform and set it to not miss a transform
- [ ] more testing


cmeissl:
most of this PR has now been implemented in #831

PR #668, opened 2022-07-12
egl: expose pub(crate) fields as public functions

i509VCB:
This makes some aspects of the EGL backend accessible from outside Smithay that may useful. Some small changes are also made like renaming `get_extensions` to `extensions`

### DIFF
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -69,11 +69,11 @@ impl Drop for EGLDisplayHandle {
 /// [`EGLDisplay`] represents an initialised EGL environment
 #[derive(Debug, Clone)]
 pub struct EGLDisplay {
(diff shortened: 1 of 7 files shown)


src/backend/egl/display.rs
```diff
@@ -405,8 +405,18 @@ impl EGLDisplay {
     }
 
     /// Returns the supported extensions of this display
-    pub fn get_extensions(&self) -> Vec<String> {
-        self.extensions.clone()
+    pub fn extensions(&self) -> &Vec<String> {
```
Drakulix, line 408: at this point why not return a slice?

i509VCB: Yeah a slice is probably better

Drakulix (approved):

PR #490, opened 2022-02-01
wayland.seat: Fix pointer grab API

vberger:
The pointer grab mecanism had a few inconsistencies in its behavior when
a grab is set or unset. THis commit uniformizes the behavior, and also
introduces a synthetic motion event when the grab is set, allowing the
newly set grab to change the focus or the pointer location if needed.

Also adjust the DnDGrab, as well as the resize and move grabs from anvil
to unset the focus while they are active, matching weston's behavior.

PR #671, opened 2022-07-19
Fix grab motion events

Drakulix:
This is basically https://github.com/Smithay/smithay/pull/490/ for 0.30...

It reintroduces `pointer.motion` calls previously missing from the grabs and also makes sure to take focus away, where appropriate.

The second commit removes `state.pointer_location` from smallvil so we catch this problem more easily in the future.
I would have given anvil the same treatment, but the cursor state and location is here much more intertwined with the rendering code and tablet code. The latter probably requires some more changes to tablet_tool at first to make it internally track its location as well.


PolyMeilex (approved):
Looks good to me

PR #672, opened 2022-07-19
drm/atomic: Re-evaluate properties for newly added connectors

Drakulix:
Apparently the drm-interface is allowed to dynamically add connectors during runtime, for example to represent daisy-chained display port connections and a few other special cases.

smithay currently does not handle this correctly in the atomic backend, as it assumes a static set of resources to cache their properties.

The first commit fixes the enumeration of properties for framebuffers. They are created by processes and inherently dynamic in nature. It was luckily never used in the code, but that piece of code makes no sense and shouldn't exist. Properties of framebuffers need to be queried from the drm-interface, when they are needed and should not be cached (at least not globally).

The second commit mostly shifts methods around to make them more accessible, the big change is that `prop_mapping` is now also encapsulated by a ReadWrite-lock. This is a little unfortunate given that `build_request` needs to access it and that this function is in the hot-path of both `commit` and `page_flip`. Extra care has been taken to make sure locks are only taken when needed (especially write locks are very few and far between) to hopefully make the performance costs negligible.

If proven necessary the read-write lock can be replaced by an append-only HashMap implementation such as provided by [`elsa`](crates.io/crates/elsa), but the drm-surface already contains quite a few locks and given the recent performance optimizations around these in stable rust I can hardly imagine this being an actual problem until proven otherwise by a profiler.

~Marked as Draft until actually tested on real hardware, but reviews are already very welcome.~

### DIFF
--- a/src/backend/drm/device/atomic.rs
+++ b/src/backend/drm/device/atomic.rs
@@ -26,7 +26,6 @@ type OldState = (
 pub type Mapping = (
     HashMap<connector::Handle, HashMap<String, property::Handle>>,
     HashMap<crtc::Handle, HashMap<String, property::Handle>>,
-    HashMap<framebuffer::Handle, HashMap<String, property::Handle>>,
     HashMap<plane::Handle, HashMap<String, property::Handle>>,
 );
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


Drakulix:
Can't seem to find any regression with anvil, I think this is good to go.

PolyMeilex (approved):
I can't really judge the drm correctness, but everything looks sane.

PR #682, opened 2022-08-01
vulkan: remove Vulkan1X features/properties from PhysicalDevice

i509VCB:
The Vulkan1X types are not ideal for our level of abstraction nor too idomatic. Smithay intends to support higher versions of Vulkan like a superset of existing functionality. However the Vulkan1X types cannot be chained with their individual parts without validation errors. There is also the fact that these types are not available until Vulkan 1.2, and the Vulkan11 structures are actually provided in Vulkan 1.2, not Vulkan 1.1. Therefore for the sake of cleaning up the codebase, the Vulkan1X types now need to be obtained some other way if desired.

### DIFF
--- a/src/backend/vulkan/mod.rs
+++ b/src/backend/vulkan/mod.rs
@@ -74,7 +74,7 @@ use std::{
 use ash::{
     extensions::ext::DebugUtils,
     prelude::VkResult,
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
Seems logical

PR #683, opened 2022-08-01
vulkan-alloc: fix validation error related to empty usage flags

i509VCB:
Empty image usage flags are not allowed in Vulkan. This pull request changes VulkanAllocator to panic if it's default usage flags are empty.

VulkanAllocator::is_format_supported and VulkanAllocator::create_buffer_with_usage now return false and fail respectively if the usage flags are empty.

### DIFF
--- a/src/backend/allocator/vulkan/mod.rs
+++ b/src/backend/allocator/vulkan/mod.rs
@@ -110,6 +110,7 @@ pub enum Error {
     /// 1. There is no equivalent Vulkan format for the specified format code.
     /// 2. The driver does not support any of the specified modifiers for the format code.
     /// 3. The driver does support some of the format, but the size of the buffer requested is too large.
+    /// 4. The usage is empty
     #[error("format is not supported")]
     UnsupportedFormat,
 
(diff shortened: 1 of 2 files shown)


src/backend/allocator/vulkan/mod.rs
```diff
@@ -184,13 +185,20 @@ impl VulkanAllocator {
...
     pub fn new(phd: &PhysicalDevice, default_usage: ImageUsageFlags) -> Result<VulkanAllocator, Error> {
         // Panic if the instance version is too high
         if phd.instance().api_version() > Self::MAX_INSTANCE_VERSION {
             panic!("Exceeded maximum instance api version for VulkanAllocator (1.3 max)")
         }
 
+        // VUID-VkPhysicalDeviceImageFormatInfo2-usage-requiredbitmask
+        // At least one image usage flag must be specified.
+        if default_usage.is_empty() {
+            panic!("Default usage flags for allocator are empty")
```
Drakulix, line 199: I mean this error is likely fatal somewhere higher up, but couldn't we just return an error in this case? What is the rationale behind panic'ing here?

i509VCB: My rationale was that generally you'd know ahead of time what default usages you would need to support.

Drakulix: Yeah that seems reasonable

Drakulix (approved):

PR #685, opened 2022-08-02
drm: fix test buffer drop for use_plane

cmeissl:
`use_plane` does not work because the created test buffer for the plane implicitely drops the test buffer of the primary plane which results in an `InvalidArgument` error during the commit.

This PR moves the test buffer out of the mutex into a `TestBuffer` struct that will clean up the test buffer on `Drop`.
Additionally it fixes a potential memory leak in `create_test_buffer` where the `DumbBuffer` would not be cleaned up
if `add_framebuffer` fails.

### DIFF
--- a/src/backend/drm/surface/atomic.rs
+++ b/src/backend/drm/surface/atomic.rs
@@ -123,7 +123,6 @@ pub struct AtomicDrmSurface<A: AsRawFd + 'static> {
     prop_mapping: RwLock<Mapping>,
     state: RwLock<State>,
     pending: RwLock<State>,
-    test_buffer: Mutex<Option<(DumbBuffer, framebuffer::Handle)>>,
     pub(crate) logger: ::slog::Logger,
 }
 
@@ -170,7 +169,6 @@ impl<A: AsRawFd + 'static> AtomicDrmSurface<A> {
             prop_mapping: RwLock::new(prop_mapping),
             state: RwLock::new(state),
             pending: RwLock::new(pending),
-            test_buffer: Mutex::new(None),
             logger,
         };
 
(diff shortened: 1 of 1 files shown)


Drakulix (commented):
So this would mean test-buffers are not cached and re-used anymore, correct?

Right now we are only testing commits for configuration changes (and not on every page-flip, if I am not mistaken), so the performance hit would be negligible in this case.

But once we start using overlay planes, I am not sure, that we won't do a a lot tests to figure out which configuration is supported. Do we maybe want to keep the cost lower for this case?

**Note:** This is a total open question, I am not suggesting that it needs to be cached, I have never benchmarked this.

cmeissl:
> So this would mean test-buffers are not cached and re-used anymore, correct?
> 
> Right now we are only testing commits for configuration changes (and not on every page-flip, if I am not mistaken), so the performance hit would be negligible in this case.
> 
> But once we start using overlay planes, I am not sure, that we won't do a a lot tests to figure out which configuration is supported. Do we maybe want to keep the cost lower for this case?
> 
> **Note:** This is a total open question, I am not suggesting that it needs to be cached, I have never benchmarked this.

Regarding the caching I am unsure what you mean. Before the change `create_test_buffer` was called in the same places and always created a new buffer, it just destroyed the existing buffer in the mutex after creating a new one. So the behavior should not be altered by these changes. (Please correct me if I am wrong here, maybe I missed some code part :smile:)

Actually I think the current api will not be sufficient for the overlay planes, but I can not tell for sure. So I would just like to fix the current bug which prevents using `use_plane` at all and then continue to find a better way to deal with planes when I have a PoC running which actually uses the planes (instead of the current mess I have done to always assign the top most surface to a plane :grimacing:)

Drakulix:
I yeah, the mutex was supposed to cache it, but right, it simply doesn't...
And performance is fine, so.. lets merge this.

Drakulix (approved):

PR #686, opened 2022-08-03
MultiGPU fixes round three

Drakulix:
Firefox strikes at it again...

First three commits fix code in the multigpu renderer, mostly related to usage of early_import.
- 38f3cf5480df7fc367d691642c4e0735f9bf6ca9 Adds a new check to make sure early_import does nothing if there was no new buffer. Otherwise it might clear the *internal* textures and we have nothing to render, because the `on_commit_buffer_handler` still has the empty `MultiTexture` and won't call `import_surface_tree`.
- 7e6473f1ca892586974ed1a522b95f00888fd37b Makes the MultiTexture reset it's internal textures, when the size changes. Otherwise the multigpu renderer could end up trying to update a texture with different dimensions, resulting in rendering errors
- 26454a0cccdd21bae9c52f5e356fd111b9cbdaf9 Clears the mappings on consequent imports (before they are actually imported), that overlap. Newer imports thus replace old mappings that are contained in the new ones. Otherwise unnecessary (and expensive) imports might happen and we can run into some assertions.
-  6e917d7c5ef78f6fa07a43c7069e32242039b6b8 Lets `import_dmabuf_internal` propagate more errors, so they don't are completely ignored.
- 0257d85c1cf10dcf34c49361a825fe563044d302 Adds a bunch of new logging statements I found helpful to fix the above errors.

### DIFF
--- a/src/backend/renderer/multigpu/mod.rs
+++ b/src/backend/renderer/multigpu/mod.rs
@@ -1153,7 +1153,18 @@ impl MultiTexture {
             .filter(|(old_src, _)| *old_src == source)
             .map(|(_, mappings)| mappings)
             .unwrap_or_default();
-        mappings.extend(new_mappings.map(|(r, m)| (r, Box::new(m) as Box<_>)));
+
+        // don't keep old mappings that are superseeded by new ones
+        let new_mappings = new_mappings
+            .map(|(r, m)| (r, Box::new(m) as Box<_>))
+            .collect::<Vec<_>>();
+        mappings.retain(|(region, _)| {
+            !new_mappings
+                .iter()
+                .any(|(new_region, _)| new_region.contains_rect(*region))
+        });
+        mappings.extend(new_mappings);
+
         textures.insert(
             render,
             GpuSingleTexture {
--- a/src/backend/renderer/multigpu/mod.rs
+++ b/src/backend/renderer/multigpu/mod.rs
@@ -1109,6 +1109,13 @@ impl MultiTexture {
         <<A::Device as ApiDevice>::Renderer as Renderer>::TextureId: 'static,
     {
         let mut tex = self.0.borrow_mut();
+        slog::trace!(
+            crate::slog_or_fallback(None),
+            "Inserting into: {:p} for {:?}: {:?}",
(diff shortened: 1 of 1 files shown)


src/backend/renderer/multigpu/mod.rs
```diff
@@ -360,53 +360,38 @@ impl<A: GraphicsApi> GpuManager<A> {
...
-                            });
-
-                        if let Err(err) =
-                            self.early_import_buffer(source, target, buffer, states, &*buffer_damage)
-                        {
-                            result = Err(err);
+                    if data.textures.is_empty() {
+                        // Import a new buffer if available
+                        if let Some(buffer) = data.buffer.as_ref() {
+                            let buffer_damage = data.damage.iter().take(1).flatten().cloned().fold(
```
i509VCB, line 366: What's the reason for using `take` here? The entire fold for damage will only be run once since `take(1)` will only yield one element?

Drakulix: Yes correct. The early_import cannot know, if we have rendered the last commit. So it may very well be, that we require more damage (possibly also depending on the output/space it is rendered on), but we try to optimistically upload parts of the texture as an optimization. So doing a very defensive import here of everything we can upload kinda defeats the purpose.

If it ends up being to few parts copied, the multigpu-code will import the rest (see call to `import_missing`), but most of the time we render in lock-step with the application, given that hardware accelerated apps usually wait for frame-events. In these cases this should be sufficient, so we really want only one frame of damage.

i509VCB: I see, might be worth documenting that behavior as those semantics aren't really described there.

Drakulix: Well yeah, the whole renderer is probably lacking in terms of docs. but sure. I will add this bit, better to start somewhere I suppose.

i509VCB (approved):

src/backend/renderer/multigpu/mod.rs
Drakulix: done

PR #688, opened 2022-08-04
anvil: Fix crash on scroll

Drakulix:
Fixes problem caused by #666 :imp:


PolyMeilex (approved):
Looks straight forward enough

PR #692, opened 2022-08-11
feat(examples): use detected user terminal instead of weston-terminal

oknozor:


anvil/src/input_handler.rs
```diff
@@ -819,3 +819,15 @@ fn process_keyboard_shortcut(modifiers: ModifiersState, keysym: Keysym) -> Optio
         None
     }
 }
+
+fn detect_terminal() -> String {
+    use std::fs::read_link;
+
+    const SYMLINK: &str = "/usr/bin/x-terminal-emulator";
```
Drakulix, line 826: This is not a portable way to detect the terminal. Fedora for example does not have this symlink.
I think the best open here would be to read the $TERMINAL environment variable.

anvil/src/input_handler.rs
```diff
@@ -819,3 +819,15 @@ fn process_keyboard_shortcut(modifiers: ModifiersState, keysym: Keysym) -> Optio
...
+
+    if let Ok(found) = read_link(SYMLINK) {
+        return read_link(&found).unwrap_or(found).to_string_lossy().to_string();
+    }
+
+    "/usr/bin/gnome-terminal".to_string()
```
Drakulix, line 832: Falling back to gnome-terminal is also a no-go.

Many people try anvil on a separate tty, while their main session is still running. If gnome-terminal is already running on the main session, this will only start another window there, not in anvil. Also I would like to avoid having a dependency on a specific desktop environment in anvil.

Anvil is meant as a demo of all of smithays features, not as a full-blown desktop environment. As such relying on test programs like `weston-terminal` (much like X11 examples used to use `xterm`) is perfectly fine imo.

examples/minimal.rs
```diff
@@ -143,7 +143,7 @@ pub fn run_winit() -> Result<(), Box<dyn std::error::Error>> {
         .unwrap();
 
     std::env::set_var("WAYLAND_DISPLAY", "wayland-5");
-    std::process::Command::new("weston-terminal").spawn().ok();
+    std::process::Command::new(detect_terminal()).spawn().ok();
```
Drakulix, line 146: The *minimal* example does not need any terminal detection imo. It should be as concise and *minimal* as possible.

smallvil/src/main.rs
```diff
@@ -49,3 +49,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    use std::fs::read_link;
+
+    const SYMLINK: &str = "/usr/bin/x-terminal-emulator";
+
+    if let Ok(found) = read_link(SYMLINK) {
+        return read_link(&found).unwrap_or(found).to_string_lossy().to_string();
+    }
+
+    "/usr/bin/gnome-terminal".to_string()
+}
```
Drakulix, line 63: Same goes for smallvil. While this is certainly nice to have (and I don't have any objections for this feature in general for anvil), I don't think smallvil is a good fit.

Drakulix (changes requested):

PR #693, opened 2022-08-11
fix: freeze on non-zero egl buffer age

oknozor:


PolyMeilex:
Passing 0 to render_output means re-render everything as far as I understand, so this does not feel right to me.
Space code:
```rs
        if age > 0 && state.old_damage.len() >= age {
            // We do not need even older states anymore
            state.old_damage.truncate(age);
            damage.extend(state.old_damage.iter().flatten().copied());
        } else {
            // just damage everything, if we have no damage
            damage = vec![output_geo];
        }
 ```

Drakulix:
The original issue this was supposed to fix was deal with via #694

PR #696, opened 2022-08-11
smallvil: Simplified flag parsing

HeavyRain266:
I found it to be cleaner and easier to understand.

**NOTE**: On my machine, `weston-terminal` doesn't start without `--client` flag.


PolyMeilex (approved):

PolyMeilex:
Clippy is complaining after recent rust update, will fix it in another PR.

PR #698, opened 2022-08-13
egl: Add EGLDisplay::from_raw and EGLContext::from_raw

technobaboo:
Confirmed working using this example:

```rs
struct EGLRawHandles {
	display: *const c_void,
	config: *const c_void,
	context: *const c_void,
}
```

```rs
let mut renderer = unsafe {
	Gles2Renderer::new(
		EGLContext::from_raw(
			egl_raw_handles.display,
			egl_raw_handles.config,
			egl_raw_handles.context,
			log.clone(),
		)?,
		log.clone(),
	)?
};
let smithay_tex_image =
	image::io::Reader::with_format(std::io::Cursor::new(STARDUST_PNG), image::ImageFormat::Png)
		.decode()
		.unwrap();
let smithay_tex = renderer
	.import_memory(
		&smithay_tex_image.to_rgba8(),
		(
			smithay_tex_image.width() as i32,
			smithay_tex_image.height() as i32,
		)
			.into(),
		false,
	)
	.unwrap();
```
https://github.com/StardustXR/stardust-assets/blob/main/icon.png
image
(The gamma issue is due to the renderer I am using, https://github.com/StereoKit/StereoKit)

### DIFF
--- a/src/backend/egl/context.rs
+++ b/src/backend/egl/context.rs
@@ -44,7 +44,7 @@ impl EGLContext {
     ///
     /// # Safety
     ///
-    /// - Not using the system default EGL library (`dlopen("libEGL.so")`) as loaded by smithay might cause undefined behavior
+    /// - The context must be created from the system default EGL library (`dlopen("libEGL.so")`)
     /// - Display, config, and context handles must be externally managed to ensure they do not become invalid before the compositor is shut down
     pub unsafe fn from_raw<L>(
         display: *const c_void,
--- a/src/backend/egl/context.rs
+++ b/src/backend/egl/context.rs
@@ -5,6 +5,8 @@ use std::{
     sync::{atomic::Ordering, Arc},
 };
 
+use libc::c_void;
+
 use super::{ffi, wrap_egl_call, Error, MakeCurrentError};
 use crate::{
     backend::{
@@ -27,6 +29,7 @@ pub struct EGLContext {
     config_id: ffi::egl::types::EGLConfig,
     pixel_format: Option<PixelFormat>,
     user_data: Arc<UserDataMap>,
+    externally_managed: bool,
 }
 
 // SAFETY: A context can be sent to another thread when this context is not current.
@@ -37,6 +40,36 @@ pub struct EGLContext {
 unsafe impl Send for EGLContext {}
 
 impl EGLContext {
+    /// Creates a new `EGLContext` from raw handles.
+    ///
+    /// # Safety
+    ///
+    /// This function is marked unsafe, because the context cannot be made current on another thread without
+    /// being unbound again (see [`EGLContext::unbind`]).
+    pub unsafe fn from_raw<L>(
+        display: *const c_void,
+        config_id: *const c_void,
+        context: *const c_void,
+        log: L,
+    ) -> Result<EGLContext, Error>
+    where
+        L: Into<Option<::slog::Logger>>,
+    {
+        let log = crate::slog_or_fallback(log.into()).new(o!("smithay_module" => "backend_egl"));
+
+        let display = EGLDisplay::from_raw(display, config_id, log)?;
+        let pixel_format = display.get_pixel_format(config_id)?;
+
+        Ok(EGLContext {
+            context,
+            display,
+            config_id,
+            pixel_format: Some(pixel_format),
+            user_data: Arc::new(UserDataMap::default()),
+            externally_managed: true,
+        })
+    }
+
     /// Creates a new configless `EGLContext` from a given `EGLDisplay`
     pub fn new<L>(display: &EGLDisplay, log: L) -> Result<EGLContext, Error>
     where
@@ -181,6 +214,7 @@ impl EGLContext {
             } else {
                 Arc::new(UserDataMap::default())
             },
+            externally_managed: false,
         })
     }
 
@@ -289,11 +323,13 @@ impl EGLContext {
 
 impl Drop for EGLContext {
     fn drop(&mut self) {
-        unsafe {
-            // We need to ensure the context is unbound, otherwise it egl stalls the destroy call
-            // ignore failures at this point
-            let _ = self.unbind();
-            ffi::egl::DestroyContext(**self.display.get_display_handle(), self.context);
+        if !self.externally_managed {
+            unsafe {
+                // We need to ensure the context is unbound, otherwise it egl stalls the destroy call
+                // ignore failures at this point
+                let _ = self.unbind();
+                ffi::egl::DestroyContext(**self.display.get_display_handle(), self.context);
+            }
         }
     }
 }
--- a/src/backend/egl/context.rs
+++ b/src/backend/egl/context.rs
@@ -44,8 +44,8 @@ impl EGLContext {
     ///
     /// # Safety
     ///
-    /// This function is marked unsafe, because the context cannot be made current on another thread without
-    /// being unbound again (see [`EGLContext::unbind`]).
+    /// - Not using the system default EGL library (`dlopen("libEGL.so")`) as loaded by smithay might cause undefined behavior
+    /// - Display, config, and context handles must be externally managed to ensure they do not become invalid before the compositor is shut down
     pub unsafe fn from_raw<L>(
         display: *const c_void,
         config_id: *const c_void,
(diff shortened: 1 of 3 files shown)


src/backend/egl/display.rs
```diff
@@ -212,6 +201,100 @@ impl EGLDisplay {
...
+        display: *const c_void,
+        config: *const c_void,
+        logger: L,
+    ) -> Result<EGLDisplay, Error>
+    where
+        L: Into<Option<::slog::Logger>>,
+    {
+        let log = crate::slog_or_fallback(logger.into()).new(o!("smithay_module" => "backend_egl"));
+
+        let dp_extensions = ffi::make_sure_egl_is_loaded()?;
```
i509VCB, line 215: Hmm, this brings up a good question. How can we assert an external context and smithay use the exact same instance of egl?

src/backend/egl/display.rs
```diff
@@ -212,6 +201,100 @@ impl EGLDisplay {
...
+
+        let extensions = EGLDisplay::get_extensions(egl_version, display)?;
+        info!(log, "Supported EGL display extensions: {:?}", extensions);
+
+        let (dmabuf_import_formats, dmabuf_render_formats) =
+            get_dmabuf_formats(&display, &extensions, &log).map_err(Error::DisplayCreationError)?;
+
+        let egl_api =
+            wrap_egl_call(|| ffi::egl::QueryAPI()).map_err(|_| Error::OpenGlesNotSupported(None))?;
+        if egl_api != ffi::egl::OPENGL_ES_API {
```
i509VCB, line 252: @drakulix, this condition isn't strictly needed?

src/backend/egl/context.rs
```diff
@@ -37,6 +40,36 @@ pub struct EGLContext {
 unsafe impl Send for EGLContext {}
 
 impl EGLContext {
+    /// Creates a new `EGLContext` from raw handles.
+    ///
+    /// # Safety
+    ///
+    /// This function is marked unsafe, because the context cannot be made current on another thread without
+    /// being unbound again (see [`EGLContext::unbind`]).
+    pub unsafe fn from_raw<L>(
```
i509VCB, line 49: We need to explain some more requirements:

Does the display need to live for the lifetime of the returned object.
Does the config need to live for the lifetime of the context?
Does the context need to live for the lifetime of the object?

src/backend/egl/display.rs
```diff
@@ -212,6 +201,100 @@ impl EGLDisplay {
         })
     }
 
+    /// Create a new [`EGLDisplay`] from an already initialized EGLDisplay and EGLConfig handle
+    pub unsafe fn from_raw<L>(
```
i509VCB, line 205: Similar comment to context about lifetimes of pointers.

src/backend/egl/display.rs
```diff
@@ -212,6 +201,100 @@ impl EGLDisplay {
...
+        display: *const c_void,
+        config: *const c_void,
```
i509VCB, line 207: Guard against null pointers if applicable.

src/backend/egl/context.rs
```diff
@@ -37,6 +40,36 @@ pub struct EGLContext {
...
+        display: *const c_void,
+        config_id: *const c_void,
+        context: *const c_void,
```
i509VCB, line 52: Guard against null pointers if applicable.

technobaboo: idk if it is applicable, you can put a null pointer in but if you do that is your responsibility since it's an unsafe function?

i509VCB: Well we can assert a pointer is not null at least.

src/backend/egl/display.rs
technobaboo: If you're putting the context into smithay wouldn't that be assured? or do you want a function that can tell you if an egl context is the same as the one smithay is using?

src/backend/egl/context.rs
technobaboo: From what I understand, display handle must outlive EGLDisplay, and config + context handles must outlive EGLContext. I don't know how to format safety docs for those though.

src/backend/egl/context.rs
technobaboo: sure, should I make a new error type for that?

i509VCB: Panicking is fine via the assert macros

src/backend/egl/display.rs
i509VCB: This is probably something @Drakulix would know about.

src/backend/egl/context.rs
i509VCB: Pretty much just some bullet points.

src/backend/egl/display.rs
Drakulix: We can't. It might be that the original library dynamically loaded a different `libEGL`.
There is not really something we can do about it, so we should just add another bullet point to the safety docs.

src/backend/egl/display.rs
Drakulix: On mesa? No. On other drivers? Yes.
OpenGL ES can behave subtly differently and most importantly `GL_OES_EGL_image` is a OpenGL ES only protocol.
Mesa does expose it for desktop GL contexts as well, but other drivers do not. Therefore we strictly need an ES context.

So yes, this check is necessary.

src/backend/egl/display.rs
technobaboo: I didn't see this, but I'll add it in! Should it be something like

- Must use the same EGL library smithay does

be good? or what should I put to be more specific?

src/backend/egl/display.rs
technobaboo: I'm wondering if this is the correct thing to even do, would usize or u32 be better?

src/backend/egl/display.rs
Drakulix: > Must use the same EGL library smithay does

Then we will get questions about "what library does smithay use". :sweat_smile: 
So I would go with something like:

> Not using the system default EGL library (`dlopen("libEGL.so")`) as loaded by smithay might cause undefined behavior

src/backend/egl/context.rs
```diff
@@ -37,6 +40,39 @@ pub struct EGLContext {
 unsafe impl Send for EGLContext {}
 
 impl EGLContext {
+    /// Creates a new `EGLContext` from raw handles.
+    ///
+    /// # Safety
+    ///
+    /// - Not using the system default EGL library (`dlopen("libEGL.so")`) as loaded by smithay might cause undefined behavior
```
i509VCB, line 47: Stylistic changes to use less passive language, the general idea is fine

```suggestion
    /// - The context must be created from the system default EGL library (`dlopen("libEGL.so")`)
```

src/backend/egl/context.rs
```diff
@@ -37,6 +40,39 @@ pub struct EGLContext {
...
+    /// - Display, config, and context handles must be externally managed to ensure they do not become invalid before the compositor is shut down
```
i509VCB, line 48:
```suggestion
    /// - The `display`, `config`, and `context` must be valid for the lifetime of the returned context.
```

src/backend/egl/display.rs
```diff
@@ -212,6 +201,108 @@ impl EGLDisplay {
         })
     }
 
+    /// Create a new [`EGLDisplay`] from an already initialized EGLDisplay and EGLConfig handle
+    ///
+    /// # Safety
+    ///
+    /// - Not using the system default EGL library (`dlopen("libEGL.so")`) as loaded by smithay might cause undefined behavior
```
i509VCB, line 208:
```suggestion
    /// - The display must be created from the system default EGL library (`dlopen("libEGL.so")`)
```

src/backend/egl/display.rs
```diff
@@ -212,6 +201,108 @@ impl EGLDisplay {
...
+    /// - Display and config handles must be externally managed to ensure they do not become invalid before the compositor is shut down
```
i509VCB, line 209: The handles created from the display (`get_display_handle`) need to be mentioned in the docs, the suggestion includes that

```suggestion
    /// - The `display` and `config` must be valid for the lifetime of the returned display and any handles created by this display (using [`EGLDisplay::get_display_handle`])
```

i509VCB (approved):

Drakulix (approved):
I think it's fine now. Thanks a lot!

PR #700, opened 2022-08-16
Feature/output renderer

cmeissl:
DO NOT MERGE


src/backend/renderer/output/element/mod.rs
```diff
@@ -0,0 +1,443 @@
...
+}
+
+/// A single render element
+pub trait RenderElement<R: Renderer> {
+    /// Get the unique id of this element
+    fn id(&self) -> &Id;
+    /// Get the current commit position of this element
+    fn current_commit(&self) -> usize;
+    /// Get the location relative to the output
+    fn location(&self, scale: Scale<f64>) -> Point<i32, Physical>;
```
Drakulix, line 30: Given both location and geometry are using Physical coordinates now, could this have a default impl, that just calls `self.geometry().loc`?

src/backend/renderer/output/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+#[derive(Debug)]
+pub struct OutputRender {
+    output: Output,
+    last_states: VecDeque<OutputRenderState>,
+    empty_state: OutputRenderState,
+}
+
+impl OutputRender {
+    /// Initialize a new [`OutputRender`]
+    pub fn new(output: &Output) -> Self {
```
Drakulix, line 40: At some point I would like to make this generic.

We essentially only need size, scale and transform from the output.
Storing the output has the nice advantage, that this updates with said output, but it should be possible for users to create this struct from a fixed size/scale/transform.

Maybe we should also rename this to `DamageTrackedRenderer` or something.

src/backend/renderer/utils.rs
```diff
@@ -471,200 +488,3 @@ where
...
-///
-/// - `scale` needs to be equivalent to the fractional scale the rendered result should have.
-/// - `location` is the position the surface should be drawn at.
-/// - `damage` is the set of regions that should be drawn relative to the same origin as the location.
-///
-/// Note: This element will render nothing, if you are not using
-/// [`crate::backend::renderer::utils::on_commit_buffer_handler`]
-/// to let smithay handle buffer management.
-#[allow(clippy::too_many_arguments)]
-pub fn draw_surface_tree<R, S>(
```
Drakulix, line 492: Do we have to kill this function? Like how would a user render a surface-tree without the OutputRenderer now?
They would have to create the RenderElements and manually call draw on them, right?
So we could keep `draw_surface_tree` (and potentially `draw_layer_surface` or `draw_window`) as a thin wrapper around that to make custom drawing more viable.

src/desktop/space/window.rs
```diff
@@ -1,21 +1,19 @@
...
     collections::HashMap,
 };
 
+use super::SpaceElement;
+
 #[derive(Default)]
 pub struct WindowState {
     pub location: Point<i32, Logical>,
+    // TODO: How to handle that?
     pub drawn: bool,
```
Drakulix, line 17: We don't have to handle this anymore (just drop it), it was only used for some ill-guided send_frame *optimization*, that we have removed already.

src/backend/renderer/output/element/mod.rs
```diff
@@ -0,0 +1,443 @@
...
+    /// A texture
+    External(&'a R::TextureId),
+}
+
+/// A single render element
+pub trait RenderElement<R: Renderer> {
+    /// Get the unique id of this element
+    fn id(&self) -> &Id;
+    /// Get the current commit position of this element
+    fn current_commit(&self) -> usize;
```
i509VCB, line 28: Is this a value which always increments or is it maxing out at 4 or so?

If it increments forever, assuming 144hz it would take 56 years to reach usize::MAX on 32-bit systems

src/backend/renderer/utils.rs
cmeissl: good idea, I will try to keep them as a wrapper

src/backend/renderer/output/element/mod.rs
cmeissl: That depends on the implementor, the new `OutputRender` type just uses the value to feed it into `damage_since`.
The default impl for `damage_since` just checks if the commit is different and will then report the whole size as damaged.
But yeah, this is possibly an ever increasing value, in one implementation I just used `wrapping_add(1)`.
I want to add a type for damage tracked texture elements which would handle that internally, but I also want to change to usize to a custom type which increments interanally (probably with a `wrapping_add`.

src/backend/renderer/output/mod.rs
cmeissl: Do we really have to store the size/scale/transform combo on creation? We could also make them part of the `render_output` function and only store them in the last state for checking if they changed. Then it would be clear from the API side and it would not be possible to forget updating them.

src/desktop/space/window.rs
cmeissl: :+1: done

src/backend/renderer/utils.rs
cmeissl: I added a small wrapper to `draw_render_elements` and re-added `draw_surface_tree`, `draw_window` and `draw_layer_surface`

src/backend/renderer/output/element/mod.rs
cmeissl: :+1: done

src/backend/renderer/output/mod.rs
Drakulix: I don't want to store them unconditionally. I just want to have the option to either initialize the renderer with an output (and have them update automatically) or initialize it statically.

PR #705, opened 2022-08-19
Make Space generic over its elements

Drakulix:
TODO:

- [x] Rebase
- [x] Bugfix
- [x] Get all of anvil, smallvil, examples, wlcs_anvil working again
- [x] Cleanup

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -842,7 +842,7 @@ fn render_surface<'a>(
 ) -> Result<bool, SwapBuffersError> {
     surface.surface.frame_submitted()?;
 
-    let output_geometry = space.output_geometry(&output).unwrap();
+    let output_geometry = space.output_geometry(output).unwrap();
 
     let (dmabuf, age) = surface.surface.next_buffer()?;
     renderer.bind(dmabuf)?;
@@ -938,7 +938,7 @@ fn render_surface<'a>(
 
     #[cfg(not(feature = "debug"))]
     let render_res = render_output(
-        &output,
+        output,
         space,
         &*elements,
         renderer,
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -918,7 +918,7 @@ fn render_surface<'a>(
         {
             surface.fps_element.update_fps(surface.fps.avg().round() as u32);
             surface.fps.tick();
-            elements.push(CustomRenderElements::Fps(fps_element.clone()));
+            elements.push(CustomRenderElements::Fps(surface.fps_element.clone()));
         }
     }
 
@@ -926,7 +926,7 @@ fn render_surface<'a>(
     // TODO we can pass the damage rectangles inside a AtomicCommitRequest
     #[cfg(feature = "debug")]
     let render_res = render_output(
-        &output,
+        output,
         space,
         &elements,
         renderer,
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -728,6 +728,7 @@ impl AnvilState<UdevData> {
                 &mut surfaces_iter
             };
 
+        let mut outputs = Vec::new();
         for (&crtc, surface) in to_render_iter {
             // TODO get scale from the rendersurface when supporting HiDPI
             let frame = self
@@ -814,14 +815,13 @@ impl AnvilState<UdevData> {
                     .expect("failed to schedule frame timer");
             }
 
+            outputs.push(output.clone());
+        }
+
+        std::mem::drop(surfaces);
+        for output in outputs {
             // Send frame events so that client start drawing their next frame
-            let space = &self.space;
-            let start_time = &self.start_time;
-            space.elements().for_each(|window| {
-                if space.outputs_for_element(window).contains(&output) {
-                    window.send_frame(start_time.elapsed().as_millis() as u32)
-                }
-            });
+            self.send_frames(&output);
         }
     }
 }
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -848,13 +848,25 @@ fn render_surface<'a>(
     renderer.bind(dmabuf)?;
 
     let mut elements: Vec<CustomRenderElements<_>> = Vec::new();
-    // set cursor
     if output_geometry.to_f64().contains(pointer_location) {
-        let (ptr_x, ptr_y) = pointer_location.into();
-        let ptr_location = Point::<i32, Logical>::from((ptr_x as i32, ptr_y as i32)); // - output_geometry.loc;
         let scale = Scale::from(output.current_scale().fractional_scale());
+        let cursor_hotspot = if let CursorImageStatus::Surface(ref surface) = cursor_status {
+            compositor::with_states(surface, |states| {
+                states
+                    .data_map
+                    .get::<Mutex<CursorImageAttributes>>()
+                    .unwrap()
+                    .lock()
+                    .unwrap()
+                    .hotspot
+            })
+        } else {
+            (0, 0).into()
+        };
+        let cursor_pos = pointer_location - output_geometry.loc.to_f64() - cursor_hotspot.to_f64();
+        let cursor_pos_scaled = cursor_pos.to_physical(scale).to_i32_round();
 
-        pointer_element.set_position(ptr_location);
+        // set cursor
         pointer_element.set_texture(pointer_image.clone());
 
         // draw the cursor as relevant
@@ -871,8 +883,7 @@ fn render_surface<'a>(
             pointer_element.set_status(cursor_status.clone());
         }
 
-        elements
-            .extend(pointer_element.render_elements(ptr_location.to_physical_precise_round(scale), scale));
+        elements.extend(pointer_element.render_elements(cursor_pos_scaled, scale));
 
         // draw input method surface if any
         let rectangle = input_method.coordinates();
@@ -892,22 +903,9 @@ fn render_surface<'a>(
         {
             if let Some(wl_surface) = dnd_icon.as_ref() {
                 if wl_surface.alive() {
-                    let cursor_hotspot = if let CursorImageStatus::Surface(ref surface) = cursor_status {
-                        compositor::with_states(surface, |states| {
-                            states
-                                .data_map
-                                .get::<Mutex<CursorImageAttributes>>()
-                                .unwrap()
-                                .lock()
-                                .unwrap()
-                                .hotspot
-                        })
-                    } else {
-                        (0, 0).into()
-                    };
                     elements.extend(AsRenderElements::<UdevRenderer<'a>>::render_elements(
-                        &SurfaceTree::from_surface(wl_surface, ptr_location - cursor_hotspot),
-                        ptr_location.to_physical_precise_round(scale),
+                        &SurfaceTree::from_surface(wl_surface, cursor_pos.to_i32_round()),
+                        cursor_pos_scaled,
                         scale,
                     ));
                 }
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -658,7 +658,7 @@ impl AnvilState<UdevData> {
             );
 
             // fixup window coordinates
-            crate::shell::fixup_positions(&display.handle(), &mut self.space);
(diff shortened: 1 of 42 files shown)


anvil/src/winit.rs
```diff
@@ -88,21 +89,18 @@ impl Backend for WinitData {
...
-        TextureRenderElement<<R as Renderer>::TextureId>,
-    ];
-    Pointer=&'a PointerElement<<R as Renderer>::TextureId>,
-    SurfaceTree=SurfaceTree,
+#[cfg(feature = "debug")]
+smithay::backend::renderer::output::element::render_elements! {
+    pub CustomRenderElements<R>;
+    Pointer=PointerRenderElement<R>,
+    Surface=WaylandSurfaceRenderElement,
+    Fps=FpsElement<<R as Renderer>::TextureId>,
```
i509VCB, line 103: This could be guarded with a cfg attribute instead of duplicating the declaration?

Drakulix: Currently the macro doesn't allow that it seems, but it is on my todo-list

cmeissl: Yes, but I have opted to use `SpaceRenderElements` in case you do not need custom render elements, just to show-case both possibilities.

src/desktop/space/mod.rs
```diff
@@ -32,43 +27,49 @@ mod window;
...
+    location: Point<i32, Logical>,
+    outputs: HashSet<Output>,
+}
+
 /// Represents two dimensional plane to map windows and outputs upon.
 #[derive(Debug)]
-pub struct Space {
+pub struct Space<E: SpaceElement> {
     pub(super) id: usize,
     // in z-order, back to front
```
cmeissl, line 46: Not sure if we still need the elements in back to front z-order. The new renderer abstractions expect them in front to back order and input handling also probably works that way.

cmeissl (commented):
This looks really nice overall :+1:

anvil/src/winit.rs
```diff
@@ -345,7 +332,10 @@ pub fn run_winit(log: Logger) {
         }
 
         // Send frame events so that client start drawing their next frame
-        state.space.send_frames(start_time.elapsed().as_millis() as u32);
+        state
+            .space
+            .elements()
+            .for_each(|window| window.send_frame(start_time.elapsed().as_millis() as u32));
```
cmeissl, line 338: small nit, we could move that function so that we can share the logic between the backends

anvil/src/x11.rs
```diff
@@ -300,9 +275,28 @@ pub fn run_x11(log: Logger) {
                 cursor_visible = true;
             }
 
-            pointer_element.set_position((x as i32, y as i32));
+            let scale = Scale::from(output.current_scale().fractional_scale());
+            let cursor_pos = state.pointer_location;
+            let cursor_pos_scaled = state.pointer_location.to_physical(scale).to_i32_round();
+
+            pointer_element.set_position(cursor_pos.to_i32_round());
```
cmeissl, line 282: Seems like the `PointerElement` position is now unused because it is no longer is a space element.

anvil/src/udev.rs
```diff
@@ -876,14 +871,44 @@ fn render_surface<'a>(
             pointer_element.set_status(cursor_status.clone());
         }
 
-        elements.push(CustomSpaceElements::PointerElement(pointer_element));
+        elements
+            .extend(pointer_element.render_elements(ptr_location.to_physical_precise_round(scale), scale));
```
cmeissl, line 875: I am not sure if this will work on a multi output setup. The `PointerElement` was a `SpaceElement` before so that it can calculate the `Physical` location on the output. IIRC the pointer location is in `Logical` space.

anvil/src/winit.rs
```diff
@@ -247,34 +229,6 @@ pub fn run_winit(log: Logger) {
 
             pointer_element.set_position(state.pointer_location.to_i32_round());
```
cmeissl, line 230: same here, position is unused

src/desktop/space/element.rs
```diff
@@ -34,53 +35,95 @@ impl From<RenderZindex> for Option<u8> {
...
+    /// Returns the geometry of this element.
+    ///
+    /// Defaults to be equal to it's bounding box.
+    fn geometry(&self) -> Rectangle<i32, Logical> {
+        self.bbox()
+    }
+    /// Returns the bounding box of this element
+    fn bbox(&self) -> Rectangle<i32, Logical>;
+    /// Returns whenever a given point inside this element will be able to receive input
+    fn input_region(&self, point: &Point<f64, Logical>) -> bool;
```
cmeissl, line 48: The name `input_region` sounds like it returns some kind of input region, maybe we can find a better name

src/desktop/space/mod.rs
```diff
@@ -469,62 +343,29 @@ impl Space {
...
-                .collect::<Vec<_>>(),
-        );
-
-        space_elements.extend(
-            self.windows()
-                .rev()
-                .map(SpaceElements::Window)
-                .collect::<Vec<_>>(),
-        );
+        space_elements.extend(self.elements.iter().rev().map(SpaceElements::Element));
```
cmeissl, line 348: as we now use near to far for the render element we could possibly get rid of the `rev()` if we also store the element from near to far

src/desktop/layer.rs
```diff
@@ -570,16 +579,15 @@ where
     S: Into<Scale<f64>>,
     P: Into<Point<f64, Physical>>,
 {
-    let location = location.into();
-    let scale = scale.into();
-
-    let elements = SpaceElement::<R, WaylandSurfaceRenderElement>::render_elements(
-        layer,
-        location.to_i32_round(),
+    draw_surface_tree(
```
cmeissl, line 582: Is this intentionally? It now no longer includes the popups

cmeissl (changes requested):

anvil/src/x11.rs
```diff
@@ -300,9 +275,28 @@ pub fn run_x11(log: Logger) {
...
             pointer_element.set_status(cursor_guard.clone());
-            custom_space_elements.push(CustomSpaceElements::Pointer(&pointer_element));
+            elements.extend(pointer_element.render_elements(cursor_pos_scaled, scale));
```
cmeissl, line 284: the pointer element now no longer uses the hotspot, resulting on wrong position of the cursor sometimes

src/desktop/space/mod.rs
```diff
@@ -538,55 +379,92 @@ pub enum OutputError {
...
 }
 
-render_elements! {
+impl<E: IsAlive> IsAlive for InnerElement<E> {
+    fn alive(&self) -> bool {
+        self.element.alive()
+    }
+}
+
+impl<E: SpaceElement> SpaceElement for InnerElement<E> {
```
cmeissl, line 388: The default `refresh` function is not forwarded to the inner element

anvil/src/winit.rs
Drakulix: Did that and then forgot to push it..
https://github.com/Smithay/smithay/pull/705/commits/df64e7ebdad6be62e755600114c435c1bf4cfad3

anvil/src/udev.rs
Drakulix: Ok, so I have to substract the output location like before. Will fix that

src/desktop/space/element.rs
Drakulix: Possibly something like `in_input_region`?

src/desktop/layer.rs
Drakulix: good catch. no, not intentional

src/desktop/space/mod.rs
```diff
@@ -538,55 +379,92 @@ pub enum OutputError {
...
+        let mut geo = self.element.geometry();
+        geo.loc += self.location;
+        geo
+    }
+
+    fn bbox(&self) -> Rectangle<i32, Logical> {
+        let mut bbox = self.element.bbox();
+        bbox.loc += self.location;
+        bbox
+    }
```
cmeissl, line 399: This is not working as intended and will result in wrong locations of the elements in the space. The location on screen should be location - geometry.loc iirc.
So maybe we can leave `geometry` as is and only ofset the `bbox`, then in the space during collecting the render element do something like `e.bbox().loc - e.geometry().loc`.

src/desktop/space/element.rs
Drakulix: went for `is_in_input_region`

src/desktop/space/mod.rs
```diff
@@ -385,38 +400,24 @@ impl<E: IsAlive> IsAlive for InnerElement<E> {
...
 }
 
-impl<E: SpaceElement> SpaceElement for InnerElement<E> {
+impl<E: SpaceElement> InnerElement<E> {
+    // the inner geometry of the element in space coordinates
     fn geometry(&self) -> Rectangle<i32, Logical> {
         let mut geo = self.element.geometry();
-        geo.loc += self.location;
+        geo.loc = self.location;
+        geo.size += geo.loc.to_size();
```
cmeissl, line 408: This will increase the size of the geometry by the element location, I don't think that is what we want.

Beside that I don't get why this is needed or what it tries to solve. For `weston-terminal` the `element.geometry()`  will be (32,32)x(742,517). If we place the window at (0,0) it will be renderer at (-32,32) with a size of (806,581), which seems correct. For calculating the occupied region in the space, ignoring anything outside of the window geometry, we would use the geometry as defined by the xdg protocol. If we want to have that space relative we should end up with (0,0)x(742,517). There is no need to extend the size by anything imo, everything outside of the original size should not be considered for layouting. So the visible region of the window in a space would be (self.location)x(self.element.geometry.size())

anvil/src/udev.rs
```diff
@@ -893,17 +916,17 @@ fn render_surface<'a>(
         {
             surface.fps_element.update_fps(surface.fps.avg().round() as u32);
             surface.fps.tick();
+            elements.push(CustomRenderElements::Fps(surface.fps_element.clone()));
         }
     }
 
     // and draw to our buffer
     // TODO we can pass the damage rectangles inside a AtomicCommitRequest
     #[cfg(feature = "debug")]
```
cmeissl, line 925: This is probably some left-over, both `cfg` parts do the same.

anvil/src/udev.rs
```diff
@@ -893,17 +916,17 @@ fn render_surface<'a>(
         {
             surface.fps_element.update_fps(surface.fps.avg().round() as u32);
             surface.fps.tick();
+            elements.push(CustomRenderElements::Fps(surface.fps_element.clone()));
```
cmeissl, line 919: Not sure if this is intentionally, but putting the fps element inside the cursor check we only get the fps overlay on a output where we currently have a cursor.

src/desktop/space/mod.rs
```diff
@@ -91,18 +87,20 @@ impl Space {
...
-            state.z_index = z_index;
+        if let Some(pos) = self.elements.iter().position(|inner| inner.element == element) {
+            self.elements.swap_remove(pos);
         }
-        self.insert_window(window, activate);
+
+        let inner = InnerElement {
+            element,
+            location: location.into(),
+            outputs: HashSet::new(),
```
cmeissl, line 101: This does not seem right, `map_element` is also called during moving a window. So each move will reset the output list and after that `refresh` will always send an `enter`, but never a `leave`.

src/desktop/space/mod.rs
Drakulix: Yeah right. I was for some reason expecting the geometry to be represented by two points instead of a point and a size. Of course the size doesn't change by translating. right.

Drakulix: fixed

anvil/src/udev.rs
Drakulix: fixed

anvil/src/udev.rs
Drakulix: fixed

src/desktop/space/mod.rs
Drakulix: fixed

cmeissl (approved):

PR #720, opened 2022-08-26
Wayland/virtual keyboard

rano-oss:
This adds support for virtual keyboard, ~~it needs a new version of wayland-protocols-misc for it to be able to be merged. The pr probably needs some reviewing first anyway. ;)~~
edit: Everything is tested and implemented, should be ready :')


src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -0,0 +1,183 @@
...
+        _client: &Client,
+        _: &ZwpVirtualKeyboardV1,
+        request: zwp_virtual_keyboard_v1::Request,
+        data: &VirtualKeyboardUserData<D>,
+        _dh: &DisplayHandle,
+        _data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            zwp_virtual_keyboard_v1::Request::Keymap { format, fd, size } => {
+                if format == 1 {
```
i509VCB, line 77: I guess the protocol does not mark the format u32 as an enum...

rano-oss: Unfortunately not

i509VCB: We should probably open an issue for that then where ever this protocol lives

HeavyRain266: @i509VCB This is part of wayland-protocols

i509VCB: No? This was posted on the mailing list but never integrated into wayland-protocols. I think purism is the source of the protocol.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+/// User data of ZwpVirtualKeyboardV1 object
+#[derive(Debug)]
+pub struct VirtualKeyboardUserData<D: SeatHandler> {
+    pub(super) handle: VirtualKeyboardHandle,
+    pub(crate) seat: Seat<D>,
+}
+
+impl<D> Dispatch<ZwpVirtualKeyboardV1, VirtualKeyboardUserData<D>, D> for VirtualKeyboardManagerState
```
i509VCB, line 60: - A ZwpVirtualKeyboardV1 has a strong reference to the VirtualKeyboardUserData.
- A VirtualKeyboardUserData contains a `VirtualKeyboard`
- A VirtualKeyboard contains a `Vec<ZwpVirtualKeyboardV1>`

I think it's pretty likely we have a reference cycle which will leak memory.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
i509VCB: I guess we should at least note why this is weird.

Something like:
```
This should be wl_keyboard::KeymapFormat::XkbV1, but the protocol does not state the parameter is an enum.
```

PolyMeilex: In addition to mailing list it also has MR open now: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/11

i509VCB: The MR indeed has the enum label

rano-oss: The alternative would be to use the new version, basically being a forerunner?

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
rano-oss: VirtualKeyboard now contains just a u8, I wasn't using the reference for anything anyway, just needed to count virtual keyboard instances.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
rano-oss: added comment :)

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -0,0 +1,205 @@
...
+                        if inner.old_keymap.is_none() {
+                            let old_keymap = CString::new(old_keymap)
+                                .expect("Keymap should not contain interior null bytes");
+                            inner.old_keymap = Some(KeymapFile::new(old_keymap, log));
+                        }
+                        internal.keymap = new_keymap;
+                        let known_kbds = &keyboard_handle.arc.known_kbds;
+                        for kbd in &*known_kbds.lock().unwrap() {
+                            kbd.keymap(KeymapFormat::XkbV1, fd, size);
+                        }
```
PolyMeilex, line 119: Not saying if this is right or not, but what does happen if someone presses a key on a physical keyboard while the virtual keyboard is open? Won't the map be wrong?

rano-oss: yes, it will be the keymap from the virtual keyboard. So yes the map will be "wrong". The alternative would be to keep all keymaps and set the keymaps on all clients before sending the key.... That would be a lot of work.
In sway they only change it when it is different. My take on this, is that in practice virtual keyboards will set the same keymap as the one set on the compositor(aka no change), also input methods will probably set a different one in many cases but that will be intended behavior and thus the physical keyboard will output the "correct" map.
I do set the keymap back to the original one when the virtual one is dropped, which I would guess at least makes it somewhat predictable...
I am open for suggestions :smile:

src/wayland/seat/keyboard.rs
```diff
@@ -124,7 +124,7 @@ where
     }
 }
 
-fn for_each_focused_kbds<D: SeatHandler + 'static>(
+pub fn for_each_focused_kbds<D: SeatHandler + 'static>(
```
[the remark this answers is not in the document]
rano-oss: yes, you are correct :sweat_smile:

rano-oss:
Seems that github actions is not allowing clippy to run properly here

PolyMeilex:
> Seems that github actions is not allowing clippy to run properly here?

Smithay build broke after new wayland-rs 0.30 beta got released, proper fix is #752

rano-oss:
I guess I will wait for that then :smiling_face_with_tear:

src/input/keyboard/mod.rs
```diff
@@ -436,6 +436,29 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
...
+
+        use std::os::unix::io::AsRawFd;
+        use wayland_server::{protocol::wl_keyboard::KeymapFormat, Resource};
+        let known_kbds = &self.arc.known_kbds;
+        for kbd in &*known_kbds.lock().unwrap() {
+            let res = keymap_file.with_fd(kbd.version() >= 7, |fd, size| {
+                kbd.keymap(KeymapFormat::XkbV1, fd.as_raw_fd(), size as u32)
+            });
+            if let Err(e) = res {
+                debug!(logger,
```
Drakulix, line 454: I think this should rather be a warning

Drakulix (commented):
Looks mostly fine from a code point of view, but I am not sure on the behavior from the protocol side.

Seems like keymap handling is real hassle and I don't know, if this is handled in the best way. I would like input from @i509VCB or @PolyMeilex before merging this.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
PolyMeilex: Sounds like a reasonable compromise, I'm fine with that.

PolyMeilex (approved):
My concerns about keymap handling were addressed so it's good to go as far as I'm concerned.

src/input/keyboard/mod.rs
rano-oss: Should I change this or merge and create a new shorter pr? :sweat_smile:

rano-oss: Figured I might as well :smile_cat:

Drakulix:
Sadly you broke CI now. 😅
I'll merge once its green again.

rano-oss:
Just a minor hiccup :see_no_evil:

PR #721, opened 2022-08-27
Add destroy hooks to XDG/Layer shell handlers

chrisduerr:
This adds `destroyed` methods for handling the death of XDG
toplevel, XDG popup and layer shell surfaces.

Providing these methods for the handlers directly allows for easy state
access and should simplify the likely different handling for the
individual shell surface types.

### DIFF
--- a/src/wayland/shell/wlr_layer/handlers.rs
+++ b/src/wayland/shell/wlr_layer/handlers.rs
@@ -308,18 +308,21 @@ where
     }
 
     fn destroyed(
-        _state: &mut D,
+        state: &mut D,
         _client_id: wayland_server::backend::ClientId,
         object_id: wayland_server::backend::ObjectId,
         data: &WlrLayerSurfaceUserData,
     ) {
         data.alive_tracker.destroy_notify();
+
         // remove this surface from the known ones (as well as any leftover dead surface)
         data.shell_data
             .known_layers
             .lock()
             .unwrap()
             .retain(|other| other.shell_surface.id() != object_id);
+
+        WlrLayerShellHandler::destroyed(state);
     }
 }
 
--- a/src/wayland/shell/wlr_layer/handlers.rs
+++ b/src/wayland/shell/wlr_layer/handlers.rs
@@ -316,13 +316,14 @@ where
         data.alive_tracker.destroy_notify();
 
         // remove this surface from the known ones (as well as any leftover dead surface)
-        data.shell_data
-            .known_layers
-            .lock()
-            .unwrap()
-            .retain(|other| other.shell_surface.id() != object_id);
-
-        WlrLayerShellHandler::destroyed(state);
+        let layers = &mut data.shell_data.known_layers.lock().unwrap();
+        if let Some(index) = layers
+            .iter()
+            .position(|layer| layer.shell_surface.id() == object_id)
+        {
+            let layer = layers.remove(index);
+            WlrLayerShellHandler::layer_destroyed(state, layer);
+        }
     }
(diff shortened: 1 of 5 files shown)


src/wayland/shell/wlr_layer/mod.rs
```diff
@@ -246,6 +246,9 @@ pub trait WlrLayerShellHandler {
 
     /// A surface has acknowledged a configure serial.
     fn ack_configure(&mut self, surface: wl_surface::WlSurface, configure: LayerSurfaceConfigure) {}
+
+    /// A layer surface was destroyed.
+    fn destroyed(&mut self) {}
```
i509VCB, line 251: Might be useful to know what surface was destroyed.

src/wayland/shell/xdg/mod.rs
```diff
@@ -870,6 +870,12 @@ pub trait XdgShellHandler {
     ///     is acknowledged by the client. See xdg_popup.repositioned for details.
     ///     The token itself is opaque, and has no other special meaning.
     fn reposition_request(&mut self, surface: PopupSurface, positioner: PositionerState, token: u32) {}
+
+    /// A toplevel surface was destroyed.
+    fn toplevel_destroyed(&mut self) {}
+
+    /// A popup surface was destroyed.
+    fn popup_destroyed(&mut self) {}
```
i509VCB, line 878: Similarly, nice to know which toplevel or popup was destroyed.

src/wayland/shell/wlr_layer/handlers.rs
```diff
@@ -308,18 +308,21 @@ where
...
         data.alive_tracker.destroy_notify();
+
         // remove this surface from the known ones (as well as any leftover dead surface)
         data.shell_data
             .known_layers
             .lock()
             .unwrap()
             .retain(|other| other.shell_surface.id() != object_id);
+
+        WlrLayerShellHandler::destroyed(state);
```
PolyMeilex, line 325: Well I think it would be helpful to send the object removed during retain to the user (perhaps by replacing it with remove(), as there is always just one surface with said objectid) 

And same for xdg as well.

PolyMeilex (changes requested):

src/wayland/shell/wlr_layer/mod.rs
chrisduerr: I was actually going to do that but then noticed it isn't passed to the destroyed handler, but looks like it's part of the surface data. Should be easy enough.

src/wayland/shell/wlr_layer/handlers.rs
```diff
@@ -308,18 +308,22 @@ where
...
-            .lock()
-            .unwrap()
-            .retain(|other| other.shell_surface.id() != object_id);
+        let layers = &mut data.shell_data.known_layers.lock().unwrap();
+        if let Some(index) = layers
+            .iter()
+            .position(|layer| layer.shell_surface.id() == object_id)
+        {
+            let layer = layers.remove(index);
+            WlrLayerShellHandler::layer_destroyed(state, layer);
```
PolyMeilex, line 325: Is `layers` MutexGuard dropped here by rust automagically? If not, I think you should `mem::drop` it to reduce the possibility of dead locks, not sure where the Mutex is used atm, and you never know where it will be used in future.

PolyMeilex (approved):

i509VCB (approved):

PolyMeilex:
Could you rebase? It should fix the CI build.

PR #722, opened 2022-08-27
Fix new seat module ergonomics

chrisduerr:
The recent seat handling changes made in 78ebaa6 have added some
friction to the way input and specifically the keyboard is handled. This
patch aims at making it easier to migrate to the new design.

To allow for easy access to the compositor state, the state is now
passed to the closure in the keyboard's `input` function. While this
makes sense as the simplest solution to this problem, a possible
alternative would be to move this function to the seat handler in the
future.


PolyMeilex:
I encountered the same problem in my compositor, so I'm in favor of `&mut D`.

Anvil and Smallvil need to be updated tho.

vberger (approved):

PR #725, opened 2022-08-29
space: Fix custom elements

Drakulix:
By mistake the RenderElements of the custom elements had to be borrowed, which in reality didn't work out (see e.g. https://github.com/Smithay/smithay/pull/726).
So now we *wrap* it and get rid of the lifetime instead.

@cmeissl Please review


cmeissl:
LGTM 👍

PR #726, opened 2022-08-29
[POC] anvil: Server-Side decorations

Drakulix:
Not intended for merge, patch is way to rough.
Just to show off the possibilities of the new generic space code.


Drakulix:
Done in #570

Issue #641, opened 2022-06-20
desktop: subsurface destruction artifacts


cmeissl:
When a sub-surface is destroyed the old geometry is not damaged.
This results in artificats where parts of the gone sub-subsurface are still visible on-screen.

Drakulix:
yeah true. but the basic idea should work for the most part, I hope. Because the destruction_callback is not giving us much more options than using something like a channel or just store the geometry inside the `SurfaceData`.

cmeissl:
Not sure if this works for all subsurfaces. There could be a sub-subsurface where the parent subsurface gets unmapped. This should also use the geometry of the sub-subsurface as damage. So we should also damage on unmap and use the complete (previous?) surface tree geometry for the damage.

Drakulix:
> Not sure if this works for all subsurfaces. There could be a sub-subsurface where the parent subsurface gets unmapped. This should also use the geometry of the sub-subsurface as damage. So we should also damage on unmap and use the complete (previous?) surface tree geometry for the damage.

Well in that case we probably need a dirty_flag for the `Window` (maybe an AtomicBool) and diff the surface_tree on the next accumulated_damage call, if that is true, or something like that.

PR #727, opened 2022-08-29
Split rendering code from Space and make it generic

Drakulix:
Work-in-progress PR for splitting of the damage-tracked rendering from Space and making Space generic (based on the previous work in https://github.com/Smithay/smithay/pull/689.

Fixes https://github.com/Smithay/smithay/issues/641.

TODO:
- [x] Make `render_elements!` accept trait bounds for the generic renderer case
- [x] Fix the mess that is the `wayland_frontend` feature in the `desktop` module right now by introducing a new `desktop`-feature
- [x] Fix commit_count and damage being a little awkward for the `TextureRenderElement`
- [x] Fix surface tracking for output_enter/leave events
  - Current layer-shell subsurfaces aren't tracked correctly
  - Removed surface's ObjectId's aren't cleaned up properly from the output's state (to be addressed by replacing the ObjectIds with Weak handles from https://github.com/Smithay/wayland-rs/pull/560)


bwidawsk (approved):

src/backend/drm/surface/atomic.rs
```diff
@@ -856,6 +856,11 @@ impl<A: AsRawFd + 'static> AtomicDrmSurface<A> {
 
         // and finally the others
         for plane_info in planes {
+            req.add_property(
+                plane_info.handle,
+                plane_prop_handle(&*prop_mapping, plane_info.handle, "CRTC_ID")?,
+                property::Value::CRTC(Some(self.crtc)),
+            );
```
bwidawsk, line 863: I don't see anything wrong with this, but, why is the primary plane handled separately. It would seem you could add primary to the planes slice and kill off some duplicated code.

src/backend/drm/device/mod.rs
```diff
@@ -143,6 +146,13 @@ impl<A: AsRawFd + 'static> DrmDevice<A> {
             .get_driver_capability(DriverCapability::MonotonicTimestamp)
             .unwrap_or(0)
             == 1;
+        let cursor_width = dev
+            .get_driver_capability(DriverCapability::CursorWidth)
+            .unwrap_or(64);
+        let cursor_height = dev
+            .get_driver_capability(DriverCapability::CursorHeight)
+            .unwrap_or(64);
+        let cursor_size = Size::from((cursor_width as u32, cursor_height as u32));
```
bwidawsk, line 155: Shouldn't it fallback to software cursor if the w/h can't be found?

src/backend/drm/device/mod.rs
```diff
@@ -14,6 +14,8 @@ use nix::sys::stat::fstat;
 
 pub(super) mod atomic;
 pub(super) mod legacy;
+use crate::utils::{Physical, Size};
```
bwidawsk, line 17: I notice the rest of the file uses the qualified path for internal crates rather than use. Not sure why exactly...

src/backend/renderer/utils.rs
```diff
@@ -325,10 +330,14 @@ pub fn on_commit_buffer_handler(surface: &WlSurface) {
...
 
+/// Defines a view into the surface
 #[derive(Debug, Default, PartialEq, Clone, Copy)]
-pub(crate) struct SurfaceView {
+pub struct SurfaceView {
+    /// The logical source used for cropping
     pub src: Rectangle<f64, Logical>,
+    /// The logical destination size used for scaling
     pub dst: Size<i32, Logical>,
+    /// The logical offset for a sub-surface
```
bwidawsk: Hard to tell at this point but maybe this should become pub(crate) and then add getters() for these fields.

src/backend/renderer/utils.rs
```diff
@@ -191,7 +191,19 @@ impl RendererSurfaceState {
         }
     }
 
-    pub(crate) fn damage_since(&self, commit: Option<usize>) -> Vec<Rectangle<i32, BufferCoord>> {
+    /// Get the current commit position of this surface
+    ///
+    /// The position should be saved after calling [`damage_since`] and
+    /// provided as the commit in the next call.
```
bwidawsk, line 188: Sounds like this should be returned with damage_since: `-> (Vec<Rectangle<i32, BufferCoord>>, usize)

anvil/Cargo.toml
```diff
@@ -35,7 +35,8 @@ features = [ "composite" ]
 gl_generator = "0.14"
 
 [features]
-default = [ "egl", "winit", "x11", "udev", "logind", "xwayland" ]
+#default = [ "egl", "winit", "x11", "udev", "logind", "xwayland" ]
+default = [ "egl", "x11" ]
```
bwidawsk: I don't understand this

i509VCB: Probably testing is why most things are disabled temporarily

src/backend/drm/surface/atomic.rs
cmeissl: Yes, I did this as part of #760 I just wanted to keep the scope of this PR cleaner.

src/backend/drm/device/mod.rs
cmeissl: This change could be moved to #760 as it is not used in this PR.
It only tells you the size of the hardware cursor, actually using a hardware cursor is handled in #760 (but only for atomic)

src/backend/renderer/utils.rs
cmeissl: Yeah, good point, this is more an atomic operation so it seems like a good idea to return both

anvil/Cargo.toml
Drakulix: Yeah, that is why this is still a draft PR. It is not completely ready ^^

src/backend/renderer/damage.rs
```diff
@@ -0,0 +1,525 @@
...
+    last_commit: CommitCounter,
+    last_geometry: Rectangle<i32, Physical>,
+    last_z_index: usize,
+}
+
+#[derive(Debug, Default)]
+struct RendererState {
+    size: Option<Size<i32, Physical>>,
+    elements: IndexMap<Id, ElementState>,
+    old_damage: VecDeque<Vec<Rectangle<i32, Physical>>>,
```
bwidawsk, line 190: Could/should this be a Vec and use push() and drain() instead of push_front() and truncate()?

src/backend/renderer/damage.rs
```diff
@@ -0,0 +1,525 @@
...
+#[derive(Debug, Clone)]
+pub enum DamageTrackedRendererMode {
+    /// Automatic mode based on a output
+    Auto(Output),
+    /// Static mode
+    Static {
+        /// Size of the static output
+        size: Size<i32, Physical>,
+        /// Scale of the static output
+        scale: Scale<f64>,
```
bwidawsk, line 203: Would `Size<i32, Logical>` be more apt?

src/backend/renderer/damage.rs
```diff
@@ -0,0 +1,525 @@
...
+
+/// Errors thrown by [`DamageTrackedRenderer::render_output`]
+#[derive(thiserror::Error)]
+pub enum DamageTrackedRendererError<R: Renderer> {
+    /// The provided [`Renderer`] returned an error
+    #[error(transparent)]
+    Rendering(R::Error),
+    /// The given [`Output`] has no mode set
+    #[error(transparent)]
+    OutputNoMode(#[from] OutputNoMode),
```
bwidawsk, line 248: Real curious to see how this can happen

src/backend/renderer/damage.rs
```diff
@@ -0,0 +1,525 @@
...
+                transform,
+            } => Ok((size, scale, transform)),
+        }
+    }
+}
+
+/// Damage tracked renderer for a single output
+#[derive(Debug)]
+pub struct DamageTrackedRenderer {
+    mode: DamageTrackedRendererMode,
```
bwidawsk, line 236: Maybe hold on to the output rather than the mode?

src/backend/renderer/damage.rs
```diff
@@ -0,0 +1,525 @@
...
+        &self.mode
+    }
+
+    /// Render this output
+    pub fn render_output<E, R>(
+        &mut self,
+        renderer: &mut R,
+        age: usize,
+        elements: &[E],
+        clear_color: [f32; 4],
```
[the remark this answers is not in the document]
i509VCB: Rounded edges?

src/backend/renderer/damage.rs
i509VCB: If this is used like a ring buffer of sorts, a VecDeque may be more appropriate

src/backend/renderer/damage.rs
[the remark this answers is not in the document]
cmeissl: Yes, I use 0.0 in my drm backend cause I don't need a background color, just black from crtc is good. I plan to check for all black in the drm renderer for some small optimizations.

src/backend/renderer/damage.rs
cmeissl: The output itself can return an option for the mode

src/backend/renderer/damage.rs
cmeissl: I am not sure what you mean, the mode has been introduced to be able to use the renderer explicitly without specifying an output.

anvil/src/drawing.rs
```diff
@@ -1,158 +1,126 @@
...
-    }
+render_elements! {
+    pub PointerRenderElement<R> where
+        R: ImportAll;
+    Surface=WaylandSurfaceRenderElement,
+    Texture=TextureRenderElement<<R as Renderer>::TextureId>,
 }
 
-impl<R> RenderElement<R> for PointerElement<<R as Renderer>::TextureId>
+impl<T: Texture + Clone + 'static, R> AsRenderElements<R> for PointerElement<T>
```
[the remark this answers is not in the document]
Drakulix: Why wrong? The PointerElement wants to own the texture (and not get a reference for example), so why shouldn't the requirement be static?

anvil/src/drawing.rs
```diff
@@ -1,158 +1,126 @@
...
+                        PointerRenderElement::<R>::from(TextureRenderElement::from_texture_buffer(
+                            location.to_f64(),
+                            texture,
+                            None,
+                            None,
+                        ))
+                        .into(),
+                    ]
+                } else {
+                    vec![]
```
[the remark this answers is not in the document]
cmeissl: Not really, this depends on the backend. X11 and winit do not provide a custom default pointer, but instead rely on the default pointer from the host compositor.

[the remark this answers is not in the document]
Drakulix: Well, if the user never calls set_texture, then rendering nothing is certainly expected imo.

anvil/src/shell.rs
```diff
@@ -345,10 +351,18 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
...
-        self.space.commit(surface);
+        if !is_sync_subsurface(surface) {
+            let mut root = surface.clone();
+            while let Some(parent) = get_parent(&root) {
+                root = parent;
+            }
+            if let Some(window) = self.space.elements().find(|w| w.toplevel().wl_surface() == &root) {
+                window.on_commit();
+            }
+        }
```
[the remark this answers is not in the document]
Drakulix: sync_subsurfaces are expected to be handled *in sync* to their parent surface. So we shouldn't update the state on their commit, but on their parent's commit. (Which is done internally by walking the surface tree.)

src/backend/renderer/damage.rs
Drakulix: Yeah, that was an requirement for the api I brought up. E.g. if you want to render a region of an output for capture, you might want to use this api without having to create an entire virtual output. The mode can optionally hold on to an output and I expect most use-cases to be covered by that scenario, but this just makes the api for flexible for those, that need it.

src/backend/renderer/damage.rs
Drakulix: Newly created Outputs in smithay don't have any mode set.

src/backend/renderer/damage.rs
cmeissl: To add, I already use this to draw the cursor to a buffer used for the cursor plane.

src/output.rs
```diff
@@ -164,6 +164,8 @@ pub(crate) struct Inner {
     pub(crate) description: String,
     #[cfg(feature = "wayland_frontend")]
     pub(crate) instances: Vec<WlOutput>,
+    #[cfg(feature = "wayland_frontend")]
+    pub(crate) handle: Option<WeakHandle>,
```
PolyMeilex, line 168: Could we get a comment describing that this is set and used by wayland::output, perhaps even group it together with `instances` and  `xdg_output` on the bottom of the struct definition, so the comment can apply to all of them?

Asking mostly because It took me a sec to guess that this is most likely set to Some by Wayland module, so someone not as familiar with the codebase might be quite confused by this.

PolyMeilex (changes requested):
It's way too big for me to do a proper line by line review, but after reading the whole thing I don't see anything that would raise my concerns, so aside from that one comment LGTM

src/desktop/space/mod.rs
```diff
@@ -358,826 +317,262 @@ impl Space {
...
+    E: SpaceElement + PartialEq + AsRenderElements<R>,
+>(
+    output: &Output,
+    renderer: &mut R,
+    age: usize,
+    spaces: &[&'a Space<E>],
+    custom_elements: &'a [C],
+    damage_tracked_renderer: &mut DamageTrackedRenderer,
+    clear_color: [f32; 4],
+    log: &slog::Logger,
```
PolyMeilex, line 557: This should be `Into<Option<Logger>>` I suppose

src/backend/renderer/output/element/mod.rs
```diff
@@ -519,24 +619,222 @@ pub use render_elements;
 use self::{surface::WaylandSurfaceRenderElement, texture::TextureRenderElement};
 
 render_elements! {
```
i509VCB: Shouldn't these test elements be in their own test module?

anvil/src/shell.rs
```diff
@@ -480,26 +495,27 @@ impl<BackendData: Backend> XdgShellHandler for AnvilState<BackendData> {
 
         let start_data = pointer.grab_start_data().unwrap();
 
+        let window = self
+            .space
+            .elements()
+            .find(|window| window.toplevel().wl_surface() == surface.wl_surface())
+            .unwrap()
+            .clone();
```
i509VCB, line 503: We seem to have a lot of these in Anvil. Maybe we could have a function to do this within anvil.

src/backend/renderer/element/mod.rs
```diff
@@ -0,0 +1,1355 @@
...
+    }
+
+    fn underlying_storage(&self, renderer: &R) -> Option<UnderlyingStorage<'_, R>> {
+        self.0.underlying_storage(renderer)
+    }
+}
+
+#[cfg(all(test, feature = "renderer_gl"))]
+#[allow(dead_code)]
+mod tests {
```
i509VCB, line 1077: This file is getting kind of long. We should probably move these tests to their own file.

bwidawsk (commented):
I'm interested in this functionality and I believe the code is good enough to merge and fix later as needed.

src/desktop/space/mod.rs
```diff
@@ -358,826 +316,263 @@ impl Space {
...
+    #[cfg(feature = "wayland_frontend")] R: Renderer + ImportAll,
+    #[cfg(not(feature = "wayland_frontend"))] R: Renderer,
+    C: RenderElement<R>,
+    E: SpaceElement + PartialEq + AsRenderElements<R>,
+    L: Into<Option<slog::Logger>>,
+>(
+    output: &Output,
+    renderer: &mut R,
+    age: usize,
+    spaces: &[&'a Space<E>],
```
bwidawsk, line 553: &[&Space] seems difficult to use from API perspective. From my perspective, I'd like &[Space], but even [&Space] seems a bit more idiomatic.

Drakulix: `&[Space]` would mean you have to have ownership over the spaces inside the collection.
If you don't happen to store them in a Vec<Space> anyway, this is cumbersome for the users.

Doing an ad-hoc array is way easier in that case and quite cheap: `[&space1, &space2]`.

`[&Space]` would satisfy the constrain above, but also forces downstream to create an ad-hoc array, which is annoying if you already have your references in another collection.
But on the other hand it is trivial to cast an array into a slice: `&[&space1, &space2]`.

That said, I think we can do better here.
@cmeissl Shouldn't `impl Iterator<Item=&Space>` also work here?

cmeissl: Yeah, I think so, probably even `impl IntoIterator<Item=&Space>` should work. I will take a look.

cmeissl: @bwidawsk I guess this is resolved now?

anvil/src/drawing.rs
```diff
@@ -1,158 +1,126 @@
...
+            }
+        }
     }
 }
 
 #[cfg(feature = "debug")]
 pub static FPS_NUMBERS_PNG: &[u8] = include_bytes!("../resources/numbers.png");
 
 #[cfg(feature = "debug")]
+#[derive(Clone)]
```
i509VCB, line 99: I

Drakulix: Yeah, cloning the commit counter and the id is not really efficient, but they probably get redrawn a bunch anyway and it's only in debug mode... but I suppose we could fix this?

i509VCB (approved):

src/backend/renderer/element/memory.rs
```diff
@@ -0,0 +1,614 @@
...
+    /// If you previously set opaque regions you should update
+    /// them with [`RenderContext::update_opaque_regions`]
+    pub fn resize(&mut self, size: impl Into<Size<i32, Buffer>>) {
+        self.buffer.resize(size);
+    }
+
+    /// Draw to the buffer
+    pub fn draw<F>(&mut self, f: F)
+    where
+        F: FnOnce(&mut [u8]) -> Vec<Rectangle<i32, Buffer>>,
```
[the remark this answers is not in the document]
Drakulix: @cmeissl Yeah, this function could indeed use some better docs... Suggestions?

anvil/src/drawing.rs
Drakulix: sure

PR #730, opened 2022-08-31
output: Split wayland::output

Drakulix:
Splits the wayland::output module into wayland specific parts and a new toplevel output-module, that can be used without any wayland functionality.

After my recent changes to the input system, I don't expect this change to be very controversial, but here are some motivations for it anyway.

The `Output` struct has got a second use-case apart from advertising to wayland_clients with our inclusion of the desktop-abstractions as a rendering helper. It represents a view into the compositors grid, that shall be rendered.

With the upcoming generic-space rework, there is no reason for the `Space` and it's rendering helpers for damage-tracking to depend on the `wayland_frontend` feature anymore. But because the rendering depends on the `Output` as a helper , there needs to be a way to use the `Output` without `wayland_frontend` to allow this.

I also think in general we should split responsibilities and make sure all parts of smithay, that have any use-case outside of the wayland-protocol are also usable without our wayland frontend.

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -15,6 +15,7 @@ use smithay::{
         keyboard::{keysyms as xkb, FilterResult, Keysym, ModifiersState},
         pointer::{AxisFrame, ButtonEvent, MotionEvent},
     },
+    output::Scale,
     reexports::wayland_server::{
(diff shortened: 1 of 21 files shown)


i509VCB (approved):

PolyMeilex (approved):
Sounds reasonable, and looks good 👍

PR #733, opened 2022-09-01
Dev/refactor damage

cmeissl:
@Drakulix please review, this PR tries to address `Fix commit_count and damage being a little awkward for the TextureRenderElement`

todos:
- [x] add some docs
- [x] use `Size` instead of `usize`  in `memory` module
- [x] move `opaque_regions`  to buffer
- [x]  crop & scale for opaque regions
- [ ] can we handle HiDPI better than with a `i32` buffer scale?


src/backend/renderer/element/memory.rs
```diff
@@ -0,0 +1,278 @@
...
+#[derive(Debug, Default)]
+struct MemoryRenderBufferInner {
+    mem: Vec<u8>,
+    damage_tracker: DamageTracker<i32, Buffer>,
+    textures: HashMap<(TypeId, usize), Box<dyn std::any::Any>>,
+    renderer_seen: HashMap<(TypeId, usize), CommitCounter>,
+}
+
+impl MemoryRenderBufferInner {
+    fn from_memory(mem: &[u8]) -> Self {
```
Drakulix, line 30: Maybe add a little todo here, that we want to allow specifying the color format in the future. (Right now the import/update_buffer calls don't support that anyway.)

src/backend/renderer/element/memory.rs
```diff
@@ -0,0 +1,278 @@
...
+
+impl<'a> Drop for RenderContext<'a> {
+    fn drop(&mut self) {
+        self.buffer.damage_tracker.add(&self.damage);
+    }
+}
+
+/// A render element for [`MemoryRenderBuffer`]
+#[derive(Debug)]
+pub struct MemoryRenderBufferRenderElement {
```
Drakulix, line 167: Getting some serious Java-vibes from these names :joy: 
(But I don't necessarily have a better idea and it is descriptive at least.)

Drakulix (approved):
LGTM overall

src/backend/renderer/element/memory.rs
cmeissl: I refactored the `memory` again to already include the `Size` of the buffer instead of the raw `usize` and added a `TODO` for the format

src/backend/renderer/element/texture.rs
```diff
@@ -1,84 +1,695 @@
...
+//! # Why use this implementation
+//!
+//! This module provides the following elements
+//!
+//! ## Static texture
+//!
+//! [`TextureBuffer`] represents a static texture which is expected to not change.
+//! It is possible to either use a [`pre-existing texture`](TextureBuffer::from_texture) or to create the texture
+//! from [`RGBA memory`](TextureBuffer::from_memory).
+//! The [`TextureBuffer`] can be used in the smithay pipeline by using [`TextureRenderElement::from_texture_buffer`].
```
Drakulix, line 12: Same problem as above, though not as bad, because this is already describes the buffer as static and the element to be explicitly part of the pipeline.

Drakulix (approved):
Minor complaints on the docs, otherwise I think this is good to go.

CI failure was caused by a mistake in the rebase. Should be fixed now on `dev/space-renderer-split` and should resolve itself, after another quick rebase.

src/backend/renderer/element/memory.rs
```diff
@@ -0,0 +1,483 @@
...
+//!     Point::default(),
+//!     Size::from((WIDTH, HEIGHT)),
+//! )]));
+//!
+//! // We explicitly drop the context here to make the borrow checker happy
+//! std::mem::drop(render_context);
+//!
+//! // Create a render element from the buffer
+//! let location = Point::from((100.0, 100.0));
+//! let render_element = MemoryRenderBufferRenderElement::from_buffer(
```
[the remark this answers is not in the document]
cmeissl: I extended the docs to clarify the usage.

src/backend/renderer/element/texture.rs
cmeissl: Changed the docs and examples to show the intended usage

src/backend/renderer/element/memory.rs
```diff
@@ -0,0 +1,483 @@
...
+
+#[derive(Debug, Default)]
+struct MemoryRenderBufferInner {
+    mem: Vec<u8>,
+    size: Size<i32, Buffer>,
+    scale: i32,
+    transform: Transform,
+    opaque_regions: Option<Vec<Rectangle<i32, Buffer>>>,
+    damage_tracker: DamageTracker<i32, Buffer>,
+    textures: HashMap<(TypeId, usize), Box<dyn std::any::Any>>,
```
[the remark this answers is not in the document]
cmeissl: Yes, that would be good, I can look into this, but would also prefer a separate PR.

PR #736, opened 2022-09-04
anvil: xkbcommon library update

akane6704:


PolyMeilex (approved):
I don't see any obvious regressions, so LGTM.

PolyMeilex (changes requested):
Actually smithay dep should be updated as well: here https://github.com/Smithay/smithay/blob/003726dd7430629824822850349fbaddc32e1808/Cargo.toml#L62

akane6704:
> Actually smithay dep should be updated as well: here
> 
> https://github.com/Smithay/smithay/blob/003726dd7430629824822850349fbaddc32e1808/Cargo.toml#L62

Done!

akane6704:
np!

PR #743, opened 2022-09-06
desktop: Decouple feature-set from wayland_frontend

Drakulix:


src/desktop/mod.rs
```diff
@@ -49,13 +49,20 @@
 //! to manage client buffers to do so. If you plan to use the provided drawing functions, you need to use
 //! [`on_commit_buffer_handler`](crate::backend::renderer::utils::on_commit_buffer_handler).
 
+#[cfg(feature = "wayland_frontend")]
```
i509VCB, line 52: To make the feature attribute spam less, could we put the wayland stuff into a wayland module we reexport the contents of?

Drakulix: yeah we could, but then we would have to do another feature-gated reexport of it's contents just to get the module-layout matching again...

i509VCB: Why would we need to feature gate the contents if everything in a hypothetical wayland module is already feature gated.

I was thinking of something like something like below:
```rust
#[cfg(...)]
mod wayland;

#[cfg(...)]
pub use wayland::*;
```

Drakulix:
cc @cmeissl - ready for review

PR #744, opened 2022-09-07
renderer: Add glow integration

Drakulix:
Adds a new `GlowRenderer` that wraps a `Gles2Renderer` and additionally exposes a `glow::Context` on top of the same GL context as the underlying renderer.

Can also be used in conjunction with the `MultiRenderer` and is purely optional.

### DIFF
--- a/src/backend/renderer/glow.rs
+++ b/src/backend/renderer/glow.rs
@@ -1,0 +1,381 @@
+//! Integration for using [`glow`] on top of smithays OpenGL ES 2 renderer
+
+#[cfg(feature = "wayland_frontend")]
+use crate::backend::renderer::{ImportDmaWl, ImportMemWl};
+#[cfg(all(
+    feature = "wayland_frontend",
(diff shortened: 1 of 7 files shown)


src/backend/renderer/glow.rs
```diff
@@ -0,0 +1,381 @@
...
+        let glow = unsafe {
+            renderer.make_current().unwrap();
+            Context::from_loader_function(|s| crate::backend::egl::get_proc_address(s) as *const _)
+        };
+
+        GlowRenderer {
+            gl: Gles2Hack::Owned(renderer),
+            glow: Arc::new(glow),
+            logger: log,
+        }
```
cmeissl, line 161: This could use `GlowRenderer::new` to reduce code duplication

cmeissl: Ah, yeah, missed that

PolyMeilex (approved):
LGTM

PR #745, opened 2022-09-08
Bump "nix" to 0.25.0

electron271:
Getting errors with old version of nix

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ lazy_static = "1"
 libc = "0.2.103"
 libseat= { version = "0.1.1", optional = true }
 libloading = { version="0.7.0", optional = true } 
-nix = "0.24.1"
+nix = "0.25"
 once_cell = "1.8.0"
 rand = "0.8.4"
 scopeguard = { version = "1.1.0", optional = true }


PolyMeilex:
Hi, Thanks for the PR.
I think that [drm-rs](https://github.com/Smithay/drm-rs) has to be updated first, otherwise we will end up with two versions of nix crate, and types will mismatch in some places.

electron271:
https://github.com/Smithay/drm-rs/pull/129

> Hi, Thanks for the PR. I think that [drm-rs](https://github.com/Smithay/drm-rs) has to be updated first, otherwise we will end up with two versions of nix crate, and types will mismatch in some places.
> 
> EDIT: yep, looks like it already broke in CI

electron271:
Here are the other packages that are on the old version of nix:
calloop
drm (pr sent)
drm-ffi
smithay-client-toolkit
wayland-backend
wayland-client
wayland-commons
wayland-cursor
wayland-server
wlcs_anvil
x11rb
x11rb-protocol

these are a lot of packages, this may not possible
i did a (very bad) attempt of replacing the versions, it could be a actual error in the code looking further

PolyMeilex:
> Here are the other packages that are on the old version of nix:
> calloop
> drm (pr sent)
> drm-ffi
> smithay-client-toolkit
> wayland-backend
> wayland-client
> wayland-commons
> wayland-cursor
> wayland-server
> wlcs_anvil
> x11rb
> x11rb-protocol

Thanks for the list, we control everything on that list except x11rb so the update should be relatively straight forward.

bwidawsk:
I think this can be closed

PolyMeilex:
Indeed we are on 0.26 now

PR #747, opened 2022-09-12
keyboard: refactor to improve keymapfile api and sealedfile api

technobaboo:
This is needed so Stardust can accept in keymaps, and I figured cleaning up the APIs would be nicer too since this is partially public.

### DIFF
--- a/src/input/keyboard/keymap_file.rs
+++ b/src/input/keyboard/keymap_file.rs
@@ -1,9 +1,8 @@
-#![allow(dead_code)]
-
 use crate::utils::sealed_file::SealedFile;
 use slog::error;
 use xkbcommon::xkb::{Keymap, KEYMAP_FORMAT_TEXT_V1};
 
+use std::ffi::CString;
 #[cfg(feature = "wayland_frontend")]
 use std::os::unix::prelude::RawFd;
 #[cfg(feature = "wayland_frontend")]
@@ -18,12 +17,17 @@ pub struct KeymapFile {
 
 impl KeymapFile {
     /// Turn the keymap into a string using KEYMAP_FORMAT_TEXT_V1, create a sealed file for it, and store the string
-    pub fn new(keymap: &Keymap, log: slog::Logger) -> Self {
+    pub fn new<L>(keymap: &Keymap, logger: L) -> Self
+    where
+        L: Into<Option<slog::Logger>>,
+    {
+        let logger = crate::slog_or_fallback(logger);
+        let name = CString::new("smithay-keymap").unwrap();
         let keymap = keymap.get_as_string(KEYMAP_FORMAT_TEXT_V1);
-        let sealed = SealedFile::new("smithay-keymap", &keymap);
+        let sealed = SealedFile::new(name, CString::new(keymap.as_str()).unwrap());
 
         if let Err(err) = sealed.as_ref() {
-            error!(log, "Error when creating sealed keymap file: {}", err);
+            error!(logger, "Error when creating sealed keymap file: {}", err);
         }
 
         Self {
--- a/src/input/keyboard/keymap_file.rs
+++ b/src/input/keyboard/keymap_file.rs
@@ -5,8 +5,6 @@ use xkbcommon::xkb::{Keymap, KEYMAP_FORMAT_TEXT_V1};
 use std::ffi::CString;
 #[cfg(feature = "wayland_frontend")]
 use std::os::unix::prelude::RawFd;
-#[cfg(feature = "wayland_frontend")]
-use wayland_server::protocol::wl_keyboard::WlKeyboard;
 
 /// Wraps an XKB keymap into a sealed file or stores as just a string for sending to WlKeyboard over an fd
 #[cfg(feature = "wayland_frontend")]
@@ -61,7 +59,10 @@ impl KeymapFile {
     }
 
     /// Send the keymap contained within to a WlKeyboard
-    pub fn send(&self, keyboard: &WlKeyboard) -> Result<(), std::io::Error> {
+    pub fn send(
+        &self,
+        keyboard: &wayland_server::protocol::wl_keyboard::WlKeyboard,
+    ) -> Result<(), std::io::Error> {
         use wayland_server::{protocol::wl_keyboard::KeymapFormat, Resource};
 
         self.with_fd(keyboard.version() >= 7, |fd, size| {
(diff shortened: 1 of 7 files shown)


vberger:
Please don't add `#[allow(...)]` attributes in the codebase without a good reason (and a comment explaining why they are needed). In particular, I don't think the `dead_code` warnings your PR raised are false positives: for example the reason `SealedFile::size()` was marked as dead by the compiler is because `SealedFile` is not `pub` but `pub(crate)` in your PR, which I believe is not what you intended.

src/utils/sealed_file.rs
```diff
@@ -0,0 +1,67 @@
...
+pub(crate) struct SealedFile {
+    file: File,
+    size: usize,
+}
+
+impl SealedFile {
+    pub fn new(name: &str, contents: &str) -> Result<Self, std::io::Error> {
+        let name = CString::new(name).expect("File name should not contain interior null bytes");
+        let contents =
+            CString::new(contents).expect("Sealed file contents should not contain interior null bytes");
```
PolyMeilex, line 27: Not sure if I like the fact that it now can panic behind callers back. In rust string, interior null bytes are perfectly valid, although rare, so maybe it does not matter.

I would either accept the CStr, or propagate the NullError to the caller.
SealError perhaps?
```rs
#[derive(thiserror::Error, Debug)]
pub enum SealError{
       Io(std::io::Error),
       InvalidName(std::ffi::NulError),
       InvalidContent(std::ffi::NulError),
}
```

PolyMeilex (commented):
Looks good to me in general, just some minor nitpicks

PolyMeilex:
> because `SealedFile` is not `pub` but `pub(crate)` in your PR, which I believe is not what you intended.

I'm not sure if it should be public, I can't imagine any use cases for it.

The dead code warning comes from the fact that `size()` is called only when `"wayland_frontend"` is on.
If we indeed need dead_code attribute, I would prefer it to be only on a method that causes problems, rather than a whole module.

vberger:
> I'm not sure if it should be public, I can't imagine any use cases for it.

Ah right, I guess the fact that the methods are `pub` lead me to misinterpret that the API was intended to be public.

Though my remark about `#[allow(...)]` still stands, especially regarding the module-level `#![allow(...)]` in `keymap_file.rs` which is not explained in any way.

src/input/keyboard/keymap_file.rs
```diff
@@ -1,26 +1,26 @@
...
     sealed: Option<SealedFile>,
-    keymap: CString,
+    keymap: String,
 }
 
 impl KeymapFile {
-    pub fn new(keymap: CString, log: slog::Logger) -> Self {
-        let sealed = SealedFile::new(&keymap);
+    /// Turn the keymap into a string using KEYMAP_FORMAT_TEXT_V1, create a sealed file for it, and store the string
+    pub fn new(keymap: Keymap, log: slog::Logger) -> Self {
```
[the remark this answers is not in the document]
technobaboo: oooh that's what I wanted, I didn't know how to do that

src/utils/sealed_file.rs
technobaboo: yea i was wondering what to do, i think that's good

vberger:
The `dead_code` warnings from `KeymapFile` fields come from the fact that most of its methods are gated by the `wayland_frontend` cargo feature. Is there a reason to not just gate the whole type on that feature?

technobaboo:
> The `dead_code` warnings from `KeymapFile` fields come from the fact that most of its methods are gated by the `wayland_frontend` cargo feature. Is there a reason to not just gate the whole type on that feature?

is there a reason not to gate the whole module?

technobaboo:
bumping this thread, what do i do? gate each individual method and struct?

rano-oss:
Just add `#[cfg(feature = "wayland_frontend")]` on all the methods it complains about ;)

Drakulix:
> Just add `#[cfg(feature = "wayland_frontend")]` on all the methods it complains about ;)

Yeah basically.

src/input/keyboard/mod.rs
```diff
@@ -424,13 +425,10 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
 
         info!(log, "Loaded Keymap"; "name" => internal.keymap.layouts().next());
 
-        let keymap = internal.keymap.get_as_string(xkb::KEYMAP_FORMAT_TEXT_V1);
-        let keymap = CString::new(keymap).expect("Keymap should not contain interior nul bytes");
-
         Ok(Self {
             arc: Arc::new(KbdRc {
+                keymap: KeymapFile::new(&internal.keymap, log.clone()),
```
PolyMeilex, line 430:
```suggestion
                #[cfg(feature = "wayland_frontend")]
                keymap: KeymapFile::new(&internal.keymap, log.clone()),
```

PolyMeilex: And I would do the same for the field in KbdRc struct

PolyMeilex (approved):

src/utils/sealed_file.rs
```diff
@@ -0,0 +1,66 @@
...
+};
+
+#[derive(Debug)]
+pub(crate) struct SealedFile {
+    file: File,
+    size: usize,
+}
+
+impl SealedFile {
+    #[allow(dead_code)]
```
PolyMeilex, line 24: Actually this is probably no longer needed

src/utils/sealed_file.rs
```diff
@@ -0,0 +1,66 @@
...
+        )?;
+
+        Ok(Self {
+            file,
+            size: contents.len(),
+        })
+    }
+
+    // Only used in KeymapFile which is under the wayland_frontend feature
+    #[allow(dead_code)]
```
PolyMeilex, line 56: Same here

PolyMeilex (changes requested):

src/utils/mod.rs
```diff
@@ -12,6 +12,8 @@ pub mod user_data;
 pub(crate) mod alive_tracker;
 pub use self::alive_tracker::IsAlive;
 
+pub(crate) mod sealed_file;
```
PolyMeilex, line 15:
```suggestion
#[cfg(feature = "wayland_frontend")]
pub(crate) mod sealed_file;
```

PolyMeilex (approved):

PolyMeilex:
Thanks!

PR #749, opened 2022-09-15
input: Use xkb `STATE_LAYOUT_EFFECTIVE` instead of `STATE_LAYOUT_LOCKED`

ids1024:
From xkb's docs, this should normally be used. So unless there's a specific reason for Smithay to do otherwise, this is how it should get the layout.


Drakulix:
@vberger I think you initially wrote this code. Any "specific reason" locked was used?

elinorbgr (approved):
Welp I guess I was not attentive enough when writing that, looks good.

PR #750, opened 2022-09-16
input: Add a way to add a keyboard to the seat from an `xkb::keymap`

ids1024:
This should provide a way to fix things like https://github.com/pop-os/xdg-shell-wrapper/issues/8.

Cosmic's panel is implemented as a compositor that hosts applets as clients. It should just pass through the same keymap that is provided by the host compositor, rather than trying to create one from any configuration.

This could also be useful for the Wayland backend (https://github.com/Smithay/smithay/pull/676), depending on use case. Then a compositor using the Wayland backend would respect the host compositors keymap settings, but for some testing it may be helpful to do otherwise.

Leaving this as a draft until I've tested it by incorporating it into xdg-shell-wrapper/cosmic-panel, and also making sure it passes through changes to the `group` communicated by `wl_keyboard::modifiers`.


ids1024:
This is potentially unsound since `xkb::Keymap` is refcounted (in a non-thread safe way), and Smithay's keyboard code assumes it owns the only references and can move it between threads.

PR #752, opened 2022-09-16
Fix/egl buffer segfault

cmeissl:
This PR fixes some segfaults for egl clients that have disconnected

- [9b8b86a](https://github.com/Smithay/smithay/commit/9b8b86ad43b2bf0bbed65312fcb3efd1a2d9245d) updates `wayland-rs` to `0.30.0-beta10` and makes use of `io_lifetimes`
- [05b1606](https://github.com/Smithay/smithay/commit/05b1606eb534b593789bb6fe263e77402f565bdd) fixes a race in the renderer accessing a destroyed EGL Buffer

### DIFF
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -1,12 +1,12 @@
 //! Type safe native types for safe egl initialisation
 
-use std::collections::HashSet;
 use std::ffi::CStr;
 use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::sync::Arc;
 #[cfg(all(feature = "wayland_frontend", feature = "use_system_lib"))]
 use std::sync::{Mutex, Weak};
+use std::{collections::HashSet, os::unix::prelude::AsRawFd};
 
 use libc::c_void;
 use nix::libc::c_int;
@@ -668,7 +668,7 @@ impl EGLDisplay {
         {
             out.extend(&[
                 names[i][0] as i32,
-                fd,
+                fd.as_raw_fd(),
                 names[i][1] as i32,
                 offset as i32,
                 names[i][2] as i32,
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -906,6 +906,10 @@ impl EGLBufferReader {
         &self,
         buffer: &WlBuffer,
     ) -> ::std::result::Result<EGLBuffer, BufferAccessError> {
+        if !buffer.is_alive() {
+            return Err(BufferAccessError::Destroyed);
+        }
+
         let mut format: i32 = 0;
         let query = wrap_egl_call(|| unsafe {
             ffi::egl::QueryWaylandBufferWL(
@@ -1022,6 +1026,10 @@ impl EGLBufferReader {
         &self,
         buffer: &WlBuffer,
     ) -> Option<crate::utils::Size<i32, crate::utils::Buffer>> {
+        if !buffer.is_alive() {
+            return None;
+        }
+
         let mut width: i32 = 0;
         if unsafe {
             ffi::egl::QueryWaylandBufferWL(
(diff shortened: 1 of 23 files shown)


cmeissl:
@vberger the build breaks because I also wanted to use `OwnedFd` for the `Dmabuf` handling in smithay and just used the re-export from `wayland_server`. What do you think about that issue? Should we add `io_lifetimes` as a dependecy to smithay instead of relying on the `wayland_server` re-export?

i509VCB:
I think smithay depending on io-lifetimes makes sense. We don't always have wayland-server (Wayland frontend feature)

src/lib.rs
```diff
@@ -49,6 +49,8 @@
 //! `log` crate. If not, the logs will discarded. This cargo feature is part of the default set of
 //! features of Smithay.
 
+#[doc(hidden)]
+pub extern crate io_lifetimes;
```
i509VCB, line 53: Why do we need an extern crate? This isn't 2015

cmeissl: I know, I just aligned it with `nix` just below, but we can also change both?

src/backend/egl/display.rs
```diff
@@ -906,6 +906,10 @@ impl EGLBufferReader {
         &self,
         buffer: &WlBuffer,
     ) -> ::std::result::Result<EGLBuffer, BufferAccessError> {
+        if !buffer.is_alive() {
+            return Err(BufferAccessError::NotManaged(EGLError::BadAccess));
```
cmeissl, line 910: @Drakulix @i509VCB not sure if this is the best suited error, maybe something like `BufferAccessError::Destroyed` would be a better fit?

src/wayland/data_device/mod.rs
```diff
@@ -114,7 +115,7 @@ pub trait DataDeviceHandler: Sized + ClientDndGrabHandler + ServerDndGrabHandler
     ///
     /// * `mime_type` - the requested mime type
     /// * `fd` - the fd to write into
-    fn send_selection(&mut self, mime_type: String, fd: RawFd) {}
+    fn send_selection(&mut self, mime_type: String, fd: OwnedFd) {}
```
cmeissl, line 118: @vberger is this correct to just close the fd by dropping it here

src/wayland/data_device/mod.rs
```diff
@@ -160,7 +161,7 @@ pub trait ServerDndGrabHandler {
     ///
     /// * `mime_type` - The requested mime type
     /// * `fd` - The FD to write into
-    fn send(&mut self, mime_type: String, fd: RawFd) {}
+    fn send(&mut self, mime_type: String, fd: OwnedFd) {}
```
cmeissl, line 164: @vberger same here

src/wayland/data_device/mod.rs
vberger: Looks right, the function passes FD ownership to the handler, which then decides what to do with it. :+1:

src/backend/egl/display.rs
vberger: `BufferAccessError::Destroyed` sounds better than `BufferAccessError::NotManaged` yeah

src/lib.rs
vberger: I think `nix` and `io_lifetimes` should just be `pub use`-ed in the `reexports` module.

cmeissl: I moved `io_lifetimes` to the `reexports` and removed the `pub extern crate nix`.
The `nix` crate already was in the `reexports`, but behind some feature gates. I made it available unconditonally as `nix` is not an optional crate anyway.

src/backend/egl/display.rs
cmeissl: Added `BufferAccessError:Destroyed` and updated `egl_buffer_contents`

Drakulix (approved):

PR #758, opened 2022-09-21
fix: add arm for primary_device_manager::Request::Destroy

wash2:
I noticed when using cosmic-comp that this unreachable!() arm would be triggered sometimes when closing a window. I assume that handling Destroy will help prevent that from happening.


Drakulix (approved):
Thanks, looks like I forgot this, when implementing. Nice catch!

PR #760, opened 2022-09-22
Feature/drm planes

cmeissl:
## Current limitations

- ~~no support for screen-capture~~ `DrmCompositor::render_frame` now returns a reference to everything needed to implement screencopy. Additional the returned struct implements functions for blitting the frame to a bound buffer.
- ~~cursor plane known to not work on nvidia (probably related to not using bo flags Cursor)~~ switched to gbm bo write and offscreen rendering -> works now
- ~~underlay/primary plane direct-scan out check could be relaxed~~ the primary plane can now be used for direct scan-out in more edge cases
## TODO

- [X] anvil: use `GbmBufferedSurface` as a fallback (or compile time feature)? -> The old behavior can be forced by setting the `ANVIL_DISABLE_DRM_COMPOSITOR` env variable

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -37,17 +37,17 @@ use smithay::{
             Allocator,
         },
         drm::{
-            DrmDevice, DrmDeviceFd, DrmError, DrmEvent, DrmEventMetadata, DrmNode, GbmBufferedSurface,
-            NodeType,
+            compositor::DrmCompositor, DrmDevice, DrmDeviceFd, DrmError, DrmEvent, DrmEventMetadata, DrmNode,
+            DrmSurface, GbmBufferedSurface, NodeType,
         },
         egl::{EGLContext, EGLDevice, EGLDisplay},
         libinput::{LibinputInputBackend, LibinputSessionInterface},
         renderer::{
             damage::{DamageTrackedRenderer, DamageTrackedRendererError},
-            element::{texture::TextureBuffer, AsRenderElements},
-            gles2::Gles2Renderer,
+            element::{texture::TextureBuffer, AsRenderElements, RenderElement, RenderElementStates},
+            gles2::{Gles2Renderbuffer, Gles2Renderer},
             multigpu::{gbm::GbmGlesBackend, GpuManager, MultiRenderer, MultiTexture},
-            Bind, Frame, Renderer,
+            Bind, DebugFlags, ExportMem, Offscreen, Renderer,
         },
         session::{libseat::LibSeatSession, Event as SessionEvent, Session},
         udev::{all_gpus, primary_gpu, UdevBackend, UdevEvent},
@@ -80,7 +80,7 @@ use smithay::{
         wayland_protocols::wp::presentation_time::server::wp_presentation_feedback,
         wayland_server::{backend::GlobalId, protocol::wl_surface, Display, DisplayHandle},
     },
-    utils::{Clock, DeviceFd, IsAlive, Logical, Monotonic, Point, Rectangle, Scale, Transform},
+    utils::{Clock, DeviceFd, IsAlive, Logical, Monotonic, Point, Scale, Transform},
     wayland::{
         compositor,
         input_method::{InputMethodHandle, InputMethodSeat},
@@ -110,7 +110,27 @@ pub struct UdevData {
     #[cfg(feature = "debug")]
     fps_texture: Option<MultiTexture>,
     pointer_image: crate::cursor::Cursor,
+    debug_flags: DebugFlags,
     logger: slog::Logger,
+}
+
+impl UdevData {
+    pub fn set_debug_flags(&mut self, flags: DebugFlags) {
+        if self.debug_flags != flags {
+            self.debug_flags = flags;
+
+            for (_, backend) in self.backends.iter_mut() {
+                let surfaces = backend.surfaces.borrow();
+                for (_, surface) in surfaces.iter() {
+                    surface.borrow_mut().compositor.set_debug_flags(flags);
+                }
+            }
+        }
+    }
+
+    pub fn debug_flags(&self) -> DebugFlags {
+        self.debug_flags
+    }
 }
 
 #[cfg(feature = "egl")]
@@ -143,7 +163,7 @@ impl Backend for UdevData {
             if let Some(gpu) = self.backends.get(&id.device_id) {
                 let surfaces = gpu.surfaces.borrow();
                 if let Some(surface) = surfaces.get(&id.crtc) {
-                    surface.borrow_mut().surface.reset_buffers();
+                    surface.borrow_mut().compositor.reset_buffers();
                 }
             }
         }
@@ -210,6 +230,7 @@ pub fn run_udev(log: Logger) {
         pointer_element: PointerElement::default(),
         #[cfg(feature = "debug")]
         fps_texture: None,
+        debug_flags: DebugFlags::empty(),
         logger: log.clone(),
     };
     let mut state = AnvilState::init(&mut display, event_loop.handle(), data, log.clone(), true);
@@ -269,7 +290,7 @@ pub fn run_udev(log: Logger) {
                     backend.event_dispatcher.as_source_ref().activate();
                     let surfaces = backend.surfaces.borrow();
                     for surface in surfaces.values() {
-                        if let Err(err) = surface.borrow().surface.surface().reset_state() {
+                        if let Err(err) = surface.borrow().compositor.surface().reset_state() {
                             slog::warn!(log2, "Failed to reset drm surface state: {}", err);
                         }
                     }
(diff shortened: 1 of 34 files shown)


src/backend/drm/renderer/mod.rs
```diff
@@ -0,0 +1,2309 @@
...
+    <F as ExportFramebuffer<B>>::Error: std::error::Error + Send + Sync,
+{
+    /// Initialize a new [`DrmRenderer`]
+    pub fn new<L>(
+        output: &Output,
+        surface: DrmSurface<D>,
+        mut allocator: A,
+        framebuffer_exporter: F,
+        renderer_formats: HashSet<DrmFormat>,
+        cursor_size: Size<u32, BufferCoords>,
```
bwidawsk, line 583: I think cursor_size should be an option because input/cursor isn't required for all compositors.

src/backend/drm/device/mod.rs
```diff
@@ -23,6 +25,89 @@ use legacy::LegacyDrmDevice;
...
+    claim: Arc<PlaneClaimInner>,
+}
+
+impl PlaneClaim {
+    /// The plane the claim was taken for
+    pub fn plane(&self) -> drm::control::plane::Handle {
+        self.claim.plane
+    }
+
+    /// The crct the claim was taken for
```
atseanpaul, line 62: s/crct/crtc/

cmeissl: thanks, should be fixed

src/backend/drm/renderer/mod.rs
cmeissl: The drm renderer currently tries to assign whatever fits on the cursor plane, so technically it is not required to be a cursor element. With the latest changes the cursor plane can be completely disabled by not passing it to the constructor.

i509VCB:
Conflict with anvil again

src/backend/drm/renderer/mod.rs
```diff
@@ -0,0 +1,2553 @@
...
+    surface: Arc<DrmSurface>,
+    planes: Planes,
+    damage_tracked_renderer: DamageTrackedRenderer,
+
+    framebuffer_exporter: F,
+
+    current_frame: FrameState<DrmScanoutBuffer<A::Buffer, <F as ExportFramebuffer<A::Buffer>>::Framebuffer>>,
+    #[allow(clippy::type_complexity)]
+    pending_frame: Option<(
+        FrameState<DrmScanoutBuffer<A::Buffer, <F as ExportFramebuffer<A::Buffer>>::Framebuffer>>,
```
i509VCB, line 487: I'd actually agree with clippy here. Maybe a Frame type internally used could be made

src/backend/renderer/mod.rs
```diff
@@ -227,6 +234,11 @@ pub trait Renderer {
     /// Set the filter method to be used when rendering a texture into a larger area than its size
     fn upscale_filter(&mut self, filter: TextureFilter) -> Result<(), Self::Error>;
 
+    /// Set the enabled [`DebugFlags`]
+    fn set_debug_flags(&mut self, flags: DebugFlags);
```
i509VCB, line 238: Could it be useful to check which debug flags are supported?

anvil/src/udev.rs
```diff
@@ -513,17 +539,60 @@ fn scan_connectors(
...
+            let mut planes = match surface.planes() {
+                Ok(planes) => planes,
+                Err(err) => {
+                    warn!(logger, "Failed to query surface planes: {}", err);
+                    continue;
+                }
+            };
+
+            // Using an overlay plane on a nvidia card breaks
+            if driver.name().to_string_lossy().to_lowercase().contains("nvidia")
```
i509VCB, line 562: How many driver workarounds do we plan to have? If it's going to go beyond a few I think we should adopt some sort of WorkaroundFlags

src/backend/renderer/mod.rs
```diff
@@ -757,3 +757,26 @@ pub fn buffer_dimensions(buffer: &wl_buffer::WlBuffer) -> Option<Size<i32, Buffe
         Err(_) => None,
     }
 }
+
+/// Returns if the underlying buffer is y-inverted
+///
+/// *Note*: This will only return y-inverted for buffer types known to smithay (see [`buffer_type`])
+#[cfg(feature = "wayland_frontend")]
+pub fn buffer_y_inverted(buffer: &wl_buffer::WlBuffer) -> Option<bool> {
```
bwidawsk, line 777: The way this is used, it should be sufficient to just return bool instead of Option<bool>. The None at the bottom is replaced by false and you can drop the Some returns.

src/backend/renderer/utils/wayland.rs
```diff
@@ -69,11 +113,10 @@ impl RendererSurfaceState {
                 self.buffer_scale = attrs.buffer_scale;
                 self.buffer_transform = attrs.buffer_transform.into();
 
-                if let Some(old_buffer) = std::mem::replace(&mut self.buffer, Some(buffer)) {
-                    if &old_buffer != self.buffer.as_ref().unwrap() {
-                        old_buffer.release();
-                    }
+                if !self.buffer.as_ref().map(|b| b == buffer).unwrap_or(false) {
```
bwidawsk, line 116: I think this could be more concisely (and readably IMHO) written:
`if !self.buffer.as_ref().map_or(false, |b| b == &buffer) {`

(maybe someday we get is_some_and)

src/backend/renderer/element/mod.rs
```diff
@@ -350,7 +348,7 @@ pub trait RenderElement<R: Renderer>: Element {
     ) -> Result<(), R::Error>;
 
     /// Get the underlying storage of this element, may be used to optimize rendering (eg. drm planes)
-    fn underlying_storage(&self, renderer: &R) -> Option<UnderlyingStorage<'_, R>> {
+    fn underlying_storage(&self, renderer: &mut R) -> Option<UnderlyingStorage> {
```
bwidawsk, line 351: AFAICT, it was uneccessary to make this `&mut R`. I've not tried all build options, but with default build I can get rid of it locally.

src/backend/renderer/gles2/shaders.rs
```diff
@@ -49,9 +57,17 @@ precision mediump float;
...
+uniform float tint;
 
 void main() {
-    gl_FragColor = vec4(texture2D(tex, v_tex_coords).rgb, 1.0) * alpha;
+    vec4 color;
+
+    color = vec4(texture2D(tex, v_tex_coords).rgb, 1.0) * alpha;
+
+    if (tint == 1.0)
+        color = vec4(0.0, 0.3, 0.0, 0.2) + color * 0.8;
```
bwidawsk, line 68: I think you can write this more clearly with:
`mix(gl_FragColor, vec4(0.0, 1.0, 0.0, 0.8)`

src/backend/renderer/mod.rs
i509VCB: If the buffer is not supported, probably just use a Result<book, UnmanagedResource>

cmeissl: > The way this is used, it should be sufficient to just return bool instead of Option. The None at the bottom is replaced by false and you can drop the Some returns.

The Option is used to communicate that the buffer is not managed by smithay and we can't tell if it is inverted. Anvil only uses smithay provided buffers, but downstream could introduce other buffer types.

cmeissl: > If the buffer is not supported, probably just use a `Result<bool, UnmanagedResource>`

Yes, I choose Option to align it with the other already existing buffer functions, all of them return an Option in that case. But I am not opposed to change it to a Result, we just should stay consistent.

anvil/src/udev.rs
cmeissl: I don't know, but this workaround is in an anvil only, not smithay. But I can move it to a separate function If you want?

src/backend/renderer/mod.rs
cmeissl: Something like returning the flags from something like `supported_debug_flags() -> DebugFlags`?

src/backend/drm/renderer/mod.rs
cmeissl: Yeah, I will try to move that to a type def.

src/backend/renderer/element/mod.rs
cmeissl: Yes, `&mut R` was necessary for exporting a texture to a dmabuf which got removed from this PR. But I think it would be fine to leave it `&mut R` to allow such a scenario without breaking the API again
@Drakulix any objections against making it `&mut`  as part of this PR?

Drakulix: Nope. Feel free to keep it.

src/backend/drm/renderer/mod.rs
cmeissl: @i509VCB I added some internal type defs and removed all `[allow(clippy::type_complexity)]`

src/backend/drm/renderer/mod.rs
```diff
@@ -0,0 +1,2460 @@
...
+    <F as ExportFramebuffer<<A as Allocator>::Buffer>>::Error,
+    R,
+>;
+
+type RenderFrameResult<A, F, R> =
+    Result<(Option<Vec<Rectangle<i32, Physical>>>, RenderElementStates), RenderFrameErrorType<A, F, R>>;
+
+/// Specialized renderer utilizing drm planes with support for direct scan-out
+#[derive(Debug)]
+pub struct DrmRenderer<A, F, U>
```
Drakulix, line 496: Might be a little late in the process, but I am not a huge fan of everything being a `Renderer` in smithay, with `DrmRenderer`s using `MultiRenderers`, that are wrapping `Gles2Renderer`, aided by the `DamageTrackedRenderer`.

Maybe we could name this thing `DrmCompositor` instead? (Or think about other ways to make the naming a little more self-explanatory?)

cmeissl: Tbh I am also not a fan of the renderer naming, but I am also pretty bad at naming things 😅 

Renaming it shouldn't be a big undertaking, it should only affect two commits. DrmCompositor is fine for me (everything else that I came up with would end in Manager, which is just as bad as Renderer)

cmeissl: renamed to `DrmCompositor`

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,2756 @@
...
+
+    last_scale: Scale<f64>,
+    last_transform: Transform,
+
+    element_states: IndexMap<Id, ElementFramebufferCache<<F as ExportFramebuffer<A::Buffer>>::Framebuffer>>,
+
+    debug_flags: DebugFlags,
+}
+
+const SUPPORTED_FORMATS: &[DrmFourcc] = &[DrmFourcc::Argb8888, DrmFourcc::Xrgb8888];
```
i509VCB, line 719: Is this something we would have to have as a parameter when creating a DrmCompositor?

cmeissl: I copied that part over from `GbmBufferedSurface` but forgot the comment. 

Updated to include the comment which explains the reason behind defining that formats

src/backend/renderer/mod.rs
Drakulix: For what it is worth, the `Result` is more self-explanatory. (and thus probably the better fit here)

Otherwise the function needs some docs, what `None` means.

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,3008 @@
...
+
+        // Just reset any next state, this will put
+        // any already acquired slot back to the swapchain
+        self.next_frame.take();
+
+        let current_size = self.output.current_mode().unwrap().size;
+        let output_scale = self.output.current_scale().fractional_scale().into();
+        let output_transform = self.output.current_transform();
+        let output_geometry =
+            Rectangle::from_loc_and_size((0, 0), output_transform.transform_size(current_size));
```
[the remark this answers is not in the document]
Drakulix: Or is it `Buffer`?

Drakulix (commented):
A few remarks on documentation, but otherwise I am really happy how this has turned out.

src/backend/renderer/mod.rs
Drakulix: That comment is sufficient. I would prefer all function to be changed, given we have a proper error type - that was largely my oversight when writing these functions, I think - but that doesn't necessarily need to be part of the PR. I agree consistency is key here.

src/backend/drm/compositor/mod.rs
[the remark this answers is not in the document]
cmeissl: Its physical, I will make it explicit and add some docs

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,3008 @@
...
+
+#[derive(Debug)]
+struct CursorState<G: AsFd + 'static> {
+    allocator: GbmAllocator<G>,
+    framebuffer_exporter: GbmDevice<G>,
+    previous_output_transform: Option<Transform>,
+    previous_output_scale: Option<Scale<f64>>,
+}
+
+/// Specialized renderer utilizing drm planes with support for direct scan-out
```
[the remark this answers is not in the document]
cmeissl: I re-wrote the docs and added some module docs

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,3008 @@
...
+    A: Allocator,
+    <A as Allocator>::Error: std::error::Error + Send + Sync,
+    <A as Allocator>::Buffer: AsDmabuf,
+    <A::Buffer as AsDmabuf>::Error: std::error::Error + Send + Sync + std::fmt::Debug,
+    F: ExportFramebuffer<A::Buffer>,
+    <F as ExportFramebuffer<A::Buffer>>::Framebuffer: std::fmt::Debug + 'static,
+    <F as ExportFramebuffer<A::Buffer>>::Error: std::error::Error + Send + Sync,
+    G: AsFd + Clone,
+{
+    /// Initialize a new [`DrmCompositor`]
```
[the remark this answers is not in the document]
cmeissl: done

src/backend/drm/compositor/mod.rs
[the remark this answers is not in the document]
cmeissl: done

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,3113 @@
...
+//!
+//! The [`DrmCompositor`] does so by walking the stack of provided [`Element`]s from front to back
+//! while trying to assign each element to a drm overlay plane. Each item that fails the plane test
+//! will be rendered on the primary plane using the provided [`Renderer`].
+//! Additionally it will try to assign the top most element that fit's into the cursor size (as specified
+//! by the [`DrmDevice`](crate::backend::drm::DrmDevice)) on the cursor plane. If the element can not be
+//! directly scanned out the renderer will be used to render the element into an [`Offscreen`] buffer that
+//! is copied over to the allocated gbm buffer for the cursor plane.
+//!
+//! Note: While the [`DrmCompositor`] also work on *legacy* drm the use of overlay and cursor planes is disabled in that case.
```
Drakulix, line 15:
```suggestion
//! Note: While the [`DrmCompositor`] also works on *legacy* drm the use of overlay and cursor planes is disabled in that case.
```

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,3113 @@
+//! Composition for [`Element`]s using drm planes
+//!
+//! When possible composition can be (partially) offloaded to the display driver by assigning
+//! elements to drm planes. This is especially important for latency intensive fullscreen clients
+//! like video renderer or games.
```
Drakulix, line 5:
```suggestion
//! like video renderers or games.
```

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,3113 @@
...
+//! The [`DrmCompositor`] does so by walking the stack of provided [`Element`]s from front to back
+//! while trying to assign each element to a drm overlay plane. Each item that fails the plane test
+//! will be rendered on the primary plane using the provided [`Renderer`].
+//! Additionally it will try to assign the top most element that fit's into the cursor size (as specified
+//! by the [`DrmDevice`](crate::backend::drm::DrmDevice)) on the cursor plane. If the element can not be
+//! directly scanned out the renderer will be used to render the element into an [`Offscreen`] buffer that
+//! is copied over to the allocated gbm buffer for the cursor plane.
+//!
+//! Note: While the [`DrmCompositor`] also work on *legacy* drm the use of overlay and cursor planes is disabled in that case.
+//! Direct scan-out will only work with a atomic [`DrmSurface`].
```
Drakulix, line 16:
```suggestion
//! Direct scan-out will only work with an atomic [`DrmSurface`].
```

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,3113 @@
...
+//! Currently this is limited to wayland buffers, but may be extended in the future.
+//! This module provides a default exporter based on [`gbm`] which should fit most use-cases.
+//!
+//! If a certain combination of elements works can only be determined by asking the driver by submitting
+//! a atomic commit test. If that test fails the element is scheduled to be rendered on the primary plane.
+//!
+//! ### Overlay planes
+//!
+//! The element can only be directly scanned out if it's geometry does not overlap with an already assigned
+//! element on a plane hight in the stack.
```
Drakulix, line 32:
```suggestion
//! element on a plane higher in the stack.
```

Drakulix (approved):

src/backend/drm/compositor/mod.rs
```diff
@@ -0,0 +1,3113 @@
...
+//!     PhysicalProperties {
+//!         size: Size::from((800, 600)),
+//!         make: "N/A".into(),
+//!         model: "N/A".into(),
+//!         subpixel: Subpixel::Unknown,
+//!     },
+//!     None,
+//! );
+//!
+//! # let device: DrmDevice = unsafe { MaybeUninit::uninit().assume_init() };
```
i509VCB, line 82: We should probably just use `todo!()` here, I'm a little scared by using MaybeUninit here. Plus the example is `no_run`, so it shouldn't cause any real change.

Drakulix: Yeah, I have also used `unreachable!()` in those cases.

cmeissl: Oh, I never thought about using just `todo!()` and always used some obscure ways, thanks!

PR #762, opened 2022-09-22
desktop: Fix handling for output_enter/leave events

Drakulix:
@cmeissl This is an attempt to handle all edge-cases around output_enter/leave events with the new generic space.

- `SurfaceTree` is not really mapped by anything, remove `SpaceElement`-impl and just provide `AsRenderElements` (TODO: Implement it directly for `WlSurface`?)
- `LayerSurface` is not mapped, but belongs to an `Output`, so also remove its `SpaceElement`-implementation
- `Window`s consist of multiple surfaces and can be on multiple outputs simultaniously. This means some surfaces might not be inside one outputs bounds, we thus need to provide `output_enter` and `refresh` with positioning information to allow for multiplexing this call correctly
- Additionally surfaces can be destroyed and created at any point, so `refresh` needs to recalculate outputs.

Note: The code for output-tracking for windows is pretty gross. Especially where it creates a `WeakOutput` to avoid any possible reference-cycles. This is the only piece of code requiring such a handle for now, so we might just leave that as is??
But I would clean this up significantly with a proper `WeakOutput` type. Or is that over-engineering?

Do you have some suggestions on the various approaches? And more importantly, do you see any edge-cases, where a sub-surface/popup might not receive the correct enter/leave events?

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -861,7 +861,7 @@ fn render_surface<'a>(
     ));
     input_method.with_surface(|surface| {
         elements.extend(AsRenderElements::<UdevRenderer<'a>>::render_elements(
-            &SurfaceTree::from_surface(surface, position),
+            &SurfaceTree::from_surface(surface),
             position.to_physical_precise_round(scale),
             scale,
         ));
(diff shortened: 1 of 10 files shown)


cmeissl:
How does this handle changes to the output geometry?
Providing both, the output and its geometry feels pretty redundant. Wouldn't it be enough to provide the output and its location? On output map refresh could receive the updated location and would also need to be called periodically to check for changed output size.

Not sure if it is a good idea to implement AsRenderElements for a surface. What would it return, only itself or also all children. That would also make it more prone to rounding errors, for example the proposed fractional scale protocol relies on some specific rules which need to take the parent location into account.

The pointer hack does really raise some concerns, but as long as it is an implementation detail I am fine with it.

cmeissl (approved):
LGTM

PR #763, opened 2022-09-23
ci: Enforce 1.59 MSRV

Drakulix:


PolyMeilex (approved):
1.59 sounds good to me. No objections, once GATs are a thing I would strongly consider bumping it, but now 1.59 is definitely enough.

Drakulix:
> once GATs are a thing I would strongly consider bumping it

Agreed. This PR is not about forcing 1.59 indefinitely, but rather enforcing **a** version at all and getting rid of CI failing every time a new rust version comes out and new clippy lints come in.

cmeissl:
As long as we can agree that bumping the MSRV needs some planning/coordination I am totally fine with this decission :)
Once GATs are a thing I will make sure we can meet the required MSRV.

PolyMeilex:
> As long as we can agree that bumping the MSRV needs some planning/coordination I am totally fine with this decission :) Once GATs are a thing I will make sure we can meet the required MSRV.

Yeah, I think that in the future, a PR that bumps the MSRV will be a good place where everyone will be able to voice theirs opinions and concerns, before it gets merged. It will give the whole process more structure.

vberger (approved):

i509VCB (approved):
1.59 is fine, I already use the latest rustc anyways.

PR #766, opened 2022-09-25
Replace Vec for tracking of pressed keys with HashSet

fayalalebrun:
In the previous implementation, pressed keys were stored in a list. This meant that each element in the list was not guaranteed to be unique, and if you received several key press events, the list would quickly expand in size. If the list became longer than 1024 elements (4096 bytes), sending the key map to any client would result in an error and the connection to the client being severed.

Switching to a set has the advantage of solving this issue as well as potentially speeding up the processing of any key release events, since the whole vec no longer has to be traversed.


Drakulix (approved):
LGTM

PR #771, opened 2022-09-29
Fix `egl_debug_log`

ids1024:
https://registry.khronos.org/EGL/extensions/KHR/EGL_KHR_debug.txt

As defined by the spec, `message` can be `NULL`, so this needs to handle that possibility.

The first argument is not "severity", it is an error code. The severity is instead the third argument, `messageType`.

Since message can be null, converting the error code to a string can be necessary to actually convey useful information.


Drakulix (approved):
Looks good! thanks :)

PR #772, opened 2022-10-02
output: Add WeakOutput type

Drakulix:
We have a couple of hack surrounding the internal use of `Output` inside smithay to make sure resources are getting cleaned up properly.
One of those is creating *weak output references*, as seen in the [`LayerMap`] and additionally (with a very hack hash-implementation) in #727.

Additionally I have found a bunch of use-cases for using `HashMap<Output, _>` in cosmic-comp, which would also be better served by a `WeakOutput` type.
So lets add that.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -52,7 +52,7 @@
 
 use std::{
     hash::{Hash, Hasher},
-    sync::{Arc, Mutex},
+    sync::{Arc, Mutex, Weak},
 };
 
 use slog::{info, o};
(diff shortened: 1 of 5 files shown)


src/output.rs
```diff
@@ -158,8 +158,9 @@ impl Scale {
     }
 }
 
+/// Inner output data
 #[derive(Debug)]
-pub(crate) struct Inner {
+pub struct Inner {
```
PolyMeilex, line 163: Feel free to disagree, but I'm not sure if exposing internal types is a good idea, and will look quite weird in the docs, but I understand that it removed one level of nested structs, so I suppose I'm fine with it (not really, but I'm trying to not be too biased :smile: ).

Maybe we could at least pick a less internally sounding type name, not sure what that would be tho, `OutputDataInner`?

PolyMeilex:
In general, I'm in favor of this, looks good. Just a minor nitpick

src/output.rs
PolyMeilex: Yep, OutputData struct initially was only meant to be UserData type wrapper so your change to move it to wayland module if for sure the right decision.

PolyMeilex (approved):
LGTM, WeakOutput is indeed a lot cleaner solution.

PR #773, opened 2022-10-03
detect gbm_bo_get_fd_for_plane

cmeissl:
When available or forced with feature backend_gbm_has_fd_for_plane use fd_for_plane which allows the use of multi planar formats which expose multiple planes with different fds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,13 +69,15 @@ slog-term = "2.3"
 [build-dependencies]
 gl_generator = { version = "0.14", optional = true }
 pkg-config = { version = "0.3.17", optional = true }
+cc = { version = "1.0", optional = true }
 
 [features]
 default = ["backend_drm", "backend_gbm", "backend_libinput", "backend_udev", "backend_session_logind", "backend_x11", "backend_winit", "desktop", "renderer_gl", "renderer_multi", "xwayland", "wayland_frontend", "slog-stdlog", "backend_vulkan"]
 backend_winit = ["winit", "backend_egl", "wayland-egl", "renderer_gl"]
 backend_x11 = ["x11rb", "x11rb/dri3", "x11rb/xfixes", "x11rb/present", "x11rb_event_source", "backend_gbm", "backend_drm", "backend_egl"]
 backend_drm = ["drm", "drm-ffi"]
-backend_gbm = ["gbm"]
+backend_gbm = ["gbm", "cc", "pkg-config"]
+backend_gbm_has_fd_for_plane = []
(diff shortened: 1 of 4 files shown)


Drakulix:
Why is this a draft? Does it lack testing?

cmeissl:
> Why is this a draft? Does it lack testing?

I am not totally happy with the `cc` output, currently this relies on testing if `gbm_bo_get_fd_for_plane` is defined in `gbm.h`.
If it is not defined there are some warnings in the build output which I can not silence.
But other than that I tested it in a few different build environments and everything seems to work as expected.

I was thinking about using `libloading` to load the gbm lib and test if the symbol is available, that would be more reliable and probably also remove the warning from the build log. What do you think?

Drakulix:
Hmm, I don't really mind the build output.

> I was thinking about using libloading to load the gbm lib and test if the symbol is available, that would be more reliable and probably also remove the warning from the build log.

But as you said, this should be more accurate, because it actually tests for the symbol instead for a declaration.

If you want to spent the time, I am not opposed to using libloading instead, though I wonder, if there are cases, where this makes building for platforms without the function actually more difficult.
E.g. when cross-compiling, where the build.rs runs on the host-system could this trigger a detected get_plane_fd from the host-system, even if the includes of the target-sysroot don't contain the function? (Or out-right fail, because it libloading would fail to load the library for the target architecture?)

cmeissl:
> Hmm, I don't really mind the build output.
> 
> > I was thinking about using libloading to load the gbm lib and test if the symbol is available, that would be more reliable and probably also remove the warning from the build log.
> 
> But as you said, this should be more accurate, because it actually tests for the symbol instead for a declaration.
> 
> If you want to spent the time, I am not opposed to using libloading instead, though I wonder, if there are cases, where this makes building for platforms without the function actually more difficult. E.g. when cross-compiling, where the build.rs runs on the host-system could this trigger a detected get_plane_fd from the host-system, even if the includes of the target-sysroot don't contain the function? (Or out-right fail, because it libloading would fail to load the library for the target architecture?)
> 
> (The reverse-case is not an issue, because you can always just force enable the function with the new cargo feature.)

I will give it a try, especially the cross compilation. If the build env is correctly set up pkg-config should find the target lib automatically, but I never tried to use libloading in the context of a build script 😅

cmeissl:
Yeah, no, this does not work out-of-the-box and I don't want to spend more time on finding a nicer solution when the current one seems to be sufficient.

Drakulix (approved):

PR #775, opened 2022-10-05
dev: make space render_output more ergonomic

cmeissl:
We can use IntoIterator instead of requiring a slice for render_output and space_render_elements

### DIFF
--- a/anvil/src/render.rs
+++ b/anvil/src/render.rs
@@ -62,7 +62,7 @@ where
             output,
             renderer,
             age,
-            &[space],
+            [space],
             custom_elements,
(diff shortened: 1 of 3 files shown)


Drakulix (approved):
Yes! Lets have 5 generics for this function.

bwidawsk:
> Yes! Lets have 5 generics for this function.

Most can generally be inferred though

cmeissl:
> Yes! Lets have 5 generics for this function.

Yeah, impl IntoIterator does not work in rust 1.59 with explicit generic arguments unfortunately.

PR #777, opened 2022-10-10
input: Add a way to retrieve current focus

Drakulix:
Now that the `Seat` is generic over focused elements, we have lost the option to easily receive the currently focused element.
We cannot return a reference, as both fields are behind `Mutex`. When this was still just a `WlSurface`, we could just clone it.
So lets re-introduce this option with a little trait bound.


PolyMeilex (approved):
Yep, makes sense

PR #780, opened 2022-10-13
desktop/layer: Don't require dh for cleanup

Drakulix:
~~Draft because this requires a new wayland-server release to behave correctly. (`Weak::upgrade` may succeed for dead objects on 0.30-beta.10)~~

Also updates to ~~beta.11~~ beta.12


PolyMeilex (approved):
LGTM

PR #786, opened 2022-10-19
Destruct hook state

bwidawsk:
This allows passing back the compositor state on the destruction hook. Such functionality may be useful for handling rendering operations after a surface is destroyed.

### DIFF
--- a/src/wayland/compositor/mod.rs
+++ b/src/wayland/compositor/mod.rs
@@ -105,6 +105,8 @@ mod handlers;
 mod transaction;
 mod tree;
 
+use std::any::Any;
+
 pub use self::cache::{Cacheable, MultiCache};
 pub use self::handlers::{RegionUserData, SubsurfaceCachedState, SubsurfaceUserData, SurfaceUserData};
 use self::tree::PrivateSurfaceData;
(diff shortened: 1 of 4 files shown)


src/wayland/compositor/mod.rs
```diff
@@ -415,8 +417,9 @@ pub fn add_post_commit_hook(surface: &WlSurface, hook: fn(&DisplayHandle, &WlSur
 /// Register a destruction hook to be invoked on surface destruction
 ///
 /// It'll be invoked when the surface is destroyed (either explicitly by the client or on
-/// client disconnect).
-pub fn add_destruction_hook(surface: &WlSurface, hook: fn(&SurfaceData)) {
+/// client disconnect). The compositor state is passed back as `Any` and may be downcast as needed
+/// by the callee.
+pub fn add_destruction_hook(surface: &WlSurface, hook: fn(&mut dyn Any, &SurfaceData)) {
```
i509VCB, line 422: Could smithay provide a thunk so under the hood it's &mut dyn Any for preventing further generic expansion inside the implementation but still make the api type safe?

i509VCB: Ideally smithay would track a list of closures. This function would add a closure to the internal surface destroy hook list. This internal closure would just downcast Any to D (hint 'static bound) and then call whatever function was provided.

bwidawsk: Pasting relevant IRC conversation
```rust
let thunk = Box::new(|dyn_data, surface_data| {
    destroy_callback(dyn_data.downcast_mut::<D>().expect("generic on add_destruction_hook does not match dispatched data"), surface_data);
});

// Add the type erased closure to the destructor hooks list as a bunch of Box<Fn(...)>
```

bwidawsk:
I reworked my code and no longer care about this functionality. Can be reopened if needed later.

PR #788, opened 2022-10-21
egl: Don't error check eglQueryDmaBufModifiersEXT

Drakulix:
Fixes NVIDIA's 520 driver.

### DIFF
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -799,7 +799,17 @@ fn get_dmabuf_formats(
 
     for fourcc in formats {
         let mut num = 0i32;
-        wrap_egl_call(|| unsafe {
+        // Some drivers return EGL_BAD_PARAMETER here for formats
+        // they themselves returned for the query above.. *sigh*
+        //
+        // - NVIDIA proprietary since 520
+        //
+        // So lets ignore any errors of this call on purpose(!),
+        // which will let `num` stay at `0` and handle the format
+        // as unsupported by explicit modifiers.
+        // Which is probably what the error is suppose to indicate
+        // although the spec doesn't seem to demand it...
+        match wrap_egl_call(|| unsafe {
             ffi::egl::QueryDmaBufModifiersEXT(
                 *display,
                 fourcc as i32,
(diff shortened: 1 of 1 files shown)


src/backend/egl/display.rs
```diff
@@ -799,7 +799,17 @@ fn get_dmabuf_formats(
...
+        //
+        // - NVIDIA proprietary since 520
+        //
+        // So lets ignore any errors of this call on purpose(!),
+        // which will let `num` stay at `0` and handle the format
+        // as unsupported by explicit modifiers.
+        // Which is probably what the error is suppose to indicate
+        // although the spec doesn't seem to demand it...
+        unsafe {
             ffi::egl::QueryDmaBufModifiersEXT(
```
cmeissl, line 813: Is the only error that can be returned `EGL_BAD_PARAMETER` or are we actually interessted in other errors?
Should we explicitely handle that one error and maybe also explicitely set `num = 0` in that case or is num guaranteed to be untouched if an error occured?
Or maybe we want to at least log that error?

cmeissl (approved):
LGTM apart from some minor questions

src/backend/egl/display.rs
Drakulix: You are right, some safety percussions cannot hurt.

Does this look better?

cmeissl: 👍

cmeissl (approved):

src/backend/egl/display.rs
i509VCB: The EGL specification says that the function will return EGL_FALSE on failure

src/backend/egl/display.rs
```diff
@@ -799,7 +799,17 @@ fn get_dmabuf_formats(
...
+        // Some drivers return EGL_BAD_PARAMETER here for formats
+        // they themselves returned for the query above.. *sigh*
+        //
+        // - NVIDIA proprietary since 520
+        //
+        // So lets ignore any errors of this call on purpose(!),
+        // which will let `num` stay at `0` and handle the format
+        // as unsupported by explicit modifiers.
+        // Which is probably what the error is suppose to indicate
+        // although the spec doesn't seem to demand it...
```
i509VCB, line 811: This is indeed a driver bug as it disagrees with the spec.

I checked the EGL CTS and there is only testing for whether eglGetProcAddress exposed the symbols and not whether the implementation behaves properly: https://github.com/KhronosGroup/VK-GL-CTS/search?p=2&q=EGL_EXT_image_dma_buf_import_modifiers&type=

> * If <format> is not one of the formats advertised by QueryDmaBufFormatsEXT for the same <dpy>, the error EGL_BAD_PARAMETER is generated.

Have you filed a driver bug with Nvidia @Drakulix?

I guess if we wanted to be extra sure, one of us could write a pull request for the EGL CTS and add a new test for this?

PR #789, opened 2022-10-21
Feature/obscured window check

cmeissl:


src/backend/renderer/utils/mod.rs
```diff
@@ -300,3 +300,20 @@ pub struct SurfaceView {
...
+    test: impl IntoIterator<Item = Rectangle<N, Kind>>,
+    by: impl IntoIterator<Item = Rectangle<N, Kind>>,
+) -> bool {
+    let test = test.into_iter().collect::<Vec<Rectangle<N, Kind>>>();
+
+    by.into_iter()
+        .fold(test, |visible_regions, opaque_region| {
+            visible_regions
+                .into_iter()
+                .flat_map(|visible_region| visible_region.subtract_rect(opaque_region))
```
bwidawsk, line 315: I think you could use `contains_rect` here? It might be a little more readable that way.

cmeissl: Don't think this would be correct, a later opaque region could match the rest that was not contained earlier.

anvil/src/state.rs
```diff
@@ -381,23 +382,103 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
+                {
+                    layer.send_frame(self.start_time.elapsed().as_millis() as u32);
+                    opaque_regions.extend(layer.opaque_regions_with_popups().into_iter().map(|mut rect| {
+                        rect.loc += location;
+                        rect
+                    }));
+                }
+            });
+
+        self.space.elements().rev().for_each(|window| {
```
bwidawsk, line 436: What's the reasoning for `rev()`?

cmeissl: space.elements returns the elements in back to front order, but the opaque regions check operates front to back

anvil/src/state.rs
```diff
@@ -381,23 +382,103 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
+                                rect
+                            })
+                            .collect::<Vec<_>>(),
+                    )
+                {
+                    layer.send_frame(self.start_time.elapsed().as_millis() as u32);
+                    opaque_regions.extend(layer.opaque_regions_with_popups().into_iter().map(|mut rect| {
+                        rect.loc += location;
+                        rect
+                    }));
```
bwidawsk, line 432: These are just begging to be put in a macro or a generic function...

cmeissl:
closed in favor of #792

PR #790, opened 2022-10-21
Allow usage of KHR or EXT variants for eglSwapBuffersWithDamage

Drakulix:
The old implementation was arguably buggy (checking for either extension, but always calling EXT).
This might be a little too safe, but should work with every stupid driver out there.

Additionally this moves the buffer age query after `backend.bind()`. Otherwise the surface might not be current, resulting in another error and wrong age value.

(Fixes damage tracking issues with the winit-backend on nvidia's proprietary driver.)

### DIFF
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -517,6 +517,32 @@ impl EGLDisplay {
         &self.dmabuf_import_formats
     }
 
+    /// Returns when the display supports extensions required for smithays
+    /// damage tracking helpers.
+    pub fn supports_damage(&self) -> bool {
+        self.supports_damage_impl().supported()
+    }
+
+    pub(super) fn supports_damage_impl(&self) -> DamageSupport {
+        if self.extensions.iter().any(|ext| ext == "EGL_EXT_buffer_age") {
+            if self
+                .extensions
+                .iter()
+                .any(|ext| ext == "EGL_KHR_swap_buffers_with_damage")
+            {
+                return DamageSupport::KHR;
+            } else if self
+                .extensions
+                .iter()
+                .any(|ext| ext == "EGL_EXT_swap_buffers_with_damage")
+            {
+                return DamageSupport::EXT;
+            }
+        }
+
+        DamageSupport::No
+    }
+
     /// Exports an [`EGLImage`] as a [`Dmabuf`]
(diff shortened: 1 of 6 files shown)


src/backend/egl/surface.rs
```diff
@@ -143,11 +143,13 @@ impl EGLSurface {
     pub fn swap_buffers(
         &self,
         damage: Option<&mut [Rectangle<i32, Physical>]>,
+        damage_impl: DamageSupport,
```
i509VCB, line 146: Is there a reason someone would want to select the damage implementation?

I'd think it would be more ergonomic to just select the damage method internally?

i509VCB (changes requested):

src/backend/egl/surface.rs
Drakulix: Granted. Should be fixed

src/backend/egl/display.rs
```diff
@@ -517,6 +517,28 @@ impl EGLDisplay {
         &self.dmabuf_import_formats
     }
 
+    /// Returns when the display supports extensions required for smithays
+    /// damage tracking helpers.
+    pub fn supports_damage(&self) -> DamageSupport {
```
i509VCB, line 522: To the api user, this just needs to be a boolean? Interally we would need this enum still

Drakulix: Done.  enum is still public, because it is part of the `NativeSurface` trait signature

src/backend/egl/display.rs
```diff
@@ -1110,3 +1136,24 @@ pub struct PixelFormat {
     /// is srgb enabled
     pub srgb: bool,
 }
+
+/// Denotes if damage tracking is supported.
+///
+/// Additionally notes which variant of the `EGL_*_swap_buffers_with_damage` extension was found.
+/// Prefers `KHR` over `EXT`, if both are available.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum DamageSupport {
```
i509VCB, line 1145: One final thing, does this need to be public?

i509VCB (approved):

src/backend/egl/display.rs
Drakulix: Yes, it is part of the NativeSurface trait signature.

PR #792, opened 2022-10-22
Feature/renderer element state

cmeissl:
This PR adds a `OutputRenderReport` that holds the state for each `RenderElement`. The report can be used the query the `RenderElementState` for each element provided to the `render_output` function. The state is usefull to test if an element was skipped or directly scanned out.

The `send_frame` functions in desktop now allow to gate sending the frame callbacks by a provided function.
`desktop` now also offers a function to query the primary output for a window surface.

Anvil has been updated to use the report in combination with the send frame gate function to implement a simple
frame callback throttling for surfaces that have been skipped.
Additionally this also filters the frame callbacks to that they are not driven by the surface primary output.

same goal as #789 but with `wp_presentation_feedback` in mind
~~based on #793~~ merged and rebased

Todo:
- [ ] enhance docs
- [ ] review comments
- [ ] cleanup the mess in `PrimaryScanoutOutput::update_from_output_render_report`
- [ ] better `default_primary_scanout_output_compare` implementation based on refresh rate and visible portion

### DIFF
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -1,14 +1,19 @@
 use std::{
     os::unix::prelude::AsRawFd,
     sync::{atomic::AtomicBool, Arc, Mutex},
+    time::Duration,
 };
 
 use smithay::{
+    backend::renderer::element::{default_primary_scanout_output_compare, RenderElementStates},
     delegate_compositor, delegate_data_device, delegate_input_method_manager,
     delegate_keyboard_shortcuts_inhibit, delegate_layer_shell, delegate_output, delegate_primary_selection,
     delegate_seat, delegate_shm, delegate_tablet_manager, delegate_text_input_manager, delegate_viewporter,
     delegate_xdg_activation, delegate_xdg_decoration, delegate_xdg_shell,
-    desktop::{PopupManager, Space, Window},
+    desktop::{
+        utils::{surface_primary_scanout_output, update_surface_primary_scanout_output},
+        PopupManager, Space, Window,
+    },
     input::{keyboard::XkbConfig, pointer::CursorImageStatus, Seat, SeatHandler, SeatState},
     output::Output,
     reexports::{
@@ -388,15 +393,38 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
         }
     }
 
-    pub fn send_frames(&self, output: &Output) {
+    pub fn send_frames(&self, output: &Output, render_element_states: &RenderElementStates) {
+        let time = self.start_time.elapsed();
+        let throttle = Some(Duration::from_secs(1));
+
         self.space.elements().for_each(|window| {
+            window.with_surfaces(|surface, states| {
+                update_surface_primary_scanout_output(
+                    surface,
+                    output,
+                    states,
+                    render_element_states,
+                    default_primary_scanout_output_compare,
+                )
+            });
+
             if self.space.outputs_for_element(window).contains(output) {
-                window.send_frame(self.start_time.elapsed().as_millis() as u32)
+                window.send_frame(output, time, throttle, surface_primary_scanout_output);
             }
         });
         let map = smithay::desktop::layer_map_for_output(output);
         for layer_surface in map.layers() {
-            layer_surface.send_frame(self.start_time.elapsed().as_millis() as u32)
+            layer_surface.with_surfaces(|surface, states| {
+                update_surface_primary_scanout_output(
+                    surface,
+                    output,
+                    states,
+                    render_element_states,
+                    default_primary_scanout_output_compare,
+                )
+            });
+
+            layer_surface.send_frame(output, time, throttle, surface_primary_scanout_output);
         }
     }
 }
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -34,7 +34,7 @@ use smithay::{
         libinput::{LibinputInputBackend, LibinputSessionInterface},
         renderer::{
             damage::{DamageTrackedRenderer, DamageTrackedRendererError},
-            element::{texture::TextureBuffer, AsRenderElements},
+            element::{texture::TextureBuffer, AsRenderElements, RenderElementStates},
             gles2::{Gles2Renderbuffer, Gles2Renderer},
             multigpu::{egl::EglGlesBackend, GpuManager, MultiRenderer, MultiTexture},
             Bind, Frame, Renderer,
@@ -789,8 +789,8 @@ impl AnvilState<UdevData> {
                 &mut *self.cursor_status.lock().unwrap(),
                 &self.log,
             );
-            let reschedule = match result {
-                Ok(has_rendered) => !has_rendered,
+            let reschedule = match &result {
+                Ok((has_rendered, _)) => !has_rendered,
                 Err(err) => {
                     warn!(self.log, "Error during rendering: {:?}", err);
                     match err {
@@ -844,7 +844,7 @@ fn render_surface<'a>(
     dnd_icon: &Option<wl_surface::WlSurface>,
     cursor_status: &mut CursorImageStatus,
     logger: &slog::Logger,
-) -> Result<bool, SwapBuffersError> {
+) -> Result<(bool, RenderElementStates), SwapBuffersError> {
     surface.surface.frame_submitted()?;
 
     let output_geometry = space.output_geometry(output).unwrap();
@@ -935,18 +935,18 @@ fn render_surface<'a>(
         age.into(),
         logger,
     )
-    .map(|x| x.is_some());
+    .map(|(damage, states)| (damage.is_some(), states));
 
     match render_res.map_err(|err| match err {
         DamageTrackedRendererError::Rendering(err) => err.into(),
         _ => unreachable!(),
     }) {
-        Ok(true) => {
+        Ok((true, states)) => {
             surface
                 .surface
                 .queue_buffer()
                 .map_err(Into::<SwapBuffersError>::into)?;
-            Ok(true)
+            Ok((true, states))
         }
         x => x,
     }
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -820,13 +820,15 @@ impl AnvilState<UdevData> {
                     .expect("failed to schedule frame timer");
             }
 
-            outputs.push(output.clone());
+            if let Ok((_, states)) = result {
+                outputs.push((output.clone(), states));
+            }
         }
 
         std::mem::drop(surfaces);
-        for output in outputs {
+        for (output, states) in outputs {
             // Send frame events so that client start drawing their next frame
-            self.send_frames(&output);
+            self.send_frames(&output, &states);
         }
     }
 }
--- a/src/backend/renderer/element/mod.rs
+++ b/src/backend/renderer/element/mod.rs
@@ -23,7 +23,10 @@ use std::{collections::HashMap, sync::Arc};
 #[cfg(feature = "wayland_frontend")]
 use wayland_server::{backend::ObjectId, protocol::wl_buffer, Resource};
 
-use crate::utils::{Buffer as BufferCoords, Physical, Point, Rectangle, Scale, Size, Transform};
+use crate::{
+    output::{Output, WeakOutput},
+    utils::{Buffer as BufferCoords, Physical, Point, Rectangle, Scale, Size, Transform},
+};
 
 use super::{utils::CommitCounter, Renderer};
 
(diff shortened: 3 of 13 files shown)


anvil/src/state.rs
```diff
@@ -388,15 +390,85 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
         for layer_surface in map.layers() {
-            layer_surface.send_frame(self.start_time.elapsed().as_millis() as u32)
+            layer_surface.send_frame(time, |surface, states| {
+                send_throttled_frame(surface, states, report, now, throttle)
+            });
+        }
+    }
+}
+
+fn send_throttled_frame(
```
Drakulix, line 432: I find this name a little confusing, given that the frame is only throttled, when the report shows it as skipped.

anvil/src/state.rs
```diff
@@ -388,15 +390,85 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
+    states
+        .data_map
+        .insert_if_missing_threadsafe(SurfaceFrameThrottlingState::default);
+    let send_throttled_frame = states
+        .data_map
+        .get::<SurfaceFrameThrottlingState>()
+        .unwrap()
+        .update(tp, throttle);
+
+    !throttle_frame || send_throttled_frame
```
Drakulix, line 453: Again the naming, especially with the inverse logic, makes this expression hard to parse.
Something like:
`!was_skipped || frame_overdue` or something along those lines would be easier to understand.

anvil/src/state.rs
```diff
@@ -388,15 +390,85 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
+    surface: &WlSurface,
+    states: &smithay::wayland::compositor::SurfaceData,
+    report: &OutputRenderReport,
+    tp: Instant,
+    throttle: Option<Duration>,
+) -> bool {
+    let throttle_frame = report
+        .element_render_state(surface)
+        .map(|state| state == RenderElementState::Skipped)
+        .unwrap_or(true);
```
Drakulix, line 442: Maybe event add a function `was_rendered` to the report, that returns false for both `Skipped` or not present elements.

src/desktop/space/mod.rs
```diff
@@ -666,3 +670,54 @@ where
 
     damage_tracked_renderer.render_output(renderer, age, &*render_elements, clear_color, log)
 }
+
+/// Find the primary output for a [`WlSurface`] of a window out of a set of outputs
```
Drakulix, line 674: The docs should mention (and maybe the name should be changed to reflect this), that this essentially tries to find the preferred frame rate for a given surface.

For shell-interactions "primary"-output might instead be the output of most overlap, independent of the refresh rate. Using this function for those cases would result in a bad user experience preferring high-refresh outputs over anything else.

That said the function is perfectly fine (and quite clever) for this use-case, it is just not apparent from the naming and documentation, what criteria are used to define "primary".

src/desktop/space/mod.rs
```diff
@@ -668,3 +670,54 @@ where
...
+///
+/// Returns `None` if no primary output could be determined
```
Drakulix, line 676: Maybe expand on this. In which cases can this happen? If the output is not "visible" on any output?

What visible means in this context is also questionable. Again using the output state here is quite clever, because it works independently from using either the `Space`-abstraction or driving `output_enter` and `output_leave` manually. But for downstream this is not apparent.

Drakulix (commented):
Overall looks very solid, just some remarks on readability/docs of the source code.

src/desktop/space/mod.rs
cmeissl: yeah, valid points. I will try to find a different name and document what this functiont does. Maybe it would be a good idea to also provide a function that returns the output of most overlap while at it?

Drakulix: > Maybe it would be a good idea to also provide a function that returns the output of most overlap while at it?

Likely, I have implemented such functionality in both anvil and cosmic-comp at some point, so there is likely a useful helper function

anvil/src/state.rs
cmeissl: renamed to `should_send_frame_callback`

anvil/src/state.rs
cmeissl: now uses `presented || frame_overdue`

anvil/src/state.rs
cmeissl: done, `element_was_presented` function added which returns `false` if not found or skipped

src/desktop/space/mod.rs
cmeissl: renamed to `window_surface_preferred_output` and mentioned the refresh rate in the docs.
also added `window_surface_output_most_overlap` for shell-interactions

src/desktop/space/mod.rs
cmeissl: Updated the docs to include the reason why `None` would be returned and added the requirement of `output_enter`/`output_leave`

Drakulix (approved):
LGTM

anvil/src/state.rs
```diff
@@ -388,15 +390,83 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
+            // as all further checks for the primary output
+            // will fail as well
+            if !outputs.contains(output) {
+                continue;
             }
-        });
+
+            window.send_frame(time, |surface, states| {
+                // We only want to send frame callbacks on the preferred output
+                if !window_surface_preferred_output(surface, &*outputs)
```
bwidawsk, line 411: Shouldn't the send_frame only occur after all outputs the surface is shown on have had their part of the surface presented?

src/backend/renderer/element/mod.rs
```diff
@@ -89,6 +96,50 @@ pub enum UnderlyingStorage<'a, R: Renderer> {
...
 }
 
+/// Defines the possible states of an element
+/// after rendering
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RenderElementState {
+    /// The element was rendered
+    Rendered,
+    /// The element was directly scanned out
+    ScannedOut,
```
bwidawsk, line 106: Ideally this would perhaps recorder the presented time. Perhaps `ScannedOut(Duration)`

src/backend/renderer/element/mod.rs
```diff
@@ -89,6 +96,50 @@ pub enum UnderlyingStorage<'a, R: Renderer> {
...
+    /// The element was skipped
+    Skipped,
+}
+
+/// Report for an output render operation
+#[derive(Debug, Clone)]
+pub struct OutputRenderReport {
+    /// Holds the damage for presenting the bound buffer
+    pub damage: Option<Vec<Rectangle<i32, Physical>>>,
```
bwidawsk, line 115: I'd drop fallability and just make the empty vec be equivalent to None.

src/backend/renderer/element/mod.rs
```diff
@@ -89,6 +96,50 @@ pub enum UnderlyingStorage<'a, R: Renderer> {
...
+    ///
+    /// Return `None` if the element was not part of the output render operation
+    pub fn element_render_state(&self, id: impl Into<Id>) -> Option<RenderElementState> {
+        self.element_render_states.get(&id.into()).copied()
+    }
+
+    /// Returns whether the element with the specified id was presented
+    ///
+    /// Returns `false` if the element with the id was not found or skipped
+    pub fn element_was_presented(&self, id: impl Into<Id>) -> bool {
```
bwidawsk, line 136: This name is confusing IMO. Presented should mean exactly that the state is ScannedOut. I think just `element_was_skipped` should accomplish the same thing afaict.

anvil/src/state.rs
cmeissl: Why? That would limit the frame callbacks to the lowest refresh rate of all outputs it is shown.
The opposite is what you want most of the time, use the output with the highest refresh rate to drive the frame callbacks.

src/backend/renderer/element/mod.rs
cmeissl: Okay, totally my fault, this is named pretty badly. That only means it has been assigned to a plane for scan out.

src/backend/renderer/element/mod.rs
cmeissl: Right, we could also check for an empty vec

src/backend/renderer/element/mod.rs
cmeissl: Same as above, the state represents the rendering and not the scan out, again probably poorly named by me. The check should explicitly return true if the element was not skipped irregardless if the element has been sampled or assigned to a plane.

anvil/src/state.rs
bwidawsk: Looking at this from a power efficiency standpoint, there's no point in starting the client's next draw until all parts of the last one have been displayed.

Suppose you have 2 outputs side by side where the window is spanned across both, one is 120Hz the other is 30Hz. Why would you kick off the next draw faster than 30Hz?

cmeissl: Because 30 fps will look pretty bad on a 120Hz monitor? Or are least not so nice how it could look like, I mean what is the point in running a monitor with that refresh rate when you limit clients then? The output with the lower refresh rate would sample the buffers at it's refresh rate.
But it may be a valid point for your use case and this is why I struggle what should be put in smithay and what in anvil.

bwidawsk: Summing up IRC conversation. This is making a choice between power (and overall time where the windows remain in sync), and performance. For the sake of making progress, let's put a pin in it, since I think coverage is a good general heuristic. Perhaps in the future an argument can be added for providing a strategy:
coverage
slowest
fastest

cmeissl:
Just noticed that this architecture in the desktop module will force running the primary scan-out output selection twice when integrating wp_presentation. Maybe a two step approach would be a better fit where the first step runs the selection and the second uses the selected output. So with wp_presentation we would do select output, send frames and send feedback per window? Every function would need to walk the whole surface tree. Or to use a single function instead of send_frames that does send both, frame callbacks and feedback? @Drakulix @bwidawsk opinions?

cmeissl:
> > Or to use a single function instead of send_frames that does send both, frame callbacks and feedback?
> 
> I think I am in favour of this. Generally I would expect every compositor to want to support wp_presentation anyway and not fake any values. (Similar to wp_viewporter.) So coupling the code with this protocol a bit more, should not be problematic, as long as downstream could still decide to not initialize the provided WpPresentation global handler.

Yeah, sounds good. Both things should probably be sent on vblank anyway. If downstream does not initialize wp_presentation there should be no feedback callbacks to send, so that should be fine.

Drakulix (approved):

src/desktop/wayland/layer.rs
```diff
@@ -562,14 +563,40 @@ impl LayerSurface {
...
+        for (popup, _) in PopupManager::popups_for_surface(surface) {
+            let surface = popup.wl_surface();
+            send_frames_surface_tree(surface, output, time, throttle, primary_scan_out_output);
+        }
+    }
 
-        send_frames_surface_tree(wl_surface, time);
-        for (popup, _) in PopupManager::popups_for_surface(wl_surface) {
-            send_frames_surface_tree(popup.wl_surface(), time);
+    /// Run a closure on all surfaces in this layer
```
Drakulix, line 589:
```suggestion
    /// Run a closure on all surfaces in this layer (including it's popups if [`PopupManager`] is used)
```

src/desktop/wayland/window.rs
```diff
@@ -218,14 +220,40 @@ impl Window {
...
+        let surface = self.0.toplevel.wl_surface();
+
+        send_frames_surface_tree(surface, output, time, throttle, primary_scan_out_output);
+        for (popup, _) in PopupManager::popups_for_surface(surface) {
+            let surface = popup.wl_surface();
+            send_frames_surface_tree(surface, output, time, throttle, primary_scan_out_output);
+        }
+    }
+
+    /// Run a closure on all surfaces in this window
```
Drakulix, line 246:
```suggestion
    /// Run a closure on all surfaces in this window (including it's popups, if [`PopupManager`] is used)
```

Drakulix (commented):
Little nitpick for the docs.

anvil/src/udev.rs
```diff
@@ -935,18 +937,18 @@ fn render_surface<'a>(
         age.into(),
         logger,
     )
-    .map(|x| x.is_some());
+    .map(|(damage, report)| (damage.is_some(), report));
```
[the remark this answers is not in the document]
cmeissl: yes, thx

src/desktop/wayland/layer.rs
cmeissl: :+1:

src/desktop/wayland/window.rs
cmeissl: :+1:

Drakulix (approved):

PR #794, opened 2022-10-24
Feature/wp presentation

cmeissl:
An attempt to support `wp_presentation`

based on #792 
depends on [wayland-rs#574](https://github.com/Smithay/wayland-rs/pull/574)

### DIFF
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -7,11 +7,15 @@ use std::{
 use smithay::{
     backend::renderer::element::{default_primary_scanout_output_compare, RenderElementStates},
     delegate_compositor, delegate_data_device, delegate_input_method_manager,
-    delegate_keyboard_shortcuts_inhibit, delegate_layer_shell, delegate_output, delegate_primary_selection,
-    delegate_seat, delegate_shm, delegate_tablet_manager, delegate_text_input_manager, delegate_viewporter,
-    delegate_virtual_keyboard_manager, delegate_xdg_activation, delegate_xdg_decoration, delegate_xdg_shell,
+    delegate_keyboard_shortcuts_inhibit, delegate_layer_shell, delegate_output, delegate_presentation,
+    delegate_primary_selection, delegate_seat, delegate_shm, delegate_tablet_manager,
+    delegate_text_input_manager, delegate_viewporter, delegate_virtual_keyboard_manager,
+    delegate_xdg_activation, delegate_xdg_decoration, delegate_xdg_shell,
     desktop::{
-        utils::{surface_primary_scanout_output, update_surface_primary_scanout_output},
+        utils::{
+            surface_presentation_feedback_flags_from_states, surface_primary_scanout_output,
+            update_surface_primary_scanout_output, OutputPresentationFeedback,
+        },
         PopupManager, Space, Window,
     },
     input::{keyboard::XkbConfig, pointer::CursorImageStatus, Seat, SeatHandler, SeatState},
@@ -28,7 +32,7 @@ use smithay::{
             Display, DisplayHandle, Resource,
         },
     },
-    utils::{Logical, Point},
+    utils::{Clock, Logical, Monotonic, Point},
     wayland::{
         compositor::CompositorState,
         data_device::{
@@ -40,6 +44,7 @@ use smithay::{
             KeyboardShortcutsInhibitHandler, KeyboardShortcutsInhibitState, KeyboardShortcutsInhibitor,
         },
         output::OutputManagerState,
+        presentation::PresentationState,
         primary_selection::{set_primary_focus, PrimarySelectionHandler, PrimarySelectionState},
         seat::WaylandFocus,
         shell::{
@@ -106,6 +111,7 @@ pub struct AnvilState<BackendData: 'static> {
     pub xdg_activation_state: XdgActivationState,
     pub xdg_decoration_state: XdgDecorationState,
     pub xdg_shell_state: XdgShellState,
+    pub presentation_state: PresentationState,
 
     pub dnd_icon: Option<WlSurface>,
     pub log: slog::Logger,
@@ -116,7 +122,7 @@ pub struct AnvilState<BackendData: 'static> {
     pub cursor_status: Arc<Mutex<CursorImageStatus>>,
     pub seat_name: String,
     pub seat: Seat<AnvilState<BackendData>>,
-    pub start_time: std::time::Instant,
+    pub clock: Clock<Monotonic>,
 
     #[cfg(feature = "xwayland")]
     pub xwayland: XWayland,
@@ -264,6 +270,7 @@ delegate_xdg_decoration!(@<BackendData: Backend + 'static> AnvilState<BackendDat
 
 delegate_xdg_shell!(@<BackendData: Backend + 'static> AnvilState<BackendData>);
 delegate_layer_shell!(@<BackendData: 'static> AnvilState<BackendData>);
+delegate_presentation!(@<BackendData: 'static> AnvilState<BackendData>);
 
 impl<BackendData: Backend + 'static> AnvilState<BackendData> {
     pub fn init(
@@ -273,6 +280,8 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
         log: slog::Logger,
         listen_on_socket: bool,
     ) -> AnvilState<BackendData> {
+        let clock = Clock::new().expect("failed to initialize clock");
+
         // init wayland clients
         let socket_name = if listen_on_socket {
             let source = ListeningSocketSource::new_auto(log.clone()).unwrap();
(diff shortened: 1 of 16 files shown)


anvil/src/udev.rs
```diff
@@ -700,6 +719,62 @@ impl AnvilState<UdevData> {
...
+        std::mem::drop(surfaces);
+
+        let tp = metadata.as_ref().and_then(|metadata| match metadata.time {
+            smithay::backend::drm::DrmEventTime::Monotonic(tp) => Some(tp),
+            smithay::backend::drm::DrmEventTime::Realtime(_) => None,
+        });
+        let seq = metadata.as_ref().map(|metadata| metadata.sequence).unwrap_or(0);
+
+        let (tv_sec, tv_nsec) = if let Some(tp) = tp {
+            let zero = unsafe { std::mem::zeroed::<std::time::Instant>() };
```
Drakulix, line 768: I have used this hack before, but this needs a SAFETY note, that explains, that this relies on the internal representation of `std::time::Instant` on unix.

anvil/src/udev.rs
```diff
@@ -700,6 +719,62 @@ impl AnvilState<UdevData> {
...
+        };
+
+        let report = match surface.borrow_mut().output_render_report.take() {
+            Some(report) => report,
+            None => return,
+        };
+        std::mem::drop(surfaces);
+
+        let tp = metadata.as_ref().and_then(|metadata| match metadata.time {
+            smithay::backend::drm::DrmEventTime::Monotonic(tp) => Some(tp),
```
Drakulix, line 762: This case should also set the `hw_clock` (and possibly even the `hw_completion`) flags of the protocol, no?

Drakulix (commented):
Looks good so far! Couple of remarks

anvil/src/udev.rs
cmeissl: yeah, right. I definitely do not like that at all and atm the only thing we do is to convert the `Duration` in the source to an `Instant` and then back to a `Duration` here, so maybe we should remove the hack in the source and find a different solution?

anvil/src/udev.rs
cmeissl: yeah, I think both should be set in this case. I think the best way would be make the ***base*** flags a parameter of `send_frames` so that the backend can provide them. For drm that would be `vsync | hw_clock | hw_completion` if we get a monotonic metadata. If we have no metadata or not a monotonic time we could fall back to `clock_gettime` and only send `vsync`. Does that sound good?

anvil/src/udev.rs
Drakulix: Yeah, we should just fix this to return a `Duration` instead of doing this hack twice: https://github.com/cmeissl/smithay/blob/feature/wp_presentation/src/backend/drm/device/mod.rs#L390

anvil/src/state.rs
```diff
@@ -418,14 +436,55 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
+                        flags |= wp_presentation_feedback::Kind::ZeroCopy;
+                    }
+                    flags
+                })
+            })
+        }
+    }
+}
+
+pub fn clock_get_time(clk_id: libc::clockid_t) -> Result<libc::timespec, std::io::Error> {
```
Drakulix, line 478: Do we need this function? `std::time::Instant` is defined as using `clock_gettime` with the monotonic clock.
https://doc.rust-lang.org/nightly/std/time/struct.Instant.html#underlying-system-calls

cmeissl: Yeah, but we need tv_sec and tv_nsec for wp_presentation_feedback and I don't see how we could get that from Instant without some hacks (looking at you std::mem::zeroed).
The client has to be able to use the same clk_id to calculate time diffs, so we need to use the same base AFAIK.

Drakulix: Yeah, this is quite annoying. `std::time::Instant` really needs some function, that allow for C-interop. Preferably just a `Instant::ZERO` to get a zero-base, but that probably doesn't mean much on other systems.

cmeissl: I don't think this would be helpful, the intended use of `Instant` is just not well suited for our use-case. The fact that it uses `timespec` under the hood is just an implementation detail. Beside that `Instant` explicitely does not define a base so it is unsuitable for transporting a timepoint which requires a known base.

.github/workflows/ci.yml
```diff
@@ -86,6 +86,11 @@ jobs:
         with:
           command: update
           args: -p log --precise 0.4.14
+      - name: Downgrade bytemuck
```
i509VCB, line 89: These downgrade steps really do make our buildscripts quite ugly... Maybe we just want to force an explicit version of bytemuck in Smithay?

Although I believe a dependency has this set.

src/backend/allocator/swapchain.rs
```diff
@@ -107,6 +109,26 @@ impl<B: Buffer> Deref for Slot<B> {
     }
 }
 
+impl<B: Buffer> Clone for Slot<B> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
```
Drakulix, line 116: Hmm, no allowing users to clone the `Slot` was a very deliberate design decision, as most of the `Slot`s magic happens on drop. I wanted to avoid issues where users were unintentionally kept more references around and thus missed freeing the slot at the right time.

Does this need to be public (I don't see any usage of it in anvil, if I am not mistaken)? Or could it be an impl-function with proper docs to convey the use-case of this?

src/backend/allocator/swapchain.rs
```diff
@@ -107,6 +109,26 @@ impl<B: Buffer> Deref for Slot<B> {
...
+    fn export(&self) -> Result<super::dmabuf::Dmabuf, Self::Error> {
+        let maybe_dmabuf = self.userdata().get::<Dmabuf>();
+        if maybe_dmabuf.is_none() {
+            let dmabuf = (**self).export()?;
+            self.userdata().insert_if_missing(|| dmabuf);
+        }
+
+        Ok(self.userdata().get::<Dmabuf>().cloned().unwrap())
+    }
+}
```
Drakulix, line 130: So `Slot`s transparently cache dmabufs? Sounds useful!

anvil/src/udev.rs
```diff
@@ -700,6 +709,119 @@ impl AnvilState<UdevData> {
...
+            }
+        };
+
+        if schedule_render {
+            let output_refresh = match output.current_mode() {
+                Some(mode) => mode.refresh,
+                None => return,
+            };
+            let repaint_delay =
+                Duration::from_millis(((1_000_000f32 / output_refresh as f32) * 0.6f32) as u64);
```
Drakulix, line 808: So the delay is longer, if the refresh-cycle is longer? That seems not very useful for very high refresh-rates, where we might select a too short duration and thus miss every other frame.

An optimal implementation would likely keep the last n rendering-times and do a sliding window average. That is overkill for anvil, but I feel like a bad example is worse than no example. Any ideas how we could make this less stupid without making it perfect?

Drakulix (commented):
Overall I am super happy how this turned out!

src/backend/allocator/swapchain.rs
cmeissl: This is a really good point. In fact this change would break the contract, Drop will always set acquired to false for each clone. I will revert that change as it is no longer necessary, if I am not mistaken I only added it for the frame experiment I already removed again.

anvil/src/udev.rs
cmeissl: I choose this value cause I expect it to work fine for refresh rates up to 120Hz. For 120 Hz the compositor would have a bit more than 3 ms to render. My basic assumption for that was that if a user is able to drive such a high refresh rate the hardware would also be most likely capable of rendering the frame in a shorter period. On lower refresh rates a client would have more time to render. Like you said it is quite hard to select a reasonable delay. Even a sliding window with average has some limitations with spikes for example. I am not a fan of solely relying on some configuration parameter either because the topic may be hard to understand for an average user. On the other hand every hard coded value would be based on some specifc assumptions like the performance of the machine it was developed on. I would have skipped implementing the delay at all, but without it wp_presentation is rather useless imo as it would even increase the latency for some clients.

Currently I have no idea how that could be made better without introducing a dynamically calculated delay based on past render times and also some prediction.
But it definitely needs some comment describing why we want to delay the repaint.

Drakulix: Seems reasonable. Once some docs are added, I would consider this resolved.

src/backend/allocator/swapchain.rs
cmeissl: done

anvil/src/udev.rs
cmeissl: done

Drakulix (approved):

PR #795, opened 2022-10-24
renderer: dtr skip elements hidden on output

cmeissl:
we can skip elements that are completely hidden on our output, not only if they are completely hidden.
without that change the dtr would only skip a element if its complete geometry was hidden, but we
are only interessted in geometry that intersects with the output.


Drakulix (approved):
:+1:

PR #796, opened 2022-10-28
shm: writeback support

Drakulix:
Draft PR, because I still need to test the new functionality in a compositor, but ready for review and anvil still runs fine!

### DIFF
--- a/src/wayland/shm/mod.rs
+++ b/src/wayland/shm/mod.rs
@@ -80,7 +80,11 @@
 //!         /* The client supplied invalid content specification for this buffer,
 //!            and was killed.
 //!          */
-//!     }
+//!     },
+//!     Err(BufferAccessError::NotReadable) => {
+//!         /* The client has not allowed reads to this buffer */
+//!     },
+//!     Err(BufferAccessError::NotWritable) => unreachable!("cannot be triggered by with_buffer_contents"),
 //! }
 //! # }
 //! ```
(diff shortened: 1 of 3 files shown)


src/wayland/shm/mod.rs
```diff
@@ -219,6 +219,35 @@ where
...
+/// If the buffer is managed by the provided `ShmGlobal`, its contents are
+/// extracted and the closure is extracted with them:
+///
+/// - The first argument is a data slice of the contents of the pool
+/// - The second argument is the specification of this buffer is this pool
+///
+/// If the buffer is not managed by the provided `ShmGlobal`, the closure is not called
+/// and this method will return `Err(BufferAccessError::NotManaged)` (this will be the case for an
+/// EGL buffer for example).
+pub fn with_buffer_contents_mut<F, T>(buffer: &wl_buffer::WlBuffer, f: F) -> Result<T, BufferAccessError>
```
i509VCB, line 233: So we want to guard against the future access flags in wl_shm V3? Not every error will end up being NotManaged in the future

Drakulix: Do you have some reference, what might happen in the future?

i509VCB: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/161/ is the proposal

eli5: In version 3, a new request is added which allows specifying what the server can do with the buffer (read and or write). For v2 and v1, assume the buffer is always readable and writable.

I believe there are some semantics changing related to resizing, but those are out to scope for this pull request.

Drakulix: Ok, so we want a `not-writable` error type, got it.

Drakulix: Added new Error-variants for non-readable or non-writable buffers.

i509VCB (approved):
Looks good, actually wiring up the permissions for read/write flags is probably for a different day.

PR #797, opened 2022-11-01
wayland/seat: Add method to retrieve wl_seats for client

Drakulix:


src/wayland/seat/mod.rs
```diff
@@ -186,6 +186,19 @@ impl<D: SeatHandler + 'static> Seat<D> {
             .map(|arc| Self { arc })
     }
 
+    /// Retrieves [`WlSeat`](wl_seat::WlSeat) resources for a given client
+    pub fn resources_for_client(&self, client: &ClientId) -> Vec<WlSeat> {
```
cmeissl, line 190: We probably should be consistent with Output, which provides a with_client_outputs function that takes a closure.

cmeissl (changes requested):

src/wayland/seat/mod.rs
PolyMeilex: I aggree, not saing that one signature or name is better than the other, just that it would be nice to have them be similar.

cmeissl: @vberger pointed out that ClientId was not meant to be used directly when I wanted to change Output to take a ClientId. With this [commit](https://github.com/Smithay/wayland-rs/commit/13f6a9beb05eabac56aa372e1d38b12bfa987982) (release pending) this should also be no longer necessary.

Drakulix: k, changing it again

Drakulix:
> Random question, why are we using callbacks for this, both in output and seat we are collecting cloned instances to Vec before looping over them and calling the CB, couldn't we just return that Vec in both cases?

I think the callbacks were used for not having to clone the contents, but the problem with that approach is, that it locks the inner of output/seat, which caused deadlocks in the past. (Turns out if you interact with the wayland objects, you usually also want to interact with the representation...)
So yeah, we could make it return a `Vec` easily. Would you prefer that?

PolyMeilex:
> So yeah, we could make it return a `Vec` easily. Would you prefer that?

Yep, I think that would be better, it's easier to do find/filter/map and all other iterator goodies that way

PolyMeilex (approved):

PR #798, opened 2022-11-01
input/keyboard: Add way to access currently pressed keys

Drakulix:
This is e.g. useful to simulate `KeyboardTarget::enter` events, when internally multiplexing a `KeyboardTarget` implementation.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -580,6 +580,24 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
         );
     }
 
+    /// Iterate over the keysyms of the currently pressed keys.
+    pub fn with_pressed_keysyms<F>(&self, mut f: F)
+    where
+        F: FnMut(KeysymHandle<'_>),
+    {
+        let guard = self.arc.internal.lock().unwrap();
+        {
+            for code in guard.pressed_keys.iter() {
+                let handle = KeysymHandle {
+                    keycode: code + 8,
+                    state: &guard.state,
+                    keymap: &guard.keymap,
+                };
+                f(handle);
+            }
+        }
+    }
+
     /// Check if keyboard has focus
     pub fn is_focused(&self) -> bool {
         self.arc.internal.lock().unwrap().focus.is_some()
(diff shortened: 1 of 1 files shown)


src/input/keyboard/mod.rs
```diff
@@ -556,6 +556,24 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
         );
     }
 
+    /// Iterate over the keysyms of the currently pressed keys.
+    pub fn with_pressed_keysyms<F>(&self, mut f: F)
+    where
+        F: FnMut(KeysymHandle<'_>),
```
i509VCB, line 562: Would it be more useful to have the closure provided an iterator rather than running multiple times?

i509VCB, line 562: Should this have a return value?

Drakulix: Not sure, I just followed the other apis. Due to borrowing state behind a Mutex for the KeysymHandle, we sadly can't simply return an Iterator...

I suppose this is a more general question about how we want to approach the few `with_` functions, that we have?

Drakulix: Maybe for the case of running the closure once with an iterator, but the semantics of merging results for running this multiple times, seem to be unnecessary complex.

Drakulix: ping @i509VCB

src/input/keyboard/mod.rs
```diff
@@ -580,6 +580,24 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
         );
     }
 
+    /// Iterate over the keysyms of the currently pressed keys.
+    pub fn with_pressed_keysyms<F>(&self, mut f: F)
```
bwidawsk, line 584: Without seeing the user, it's hard to say but it seems like this shouldn't be `mut`

Drakulix: Sure, it needs to be. It's an `FnMut`.

Drakulix:
Replaced with https://github.com/Smithay/smithay/pull/880

PR #800, opened 2022-11-02
renderer: Add `Blit`-trait and implementations

Drakulix:
Adds a new trait `Blit` that can be used to do copies between framebuffers.
These are usually more performant, than exporting the source framebuffer and rendering the resulting texture to the target framebuffer, if supported.

### DIFF
--- a/src/backend/egl/context.rs
+++ b/src/backend/egl/context.rs
@@ -228,7 +228,16 @@ impl EGLContext {
     /// This function is marked unsafe, because the context cannot be made current on another thread without
     /// being unbound again (see [`EGLContext::unbind`]).
     pub unsafe fn make_current(&self) -> Result<(), MakeCurrentError> {
-        self.make_current_with_draw_and_read_surface(None, None)
+        wrap_egl_call(|| {
+            ffi::egl::MakeCurrent(
+                **self.display.get_display_handle(),
+                ffi::egl::NO_SURFACE,
+                ffi::egl::NO_SURFACE,
+                self.context,
+            )
+        })
+        .map(|_| ())
+        .map_err(Into::into)
     }
 
     /// Makes the OpenGL context the current context in the current thread with a surface to
@@ -239,7 +248,7 @@ impl EGLContext {
     /// This function is marked unsafe, because the context cannot be made current on another thread without
     /// being unbound again (see [`EGLContext::unbind`]).
     pub unsafe fn make_current_with_surface(&self, surface: &EGLSurface) -> Result<(), MakeCurrentError> {
-        self.make_current_with_draw_and_read_surface(Some(surface), Some(surface))
+        self.make_current_with_draw_and_read_surface(surface, surface)
     }
 
     /// Makes the OpenGL context the current context in the current thread with surfaces to
@@ -251,15 +260,11 @@ impl EGLContext {
     /// being unbound again (see [`EGLContext::unbind`]).
     pub unsafe fn make_current_with_draw_and_read_surface(
         &self,
-        draw_surface: Option<&EGLSurface>,
-        read_surface: Option<&EGLSurface>,
+        draw_surface: &EGLSurface,
+        read_surface: &EGLSurface,
     ) -> Result<(), MakeCurrentError> {
-        let draw_surface_ptr = draw_surface
-            .map(|s| s.surface.load(Ordering::SeqCst))
-            .unwrap_or(ffi::egl::NO_SURFACE as _);
-        let read_surface_ptr = read_surface
-            .map(|s| s.surface.load(Ordering::SeqCst))
-            .unwrap_or(ffi::egl::NO_SURFACE as _);
+        let draw_surface_ptr = draw_surface.surface.load(Ordering::SeqCst);
+        let read_surface_ptr = read_surface.surface.load(Ordering::SeqCst);
         wrap_egl_call(|| {
             ffi::egl::MakeCurrent(
                 **self.display.get_display_handle(),
--- a/src/backend/egl/context.rs
+++ b/src/backend/egl/context.rs
@@ -221,27 +221,6 @@ impl EGLContext {
         })
     }
 
-    /// Makes the OpenGL context the current context in the current thread with a surface to
-    /// read/write to.
-    ///
-    /// # Safety
-    ///
-    /// This function is marked unsafe, because the context cannot be made current on another thread without
-    /// being unbound again (see [`EGLContext::unbind`]).
-    pub unsafe fn make_current_with_surface(&self, surface: &EGLSurface) -> Result<(), MakeCurrentError> {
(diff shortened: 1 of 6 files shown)


src/backend/renderer/mod.rs
```diff
@@ -568,6 +568,36 @@ pub trait ExportDma: Renderer {
...
+    /// The currently bound target might get unbound during this operation and not restored.
+    ///
+    /// This function *may* fail, if (but not limited to):
+    /// - The source framebuffer is not readable
+    /// - The destination framebuffer is not writable
+    /// - `src` is out of bounds for the source framebuffer
+    /// - `dst` is out of bounds for the destination framebuffer
+    /// - `src` and `dst` sizes are different and interpolation id not supported by this renderer.
+    /// - source and target framebuffer are the same, and `src` and `dst` overlap
+    fn blit(
```
i509VCB: Is blit always going to be damageless?

src/backend/egl/context.rs
```diff
@@ -221,39 +221,50 @@ impl EGLContext {
         })
     }
 
+    /// Makes the OpenGL context the current context in the current thread with no surface bound.
+    ///
+    /// # Safety
+    ///
+    /// This function is marked unsafe, because the context cannot be made current on another thread without
+    /// being unbound again (see [`EGLContext::unbind`]).
+    pub unsafe fn make_current(&self) -> Result<(), MakeCurrentError> {
```
i509VCB, line 230: Do we need a way to make the context not current if we expose this?

src/backend/renderer/gles2/mod.rs
```diff
@@ -1627,6 +1632,108 @@ impl Offscreen<Gles2Renderbuffer> for Gles2Renderer {
...
+{
+    fn blit(
+        &mut self,
+        from: B1,
+        to: B2,
+        src: Rectangle<i32, Physical>,
+        dst: Rectangle<i32, Physical>,
+        filter: TextureFilter,
+    ) -> Result<(), <Self as Renderer>::Error> {
+        // glBlitFramebuffer is sadly only available for GLES 3.0 and higher
```
i509VCB, line 1646: Would be nice to state the features gated by GL versions and features.

src/backend/renderer/gles2/mod.rs
```diff
@@ -1627,6 +1632,108 @@ impl Offscreen<Gles2Renderbuffer> for Gles2Renderer {
...
+            Gles2Target::Image { ref buf, .. } => unsafe {
+                self.gl.BindFramebuffer(ffi::READ_FRAMEBUFFER, buf.fbo)
+            },
+            Gles2Target::Texture { ref fbo, .. } => unsafe {
+                self.gl.BindFramebuffer(ffi::READ_FRAMEBUFFER, *fbo)
+            },
+            Gles2Target::Renderbuffer { ref fbo, .. } => unsafe {
+                self.gl.BindFramebuffer(ffi::READ_FRAMEBUFFER, *fbo)
+            },
+            _ => {}
```
i509VCB, line 1684: Does this unhandled variant need to be handled or will it never be reached?

src/backend/renderer/mod.rs
Drakulix: Blit will either completely succeed or not. So either the damage done to the target framebuffer is equal to `dst` or not.

src/backend/renderer/gles2/mod.rs
```diff
@@ -1627,6 +1632,108 @@ impl Offscreen<Gles2Renderbuffer> for Gles2Renderer {
...
+        }
+
+        let status = unsafe { self.gl.CheckFramebufferStatus(ffi::FRAMEBUFFER) };
+        if status != ffi::FRAMEBUFFER_COMPLETE {
+            let _ = self.unbind();
+            return Err(Gles2Error::FramebufferBindingError);
+        }
+
+        let errno = unsafe {
+            let _ = self.gl.GetError(); // clear flag before
```
i509VCB, line 1706: Don't we need to drain the error in a loop?

src/backend/renderer/multigpu/mod.rs
```diff
@@ -1916,3 +1916,45 @@ where
             .map_err(Error::Render)
     }
 }
+
+impl<'a, 'b, R: GraphicsApi, T: GraphicsApi, Target, B1, B2> Blit<B1, B2>
```
i509VCB: I assume blits cannot cross the device boundary.

src/backend/egl/context.rs
Drakulix: `EGLContext::unbind` exists. (and is mentioned in the docs for exactly this function)

This also was already public btw.

src/backend/renderer/gles2/mod.rs
Drakulix: I have quite a few renderer-specific doc comments on some trait-implementations for the `Gles2Renderer`, sadly rustdoc doesn't render any of them: https://smithay.github.io/smithay/smithay/backend/renderer/gles2/struct.Gles2Renderer.html#trait-implementations ...

src/backend/renderer/gles2/mod.rs
Drakulix: Can't be reached, because the only target not covered is the `Surface`, which is handled above (because it needs to be part of `make_current`).

src/backend/renderer/gles2/mod.rs
Drakulix: https://docs.gl/es2/glGetError

Seems you are right.

src/backend/renderer/multigpu/mod.rs
Drakulix: Blits happen between frame buffers, so they always affect the "target" device in the multigpu context and through the `Bind`-requirement (that also only affects the target), it is not possible to copy from frame buffers of another device right now.

Theoretically, we could support the api for arbitrary frame buffers, by requiring (`ExportDma` and `ImportDma`) or (`ExportMem` and `ImportMem`) instead, but I don't see a way to provide this without making the case this is supposed to optimize slower without specialization.
Given downstream can already manually use `Export/Import[Dma/Mem]` for this use-case, I opted to only implement the direct frame buffer to frame buffer copy.

src/backend/renderer/gles2/mod.rs
Drakulix: fixed

src/backend/renderer/gles2/mod.rs
Drakulix: Added a comment to explain this.

Drakulix:
rebased without changes

Drakulix:
ping @i509VCB

i509VCB (approved):
This looks fine for GL at least. I'm not sure how Vulkan would look with this API if even implemented.

PR #807, opened 2022-11-12
winit: Support using the glow-renderer

Drakulix:


PolyMeilex (approved):
Looks straight forward.

PR #811, opened 2022-11-14
element: Allow errors to escape from  `RenderContext::draw`

Drakulix:
cc @cmeissl 

Allows errors to propage from inside the `RenderContext::draw` call.


cmeissl:
Would it be a good idea to apply the same change to the memory buffer too?

Drakulix:
> Would it be a good idea to apply the same change to the memory buffer too?

Good idea. Done

cmeissl (approved):
LGTM

PolyMeilex (approved):
LGTM

PR #814, opened 2022-11-15
bump MSRV to 1.65.0

cmeissl:


Drakulix (approved):
May clippy spare our souls

PR #816, opened 2022-11-16
GAT Frame

Drakulix:
This got a little out of hand... but this is what I initially imagined would have been the better API for `Renderer`/`Frame`. With GATs we finally can do it.

Notable changes:
- `Renderer::render` does not rely on a closure anymore, but returns a `Frame` instead.
- `Frame` does not implement the same traits as the renderers and blocks access to the renderer during it's lifetime. If that proves cumbersome, we can implement the traits without a problem, just a good pile of boilercode.
- This actually encourages a cleaner separation between resource management and actual rendering, which I personally see as a positiv (it is limiting however, as can be seen by the MultiRenderer, which needs both for cleaning up the Frame).
- Most hacks around wrapping `Renderer::render` calls are now gone (`GlowRenderer::render` and `MultiRenderer::render`, though `MultiRenderer` gained some new unsafe code).
- It gives downstream more options for passing a frame around and potentially even storing it.

Open for discussion.

~~Draft until tested.~~

### DIFF
--- a/src/backend/renderer/mod.rs
+++ b/src/backend/renderer/mod.rs
@@ -135,6 +135,10 @@ pub trait Frame {
     /// Texture Handle type used by this renderer.
     type TextureId: Texture;
 
+    /// Returns an id, that is unique to all renderers, that can use
+    /// `TextureId`s originating from any of these renderers.
+    fn id(&self) -> usize;
+
     /// Clear the complete current target with a single given color.
     ///
     /// The `at` parameter specifies a set of rectangles to clear in the current target. This allows partially
@@ -189,6 +193,18 @@ pub trait Frame {
 
     /// Output transformation that is applied to this frame
     fn transformation(&self) -> Transform;
+
+    /// Finish this [`Frame`] returning any error that may happen during any cleanup.
+    ///
+    /// Dropping the frame instead may result in any of the following and is implementation dependent:
+    /// - All actions done to the frame vanish and are never executed
+    /// - A partial renderer with undefined framebuffer contents occurs
+    /// - All actions are performed as normal without errors being returned.
+    ///
+    /// Leaking the frame instead will leak resources and can cause any of the previous effects.
+    /// Leaking might make the renderer return Errors and force it's recreation.
+    /// Leaking may not cause otherwise undefined behavior and program execution will always continue normally.
+    fn finish(self) -> Result<(), Self::Error>;
 }
 
 /// Abstraction of commonly used rendering operations for compositors.
@@ -198,7 +214,9 @@ pub trait Renderer {
     /// Texture Handle type used by this renderer.
     type TextureId: Texture;
     /// Type representing a currently in-progress frame during the [`Renderer::render`]-call
-    type Frame: Frame<Error = Self::Error, TextureId = Self::TextureId>;
+    type Frame<'a>: Frame<Error = Self::Error, TextureId = Self::TextureId> + 'a
+    where
+        Self: 'a;
 
     /// Returns an id, that is unique to all renderers, that can use
     /// `TextureId`s originating from any of these renderers.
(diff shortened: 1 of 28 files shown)


src/backend/renderer/multigpu/mod.rs
```diff
@@ -831,188 +843,194 @@ where
...
-        let (res, damage) = res?;
-        let mut damage = damage
-            .into_iter()
-            .map(|rect| {
-                rect.to_logical(1)
-                    .to_buffer(1, dst_transform, &size.to_logical(1))
-            })
-            .collect::<Vec<_>>();
+            // now the frame is gone, lets use our unholy ptr till the end of this call:
+            let render = unsafe { &mut *self.render };
```
i509VCB, line 883: Properly defending our unsafe use:

```rust
// SAFETY:
// - The renderer will never be invalid because the lifetime of the frame must be shorter than the renderer.
// - The pointer can't be aliased because of the following:
//   - Creating a frame requires an `&mut` reference to the renderer, making the mutable borrow safe.
//   - The mutable reference is used in a function which mutably borrows the frame, that either being `.finish()` (which takes ownership of the frame) or dropping the frame.
```

src/backend/renderer/multigpu/mod.rs
```diff
@@ -618,74 +618,90 @@ impl<'a, 'b, R: GraphicsApi, T: GraphicsApi, Target> AsMut<<R::Device as ApiDevi
...
-            RenderDevice::Device(dev) => dev.renderer_mut(),
-            RenderDevice::Renderer(renderer, _) => unsafe { &mut **renderer },
-        }
-    }
-}
-
 /// [`Frame`](super::Frame) implementation of a [`MultiRenderer`].
-#[derive(Debug)]
-pub struct MultiFrame<R: GraphicsApi, T: GraphicsApi> {
+pub struct MultiFrame<'a, 'b, 'c, R: GraphicsApi + 'c, T: GraphicsApi, Target>
```
i509VCB, line 622: These lifetime names are meaningless to me.

i509VCB (commented):
The changes are pretty good looking, just a few things around the place.

src/backend/renderer/mod.rs
```diff
@@ -198,7 +214,9 @@ pub trait Renderer {
     /// Texture Handle type used by this renderer.
     type TextureId: Texture;
     /// Type representing a currently in-progress frame during the [`Renderer::render`]-call
-    type Frame: Frame<Error = Self::Error, TextureId = Self::TextureId>;
+    type Frame<'a>: Frame<Error = Self::Error, TextureId = Self::TextureId> + 'a
```
i509VCB, line 217: do we want to use a more descriptive lifetime name than `'a`?

src/backend/renderer/mod.rs
```diff
@@ -189,6 +193,18 @@ pub trait Frame {
...
+    /// Finish this [`Frame`] returning any error that may happen during any cleanup.
+    ///
+    /// Dropping the frame instead may result in any of the following and is implementation dependent:
+    /// - All actions done to the frame vanish and are never executed
+    /// - A partial renderer with undefined framebuffer contents occurs
+    /// - All actions are performed as normal without errors being returned.
+    ///
+    /// Leaking the frame instead will leak resources and can cause any of the previous effects.
+    /// Leaking might make the renderer return Errors and force it's recreation.
+    /// Leaking may not cause otherwise undefined behavior and program execution will always continue normally.
```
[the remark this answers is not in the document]
Drakulix: done

src/backend/renderer/multigpu/mod.rs
Drakulix: Changed into `'render, 'target, 'frame` for the whole module.

src/backend/renderer/mod.rs
Drakulix: Used `'frame`

i509VCB (approved):
lgtm

PR #817, opened 2022-11-17
anvil/udev: Don't delay frames for secondary gpus

Drakulix:
@cmeissl


cmeissl (approved):
LGTM

PolyMeilex (approved):
Looks good, not sure what's going on with documentation build in CI

Drakulix:
> Looks good, not sure what's going on with documentation build in CI

It think that happens, because calloop was just updated, so 0.10.2 is not in our list of packages anymore, but 0.10.3 is picked by Cargo...
I will handle that in a separate PR.

PR #819, opened 2022-11-17
Remove `io-lifetimes` dependency

ids1024:
Since Smithay now requires Rust 1.65.0, and io-lifetimes reexports these types from std on recent Rust versions, this dependency is redundant.


PolyMeilex (approved):
Sounds like a good idea :+1:

i509VCB (approved):

PR #820, opened 2022-11-18
renderer: fix export_texture

cmeissl:
eglCreateImage expects the attributes to be of type EGLAttrib which is platform dependent using EGLEnum types will 
 populate an array with the wrong type which causes a segfault when calculating the attrib list len in the mesa egl code


Drakulix (approved):
Wow, nice catch

cmeissl:
> Wow, nice catch

yeah, was quite puzzling looking at gdb. took me a few minutes to realize what causes the segfault

PR #822, opened 2022-11-23
egl: log the format when eglQueryDmaBufModifiersEXT fails

cmeissl:


Drakulix:
On my system at least those errors appear in the EGL_KHR_DEBUG output anyway. Do we need this?

cmeissl:
> On my system at least those errors appear in the EGL_KHR_DEBUG output anyway. Do we need this?

I felt like it was a good addition as I already added a `dbg!` twice now, `EGL_KHR_DEBUG` is missing the fourcc at least on my system.

Drakulix (approved):
It surely doesn't hurt

PR #824, opened 2022-11-25
Fix/multigpu alpha clear

cmeissl:
Maybe we find a better solution, but that at least seems to make it work


cmeissl:
superseded by #844

PR #826, opened 2022-11-30
renderer: Add `with_context` not only to `Frame`

Drakulix:


src/backend/renderer/gles2/mod.rs
```diff
@@ -1684,6 +1684,21 @@ impl Gles2Renderer {
...
+    /// as such you should reset everything you change back to its previous value
+    /// or check the source code of the version of Smithay you are using to ensure
+    /// your changes don't interfere with the renderer's behavior.
+    /// Doing otherwise can lead to rendering errors while using other functions of this renderer.
+    pub fn with_context<F, R>(&mut self, func: F) -> Result<R, Gles2Error>
+    where
+        F: FnOnce(&ffi::Gles2) -> R,
+    {
+        Ok(func(&self.gl))
+    }
```
PolyMeilex, line 1701: Just a small question, before `with_context` got removed from `Gles2Renderer` we used to also call `make_current` before calling the closure, was this behavior intentionally omitted this time around?

Just for reference taking about this:
```rs
    pub fn with_context<F, R>(&mut self, func: F) -> Result<R, Gles2Error>
    where
        F: FnOnce(&mut Self, &ffi::Gles2) -> R,
    {
        self.make_current()?;
        let gl = self.gl.clone();
        Ok(func(self, &gl))
    }
```

Drakulix: Hmm, good call. For the Frame it is safe too assume, that the context is current, but not for the renderer. I will update this.

PolyMeilex (approved):

PR #831, opened 2022-12-05
Crop & Scale (+Window Preview)

cmeissl:


src/backend/renderer/element/mod.rs
```diff
@@ -116,20 +116,20 @@ pub struct RenderElementState {
     /// Holds the visible portion of the element on the output
     ///
     /// Note: If the presentation_state is [`RenderElementPresentationState::Skipped`] this will be zero.
-    pub visible_portion: Size<i32, Physical>,
+    pub visible_portion: usize,
```
Drakulix, line 120: I get, that we don't need a Rectangle (or multiple) here, but could be add a doc comment explaining, why this is an usize and what exactly it represents?

src/backend/renderer/element/utils.rs
```diff
@@ -0,0 +1,502 @@
...
+
+impl<E: Element> CropRenderElement<E> {
+    /// Create a cropping render element for an existing element
+    ///
+    /// The crop rect is expected to be relative to the same origin the element is relative to.
+    /// It can extend outside of the element geometry but the resulting geometry will always
+    /// be equal or smaller than the geometry of the original element.
+    ///
+    /// The scale is used to calculate the scaling from physical space to buffer space and must
+    /// match the scale provided to the [`Element`] functions.
```
Drakulix, line 122: *must* sounds very strong. What happens if we pass a scale of 2 (like the underlying element), but then render a cropped window on outputs of scales 1 and 2? Then this description is ambiguous. 

We should rather document what happens if the scale does not match the buffer scale. Or if possible get rid of the scale and specify the `crop_rect` in `Buffer`-coordinate space?

Drakulix (commented):
Some initial thoughts. Overall this looks really fleshed out. Nice work!

src/backend/renderer/element/mod.rs
cmeissl: Sure, I will add a comment explaining what it represents.

src/backend/renderer/element/utils.rs
cmeissl: > _must_ sounds very strong. What happens if we pass a scale of 2 (like the underlying element), but then render a cropped window on outputs of scales 1 and 2? Then this description is ambiguous.
> 
> We should rather document what happens if the scale does not match the buffer scale. Or if possible get rid of the scale and specify the `crop_rect` in `Buffer`-coordinate space?

We do not have access to the buffer space here as the `Element`-Trait does not expose the buffer scale. The intent is that downstream can just specify a rect on screen that is used for cropping, this makes it easier to crop a whole surface tree for example. I think it should be possible to move the logic to the element logic instead at construction time.

Regarding different output scales. I don't think we support that anyway, both outputs create their own element tree already as the positioning already depends on the scale.

cmeissl: Ok, no, moving the logic to run-time instead of creation time is not possible, I will change the docs a bit instead.

src/backend/renderer/element/mod.rs

src/backend/renderer/element/utils.rs
cmeissl: relaxed the scaling part and clarified the assumptions in the docs

Drakulix (approved):
Looks good to me! Though it needs a rebase

cmeissl:
rebased on master

PR #832, opened 2022-12-07
anvil: integrate renderdoc in debug feature

cmeissl:
this enables renderdoc on the x11 backend if the debug feature is enabled


Drakulix:
Any particular reason this is a draft PR? Imo we can merge this even without support for every single backend, given it is a debug feature anyway.

cmeissl:
> Any particular reason this is a draft PR? Imo we can merge this even without support for every single backend, given it is a debug feature anyway.

No, no particular reason

Drakulix (approved):

PR #836, opened 2022-12-13
Fix IME not closing on application exit

### LINKED ISSUE #835: Text input doesn't send `leave` for dead surface
When a surface dies which currently has text input focus, it will never generate a `zwp_text_input_v3::leave` event. Because of that IME clients might not realize that they should close themselves.

The issue seems to be with the `with_focused_text_input` call here:
https://github.com/Smithay/smithay/blob/bb08df51681e69662b7c73f981d2936429badc05/src/wayland/text_input/text_input_handle.rs#L76-L78

More concretely, the `with_focused_text_input` has two hurdles that need to be overcome to fix this. The first issue is that it checks for liveliness itself:
https://github.com/Smithay/smithay/blob/bb08df51681e69662b7c73f981d2936429badc05/src/wayland/text_input/text_input_handle.rs#L30

I'm pretty sure that code is completely useless anyway. It seems like a minor optimization but realistically it shouldn't optimize much at all. And it shouldn't be necessary since `same_client_as` always returns `false` when any of the surfaces are dead anyway. And there lies the bigger issue, since this is required before the function parameter is called:
https://github.com/Smithay/smithay/blob/bb08df51681e69662b7c73f981d2936429badc05/src/wayland/text_input/text_input_handle.rs#L34-L36

I'm sure the `same_client_as` check has a good reason to perform a liveliness check, otherwise I doubt it would be present in the method. So it seems like the only solution would be to **not** check for this on leave? But that would likely cause generation of incorrect `leave` events.

My current patch to get it working just ignores the client stuff since it should mostly be irrelevant for me anyway:

```diff
diff --git a/src/wayland/text_input/text_input_handle.rs b/src/wayland/text_input/text_input_handle.rs
index b42402d7264..0141bb772f9 100644
--- a/src/wayland/text_input/text_input_handle.rs
+++ b/src/wayland/text_input/text_input_handle.rs
@@ -73,9 +73,11 @@ impl TextInputHandle {
         let same = inner.focus.as_ref() == focus;
         if !same {
             focus_changed_hook();
-            inner.with_focused_text_input(|ti, surface, _serial| {
-                ti.leave(surface);
-            });
+            if let Some(ref surface) = inner.focus {
+                for ti in inner.instances.iter() {
+                    ti.instance.leave(surface);
+                }
+            }

             inner.focus = focus.cloned();
```

However I doubt that this patch is suitable for upstream, so I'd be curious how this should generally be handled. I'm sure that there's already some method in place somewhere else that helps with this kind of stuff, so I didn't want to send an obviously broken patch.

chrisduerr:
This fixes an issue with the text_input/input_method protocols where the application requesting IME wouldn't notify the IME client about its death, keeping the IME client open after the application was closed.

With this patch whenever the `text_input` object is destroyed, the `deactivate` event is automatically emitted and committed ensuring the IME client is properly updated.

Closes #835.

### DIFF
--- a/src/wayland/text_input/text_input_handle.rs
+++ b/src/wayland/text_input/text_input_handle.rs
@@ -162,6 +162,12 @@ where
     }
 
     fn destroyed(_state: &mut D, _client: ClientId, ti: ObjectId, data: &TextInputUserData) {
+        // Ensure IME is deactivated when text input dies.
+        data.input_method_handle.with_instance(|input_method| {
+            input_method.deactivate();
+            input_method.done();
+        });
+
         data.handle
             .inner
             .lock()


Drakulix (approved):
LGTM

PR #844, opened 2022-12-22
multigpu: Fix broken gpu-copy

Drakulix:
Latest nvidia driver can somehow import the `Gles2Renderbuffer` on my system, which has an `Invalid`-modifier and is therefore garbage for the nvidia gpu as it is tiled.
So lets explicitly check for supported formats.

- Also includes commits from #841.
- Contains commit to move the generic buffer type of the `Allocator` trait into an associated generic type. Simplifies code all over the place.

Additionally prepares for more sensible gpu-copies in the future:
- Fixes missing cleanup of `GpuManager::dma_source`
- Caches offscreen-buffer
- Attempts to import the offscreen-buffer on the target gpu **on creation** (instead of after rendering, which is expected to fail in most cases)
- Caches resulting texture with the offscreen buffer

In the future we can hopefully do what #842 sets out to do and use a GBM `Allocator` instead of the `Offscreen`-trait, which allocates through the graphics-api.
Currently that crashes on nvidia-gpus, which are by far the most interesting use-case for this feature given most multi-gpu systems are likely nVidia Optimus laptops.
To fix that we need to either test and confirm, that the `VulkanAllocator` is working for this use-case or provide a `GlAllocator` as an alternative, that can only provide `Invalid`-buffers and can then replicate how the `MultiRenderer` works today.
(Also hopefully gbm buffers won't hard crash the nvidia driver in the future.)

### DIFF
--- a/src/backend/allocator/mod.rs
+++ b/src/backend/allocator/mod.rs
@@ -56,7 +56,9 @@ pub trait Buffer {
 }
 
 /// Interface to create Buffers
-pub trait Allocator<B: Buffer> {
+pub trait Allocator {
+    /// Buffer type produced by this allocator
+    type Buffer: Buffer;
     /// Error type thrown if allocations fail
     type Error: std::error::Error;
 
@@ -67,12 +69,13 @@ pub trait Allocator<B: Buffer> {
         height: u32,
         fourcc: Fourcc,
         modifiers: &[Modifier],
-    ) -> Result<B, Self::Error>;
+    ) -> Result<Self::Buffer, Self::Error>;
 }
 
 // General implementations for interior mutability.
 
-impl<A: Allocator<B>, B: Buffer> Allocator<B> for Arc<Mutex<A>> {
+impl<A: Allocator> Allocator for Arc<Mutex<A>> {
+    type Buffer = A::Buffer;
     type Error = A::Error;
 
     fn create_buffer(
@@ -81,13 +84,14 @@ impl<A: Allocator<B>, B: Buffer> Allocator<B> for Arc<Mutex<A>> {
         height: u32,
         fourcc: Fourcc,
         modifiers: &[Modifier],
-    ) -> Result<B, Self::Error> {
+    ) -> Result<Self::Buffer, Self::Error> {
         let mut guard = self.lock().unwrap();
         guard.create_buffer(width, height, fourcc, modifiers)
     }
 }
 
-impl<A: Allocator<B>, B: Buffer> Allocator<B> for Rc<RefCell<A>> {
+impl<A: Allocator> Allocator for Rc<RefCell<A>> {
+    type Buffer = A::Buffer;
     type Error = A::Error;
 
     fn create_buffer(
@@ -96,12 +100,13 @@ impl<A: Allocator<B>, B: Buffer> Allocator<B> for Rc<RefCell<A>> {
         height: u32,
         fourcc: Fourcc,
         modifiers: &[Modifier],
-    ) -> Result<B, Self::Error> {
+    ) -> Result<Self::Buffer, Self::Error> {
         self.borrow_mut().create_buffer(width, height, fourcc, modifiers)
     }
 }
 
-impl<B: Buffer, E: std::error::Error> Allocator<B> for Box<dyn Allocator<B, Error = E> + 'static> {
+impl<B: Buffer, E: std::error::Error> Allocator for Box<dyn Allocator<Buffer = B, Error = E> + 'static> {
+    type Buffer = B;
     type Error = E;
 
     fn create_buffer(
--- a/src/backend/renderer/multigpu/mod.rs
+++ b/src/backend/renderer/multigpu/mod.rs
@@ -41,14 +41,13 @@ use std::{
     convert::{AsMut, AsRef},
     fmt,
     rc::Rc,
-    sync::Mutex,
 };
 
 #[cfg(feature = "wayland_frontend")]
 use crate::wayland::{dmabuf::get_dmabuf, shm};
 use crate::{
     backend::{
-        allocator::{dmabuf::WeakDmabuf, Buffer as BufferTrait, Format},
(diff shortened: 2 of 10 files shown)


src/backend/allocator/mod.rs
```diff
@@ -67,12 +69,13 @@ pub trait Allocator<B: Buffer> {
...
         modifiers: &[Modifier],
-    ) -> Result<B, Self::Error>;
+    ) -> Result<Self::Buffer, Self::Error>;
 }
 
 // General implementations for interior mutability.
 
-impl<A: Allocator<B>, B: Buffer> Allocator<B> for Arc<Mutex<A>> {
+impl<A: Allocator> Allocator for Arc<Mutex<A>> {
+    type Buffer = A::Buffer;
```
i509VCB, line 78: Actually this brings up a good question. If a buffer implements AsDmabuf, exporting it might not be synchronized. In the vulkan allocator, we export directly in the buffer, but it is safe as we have an Arc<Device> inside.

However for other backends this may vary.

src/backend/renderer/multigpu/mod.rs
```diff
@@ -1968,8 +2053,8 @@ where
 impl<'a, 'b, R: GraphicsApi, T: GraphicsApi, Target, BlitTarget> Blit<BlitTarget>
     for MultiRenderer<'a, 'b, R, T, Target>
 where
-    <R::Device as ApiDevice>::Renderer: Blit<BlitTarget>,
-    <T::Device as ApiDevice>::Renderer: Blit<BlitTarget>,
+    <R::Device as ApiDevice>::Renderer: Blit<BlitTarget> + Bind<BlitTarget>,
+    <T::Device as ApiDevice>::Renderer: Blit<BlitTarget> + Bind<BlitTarget>,
```
i509VCB, line 2057: Is this not a duplicate trait bound?

src/backend/renderer/multigpu/mod.rs
```diff
@@ -1310,6 +1381,7 @@ where
     <T::Device as ApiDevice>::Renderer: ImportDma + ImportMem,
     <<R::Device as ApiDevice>::Renderer as Renderer>::Error: 'static,
     <<T::Device as ApiDevice>::Renderer as Renderer>::Error: 'static,
+    Target: Clone,
```
i509VCB, line 1384: What introduced this Clone bound btw?

src/backend/allocator/mod.rs
Drakulix: Yeah right, but we don't do a particular good job at accounting for synchronization (which is also hard, when you have multiple objects - textures, dmabufs, framebuffer - referr to the same GPU memory). So I don't think this is an issue for this PR, but more something we should have a longer discussion about.

src/backend/renderer/multigpu/mod.rs
Drakulix: Nope. Once for the render and once for the target device.

Drakulix: Nope. Once for the render and once for the target device.

src/backend/renderer/multigpu/mod.rs
Drakulix: Because we cache the renderbuffer and `Bind<Target>` consumes the buffer otherwise.

src/backend/renderer/multigpu/mod.rs
i509VCB: Oh wait I misread Blit as Bind

src/backend/renderer/multigpu/mod.rs
```diff
@@ -839,15 +887,66 @@ where
...
+                    .renderer_mut()
+                    .bind(render_buffer.clone())
+                    .map_err(Error::Render)?;
+                target.offscreen_buffer = Some((buffer_size, render_buffer));
+
+                // figure out if we can use this for a gpu-copy.
+                // *Note*: Probably not as long as this uses `Offscreen`, which allocates through the graphics api.
+                //         For proper GPU copies we want to compare `Bind::<Dmabuf>::supported_formats()` of the render-device
+                //         with `ImportDma::dmabuf_formats` of the target-device and allocate via GBM (or maybe vulkan) on
+                //         on of the devices. But that has other problems...
```
cmeissl, line 911: Small comment nit: s/on on/on one/

src/backend/renderer/multigpu/mod.rs
```diff
@@ -839,15 +887,66 @@ where
...
+        if let Some(mut target) = self.target.as_mut() {
             let buffer_size = size.to_logical(1).to_buffer(1, dst_transform);
-            let render_buffer = Offscreen::<Target>::create_buffer(self.render.renderer_mut(), buffer_size)
-                .map_err(Error::Render)?;
-            self.render
-                .renderer_mut()
-                .bind(render_buffer)
-                .map_err(Error::Render)?;
-        }
+            if target
```
cmeissl, line 892: Is the size check enough? Are the formats/modifiers guaranteed to be the same over time?

cmeissl (commented):
I really like the allocator change, makes the code more readable imo.
Maybe we also want to introduce a type alias for BufferObject<()>.

src/backend/renderer/multigpu/mod.rs
Drakulix: Dmabufs don't change format or modifiers, so by directly trying to export it we sidestep the problem, because:

- either this will succeed, then format and modifier is pinned.
- it doesn't succeed, which means we won't try again for the lifetime of the render_buffer and then it doesn't matter to us, if GL reallocates the buffer in the background. And we wouldn't want to try to export the buffer on every frame. Especially given that dmabufs are meant to be created right after creating and before use of the underlying object (as we found out from your `ExportDma` experiments).

src/backend/renderer/multigpu/mod.rs
Drakulix: done

cmeissl (approved):

PR #847, opened 2022-12-23
Ref counted FD

Drakulix:
This PR consists of two commits, that should be reviewed independently:

- 1a2c6b2 - Adds a new utility type `DeviceFd`, which is essentially a ref-counted `OwnedFd`. It also adds a `DrmDeviceFd` based on that and gets rid of most the generics in the `backend/drm`-module in the process.
- 5ae9412 - Tries to address the unsafety of the lifetime of `EGLDisplay`-objects, which are bound to their underlying object. The big issue here was the `GbmDevice`, which is only clonable, if they underlying file-type was clonable. And we didn't want to clone a raw file descriptor, but we can easily clone a ref-counted one. So now we can simply make the `EGLDisplay` take ownership of the native type.

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -3,7 +3,7 @@ use std::{
     cell::RefCell,
     collections::hash_map::{Entry, HashMap},
     convert::TryInto,
-    os::unix::io::{AsRawFd, RawFd},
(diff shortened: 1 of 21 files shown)


PolyMeilex (approved):
That's nice, looks a lot cleaner now.

PR #849, opened 2022-12-27
session: Drop everything except libseat

Drakulix:
Drops all sessions except the libseat implementation, makes libseat mandatory for anvil in the process. Trait is kept for people wanting to implement different session interfaces out of process.

This allows to simplifies the `Session` interface quite a bit. Gets rid of the notifier infrastructure, replaced by some simple routines in `anvil/src/udev.rs`.


Cargo.toml
```diff
@@ -84,8 +84,6 @@ backend_libinput = ["input"]
 backend_session = []
 backend_udev = ["udev", "input/udev"]
 backend_vulkan = ["ash", "scopeguard"]
-backend_session_logind = ["dbus", "backend_session", "pkg-config"]
```
ids1024, line 87: The `dbus` optional dependency should be unneeded now.

Drakulix: good call, removed!

src/backend/session/mod.rs
```diff
@@ -70,39 +70,15 @@ pub trait Session {
...
 }
 
-/// Signals that can be generated by a session
-///
-/// Objects that need to be notifier about activation and deactivation
-/// of devices need to be linked to the `Signaler` of the session providing
-/// these signals.
+/// Events that can be generated by a session
 #[derive(Copy, Clone, Debug)]
 pub enum Signal {
```
PolyMeilex, line 75: Kinda nitpicky, but I wonder if the name Signal / SessionSignal, is still a good choice since it is no longer a calloop signal, and just a regular event.

PolyMeilex (approved):

src/backend/session/mod.rs
Drakulix: Fixed

PR #851, opened 2022-12-30
Use pointer and length to refer to shm memory, instead of slice

ids1024:
Creating a slice into shared memory is problematic, since this is considered undefined behavior if the client mutates the memory. This is especially a problem since it was creating a slice for the whole pool, as even a correctly behaving client may mutate another buffer in the same pool.

We don't want to trust clients not to mutate buffers for soundness, even if it's not obvious what bugs or exploits this could result in. It's also quite unnecessary if all we do is convert the slice back to a pointer and pass it to OpenGL.

Memory access that could result in `SIGBUS` is probably also problematic in access from safe Rust code that expects all memory accesses to succeed.

This simply replaces the slice with a pointer and length, which for the most part (with the current renderer implementation) means we're not converting to a slice just to convert it back to a pointer again.

### DIFF
--- a/src/wayland/shm/mod.rs
+++ b/src/wayland/shm/mod.rs
@@ -63,7 +63,7 @@
 //! use smithay::wayland::shm::{with_buffer_contents, BufferData, BufferAccessError};
 //!
 //! let content = with_buffer_contents(&buffer,
-//!     |slice: &[u8], buffer_metadata: BufferData| {
+//!     |ptr: *const u8, len: usize, buffer_metadata: BufferData| {
 //!         // do something to extract the contents of the buffer
 //!     }
 //! );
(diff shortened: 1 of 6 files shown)


Drakulix:
Seems like a good idea to me. Pinging @i509VCB for some input on this proposal.

src/wayland/shm/mod.rs
```diff
@@ -207,21 +207,25 @@ impl From<UnmanagedResource> for BufferAccessError {
...
+/// The pool cannot be provided as a slice since it is shared memory that could be mutated
+/// by the client.
 ///
 /// If the buffer is not managed by the provided `ShmGlobal`, the closure is not called
 /// and this method will return `Err(BufferAccessError::NotManaged)` (this will be the case for an
 /// EGL buffer for example).
 pub fn with_buffer_contents<F, T>(buffer: &wl_buffer::WlBuffer, f: F) -> Result<T, BufferAccessError>
 where
-    F: FnOnce(&[u8], BufferData) -> T,
+    F: FnOnce(*const u8, usize, BufferData) -> T,
```
i509VCB, line 222: the whole function should probably become unsafe since using the pointer can be unsafe.

In particular we want to note that using the pointer outside the closure is likely to cause undefined behavior.

i509VCB (changes requested):

src/wayland/shm/mod.rs
ids1024: Using raw pointers is always unsafe, so at least technically there's no reason `with_buffer_contents` would need to be `unsafe`.

Making the function `unsafe` also seems like it may be sub-optimal, since it seems a closure here is then considered part of the unsafe block, so the scope of `unsafe { }` is unnecessarily expanded beyond the code where it could be an issue.

```rust
unsafe {
    with_buffer_contents(buffer, |ptr, len, data| {
        // Unsafe code can be used inside the closure without another unsafe block
        *ptr;
   });
```

i509VCB: Okay so it would be a usability issue.

Also the way you described this seems to be that the unsafe block would also apply in the closure. I'd think that seems bizzare since you'd probably want unsafe code in that closure to be seperately considered for safety. Maybe that's an issue that needs to be raised with rustc?

Drakulix: I guess "unsafe" is simply not meant for "this function has to be used carefully", but for well defined things, like dereferencing a raw pointer.
So while I am all for documenting, that this pointer should not be accessed outside of the closure, I would argue the closure itself doesn't need to be marked unsafe.

ids1024: > Also the way you described this seems to be that the unsafe block would also apply in the closure. I'd think that seems bizzare since you'd probably want unsafe code in that closure to be seperately considered for safety. Maybe that's an issue that needs to be raised with rustc?

Yeah, it's somewhat non-obvious whether or not the unsafe block would apply to the inside of the closure. Probably not something that's had much thought since unsafe functions taking closures aren't that common. I think changing something like this in an edition would be possible but need an RFC, though it's about as simple as a Rust RFC can be.

Another possibility here is to not use a function taking a callback and instead return an RAII type with methods for ptr and length. That's generally more idiomatic, though I guess the issue is that there's no way to guarantee `drop` is called since `mem::forget` is safe.

Callbacks taking a pointer like this aren't so idiomatic, but even without checking the documentation one would probably expect the pointer to only be valid until the callback returns. In general, if an API involves raw pointers you need to check the documentation before dereferencing them and especially before trying to store them and dereference later.

> I guess "unsafe" is simply not meant for "this function has to be used carefully", but for well defined things, like dereferencing a raw pointer.

Yeah, I think it's generally considered a best practice to require `unsafe` only where the actual unsafe behavior occurs. So this function shouldn't be unsafe if merely calling it without de-referencing the pointer is safe.

PolyMeilex: I have to agree with @Drakulix and @ids1024 here, calling the function itself is safe. Dereferencing the pointer is not, but that's already unsafe by default. So good docs is the best approach here.

ids1024: I've added a `# Safety` section to the doc comments, anyway. Though describing exactly how this can be safely used is a bit complicated.

PolyMeilex (approved):

PR #856, opened 2023-01-08
Expose properties of Winit*Event

egasimus:
Just changing from `pub(crate)` to `pub` to allow these structs to be created from outside the crate.

This enabled me to write an [alternate winit backend](https://github.com/egasimus/charlie/blob/rebuild/src/engine/winit/patch.rs) which shares the winit event loop between multiple windows. So far it renders a background and receives events from each window; I'm currently putting the XWayland module back together and then I'll try to get multiple views into the same screenspace from winit and then try again to port my compositor to the udev backend through this [`Engine` trait](https://github.com/egasimus/charlie/blob/rebuild/src/engine.rs#L24).

### DIFF
--- a/src/backend/winit/input.rs
+++ b/src/backend/winit/input.rs
@@ -50,10 +50,10 @@ impl Device for WinitVirtualDevice {
 /// Winit-Backend internal event wrapping `winit`'s types into a [`KeyboardKeyEvent`]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct WinitKeyboardInputEvent {
-    pub(crate) time: u32,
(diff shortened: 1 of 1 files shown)


HeavyRain266:
Links which you provided leads to private repo also from the talk on matrix, winit will be replaced with pure backends for both, X11 and Wayland because it doesn't play well with Compositor's event loop.

Drakulix:
Tbh, there is a reason these are not exposed by default, because the contents of these events is not part of the public api.

I would rather like to see a modification of the `winit`-Backend, that allows to create a new `WinitGraphicsBackend` from an existing `WinitEventLoop` basically breaking apart the `init`-methods, which always return both.

PR #861, opened 2023-01-13
Bump `udev` to 0.7 and `input` to 0.8.1

ids1024:


ids1024:
The failing CI check could be fixed by specifying the version of `udev` in `ci.yaml`, but perhaps it would be better to just release https://github.com/Smithay/input.rs and update to it, so only the latest version of the `udev` crate is used.

ids1024:
Input 0.8 fails to build with `cargo build --no-default-features --features libinput_1_14`, which is what Smithay is currently using.

Drakulix:
Input 0.8.1 is released. (*sigh*)

Drakulix (approved):
More safe-io goodness

PR #862, opened 2023-01-13
drm: fix early master lock release

cmeissl:
when a privileged `DrmDeviceFd` is cloned and dropped it also releases the master lock potential leaving the still exiting fds in a bad state


Drakulix (approved):
I guess one `Arc`-more isn't going to make a difference :laughing:

cmeissl:
> I guess one `Arc`-more isn't going to make a difference 😆

Yeah, at least we don't have an additional lock. But that fixes anvil running with the libseat built-in backend for me which was broken after the DrmDeviceFd PR

PR #866, opened 2023-01-19
egl: Add support for surfaceless

bwidawsk:
ChromeOS uses [minigbm](https://chromium.googlesource.com/chromiumos/platform/minigbm/) instead of GBM. As part of this, the EGL_MESA_platform_gbm is unavailable. To allow Smithay to work on ChromeOS, enable a minigbm flag.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -31,6 +31,7 @@ fn gl_generate() {
                 "EGL_EXT_platform_wayland",
                 "EGL_KHR_platform_gbm",
                 "EGL_MESA_platform_gbm",
+                "EGL_MESA_platform_surfaceless",
                 "EGL_EXT_platform_device",
                 "EGL_WL_bind_wayland_display",
                 "EGL_KHR_image_base",
--- a/src/backend/egl/native.rs
+++ b/src/backend/egl/native.rs
@@ -201,6 +201,23 @@ impl EGLNativeDisplay for EGLDevice {
     }
 }
 
+/// Shallow type for the EGL_PLATFORM_SURFACELESS with default EGL display
+#[derive(Debug)]
+pub struct EGLSurfacelessDisplay;
+
+impl EGLNativeDisplay for EGLSurfacelessDisplay {
+    fn supported_platforms(&self) -> Vec<EGLPlatform<'_>> {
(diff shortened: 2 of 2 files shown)


Drakulix:
The surfaceless platform does not appear to have anything todo with the `GbmDevice`, given that it just uses the `EGL_DEFAULT_DISPLAY` constant

As such I would prefer, if this PR just introduced a new `Surfaceless`-type, which implements `NativeDisplay` side-stepping the unnecessary `GbmDevice` all together.

bwidawsk:
Didn't mean to close. I will resubmit with the requested changes.

bwidawsk:
@Drakulix won't I run into an issue where the egl backend will get upset because my platform's mesa doesn't support the platform_gbm extensions if I go this route? I still need GBM (for minigbm).

Drakulix:
> @Drakulix won't I run into an issue where the egl backend will get upset because my platform's mesa doesn't support the platform_gbm extensions if I go this route? I still need GBM (for minigbm).

We only require the extensions for initializing either the `EGLDisplay` or an `EGLSurface` (though anvil doesn't use this code path) via a `GbmDevice` or `GbmSurface`. You even override them in your current patch with `PLATFORM_SURFACELESS_MESA`.
So no, as long as you use the proposed `Surfaceless` type for initializing your `EGLDisplay` nothing will complain about the missing extension. `GbmBufferedSurface` and the like don't have anything to do with `EGL` and should continue to work normally with minigbm.

But anvil won't work without any patches of course, because it expects to be able to use the `GbmDevice` to initialize it's `EGLDisplay`. If you want anvil to support minigbm, feel free to add an additional feature or environment variable flag to make it use the surfaceless platform instead.

bwidawsk:
> But anvil won't work without any patches of course, because it expects to be able to use the `GbmDevice` to initialize it's `EGLDisplay`. If you want anvil to support minigbm, feel free to add an additional feature or environment variable flag to make it use the surfaceless platform instead.

Since at the moment testing for me is quite cumbersome, I'll hold off on Anvil support if that's okay with you.

Drakulix (approved):

PR #867, opened 2023-01-20
Fixes for building on FreeBSD

ids1024:
With this I'm able to start Anvil, if I disable `xwayland`. (`UnixAddr::new_abstract` is Linux-specific.)

Handling drm device nodes is really a mess regardless of OS... and is also a mess in `libdrm`.

### DIFF
--- a/src/backend/drm/node/constants.rs
+++ b/src/backend/drm/node/constants.rs
@@ -23,17 +23,17 @@ pub const DRM_MAJOR: u64 = 226;
 #[cfg(not(target_os = "openbsd"))]
 pub const PRIMARY_NAME: &str = "card";
 
-#[cfg(target_os = "freebsd")]
+#[cfg(target_os = "openbsd")]
 pub const PRIMARY_NAME: &str = "drm";
 
 #[cfg(not(target_os = "openbsd"))]
 pub const CONTROL_NAME: &str = "controlD";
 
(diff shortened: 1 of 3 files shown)


ids1024:
Perhaps some of this DRM code should be in the `drm` crate.

i509VCB:
I believe @Drakulix had an opinion on some of this code moving to the drm crate.

I would be fine with a new crate for some of the node identification and path lookup if it doesn't directly fit into the drm crate.

Drakulix:
> I would be fine with a new crate for some of the node identification and path lookup if it doesn't directly fit into the drm crate.

I am not sure it fits the drm-crate, given that a lot of that stuff is
- ... pretty platform dependant
- ... applies to device nodes in general as opposed to just drm-devices

The idea of the drm-crate is to model the drm ioctl/kernel api, not really any more as shown by the relatively flexible `*Device`-traits. which don't care where you got that file descriptor from.

I think a separate crate might make sense here, but I am also not opposed to merge this in the meantime directly into smithay. @i509VCB Do you have any particular opinion about that?

i509VCB:
Merging this into smithay for the time being would be fine.

Drakulix (approved):
> Merging this into smithay for the time being would be fine.

Well, in that case there is no reason to not merge these fixes. The code itself is fine.

PR #868, opened 2023-01-20
Fix touch event focus

chrisduerr:
Currently the handles for a specific touch slot are always cleared on touch down, meaning that touch events are correctly propagated as long as `down` is called before any other touch events. However when motion is called after `up` without a previous `down`, then the motion is still reported to the previous client.

While this can be easily avoided in the compositor by always calling `down` first and avoiding `motion`/`up`/`cancel` without previous `down` call, it should provide a more intuitive behavior this way.


Drakulix (approved):
Seems sensible

PR #869, opened 2023-01-21
xwm: replace LATE_SURFACE_ID

Drakulix:
The current mechanism by which x11 window ids and their corresponding wl_surfaces are matched, may inject a `LATE_SURFACE_ID` client message similar to the original `WL_SURFACE_ID` message.

The problem with injecting this message is, that to do that it utilizes the same `SyncSender`, that is internally used by the `X11Source`. Given that this sender is *sync* and limited to 5 in-flight messages, queuing up additional messages from the same thread, that runs the calloop event loop and receives these messages may cause a deadlock. This can often be encountered when multiple windows are spawned by a client at the same time.

Instead of making the channel unbound this PR avoids the problem by changing the mechanism all together. Instead of injecting a message, the xwm state is accessed by an idle callback, which isn't prone to cause a deadlock.

### DIFF
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -28,6 +28,8 @@ use smithay::{
 };
 
 use crate::state::{AnvilState, Backend};
+#[cfg(feature = "xwayland")]
+use crate::CalloopData;
 
 mod element;
 mod grabs;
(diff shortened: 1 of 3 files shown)


PolyMeilex (approved):
LGTM

Kinda off topic, do you think Xwayland shell will help a bit with this convoluted event order dance?

Drakulix:
> LGTM
> 
> 
> 
> Kinda off topic, do you think Xwayland shell will help a bit with this convoluted event order dance? 

Yes, because it eliminates the race. It requires a relatively recent Xwayland version though, which why I think we will need to keep the old code path for some time anyway.

PR #872, opened 2023-01-24
Pointer constraints protocol

ids1024:
The protocol implementation should generally work, though there's some more work still needed to get the implementation in Anvil fully functional. And testing, docs, etc.


Drakulix (commented):
Looks good for a first draft!

ids1024:
I've implemented pointer confinement in Anvil; confining to just the window boundary, ignoring the region.

Handling this correctly seems complicated. The logic in wlroots is slightly complicated (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/util/region.c#L190-256), but more significantly seems to depend on the `pixman_region32_t` data structure. I don't see any crates providing something similar. (I guess other compositors are all in C and using pixman, so it's not an issue for them.)

I think this is good enough to merge now, at least given the part that's incomplete is the Anvil implementation.

Drakulix:
> Handling this correctly seems complicated. The logic in wlroots is slightly complicated (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/util/region.c#L190-256), but more significantly seems to depend on the pixman_region32_t data structure. I don't see any crates providing something similar. (I guess other compositors are all in C and using pixman, so it's not an issue for them.)

I am slightly confused, because I am pretty sure we have to deal with this for input_regions of surfaces already. So I am not sure, why the same code wouldn't work here?
It probably needs logic to properly clamp to the region, but a cheap implementation could just discard cursor movements, that would result in a place outside the region. And we already have a "is point in wl_region" test for certain.

ids1024:
Yeah, I was thinking of the clamping part. Just ignoring movement if the new position would be outside the region might work alright. Technically less pixel-perfect. Also with a really weird region that could jump over a tiny area that isn't in the region to another that is, but that's even less likely to be an issue (I'm not sure exactly how other implementations handle these cases).

`pixman_region32_t` should be a more efficient data structure in general, but hopefully applications aren't adding and subtracting 500 rectangles to a `wl_region`...

Drakulix:
> pixman_region32_t should be a more efficient data structure in general, but hopefully applications aren't adding and subtracting 500 rectangles to a wl_region...

Oh absolutely and yes the issues above *can* in theory happen. But I think implementing it this way in anvil is better, than not implementing it at all. Especially because compositors are likely going to copy&paste anvil's implementation. :sweat_smile:

ids1024:
This isn't doing any fancy pointer warping, but setting a region seems to work now, with https://github.com/Smithay/client-toolkit/pull/407.

Just testing if the pointer ends up outside the region and dropping the event, instead of clamping, seems to be working alright.

Drakulix (approved):
LGTM

PR #876, opened 2023-01-30
backend: Gesture events

ids1024:
This just defines the traits and associated types. It still needs an implementation for libinput (which is straightforward). And then to be useful an implementation of the `zwp_pointer_gestures_v1` protocol is also needed.

This seems to be consistent with how other event types are handled, but seems a little verbose. Would it be better to not have 10 new traits and 8 new associated types?

The `begin` events are all the same, as are the `end` events. `update` exists only for swipe and pinch, and has additional fields for pinch. Would it be sensible to reduce the 6 begin/end associated types to 2?

### DIFF
--- a/src/backend/libinput/mod.rs
+++ b/src/backend/libinput/mod.rs
@@ -1,3 +1,7 @@
+// TODO: Update to use new `Scroll*` events instead of now deprecated
+// `PointerAxis*`.
+#![cfg_attr(feature = "libinput_1_19", allow(deprecated))]
+
 //! Implementation of input backend trait for types provided by `libinput`
 
 use crate::backend::input::{self as backend, Axis, InputBackend, InputEvent};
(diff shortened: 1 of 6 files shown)


Drakulix:
> Would it be better to not have 10 new traits and 8 new associated types?

Nah I think, this is actually fine.

This is a one-time write cost and some backends might have just a selection of these gestures. I think for the backend the most flexible approach is the best.
We can take some freedom actually implementing support for pointer gestures though and start grouping them.

i509VCB:
Do we need to add a new `DeviceCapability` for the gestures?

ids1024:
Looks like there's already a `DeviceCapability::Gesture` variant, which maps the `libinput::DeviceCapability::Gesture`, so I guess that's already covered.

ids1024:
I've renamed things from `PointerGesture*` to just `Gesture*` to match libinput's naming and make these names not quite so excessively long.

ids1024:
The backend code here is basically done. For libinput anyway. Hold gesture support on libinput is commented since it requires a new `input` release with https://github.com/Smithay/input.rs/pull/49 and `libinput_1_19`. (Would it be okay to bump the dependency to that? Or does it need a feature?)

Drakulix:
> The backend code here is basically done. For libinput anyway. Hold gesture support on libinput is commented since it requires a new `input` release with [Smithay/input.rs#49](https://github.com/Smithay/input.rs/pull/49) and `libinput_1_19`. (Would it be okay to bump the dependency to that? Or does it need a feature?)

input.rs release 0.8.2 is on it's way.

I would like a new feature for now, I think. I generally try and keep the project buildable on debian stable by default and they seem to be on 1.16 right now: https://packages.debian.org/search?keywords=libinput&searchon=names&suite=stable&section=all

ids1024:
Seems reasonable. Though if the target is Debian stable it might be a good idea to have a CI test running that.

So call the feature `libinput_1_19` and have it non-default? And then compositors that want to use it would have to enable the feature, or have their own feature. A bit awkward.

Drakulix:
> Seems reasonable. Though if the target is Debian stable it might be a good idea to have a CI test running that.
> 
> So call the feature `libinput_1_19` and have it non-default? And then compositors that want to use it would have to enable the feature, or have their own feature. A bit awkward.

Nah, we have broken that promise in the past with libseat- and mesa-requirements, because it can be a giant maintenance burden, so I would definitely not want to swap CI.
But for libinput this is easy enough to do, as far as I know it is building currently and we don't need to make anyone's life harder, even if it ends up being a little awkward.

I think we could also give this a more descriptive name in the context of smithay like `libinput_gestures`, but that might also be confusing, given that 1.19 also added high-resolution scroll wheel support, which we also want to have in smithay. So I guess `libinput_1_19` is fine.

ids1024:
The `libinput_1_19` feature also produces warnings like:

> warning: use of deprecated struct `input::event::pointer::PointerAxisEvent`: Use `PointerEvent::Scroll*` events instead

I guess it needs to either conditionally use those as well, or just `#[allow(deprecated)]` them.

Drakulix:
> The `libinput_1_19` feature also produces warnings like:
> 
> > warning: use of deprecated struct `input::event::pointer::PointerAxisEvent`: Use `PointerEvent::Scroll*` events instead
> 
> I guess it needs to either conditionally use those as well, or just `#[allow(deprecated)]` them.

Yeah, that are the new scroll-events we don't support yet. `#[allow(deprecated)]` is fine for now in this case.

src/backend/libinput/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "libinput_1_19", allow(deprecated))]
```
Drakulix, line 1: Can you add a `TODO` comment here to explain, why this is still needed?

Drakulix (commented):
Nitpicks and could you also add a line to the `CHANGELOG`, that smithay supports processing GestureEvents now? (Even if they are not send to clients yet.)

Drakulix (approved):

PR #879, opened 2023-02-01
Xwm Selection

Drakulix:
Clipboard synchronization support for Xwayland!


anvil/src/shell/x11.rs
```diff
@@ -229,6 +239,75 @@ impl<BackendData: Backend> XwmHandler for CalloopData<BackendData> {
...
+                if surface.xwm_id().unwrap() == xwm {
+                    return true;
+                }
+            }
+        }
+        false
+    }
+
+    fn send_selection(&mut self, _xwm: XwmId, selection: SelectionType, mime_type: String, fd: OwnedFd) {
+        //OwnedFd) {
```
PolyMeilex, line 256: Looks like a stray comment.

Drakulix: fixed

src/wayland/data_device/mod.rs
```diff
@@ -99,6 +103,9 @@ use seat_data::{SeatData, Selection};
 /// Events that are generated by interactions of the clients with the data device
 #[allow(unused_variables)]
 pub trait DataDeviceHandler: Sized + ClientDndGrabHandler + ServerDndGrabHandler {
+    /// UserData attached to server-side selections
+    type SelectionUserData: Clone + Send + Sync + 'static;
```
vberger, line 107: So, in a very "thinking broad" way, would it make sense to not restrict it do server-side selections, but rather allow it for all selections, adding a `Default` bound?

Basically, this increases the realm of possibles for users, and if they only want to attach data to server-side selections they can use an `enum` to distinguish. So that would be a little less practical.

I'm just throwing the idea as it came to me, no idea if that'd actually be worth it or not.

anvil/src/shell/x11.rs
```diff
@@ -232,6 +240,39 @@ impl<BackendData: Backend> XwmHandler for CalloopData<BackendData> {
     fn move_request(&mut self, _xwm: XwmId, window: X11Surface, _button: u32) {
         self.state.move_request_x11(&window)
     }
+
+    fn new_selection(&mut self, _xwm: XwmId, selection: SelectionType, mime_types: Vec<String>) {
+        slog::info!(
```
vberger: I think this should be a `trace!` instead, no?

src/xwayland/xwm/mod.rs
```diff
@@ -284,6 +309,25 @@ pub trait XwmHandler {
...
+
+    /// A new selection was set by an X client with provided mime_types
+    fn new_selection(&mut self, xwm: XwmId, selection: SelectionType, mime_types: Vec<String>) {
+        let _ = (xwm, selection, mime_types);
+    }
+
+    /// A proviously set selection of an X client got cleared
+    fn cleared_selection(&mut self, xwm: XwmId, selection: SelectionType) {
+        let _ = (xwm, selection);
+    }
```
vberger, line 330: I feels weird to have most of those methods be optional with a default of doing nothing, but not all. Feels inconsistent to me.

anvil/src/shell/x11.rs
```diff
@@ -241,13 +243,38 @@ impl<BackendData: Backend> XwmHandler for CalloopData<BackendData> {
...
+    }
+
     fn new_selection(&mut self, _xwm: XwmId, selection: SelectionType, mime_types: Vec<String>) {
-        slog::info!(
-            self.state.log,
-            "Got Selection {:?} from X11: {:?}",
-            selection,
-            mime_types
-        );
+        info!(?selection, ?mime_types, "Got Selection from X11",);
```
vberger, line 277: I'm confused as to why this is in this PR...?

Drakulix: I think because the tracing migration was midway and during the rebase I merged this fix with the wrong commit. :sweat_smile: 
Not super harmful, but I can clean that up.

src/xwayland/xwm/mod.rs
Drakulix: Well, it is a balance of operations you need to implement to provide a somewhat usable Xwm and those that could be skipped (like clipboard synchronization). Its a narrow line for sure. Any recommendations?

vberger: I probably don't know enough about X11 selection, but my first understanding would be that if `allow_selection_access` always returns `false`, then `send_selection` would never be invoked? And so why require it to be implemented?

Probably that's not what happens though, but in that case maybe it'd be worth expanding the documentation to explain, for the optional methods, what behavior you get if you don't implement them.

src/wayland/data_device/mod.rs
Drakulix: Resolving, as we were not able to come up with a use-case.

src/xwayland/xwm/mod.rs
Drakulix: > I probably don't know enough about X11 selection, but my first understanding would be that if allow_selection_access always returns false, then send_selection would never be invoked? And so why require it to be implemented?

No that is exactly right and a good hint. I'll give `send_selection` a default impl, that panics. (Because it needs to be handled once you override allow_selection_access.)

src/xwayland/xwm/mod.rs
```diff
@@ -1259,8 +2230,154 @@ fn handle_event<D: XwmHandler>(state: &mut D, xwmid: XwmId, event: Event) -> Res
...
+}
+
+fn read_selection_callback(
+    conn: &RustConnection,
+    atoms: &Atoms,
+    fd: BorrowedFd<'_>,
+    transfer: &mut OutgoingTransfer,
+) -> Result<OutgoingAction, ReplyOrIdError> {
+    let mut buf = [0; INCR_CHUNK_SIZE];
+    let Ok(len) = nix::unistd::read(fd.as_raw_fd(), &mut buf) else {
```
i509VCB, line 2255: I assume our msrv has let else?

i509VCB (approved):
It looks fine, I don't fully understand everything so I can kind of a yesman here.

src/xwayland/xwm/mod.rs
```diff
@@ -330,6 +390,205 @@ impl Drop for X11Wm {
...
+        if let Some(token) = self.token.take() {
+            handle.remove(token);
+        }
+    }
+}
+
+impl Drop for OutgoingTransfer {
+    fn drop(&mut self) {
+        if self.token.is_some() {
+            println!("WARNING: OutgoingTransfer freed before being removed from EventLoop");
```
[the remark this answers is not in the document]
Drakulix: Good call, will fix that before merge

src/xwayland/xwm/mod.rs
Drakulix: I'll check

Drakulix: Yes, it's 1.65 according to CI, which is the first version with let else.

src/xwayland/xwm/mod.rs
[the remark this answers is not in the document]
Drakulix: fixed

PR #880, opened 2023-02-02
input: More introspection

Drakulix:
Supersedes #798 and adds some more functions to inspect the current state of a seat and the keyboard handle.

Changes `with_pressed_keysyms` to take an `FnOnce`, which will take a `Vec<KeysymHandle>` and passes on it's return value.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -580,6 +580,32 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
         );
     }
 
+    /// Iterate over the keysyms of the currently pressed keys.
+    pub fn with_pressed_keysyms<F, R>(&self, f: F)
+    where
+        F: FnOnce(Vec<KeysymHandle<'_>>) -> R,
+        R: 'static,
+    {
+        let guard = self.arc.internal.lock().unwrap();
+        {
+            let handles = guard
+                .pressed_keys
+                .iter()
+                .map(|code| KeysymHandle {
+                    keycode: code + 8,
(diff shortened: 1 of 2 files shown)


src/input/keyboard/mod.rs
```diff
@@ -707,6 +733,21 @@ impl<'a, D: SeatHandler + 'static> KeyboardInnerHandle<'a, D> {
...
+        R: 'static,
+    {
+        let handles = self
+            .inner
+            .pressed_keys
+            .iter()
+            .map(|code| self.keysym_handle(*code))
+            .collect();
+        f(handles)
+    }
```
PolyMeilex, line 749: Just to be sure, this is callback based for the sake of consistency with KeyboardHandle (non inner)?

Drakulix: yes

src/input/keyboard/mod.rs
```diff
@@ -580,6 +580,32 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
...
+                })
+                .collect::<Vec<_>>();
+            f(handles);
+        }
+    }
+
+    /// Get the current modifiers state
+    pub fn modifier_state(&self) -> ModifiersState {
+        self.arc.internal.lock().unwrap().mods_state
+    }
```
PolyMeilex, line 607: Could we get this one for Inner flavor of handle as well?

Drakulix: Sure!

PolyMeilex (approved):

PR #882, opened 2023-02-06
example: Buffer Tester

Drakulix:
Something I wanted to build for a long time to debug multi-gpu setups.

### DIFF
--- a/examples/buffer_test.rs
+++ b/examples/buffer_test.rs
@@ -1,0 +1,523 @@
+use std::{error, fmt, fs::File, os::unix::io::OwnedFd, path::PathBuf};
+
+use clap::{Args, Parser, Subcommand, ValueEnum};
+use slog::{o, Drain};
+use smithay::{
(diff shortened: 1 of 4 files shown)


examples/buffer_test.rs
```diff
@@ -0,0 +1,351 @@
...
+    export: Option<String>,
+    #[arg(short, long)]
+    renderer: Option<RendererType>,
+
+    #[arg(short, long, default_value_t = 256)]
+    width: u32,
+    #[arg(short, long, default_value_t = 256)]
+    height: u32,
+    #[arg(short, long, default_value = "Abgr8888")]
+    fourcc: String,
```
i509VCB, line 39: I believe clap allows specifying an attribute to use some function to do the `fn(String) -> T` conversion so you can store a DrmFourcc code directly here.

Drakulix: Done.

PR #886, opened 2023-02-10
Multigpu: Use real allocators for gpu->gpu copies

Drakulix:
Best reviewed commit-by-commit. Supersedes https://github.com/Smithay/smithay/pull/842, builds upon https://github.com/Smithay/smithay/pull/882.

Changes the `MultiRenderer` to use a user-provided allocator to create render buffers to cross-gpu transfers.
By default anvil now uses the `VulkanAllocator`, which fixes crashes on nvidia devices, compared to the `GbmAllocator`, which is still used as a fallback, if vulkan is not available or fails to initialize.

Additionally adds a new gbm-api for the multigpu renderer as an alternative for the egl-device based enumeration. This *should* fix kmsro devices with anvil, which now opts in to use this new backend.


anvil/src/udev.rs
```diff
@@ -323,6 +285,89 @@ pub fn run_udev(log: Logger) {
         state.device_added(&mut display, dev, path.into())
     }
 
+    if let Ok(instance) = Instance::new(Version::VERSION_1_2, None, log.clone()) {
```
i509VCB, line 288: for purposes of testing, would it be useful to have a way to ignore vulkan and directly use gbm?

anvil/src/udev.rs
```diff
@@ -323,6 +285,89 @@ pub fn run_udev(log: Logger) {
...
+        if let Some(physical_device) = PhysicalDevice::enumerate(&instance).ok().and_then(|devices| {
+            devices
+                .filter(|phd| {
+                    phd.has_device_extension(unsafe {
+                        CStr::from_bytes_with_nul_unchecked(b"VK_EXT_physical_device_drm\0")
```
i509VCB, line 293: You should be able to use `ExtPhysicalDeviceDrmFn::name` from ash instead of manually creating that string.

i509VCB (approved):
More patches will be needed of course, but it does improve the situation.

Review on the following hardware:

## M2 mac
Anvil now works when `ANVIL_DRM_DEVICE=/dev/dri/card2` is set (card2 being the display controller). Before this did not work regardless of the drm device that was set.

## amdgpu

Vulkan allocator works in both tty-udev and x11 backends.

PR #891, opened 2023-02-15
backend/gbm: Use correct member for legacy modifiers

bwidawsk:


bwidawsk:
Note that CI has a new enough mesa as to not see any change here.

Drakulix (approved):

PR #892, opened 2023-02-16
switch from slog to tracing

cmeissl:
unfinished draft PR to switch from slog to tracing


anvil/src/input_handler.rs
```diff
@@ -82,10 +83,10 @@ impl<BackendData: Backend> AnvilState<BackendData> {
                     )
                     .spawn()
                 {
-                    error!(self.log,
-                        "Failed to start program";
-                        "cmd" => cmd,
-                        "err" => format!("{:?}", e)
+                    error!(
+                        cmd,
+                        err = ?e,
```
i509VCB, line 88: io::Error implements Display, so I'm not sure why we use the Debug representation here

src/backend/allocator/vulkan/mod.rs
```diff
@@ -138,6 +139,7 @@ impl fmt::Debug for VulkanAllocator {
         f.debug_struct("VulkanAllocator")
             .field("formats", &self.formats)
             .field("images", &self.images)
+            .field("default_usage", &self.default_usage)
```
i509VCB, line 142: Hmm an idea for a clippy lint

src/backend/egl/context.rs
```diff
@@ -46,86 +47,67 @@ impl EGLContext {
...
         assert!(!display.is_null(), "EGLDisplay pointer is null");
         assert!(!config_id.is_null(), "EGL configuration id pointer is null");
         assert!(!context.is_null(), "EGLContext pointer is null");
-        let log = crate::slog_or_fallback(log.into()).new(o!("smithay_module" => "backend_egl"));
 
-        let display = EGLDisplay::from_raw(display, config_id, log)?;
+        let display = EGLDisplay::from_raw(display, config_id)?;
         let pixel_format = display.get_pixel_format(config_id)?;
 
+        let span = info_span!(parent: &display.span, "egl_context");
```
i509VCB, line 62: Would it be useful to have a way to identify two different contexts? Probably would be fine to use the pointer for representation? Or some sort of counted I'd?

src/backend/egl/context.rs
```diff
@@ -46,86 +47,67 @@ impl EGLContext {
...
     }
 
     /// Create a new configless `EGLContext` from a given `EGLDisplay` sharing resources with another context
-    pub fn new_shared<L>(display: &EGLDisplay, share: &EGLContext, log: L) -> Result<EGLContext, Error>
-    where
-        L: Into<Option<::slog::Logger>>,
-    {
-        Self::new_internal(display, Some(share), None, log)
+    pub fn new_shared(display: &EGLDisplay, share: &EGLContext) -> Result<EGLContext, Error> {
+        Self::new_internal(display, Some(share), None)
```
i509VCB, line 91: It might be useful to store the context from which objects are being shared in the span?

src/backend/renderer/gles2/mod.rs
```diff
@@ -454,20 +452,18 @@ extern "system" fn gl_debug_log(
     _severity: ffi::types::GLenum,
     _length: ffi::types::GLsizei,
     message: *const ffi::types::GLchar,
-    user_param: *mut nix::libc::c_void,
+    _: *mut nix::libc::c_void,
```
i509VCB, line 455: This could use a span for the gles renderer?

src/backend/vulkan/mod.rs
```diff
@@ -317,19 +298,13 @@ impl Instance {
                         | vk::DebugUtilsMessageTypeFlagsEXT::PERFORMANCE
                         | vk::DebugUtilsMessageTypeFlagsEXT::VALIDATION,
                 )
-                .pfn_user_callback(Some(vulkan_debug_utils_callback))
-                .user_data(*logger_ptr as *mut _);
+                .pfn_user_callback(Some(vulkan_debug_utils_callback));
```
i509VCB, line 301: Would be nice to have a span passed through the user data

i509VCB (commented):
Review for everything in the backend module. Still need to look through the rest

src/backend/vulkan/mod.rs
Drakulix: The idea was, that vulkan calls will likely be triggered through smithay functions, that will set the span anyway (which will then still be active, should the debug-callback be called from inside the vulkan driver). But I guess because we are exposing raw ash types, this might not always be the case?

src/backend/renderer/gles2/mod.rs
Drakulix: This is a non-issue, as the `Gles2Renderer` is already setting up the span, thus every GL-debug-callback will still have the span active. (I will check, that `with_context` does do that, to make sure even Raw GL commands will have it).

src/output.rs
```diff
@@ -205,13 +201,9 @@ pub(crate) type OutputData = Arc<(Mutex<Inner>, UserDataMap)>;
...
-    pub fn new<L>(name: String, physical: PhysicalProperties, logger: L) -> Output
-    where
-        L: Into<Option<::slog::Logger>>,
-    {
-        let log = crate::slog_or_fallback(logger).new(o!("smithay_module" => "output_handler"));
-
-        info!(log, "Creating new wl_output"; "name" => &name);
+    #[instrument]
+    pub fn new(name: String, physical: PhysicalProperties) -> Output {
+        info!("Creating new wl_output");
```
i509VCB, line 206: This isn't actually making a wl_output.

i509VCB (approved):
All reviewed. Sure there may be some changes to make our use of tracing nicer, but those changes can be done later.

src/output.rs
Drakulix: Good call, fixed

PR #896, opened 2023-02-20
Fix render output transforms

chrisduerr:
The transform of an output is specified in the rotation that needs to be applied to a surface for rendering. This is the exact inverse of the transform that needs to be applied to the renderer itself.

This patch inverts the transform stored on the output's mode to ensure the rendering transform correctly specifies the inverse of the surface transform.

---

If there's any other renderer/compositors, please let me know and I'll add this line there too. This was tested on my phone and not a desktop monitor.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -1388,6 +1388,11 @@ where
         let current_size = self.output.current_mode().unwrap().size;
         let output_scale = self.output.current_scale().fractional_scale().into();
         let output_transform = self.output.current_transform();
+
+        // Output transform is specified in surface-rotation, so inversion gives us the
+        // render transform for the output itself.
+        let output_transform = output_transform.invert();
+
         // Geometry of the output derived from the output mode including the transform
         // This is used to calculate the intersection between elements and the output.
         // The renderer (and also the logic for direct scan-out) will take care of the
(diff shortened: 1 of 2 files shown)


src/backend/drm/compositor/mod.rs
```diff
@@ -1388,6 +1388,11 @@ where
         let current_size = self.output.current_mode().unwrap().size;
         let output_scale = self.output.current_scale().fractional_scale().into();
         let output_transform = self.output.current_transform();
+
+        // Output transform is specified in surface-rotation, so inversion gives us the
+        // render transform for the output itself.
+        let output_transform = output_transform.invert();
```
i509VCB, line 1394: Is this a case of GL's coordinate system being flipped upright and then around the wrong way again? Having to invert the transform seems suspicious.

chrisduerr: No, as the comment states it's the difference between surface and output rotation.

chrisduerr: To elaborate: If I turn my phone 90° clockwise, that means the output's rendering needs to be flipped 90° counter-clockwise. However the surface displayed on it get the information that it's flipped clockwise since that's what the protocol specifies.

cmeissl (approved):
We re-use the `Transform` for output rotation and for `Renderer` transform.
Without some `Kind` that specifies the direction (clock wise vs counter clock wise) this can lead to unexpected behavior. So the output defines the transform as counter clock wise (compensating for the physical rotation), but the `Gles2Renderer` rotates clock-wise.

But that should be handled in a separate PR and this fix should be fine for now.

cc @Drakulix approved by me, but I have no right for merge

Drakulix (approved):

PR #899, opened 2023-02-22
Feature/dmabuf feedback

cmeissl:
### TODO 

- [ ] review docs
- [x] anvil multi-gpu feedback handling

### DIFF
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -7,7 +7,9 @@ use std::{
 use tracing::{info, warn};
 
 use smithay::{
-    backend::renderer::element::{default_primary_scanout_output_compare, RenderElementStates},
+    backend::renderer::element::{
+        default_primary_scanout_output_compare, utils::select_dmabuf_feedback, RenderElementStates,
+    },
     delegate_compositor, delegate_data_device, delegate_fractional_scale, delegate_input_method_manager,
     delegate_keyboard_shortcuts_inhibit, delegate_layer_shell, delegate_output, delegate_presentation,
     delegate_primary_selection, delegate_relative_pointer, delegate_seat, delegate_shm,
@@ -40,6 +42,7 @@ use smithay::{
             set_data_device_focus, ClientDndGrabHandler, DataDeviceHandler, DataDeviceState,
             ServerDndGrabHandler,
         },
+        dmabuf::DmabufFeedback,
         fractional_scale::{with_fractional_scale, FractionScaleHandler, FractionalScaleManagerState},
         input_method::{InputMethodManagerState, InputMethodSeat},
         keyboard_shortcuts_inhibit::{
@@ -553,10 +556,17 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
     }
 }
 
+#[derive(Debug, Copy, Clone)]
+pub struct SurfaceDmabufFeedback<'a> {
+    pub render_feedback: &'a DmabufFeedback,
+    pub scanout_feedback: &'a DmabufFeedback,
+}
+
 pub fn post_repaint(
     output: &Output,
     render_element_states: &RenderElementStates,
     space: &Space<WindowElement>,
+    dmabuf_feedback: Option<SurfaceDmabufFeedback<'_>>,
     time: impl Into<Duration>,
 ) {
     let time = time.into();
@@ -581,6 +591,16 @@ pub fn post_repaint(
 
         if space.outputs_for_element(window).contains(output) {
             window.send_frame(output, time, throttle, surface_primary_scanout_output);
+            if let Some(dmabuf_feedback) = dmabuf_feedback {
+                window.send_dmabuf_feedback(output, surface_primary_scanout_output, |surface, _| {
+                    select_dmabuf_feedback(
+                        surface,
+                        render_element_states,
+                        dmabuf_feedback.render_feedback,
+                        dmabuf_feedback.scanout_feedback,
+                    )
+                });
+            }
         }
     });
     let map = smithay::desktop::layer_map_for_output(output);
@@ -602,6 +622,16 @@ pub fn post_repaint(
         });
 
         layer_surface.send_frame(output, time, throttle, surface_primary_scanout_output);
+        if let Some(dmabuf_feedback) = dmabuf_feedback {
+            layer_surface.send_dmabuf_feedback(output, surface_primary_scanout_output, |surface, _| {
+                select_dmabuf_feedback(
+                    surface,
+                    render_element_states,
+                    dmabuf_feedback.render_feedback,
+                    dmabuf_feedback.scanout_feedback,
+                )
+            });
+        }
     }
 }
 
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -2,7 +2,10 @@
 use std::ffi::OsString;
 use std::{
     borrow::Cow,
-    collections::hash_map::{Entry, HashMap},
+    collections::{
+        hash_map::{Entry, HashMap},
+        HashSet,
+    },
     convert::TryInto,
     os::unix::io::FromRawFd,
     path::Path,
@@ -10,20 +13,17 @@ use std::{
     time::Duration,
 };
 
+use crate::state::SurfaceDmabufFeedback;
 use crate::{
     drawing::*,
     render::*,
     shell::WindowElement,
     state::{post_repaint, take_presentation_feedback, AnvilState, Backend, CalloopData},
 };
+#[cfg(feature = "egl")]
+use smithay::backend::renderer::ImportEgl;
 #[cfg(feature = "debug")]
 use smithay::backend::renderer::ImportMem;
-#[cfg(feature = "egl")]
-use smithay::{
-    backend::renderer::{ImportDma, ImportEgl},
-    delegate_dmabuf,
-    wayland::dmabuf::{DmabufGlobal, DmabufHandler, DmabufState, ImportError},
-};
 use smithay::{
     backend::{
         allocator::{
@@ -43,7 +43,7 @@ use smithay::{
             element::{texture::TextureBuffer, AsRenderElements, RenderElement, RenderElementStates},
             gles2::{Gles2Renderbuffer, Gles2Renderer},
             multigpu::{gbm::GbmGlesBackend, GpuManager, MultiRenderer, MultiTexture},
-            Bind, DebugFlags, ExportMem, Offscreen, Renderer,
+            Bind, DebugFlags, ExportMem, ImportDma, Offscreen, Renderer,
         },
         session::{
             libseat::{self, LibSeatSession},
@@ -53,6 +53,7 @@ use smithay::{
         vulkan::{version::Version, Instance, PhysicalDevice},
         SwapBuffersError,
     },
+    delegate_dmabuf,
     desktop::{
         space::{Space, SurfaceTree},
         utils::OutputPresentationFeedback,
@@ -77,12 +78,18 @@ use smithay::{
         },
         input::Libinput,
         nix::{fcntl::OFlag, sys::stat::dev_t},
-        wayland_protocols::wp::presentation_time::server::wp_presentation_feedback,
+        wayland_protocols::wp::{
+            linux_dmabuf::zv1::server::zwp_linux_dmabuf_feedback_v1,
+            presentation_time::server::wp_presentation_feedback,
+        },
         wayland_server::{backend::GlobalId, protocol::wl_surface, Display, DisplayHandle},
     },
     utils::{Clock, DeviceFd, IsAlive, Logical, Monotonic, Point, Scale, Transform},
     wayland::{
         compositor,
+        dmabuf::{
+            DmabufFeedback, DmabufFeedbackBuilder, DmabufGlobal, DmabufHandler, DmabufState, ImportError,
+        },
         input_method::{InputMethodHandle, InputMethodSeat},
     },
 };
@@ -100,7 +107,6 @@ struct UdevOutputId {
 pub struct UdevData {
     pub session: LibSeatSession,
     dh: DisplayHandle,
-    #[cfg(feature = "egl")]
     dmabuf_state: Option<(DmabufState, DmabufGlobal)>,
     primary_gpu: DrmNode,
     allocator: Option<Box<dyn Allocator<Buffer = Dmabuf, Error = AnyError>>>,
@@ -132,7 +138,6 @@ impl UdevData {
     }
 }
 
-#[cfg(feature = "egl")]
 impl DmabufHandler for AnvilState<UdevData> {
     fn dmabuf_state(&mut self) -> &mut DmabufState {
         &mut self.backend_data.dmabuf_state.as_mut().unwrap().0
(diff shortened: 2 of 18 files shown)


anvil/src/udev.rs
```diff
@@ -716,9 +751,55 @@ fn scan_connectors(
             let allocator =
                 GbmAllocator::new(gbm.clone(), GbmBufferFlags::RENDERING | GbmBufferFlags::SCANOUT);
 
+            #[cfg(feature = "egl")]
+            let (render_feedback, scan_out_feedback) = {
+                // TODO: We should use the primary node as the main device and
+                // add the renderer as a preferred target, but as the gpu manager is
+                // late bound we can't do that here
+                let render_feedback =
+                    DmabufFeedbackBuilder::new(render_node.dev_id(), render_formats.clone())
```
Drakulix, line 760: I think those should be `texture_formats`? (We are probably doing it wrong in the `cfg(not(feature = "egl"))` case as well.)

For rendering we need to sample from the provided dmabufs, no render into them.
It would really be nice, if we could use `ImportDma::supported_formats` instead, as that is more readable/obvious, but I get why that isn't possible.

anvil/src/udev.rs
```diff
@@ -716,9 +751,55 @@ fn scan_connectors(
...
+                            .iter()
+                            .flat_map(|p| surface.supported_formats(p.handle).unwrap()),
+                    )
+                    .collect::<HashSet<_>>()
+                    .intersection(&texture_formats)
+                    .copied()
+                    .collect::<Vec<_>>();
+
+                // TODO: We should use the primary node as the main device and
+                // add the renderer as a preferred target after scan-out, but as the gpu manager is
```
Drakulix, line 781: > add the renderer

`renderer`? Does that refer to the render node? I am not sure I understand this comment.

Drakulix (commented):
Looks really solid, just a couple of remarks.

anvil/src/state.rs
```diff
@@ -568,6 +571,7 @@ pub fn post_repaint(
     output: &Output,
     render_element_states: &RenderElementStates,
     space: &Space<WindowElement>,
+    dmabuf_feedback: Option<(&DmabufFeedback, &DmabufFeedback)>,
```
i509VCB, line 574: These should be separate fields.

src/wayland/dmabuf/mod.rs
```diff
@@ -112,18 +119,311 @@ use crate::{
...
+        let formats = self
+            .formats
+            .iter()
+            .copied()
+            .map(DmabufFeedbackFormat::from)
+            .collect::<Vec<_>>();
+
+        let name = CString::new("smithay-dmabuffeedback-format-table").unwrap();
+        let format_table_file = SealedFile::with_data(name, unsafe {
+            std::slice::from_raw_parts(
```
i509VCB, line 271: I think we could do this without unsafe. We can use `u32::into_bytes_ne`, and then collect all the u8 bytes.

src/wayland/dmabuf/mod.rs
```diff
@@ -146,6 +449,7 @@ impl DmabufState {
         D: GlobalDispatch<zwp_linux_dmabuf_v1::ZwpLinuxDmabufV1, DmabufGlobalData>
             + BufferHandler
             + DmabufHandler
+            + DmabufFeedbackHandler
```
i509VCB, line 452: You should probably note that with this constructor that the functions inside DmabufFeedbackHandler will not be invoked.

src/wayland/dmabuf/mod.rs
```diff
@@ -286,6 +739,14 @@ pub enum ImportError {
     InvalidFormat,
 }
 
+/// Handler trait for [`DmabufFeedback`]
+pub trait DmabufFeedbackHandler {
+    /// This function allows to override the default [`DmabufFeedback`] for a surface
```
i509VCB, line 744: What does a return value of `None` mean? Also how do we account for the surface here?

i509VCB:
Also I didn't see anything to set the main device?

Drakulix:
> Also I didn't see anything to set the main device?

I think the device is part of the `DmabufFeedback` (which you specify using the `DmabufFeedbackBuilder`). So you can replace the main-device by setting a new default feedback with `DmabufGlobal::set_default_feedback`.

src/wayland/dmabuf/mod.rs
```diff
@@ -112,18 +119,311 @@ use crate::{
...
+    /// Adds a preferred tranche to the builder
+    ///
+    /// The tranches will be reported in the order they have been added with
+    /// this function.
+    ///
+    /// Note: Formats already present in a previously added preferred tranche with the
+    /// same target device and flags will be skipped.
+    /// If all formats are already included in a previously added tranche
+    /// with the same target device and flags this tranche will be skipped.
+    pub fn add_preferred_tranche(
```
ids1024, line 212: "Preference tranche" is the term used in the protocol spec, rather than "preferred tranche".

src/wayland/dmabuf/dispatch.rs
```diff
@@ -97,23 +182,27 @@ where
...
 
         let zwp_dmabuf = data_init.init(resource, data);
 
         // Immediately send format info to the client if we are the correct version.
         //
         // These events are deprecated in version 4 of the protocol.
-        if zwp_dmabuf.version() <= 3 {
+        if zwp_dmabuf.version() < zwp_linux_dmabuf_v1::REQ_GET_DEFAULT_FEEDBACK_SINCE {
             for format in &*global_data.formats {
                 zwp_dmabuf.format(format.code as u32);
```
ids1024, line 196: I guess this results in formats that are supported with multiple modifiers being sent multiple times? Is it possible for a DRM format to only work with an explicit modifier? In which case this could also send an incorrect format.

Though this is existing code, not added by this PR.

Drakulix: Right the filtering for dmabuf < v3 is not good here.

src/wayland/dmabuf/mod.rs
cmeissl: I will clarify None, but it means to not override the default feedback which will then be sent.
And yes, I forgot to add the surface 😅

anvil/src/udev.rs
cmeissl: The idea was to announce the primary gpu as the main device, a preference tranche for scan-out and a second preference tranche for the gpu on that output. But I am not sure if this will work or have any benefit.

cmeissl:
> Also I didn't see anything to set the main device?

As @Drakulix already wrote this is enforced by the Builder. I opted to enforce the use of a main device instead of adding a separate function.

anvil/src/udev.rs
Drakulix: Ah so we would have:
- Scanout trench
- Non-scanout Trench of the optimal gpu for the multi renderer
- Non-scanout Trench of the primary gpu

Yeah, that seems reasonable. Can't we move global initialization after the udev.add_device calls to do this?

cmeissl: You mean the feedback inizialization per surface? Maybe, though we need to make the feedback optional and late initialize it then. But it should be doable.

cmeissl: Hm, not sure about hot-plugging though. Maybe we need to handle that in both planes.

anvil/src/state.rs
cmeissl: I don't think using two fields is a good idea, we need both or none. But I agree that without a name it is hard to tell what they are used for. An alternative would be a struct?

anvil/src/udev.rs
```diff
@@ -101,7 +107,7 @@ pub struct UdevData {
     pub session: LibSeatSession,
     dh: DisplayHandle,
     #[cfg(feature = "egl")]
-    dmabuf_state: Option<(DmabufState, DmabufGlobal)>,
+    dmabuf_state: Option<(DmabufState, DmabufGlobal, DmabufFeedback)>,
```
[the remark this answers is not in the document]
cmeissl: Actually this is never used again and the global already stored the feedback internally, I will remove it here.

src/wayland/dmabuf/dispatch.rs
cmeissl: should be fixed in latest version

src/wayland/dmabuf/mod.rs
cmeissl: done

src/wayland/dmabuf/mod.rs
cmeissl: done

anvil/src/udev.rs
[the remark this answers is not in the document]
cmeissl: removed

anvil/src/udev.rs
cmeissl: Good catch, I added the texture formats but forgot to use them. I think the non egl case should be fine as we only need the render formats for initializing the surface.

src/wayland/dmabuf/mod.rs
cmeissl: renamed to preference tranche

anvil/src/state.rs
cmeissl: moved to a separate struct

anvil/src/udev.rs
cmeissl: @Drakulix I implemented some late init for feedback which now can use the primary gpu

anvil/src/udev.rs
cmeissl: refactored because of late init which now uses the renderer dmabuf formats

src/wayland/dmabuf/mod.rs
cmeissl: added `DmabufFeedbackFormat::to_ne_bytes` and replaced the unsafe part with `.flat_map(DmabufFeedbackFormat::to_ne_bytes).collect()`

src/wayland/dmabuf/dispatch.rs
```diff
@@ -97,22 +182,29 @@ where
...
-                if zwp_dmabuf.version() == 3 {
-                    let modifier_hi = (Into::<u64>::into(format.modifier) >> 32) as u32;
-                    let modifier_lo = Into::<u64>::into(format.modifier) as u32;
+        if zwp_dmabuf.version() < zwp_linux_dmabuf_v1::REQ_GET_DEFAULT_FEEDBACK_SINCE {
+            for (fourcc, modifiers) in &*global_data.formats {
+                // Modifier support got added in version 3
+                if zwp_dmabuf.version() < zwp_linux_dmabuf_v1::EVT_MODIFIER_SINCE
+                    && modifiers.contains(&Modifier::Invalid)
+                {
+                    zwp_dmabuf.format(*fourcc as u32);
```
Drakulix, line 200: I don't think you are supposed to never send `format` on v3, the event was only deprecated on v4.

Drakulix (commented):
Almost there :)

src/wayland/dmabuf/dispatch.rs
cmeissl: That may be a bit under-defined in the spec, but neither `wlroots` nor `weston` send both. Both either send the format if version is less than modifier since or the modifiers otherwise. Though it seems weston sends the format if it is invalid or linear while wlroots only send the format if it includes invalid.

Drakulix: Ok sure, then lets keep it like this.

i509VCB (approved):
The content looks good.

I'd like some module documentation to describe some of the feedback related api.

src/wayland/dmabuf/mod.rs
```diff
@@ -270,6 +769,24 @@ pub enum ImportError {
     InvalidFormat,
 }
 
+/// Handler trait for [`DmabufFeedback`]
+pub trait DmabufFeedbackHandler {
```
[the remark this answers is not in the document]
cmeissl: Yeah, I think I had the invocation on some other place, but now it does not really make sense. I will move the function including the default impl to the `DmabufHandler`. This has the nice effect that this PR has no breaking change :)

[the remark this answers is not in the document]
cmeissl: Moved to `DmabufHandler` and removed `DmabufFeedbackHandler`

cmeissl:
> The content looks good.
> 
> I'd like some module documentation to describe some of the feedback related api.

I updated the module docs, can you please take a look?

anvil/src/winit.rs
```diff
@@ -129,24 +129,46 @@ pub fn run_winit() {
...
+                .unwrap();
+            let mut dmabuf_state = DmabufState::new();
+            let dmabuf_global = dmabuf_state.create_global_with_default_feedback::<AnvilState<WinitData>>(
+                &display.handle(),
+                &dmabuf_default_feedback,
+            );
+            Some((dmabuf_state, dmabuf_global, dmabuf_default_feedback))
+        }
+        Ok(None) => {
+            warn!("failed to query render node, dmabuf will be unavailable");
```
Drakulix, line 149: Couldn't we do v3 at least, if the `EGLDevice`-extensions are not available or if the renderer has no `render_node` (either because it is an old driver or because we are using llvmpipe for example?).

Drakulix: I guess this is not super important, given we still enable `wl_drm` (at least), but we could add at least a comment here about it.

anvil/src/udev.rs
```diff
@@ -799,6 +826,86 @@ enum DeviceAddError {
...
+        .add_preference_tranche(render_node.dev_id(), None, render_formats.clone())
+        .build()
+        .unwrap();
+
+    // We only provide a scan-out tranche if the drm node we selected as the preference for
+    // rendering equals the scan-out drm node
+    let can_scanout = render_node
+        .node_with_type(NodeType::Primary)
+        .and_then(|n| n.ok().map(|n| n.dev_id()))
+        == surface.device_fd().dev_id().ok();
```
Drakulix, line 887: I think this whole check is unnecessary.

Given we could scan out on the secondary gpu as well - if it has outputs - and plane_formats are filtered for render-able formats anyway, we can just advertise the scanout tranche without any restrictions.

Drakulix (commented):
Final review round in my opinion. Just a couple of nitpicks, I really like the api.

src/wayland/dmabuf/mod.rs
```diff
@@ -8,25 +8,26 @@
 //!
 //! To setup the dmabuf global, you will need to provide 2 things:
 //!
-//! - a list of the dmabuf formats you wish to support when creating the `Global` through `DmabufState::create_global`
-//! - an implementation of `DmabufHandler` to test if a dmabuf buffer can be imported by your renderer
+//! - the default [`DmabufFeedback`] containing the main device and the formats you wish to support when creating the `Global` through [`DmabufState::create_global_with_default_feedback`]
+//! - an implementation of [`DmabufHandler`] to test if a dmabuf buffer can be imported by your renderer and optional override the initial surface feedback
```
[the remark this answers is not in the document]
cmeissl: done

anvil/src/winit.rs
cmeissl: Yeah, but I added the fallback path to initialize v3 and added a note that mesa requires either v4 or wl_drm

anvil/src/udev.rs
cmeissl: right, this was added with optimus setup in mind. 
removed the check

PR #901, opened 2023-02-22
gles2: custom pixel shaders

Drakulix:
Adds first-party support for shader-based render elements.

Refactors a few things in the process:

- Combines our three fragment texture shaders into one and compiles it multiple times with corresponding `#define`s.
- Adds another debug shader permutation for `debug_flags`

TODO:

- [ ] Add the same option for custom texture shaders


PolyMeilex:
If our shaders are now nicely combined, would it make sense to move them to `.vert`, `.frag` files, and `include_str!` them, to get proper syntax highlighting and editor support?

Drakulix:
> If our shaders are now nicely combined, would it make sense to move them to `.vert`, `.frag` files, and `include_str!` them, to get proper syntax highlighting and editor support?

Sure thing https://github.com/Smithay/smithay/pull/901/commits/596c4d0b3fc7909afd09459fc2535e270de9d78d

src/backend/renderer/gles2/element.rs
```diff
@@ -0,0 +1,222 @@
...
+            &self.shader,
+            dst,
+            Some(damage),
+            self.alpha,
+            &self.additional_uniforms,
+        )
+    }
+}
+
+/// Wrapping Render element to replace the default texture shader used
```
cmeissl, line 111: It might be worth mentioning that this will prevent direct scan-out of the wrapped element. Either here on in the constructor docs.

Drakulix: Right, this element should also forcefully override `underlying_storage`, good call.

src/backend/renderer/gles2/element.rs
```diff
@@ -0,0 +1,236 @@
...
+    fn underlying_storage(&self, _renderer: &mut Gles2Renderer) -> Option<UnderlyingStorage> {
+        None
+    }
+}
+
+/// Wrapping Render element to replace the default texture shader used.
+///
+/// **Note*: This will disallow direct-scanout to happen, e.g. when using the [`DrmCompositor`].
+#[derive(Debug, Clone)]
+pub struct TextureShaderWrapperElement<E> {
```
cmeissl, line 123: We may want to override the reported `Id` for this element, this could be important if the element is dynamically switching between wrapped and non-wrapped. Reporting a different `Id` for the wrapped element would make sure it gets a full redraw.

cmeissl (commented):
I really like that approach and the flexibility it introduces :+1:

src/backend/renderer/gles2/element.rs
cmeissl: Thinking about that again, overriding the id could lead to some surprising effect. Like a wayland surface no longer getting frame callbacks. So that might be a bad idea. Maybe there is some other way we can force a redraw.

Drakulix: Dropped `TextureShaderWrapperElement` for various reasons.

cmeissl (approved):

PR #904, opened 2023-02-23
Extract WLCS functionality (based on #1160)

bwidawsk:
Compositors that wish to implement the WLCS test suite can benefit from wlcs_anvil. The two primary functionalities that are needed are:
1. WLCS FFI
2. DummyRenderer support

This PR bring both. Testing WLCS before and after this change yield the same number of Pass/Skip/Fail


src/backend/renderer/dummy.rs
```diff
@@ -1,6 +1,9 @@
+//! DummyRenderer
```
i509VCB, line 1: DummyRenderer actually isn't a very good name since it does do something (tries to access every pixel in an imported shm buffers). Maybe TestRenderer would make more sense?

wlcs_anvil/src/ffi_wrappers.rs
```diff
@@ -15,438 +13,156 @@ use wayland_sys::{
...
+    fn create_touch(&mut self) -> Option<Self::Touch> {
+        let Some(ref server) = self.server else { return None; };
+        Some(TouchHandle {
+            device_id: DEVICE_ID.fetch_add(1, Ordering::Relaxed),
+            sender: server.0.clone(),
+        })
     }
 
-    unsafe extern "C" fn get_descriptor(_: *const WlcsDisplayServer) -> *const WlcsIntegrationDescriptor {
+    fn get_descriptor(&self) -> *const WlcsIntegrationDescriptor {
```
i509VCB, line 99: Should this return a &[WlcsIntegrationDescriptor]?

wlcs_anvil/src/ffi_wrappers.rs
```diff
@@ -1,11 +1,9 @@
 use std::{
-    os::{
-        raw::{c_char, c_int},
-        unix::{
-            net::UnixStream,
-            prelude::{AsRawFd, IntoRawFd},
-        },
```
i509VCB, line 7: The entire ffi_wrappers file doesn't really have a reason to exist anymore, probably just move it to lib.rs

wlcs_anvil/src/ffi_wrappers.rs
bwidawsk: Hmm, why not just `WlcsIntegrationDescriptor` actually? What does &[] add?

i509VCB: Oh I forgot the array is inside the type.

For returning WlcsIntegrationDescriptor, you'd have to ensure the two internal pointers live as long as the owning instance of the Wlcs trait. I guess just a `&WlcsIntegrationDescriptor` would be fine

src/backend/renderer/mod.rs
```diff
@@ -43,6 +43,8 @@ pub mod element;
 
 pub mod damage;
 
+pub mod test;
```
Drakulix, line 46: Can you please add a new (not enabled by default) feature to this module? Something like `renderer_test`.

wlcs_anvil/Cargo.toml
```diff
@@ -17,3 +17,4 @@ libc = "0.2"
 memoffset = "0.6"
 cgmath = "0.18"
 nix = "0.24"
+wlcs-rs = { git = "https://gitlab.freedesktop.org/bwidawsk/wlcs-rs.git" }
```
Drakulix, line 20: I would definitely like to move this to the smithay-org before merging this PR.

Drakulix (approved):
Looks pretty good to me! I don't know if the rename from Dummy -> Test was super necessary, but it doesn't bother me either. But I would like to feature-gate the new renderer and enable it in wlcs_anvil.

wlcs_anvil/src/lib.rs
```diff
@@ -107,6 +118,145 @@ pub enum WlcsEvent {
...
+        });
+    }
+
+    fn touch_up(&mut self) {
+        let _ = self.sender.send(WlcsEvent::TouchUp {
+            device_id: self.device_id,
+        });
+    }
+
+    fn destroy(&mut self) {}
```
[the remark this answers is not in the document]
bwidawsk: These were copied from the original implementation. I can do as you recommend, but I'd do it as a follow-on patch to keep the "`mv`" history as clean as possible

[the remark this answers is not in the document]
i509VCB: Ok

PR #905, opened 2023-02-23
Split event source out of DrmDevice

PolyMeilex:
This is basically a proposal to use the same pattern as we do in session module, but for DrmDevice.
This eliminates the need for Dispatcher, and IMO makes building abstractions easier, as storing Dispatchers pollutes parent type with generics.
Also potentially less confusing for newcomers, as Dispatchers are never used in the codebase expect this one special case.


Drakulix (approved):
Sure, why not. This is a good idea :+1:

PR #908, opened 2023-02-24
Fix wlr LayerSurface dose not destroyed.

songww:


Drakulix:
This should be done automatically by `LayerMap::arrange` and `LayerMap::cleanup`, but sure, this doesn't hurt.

Drakulix (approved):

songww:
> 

I'm using envil these days. wofi's window disappears after exit, but that area still belongs to the layer created by wofi, which gets mouse and keyboard focus when clicked.

Sorry, my English is poor, this is translated using deepl.

PR #910, opened 2023-02-26
Fix overlay zpos check during rendering

chrisduerr:
This fixes an issue where Overlay planes were assumed to be below the primary plane when they both had a zpos of 0. However some devices do not report a zpos for the overlay plane which should still put it above the primary one.

To fix this, overlay planes are now assumed to be above the primary plane when their zpos is equal to that of the primary plane. This is consistent with the other `is_underlay` check in
`try_assign_overlay_plane`.


Drakulix (approved):
Looks good to me. cc @cmeissl ?

cmeissl:
> Looks good to me. cc @cmeissl ?

Yes, that was definitely wrong and should be `<` 
While that fixes this issue I have to look again at the ordering guarantees if we have no zpos.
But that is independent from this Issue.
So, LGTM!

PR #911, opened 2023-02-28
Fix build of `backend_vulkan` and `xwayland` on FreeBSD

ids1024:


i509VCB (approved):
The Vulkan half of the pull request looks fine. I'm less familiar with the xwayland half however, you'll probably need someone else to review that half.

ids1024:
I avoided changing the xwayland code in my previous PR since I wasn't sure what the abstract socket was for, but apparently on Linux X listens on both an abstract domain socket and on one that exists on the filesystem. But abstract sockets are a Linux extension so this isn't a thing elsewhere.

ids1024:
Updated with a small typo fix.

Apparently Mutter ended up disabling the abstract socket in XWayland: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1424

Not sure about other compositors, what use it has, or whether it has a negative impact on security (except the security problems that exist with X regardless).

Drakulix:
> Updated with a small typo fix.
> 
> Apparently Mutter ended up disabling the abstract socket in XWayland: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1424
> 
> Not sure about other compositors, what use it has, or whether it has a negative impact on security (except the security problems that exist with X regardless).

I think we can just drop it on linux as well?

It allows clients to connect without access to the socket-path, just the name and I think some clients even guess that `:0` exists, if they can't find a `DISPLAY` variable. Which is confusing from a UX-perspective anyway, once you are having more then one session running.

ids1024:
Actually, it seems that was reverted, and remains reverted in the current Mutter `main`: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1508

It broke snaps and some other things.

So I guess we can keep the abstract socket on Linux for now, and remove it in the future of the things that rely on it are updated.

Drakulix:
> Actually, it seems that was reverted, and remains reverted in the current Mutter main: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1508

Of course :woman_facepalming: ...

Can we make this optional at runtime at least? So that disabling it doesn't require patching smithay.

i509VCB:
If you wanted to open a pull request with the vulkan half of this I'd be happy to merge it now.

ids1024:
I don't think there's any particular reason not to merge this now. This shouldn't impact behavior on Linux, and a compile or run time option to disable abstract sockets on Linux is a separate feature that can be added later.

(Not that that's complicated, but I'm not sure how that should be added. I guess `XWayland::start` would take a bool, and the compositor can decide whether to enable this, or how to configure it?)

Drakulix:
> I guess `XWayland::start` would take a bool, and the compositor can decide whether to enable this, or how to configure it?

That is exactly, why I had in mind. But I guess we can just merge it as is.

ids1024:
Okay, rebased, added an argument to configure this, and make sure it still compiles on FreeBSD.

Drakulix (approved):

PR #912, opened 2023-02-28
logging: Be less expensive with tracing

Drakulix:
This downgrades the log-level of a bunch of spans, which annotate hot paths, many of which aren't even used for actual log messages.
Because entering and leaving spans generates internal events - even though we don't log them - this caused severe performance implications, observable using cargo-flamegraph before and after this commit.


PolyMeilex (approved):
Unused import aside, LGTM

PR #915, opened 2023-03-01
Fix/vt switch

### LINKED ISSUE #907: Anvil crashing on tty switch off of it and back
This seems to be happening for me since https://github.com/Smithay/smithay/pull/760.

```
2023-02-24T17:16:55.032058Z  WARN anvil::udev: Error during rendering: ContextLost(PrimaryPlaneTestFailed)
thread 'main' panicked at 'Rendering loop lost: Testing the primary plane failed', anvil/src/udev.rs:1230:63
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I'm not really sure how the compositor is supposed to handle losing control of DRM device on tty switch and then re-configuring planes on the switch back, but I guess that isn't being handled correctly.

And if other people aren't seeing this issue... would it be hardware related, based on what planes the DRM device offers?

cmeissl:
fixes #907

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -271,11 +271,15 @@ pub fn run_udev() {
         .insert_source(notifier, move |event, &mut (), data| match event {
             SessionEvent::PauseSession => {
                 libinput_context.suspend();
+                info!("pausing session");
+
                 for backend in data.state.backend_data.backends.values() {
                     backend.drm.pause();
                 }
             }
             SessionEvent::ActivateSession => {
+                info!("resuming session");
+
                 if let Err(err) = libinput_context.resume() {
                     error!("Failed to resume libinput context: {:?}", err);
                 }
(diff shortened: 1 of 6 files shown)


anvil/src/udev.rs
```diff
@@ -1034,7 +1044,8 @@ impl AnvilState<UdevData> {
                 warn!("Error during rendering: {:?}", err);
                 match err {
                     SwapBuffersError::AlreadySwapped => true,
-                    SwapBuffersError::TemporaryFailure(err) => matches!(
+                    // If the device has been deactivated or a vt switch is in progress do not reschedule
+                    SwapBuffersError::TemporaryFailure(err) => !matches!(
```
Drakulix, line 1048: I think this is too simplistic. E.g. after resuming the session my nvidia gpu will cause TemporaryFailures for one or two frames until operation continues as normal.

So I think we unfortunately need to differentiate here between TemporaryErrors, when the device should be active or should not be active.

Drakulix (commented):
Looks mostly good to me.

anvil/src/udev.rs
cmeissl: Ok, makes sense. If we only disable re-schedule for not active it should be fine, though it will render a few frames that will never make it to the screen. But that is not really an issue I guess.

cmeissl: done

Drakulix (approved):

PR #917, opened 2023-03-02
fix(layer): arrange layer surfaces accounting for margins

wash2:
This seems to improve arrangement of layer surfaces with margins. I tested it in anvil with a bunch of anchored layer surfaces with exclusive zones on each edge.


Drakulix (approved):
Looks like a good fix!

PR #919, opened 2023-03-04
Add from DRM mode to smithay output mode conversion

PolyMeilex:
Weston, wlroots and cosmic-comp all use the same logic for this (minus some stylistic differences), so I think it's safe to just provide default behavior.


Drakulix (approved):
Seems reasonable!

PR #920, opened 2023-03-05
Implement `Default` for `RenderElementStates`

chrisduerr:
This change makes it easier for consumers to just `mem::take` the states from the frame result, which is useful since it has a lifetime attached to the original input data.

This patch now derives all important common traits that are directly derivable, following Rust's official API guidelines: https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits


Drakulix (approved):
cc @cmeissl, but I don't see anything wrong with it.

cmeissl:
> cc @cmeissl, but I don't see anything wrong with it.

Yes, should be totally fine LGTM :+1:

PR #927, opened 2023-03-08
desktop: Remove outdated doc comments about `z_index`

ids1024:
It seems since e8c977c9 `Space::map_element` doesn't take a `z_index` argument, and `SpaceElement::z_index` doesn't take a space. So these mentions are outdated.


Drakulix (approved):
:+1:

PR #933, opened 2023-03-11
Adding rust toolchain file

TornaxO7:
In my opinion it wouldn't be bad, if smithay would include the
rust-toolchain.toml file because then it's likely guaranteed that all devs are
using the same rust-tools.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "stable"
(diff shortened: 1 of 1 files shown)


Drakulix:
Good idea, however we actually have a MSRV, which we would want to enforce. Looking at CI, this seems to be 1.65.0 currently.

TornaxO7:
Good point! Done!

PR #934, opened 2023-03-13
xwm: Fix setting `_NET_ACTIVE_WINDOW`

ids1024:
This is probably not the best way to handle this, but I'm not sure what is. It requires a new root argument to `X11Surface::new`, which is a change to the public API though this function is mainly called within Smithay itself, it seems.

Previously this was set based on the `FocusIn`/`FocusOut` events of the root window, but those always have the id of the root window. Presumably X sends the events to the focused window and all parent windows. So applications saw a change to `_NET_ACTIVE_WINDOW`, but the value was always the id of the root window. This seems to behave more correctly.

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -854,6 +854,7 @@ fn handle_event<D: XwmHandler>(state: &mut D, xwmid: XwmId, event: Event) -> Res
                 n.override_redirect,
                 Arc::downgrade(&conn),
                 xwm.atoms,
+                xwm.screen.root,
                 Rectangle::from_loc_and_size(
                     (geo.x as i32, geo.y as i32),
                     (geo.width as i32, geo.height as i32),
@@ -1099,24 +1100,6 @@ fn handle_event<D: XwmHandler>(state: &mut D, xwmid: XwmId, event: Event) -> Res
             if let Some(surface) = xwm.windows.iter().find(|x| x.window_id() == n.window) {
                 surface.update_properties(Some(n.atom))?;
             }
-        }
-        Event::FocusIn(n) => {
-            conn.change_property32(
-                PropMode::REPLACE,
-                xwm.screen.root,
-                xwm.atoms._NET_ACTIVE_WINDOW,
-                AtomEnum::WINDOW,
-                &[n.event],
-            )?;
-        }
-        Event::FocusOut(n) => {
-            conn.change_property32(
-                PropMode::REPLACE,
-                xwm.screen.root,
-                xwm.atoms._NET_ACTIVE_WINDOW,
-                AtomEnum::WINDOW,
-                &[n.event],
-            )?;
         }
         Event::ClientMessage(msg) => {
             if let Some(reply) = conn.get_atom_name(msg.type_)?.reply_unchecked()? {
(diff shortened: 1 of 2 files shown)


Drakulix (changes requested):
While I think this is a correct solution (though I find it baffling, that the FocusIn-event doesn't seem to carry the focused window then), I am unhappy, that this is moving window-management code (setting _NET_ACTIVE_WINDOW) into the `X11Surface` type. (Which is meant to be usable by custom WM implementations.)

I think I would prefer, if this would be sending a client-request from the X11Surface, like the Extended Window Manager Hints describe: https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45240717405776

That way we could focusing our own windows just like we handle clients requesting focus. (Which should hopefully work out and not turn into problems later. :sweat_smile: )

ids1024:
> I am unhappy, that this is moving window-management code (setting _NET_ACTIVE_WINDOW) into the X11Surface type. (Which is meant to be usable by custom WM implementations.)

Is this also a concern with `_NET_WM_STATE`? That's currently set in `X11Surface::change_net_state`, but EWMH states "A Client wishing to change the state of a window MUST send a _NET_WM_STATE client message to the root window".

So technically `X11Surface` shouldn't do this if it isn't part of the "window manager". But strict conforming to what is and is not the job of the window manager may not be necessary here.

Drakulix:
> Edit: Well, X11Wm seems to handle _NET_WM_STATE client messages by calling methods of the X11Surface, which then updates the property. So it should be okay, given one component is responsible for managing it.

Right, the idea here is, that an external WM could also call `X11Surface::change_net_state`, it is basically a helper.

For focus-changes this is not that easy, because the call doesn't originate from `X11Wm`, because it is part of the `KeyboardTarget` implementation.

ids1024:
>  an external WM could also call X11Surface::change_net_state, it is basically a helper.

Well, at present it couldn't, since `change_net_state` isn't `pub`.

Drakulix:
> > an external WM could also call X11Surface::change_net_state, it is basically a helper.
> 
> Well, at present it couldn't, since `change_net_state` isn't `pub`.

Oh, well I guess we could change that.

ids1024:
Sending a `_NET_ACTIVE_WINDOW` client message is easy enough, though not very useful unless `X11WM` handles it. Where things are a little less obvious.

It's expected that a EWMH WM will react to this message by 1) doing nothing 2) changing focus to the window, which includes setting this property 3) not actually changing focus but setting a flag like `_NET_WM_STATE_DEMANDS_ATTENTION`.

Presumably the WM shouldn't respond to a `_NET_ACTIVE_WINDOW` message by just setting that property, and not changing focus. But do we want any X client to actually be able to set the focused window? That seems like a behavior not all Smithay compositors, at least, would desire.

Drakulix:
> Presumably the WM shouldn't respond to a `_NET_ACTIVE_WINDOW` message by just setting that property, and not changing focus. But do we want any X client to actually be able to set the focused window? That seems like a behavior not all Smithay compositors, at least, would desire.

Yeah, that sounds like we want a `focus_request` for the `XwmHandler` trait, which works for clients sending that message, but is a little awkward to receive for our own surfaces, which should have just got focus anyway.

So I guess we need another communication channel back from the KeyboardTarget implementation to the X-WM, which is easier said than done. Maybe a custom client-message?

PR #935, opened 2023-03-14
10-bit color support

Drakulix:
Based on #901 due to shader changes.


src/backend/allocator/format.rs
```diff
@@ -66,6 +66,23 @@ macro_rules! format_tables {
             }
         }
 
+        /// Returns the transparent alternative of the specified format.
+        ///
+        /// If the format has an unused alpha channel, this may return the corresponding non-opaque format.
+        ///
+        /// Unknown formats will always return [`None`].
+        pub const fn get_transparent(
```
i509VCB, line 74: Could you add a test to ensure we aren't breaking that behavior like the other tests for the format tables?

src/backend/renderer/gles2/format.rs
```diff
@@ -0,0 +1,34 @@
...
+                    ($crate::backend::renderer::gles2::ffi::$gl_format, $crate::backend::renderer::gles2::ffi::$gl_layout) => Some($crate::backend::allocator::Fourcc::$fourcc),
+                )*
+                _ => None,
+            }
+        }
+    }
+}
+
+gl_format_table! {
+    Abgr8888 => (RGBA, UNSIGNED_BYTE),
```
i509VCB, line 28: Isn't there an Argb8888 -> something conversion the gles renderer already does?

src/wayland/shm/mod.rs
```diff
@@ -366,6 +366,147 @@ pub fn has_alpha(format: wl_shm::Format) -> bool {
...
 }
 
+macro_rules! shm_format_table {
+    (
+        $(
+            $fourcc: ident => $shm: ident
+        ),* $(,)?
+    ) => {
+        /// Convert from a [`Fourcc`](backend::allocator::Fourcc) format to a wl_shm format
+        pub const fn fourcc_to_shm_format(value: $crate::backend::allocator::Fourcc) -> Option<$crate::reexports::wayland_server::protocol::wl_shm::Format> {
```
i509VCB, line 376: Only if we had impl const...

Drakulix: Even then this doesn't work, because of the orphan rule. `Fourcc` and `wl_shm` are defined in different crates.

i509VCB: Well less so for a From impl and more so that we can just map Argb8888 and Xrgb8888 and the using wl_shm::Format::from(u32).

Drakulix: yeah, right

src/backend/allocator/format.rs
Drakulix: Added https://github.com/Smithay/smithay/pull/935/files#diff-4defc011b2baa0cfeeadfc075ce3121157bcbf5ec99b70d725a130c9961d7035R384-R401

src/backend/drm/compositor/gbm.rs
```diff
@@ -185,14 +194,19 @@ pub struct DrmError {
 }
 
 /// Attach a [`framebuffer::Handle`] to an [`BufferObject`]
-pub fn framebuffer_from_bo<T>(drm: &DrmDeviceFd, bo: &BufferObject<T>) -> Result<GbmFramebuffer, DrmError> {
+pub fn framebuffer_from_bo<T>(
+    drm: &DrmDeviceFd,
+    bo: &BufferObject<T>,
+    consider_opaque: bool,
```
cmeissl, line 200: As this actually implements a fallback path we may want to also express that in the argument name.
Maybe something like `allow_opaque_fallback`?

src/backend/renderer/element/memory.rs
```diff
@@ -261,10 +281,15 @@ impl MemoryRenderBufferInner {
         // TODO: Allow to specify the format when ImportMem
         // has been changed to allow specifying a format
```
cmeissl, line 282: This TODO can also be removed

src/backend/renderer/element/memory.rs
```diff
@@ -231,16 +246,20 @@ impl Default for MemoryRenderBufferInner {
...
         opaque_regions: Option<Vec<Rectangle<i32, Buffer>>>,
     ) -> Self {
-        // TODO: Allow to specify the format when ImportMem
-        // has been changed to allow specifying a format
         let size = size.into();
         MemoryRenderBufferInner {
-            mem: vec![0; (size.w * 4 * size.h) as usize],
+            mem: vec![
+                0;
+                (size.w * (get_bpp(format).expect("Format with unknown bits per pixel") as i32) * size.h)
```
cmeissl, line 259: Do we expect that `expect` to ever be invoked? I am not sure if returning a `Result` would be justified, but at least we should add a comment here and on `from_memory` warning about that panic if it can happen.

src/backend/renderer/mod.rs
```diff
@@ -119,13 +119,27 @@ pub trait Texture {
...
     /// Height of the texture plane
     fn height(&self) -> u32;
+
+    /// Format of the texture, if available.
+    ///
+    /// In case the format is hidden by the implementation,
+    /// it should be assumed, that the pixel representation cannot be read.
+    ///
+    /// Thus [`ExportMem::copy_texture`], if implemented, will not succeed for this texture.
+    /// Not that this does **not** mean every texture with a format is guaranteed to be copyable.
```
cmeissl, line 129: Maybe a typo here Not -> Note

src/wayland/shm/mod.rs
```diff
@@ -366,6 +378,147 @@ pub fn has_alpha(format: wl_shm::Format) -> bool {
...
+                $(
+                    $crate::reexports::wayland_server::protocol::wl_shm::Format::$shm => Some($crate::backend::allocator::Fourcc::$fourcc),
+                )*
+                _ => None,
+            }
+        }
+    }
+}
+
+shm_format_table! {
```
cmeissl, line 409: Given that most formats directly map we could extend the macro to define a drm format without specifying the shm variant and just pass the drm ident as the shm ident. But I am not sure this is really worth extending the macro.

src/backend/drm/surface/gbm.rs
```diff
@@ -445,6 +451,33 @@ impl Drop for FbHandle {
     }
 }
 
+struct BufferWrapper<'a, B>(&'a B);
```
cmeissl, line 454: We could try to promote the already `pub` `framebuffer_from_bo` function from `drm::compositor::gbm` to some drm util module and re-use the logic here. I missed that opportunity in the planes PR unfortunately.

cmeissl (commented):
Looks pretty solid :+1: 

I am not totally happy about the new parameter in `ExportFramebuffer`, but looking at the uses I also can't think about a better alternative. I thought it might be possible to define a `GbmFramebufferExporter` like you did for the gbm allocator, but we need to be able to define it dynamically for direct scan-out on the primary plane.

anvil/src/udev.rs
```diff
@@ -88,6 +88,20 @@ use smithay::{
 };
 use tracing::{debug, error, info, trace, warn};
 
+// we cannot simply pick the first supported format of the intersection of *all* formats, because:
+// - we do not want something like Abgr4444, which looses color information, if something better is available
+// - some formats might perform terribly
+// - we might need some work-arounds, if one supports modifiers, but the other does not
+//
+// So lets just pick `XRGB2101010` (10-bit) or `XRGB8888` (8-bit) for now, they are widely supported.
```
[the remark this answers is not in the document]
Drakulix: Right, needs to be updated. This is now not referring to scanout formats anymore, but only framebuffer formats.

src/backend/renderer/element/memory.rs
Drakulix: A result only makes sense if there is a way to handle this case at runtime. Given we strive to implement `get_bpp` for every possible format and the `MemoryRenderElement` is meant to be used for software rendered content, where you usually know the format during compile time, it did occur to me that there would be reason to do that. But I guess I could convert this into a proper error case?

src/wayland/shm/mod.rs
Drakulix: Honestly I like the explicit variant and its not like we will be extending that list every other day.

anvil/src/udev.rs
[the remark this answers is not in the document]
Drakulix: fixed

src/backend/drm/compositor/gbm.rs
Drakulix: done

src/backend/renderer/element/memory.rs
Drakulix: done

src/backend/renderer/mod.rs
Drakulix: fixed

src/backend/drm/surface/gbm.rs
Drakulix: Done in https://github.com/Smithay/smithay/pull/935/commits/6d32c24f8a4cc2cd6047d817d2196f9aea35e026

Drakulix:
Reverting back to a draft, because we can't really do blending inside a `Argb2101010`-buffer. So I guess the Gles2Renderer will need a `16161616` shadow-buffer....

src/backend/renderer/gles/mod.rs
```diff
@@ -1385,20 +1226,38 @@ impl Gles2Renderer {
...
+        let (format, layout) = gl_read_for_internal(internal).ok_or(GlesError::UnknownPixelFormat)?;
+
         let mut pbo = 0;
         unsafe {
             self.gl.GenBuffers(1, &mut pbo);
             self.gl.BindBuffer(ffi::PIXEL_PACK_BUFFER, pbo);
-            let size = (region.size.w * region.size.h * 4) as isize;
+            let size = (region.size.w
+                * region.size.h
+                * (gl_bpp(format, layout).ok_or(GlesError::UnsupportedPixelLayout)? / 8) as i32)
```
cmeissl, line 1259: I find that a bit hard to read with the inline `ok_or` part, maybe something like `let bpp = gl_bpp(format, layout).ok_or(GlesError::UnsupportedPixelLayout)? / 8) as i32` would increase readability.

src/backend/renderer/gles/mod.rs
```diff
@@ -1432,12 +1294,18 @@ impl ExportMem for Gles2Renderer {
...
+
         unsafe {
             self.gl.GenBuffers(1, &mut pbo);
             self.gl.BindBuffer(ffi::PIXEL_PACK_BUFFER, pbo);
             self.gl.BufferData(
                 ffi::PIXEL_PACK_BUFFER,
-                (region.size.w * region.size.h * 4) as isize,
+                (region.size.w
+                    * region.size.h
+                    * (gl_bpp(format, layout).expect("We check the format before") / 8) as i32)
```
cmeissl, line 1307: same here (if it has been checked before we can also assign it once maybe?)

src/backend/renderer/gles/texture.rs
```diff
@@ -0,0 +1,140 @@
+use super::*;
+
+/// A handle to a GLES2 texture
+#[derive(Debug, Clone)]
+pub struct GlesTexture(pub(super) Rc<GlesTextureInternal>);
+
+impl GlesTexture {
+    /// Create a GlesTexture from a raw gl texture id.
+    ///
+    /// This expects the texture to be in RGBA format to be rendered
```
cmeissl, line 10: Does this still apply? The function accepts a format so I would think it is possible to use a different format.

src/backend/renderer/glow.rs
```diff
@@ -38,16 +38,16 @@ use super::{element::RenderElement, Frame};
 #[derive(Debug)]
 /// A renderer utilizing OpenGL ES 2 and [`glow`] on top for easier custom rendering.
```
cmeissl, line 39: Still mentions ES 2

anvil/src/udev.rs
```diff
@@ -321,6 +326,14 @@ pub fn run_udev() {
...
         }
     }
+    state.shm_state.add_formats(
+        state
+            .backend_data
+            .gpus
+            .single_renderer(&primary_gpu)
+            .unwrap()
+            .shm_formats(),
+    );
```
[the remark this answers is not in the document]
Drakulix: Non issue for anvil.

Behaviour is now documented and it is now possible to select different compatible formats, when copying from a gpu. As such the multigpu-renderer can now fallback to 8-bit formats for transfers to ES-2.0-only GPUs.

src/wayland/shm/mod.rs
```diff
@@ -160,8 +160,11 @@ impl ShmState {
     /// Updates the list of formats advertised by the global.
     ///
     /// This will only affect new binds to the wl_shm global.
-    pub fn add_formats(&mut self, formats: impl IntoIterator<Item = wl_shm::Format>) {
-        self.formats.extend(formats.into_iter());
+    ///
+    /// Removing formats will cause old clients trying to create
+    /// a buffer of a now unsupported format to be killed.
+    pub fn update_formats(&mut self, formats: impl IntoIterator<Item = wl_shm::Format>) {
+        self.formats = formats.into_iter().collect::<HashSet<_>>();
```
cmeissl, line 167: This could potentially remove the mandatory formats, which feels a bit inconsitent as `ShmState::new` always adds them also mentions in the docs that the global will always advertize them.

cmeissl (commented):
This PR now includes a bunch of commits from master that are not part of the base branch which makes the review a bit hard. Maybe you want to also rebase the base branch against master?

Drakulix:
> This PR now includes a bunch of commits from master that are not part of the base branch which makes the review a bit hard. Maybe you want to also rebase the base branch against master?

Thanks for the heads up, forgot to change the base to master (the old base branch was merged in the meantime and I rebased against master). Should be fixed now.

cmeissl (approved):

PR #941, opened 2023-03-15
xwm: Fix handling of `WM_TAKE_FOCUS` windows

ids1024:
As described in ICCCM, "Locally Active Input" windows expect to receive input focus from the WM, but may then want to set the focus to another of its windows. So it is still necessary to call `set_input_focus` for these applications to work correctly.

Additionally "Globally Active Input" windows expect to receive `WM_TAKE_FOCUS` messages. So that needs to be sent to them.

I'm not sure if there are more subtleties in exactly when to set input focus and when to send `WM_TAKE_FOCUS`, but this should be more correct, at least.

Fixes issue with `emacs-lucid` described in https://github.com/pop-os/cosmic-comp/issues/92.


Drakulix (approved):
Looks good!

PR #943, opened 2023-03-17
Fix virtual keyboard crash with invalid keymap

### LINKED ISSUE #942: Running wvkbd crashes Anvil
When running [wvkbd](https://github.com/jjsullivan5196/wvkbd), Anvil prints the following warnings and panics:

```
xkbcommon: ERROR: (input string):1:61420: unrecognized token
xkbcommon: ERROR: (input string):1:61420: syntax error
xkbcommon: ERROR: Failed to parse input xkb string
```

The panic is due to an unwrap here:

https://github.com/Smithay/smithay/blob/master/src/wayland/virtual_keyboard/virtual_keyboard_handle.rs#L104

chrisduerr:
Since the keymap sent for the virtual keyboard protocol is supplied by a client, it should not be trusted. Unwrapping on it causes the compositor to crash when a malicious or buggy client tries to send the `Keymap` request.

To ensure even misbehaving clients don't cause any issues, requests with invalid keymaps are simply ignored.

This has been sparked by an issue encountered with `wvkbd`, which sends a keymap with 54 `\0` bytes at the end. While it would be possible for Smithay to clean this up, I'm not sure if it is appropriate for the compositor to work around what is arguably a client issue.

An issue has been filed with `wvkbd` here:
https://github.com/jjsullivan5196/wvkbd/issues/35

Closes #942.

### DIFF
--- a/src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
+++ b/src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
@@ -100,9 +100,14 @@ where
                             format,
                             xkb::KEYMAP_COMPILE_NO_FLAGS,
                         )
-                        .unwrap()
-                        .unwrap()
                     };
+
+                    // Ignore requests with invalid keymap attached.
+                    let new_keymap = match new_keymap {
+                        Ok(Some(new_keymap)) => new_keymap,
+                        _ => return,
+                    };
+
                     if old_keymap != new_keymap.get_as_string(xkb::FORMAT_TEXT_V1) {
                         let mut inner = data.handle.inner.lock().unwrap();
                         inner.old_keymap = Some(old_keymap);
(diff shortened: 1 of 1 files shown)


i509VCB:
> it should not be trusted

In this case the only acceptable answer is to assume we are reading like shm and guard against sigbus.

chrisduerr:
> In this case the only acceptable answer is to assume we are reading like shm and guard against sigbus.

That is probably fair to be honest. Though this patch hopefully provides a simple solution that fixes 99% of all issues, even if it might not stop a malicious actor (which with virtual keyboard is probably a little less relevant anyway).

But if you have any concrete changes to suggest, I'd be happy to make them.

Drakulix (approved):
This is definitely an improvement though and as such should be merged.

@i509VCB @chrisduerr Please someone open an issue describing further measures that can be taken against malicious clients.

vberger:
> > it should not be trusted
> 
> In this case the only acceptable answer is to assume we are reading like shm and guard against sigbus.

No, not really. Given defining a keymap like this is a rare occurrence (compared to a client submitting a buffer), we can take the slightly more expensive step of reading the contents of the FD (with a read syscall, not memmap) into a buffer and then giving that to xkbcommon for initialization. This completely sidesteps the memmap whoes.

From what I see, a raw keymap file is a worst a few tens of kilobytes in size, so we could just put a 1MB hard limit as a safety measure to avoid misbehaving clients.

PR #945, opened 2023-03-18
Improve virtual keyboard keymap handling

chrisduerr:
I wanted to immediately create this PR as a follow-up since otherwise chances are nobody would have gotten to it (arguably it's relatively insignificant). See https://github.com/Smithay/smithay/pull/943#issuecomment-1474854032 for context.

---

This patch further improves upon 7f8618c, by parsing the keymap as a string instead of letting xkbcommon memory map the provided file descriptor.

The keymap is additionally also limited to 1MiB in size, to ensure clients cannot DoS the compositor by passing a huge `size` value to the `Keymap` request.

### DIFF
--- a/src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
+++ b/src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
@@ -1,15 +1,19 @@
+use std::fs::File;
+use std::io::Read;
+use std::os::unix::io::{FromRawFd, OwnedFd};
 use std::{
     fmt::Debug,
     os::unix::io::AsRawFd,
     sync::{Arc, Mutex},
 };
 
+use tracing::debug;
 use wayland_protocols_misc::zwp_virtual_keyboard_v1::server::zwp_virtual_keyboard_v1::{
     self, ZwpVirtualKeyboardV1,
 };
 use wayland_server::{
     backend::{ClientId, ObjectId},
-    protocol::wl_keyboard::KeyState,
+    protocol::wl_keyboard::{KeyState, KeymapFormat},
     Client, DataInit, Dispatch, DisplayHandle,
 };
 use xkbcommon::xkb;
@@ -21,6 +25,9 @@ use crate::{
 };
 
 use super::VirtualKeyboardManagerState;
+
+/// Maximum keymap size. Up to 1MiB.
+const MAX_KEYMAP_SIZE: usize = 0x100000;
 
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Hash)]
 struct SerializedMods {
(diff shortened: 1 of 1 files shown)


src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -192,3 +172,62 @@ where
...
     }
 }
+
+/// Handle the zwp_virtual_keyboard_v1::keymap request.
+fn update_keymap<D>(data: &VirtualKeyboardUserData<D>, format: u32, fd: OwnedFd, size: usize)
+where
+    D: SeatHandler + 'static,
+{
+    // Only libxkbcommon compatible keymaps are supported.
+    if format != KeymapFormat::XkbV1 as u32 {
```
rano-oss, line 182: Nice!

chrisduerr: I'm actually curious why you wrote in your original PR that the format is not an enum. Looking at the protocol it seems like it is an enum and that's just how they're represented. Were you just not aware of that back then?

rano-oss: Ah well, yes and no, it is an enum, but it is not explicitly defined as an enum required to use it as such. That said, it did not occur to me that I could write it like this, and this is much more self-explanatory.

Drakulix (approved):
Looks sensible to me.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -192,3 +172,62 @@ where
...
+        return;
+    }
+
+    // Read entire keymap.
+    let mut keymap_buffer = vec![0; size];
+    let mut file = unsafe { File::from_raw_fd(fd.as_raw_fd()) };
+    if let Err(err) = file.read_exact(&mut keymap_buffer) {
+        debug!("Could not read keymap: {err}");
+        return;
+    }
```
vberger, line 199: I know I'm coming after the merge, but I suppose we could have avoided that unsafe by just using `nix::unistd::read()` instead?

More generally, can creating this `File` do stupid stuff if a malicious client sends a stupid FD?

And writing that I realize that if the client sends an fd to something that is not a file/memory mapped region, but instead a pipe for example, it could block the compositor, no?

chrisduerr: > More generally, can creating this File do stupid stuff if a malicious client sends a stupid FD?

Yeah that's fair. The docs specifically state `The fd passed in must be a valid and open file descriptor`, so passing an invalid descriptor is just UB?

> And writing that I realize that if the client sends an fd to something that is not a file/memory mapped region, but instead a pipe for example, it could block the compositor, no?

I always forget under which circumstances the `WouldBlock` error is emitted, but if it isn't for pipes then that could be an issue, yeah.

vberger: > so passing an invalid descriptor is just UB?

We cannot receive an invalid FD from the wayland socket, imo the main worry is what if the passed FD is to something else than a file or memory buffer.

About the use of `read()`, mostly my question was whether using `File::from_raw_fd()` implies more assumptions about the FD than just using `read()`, which would not be unsafe and just return an error if the FD cannot be read.

chrisduerr: > About the use of read(), mostly my question was whether using File::from_raw_fd() implies more assumptions about the FD than just using read(), which would not be unsafe and just return an error if the FD cannot be read.

Unfortunately I'm not aware of Rust internals, but even with `read()` you could get blocking.

Issue #946, opened 2023-03-18
Explicitely set the unsafe parts in unsafe functions.


TornaxO7:
This Issue is a follow-state of wrapping the explicitely wrap the unsafe parts with `unsafe` blocks in order to fix https://github.com/Smithay/smithay/issues/756.

I'll create smaller PRs which document and fix the parts separately:
- [ ] `src/wayland/shm/pool.rs` https://github.com/Smithay/smithay/pull/939
- [ ] `src/utils/fd.rs` https://github.com/Smithay/smithay/pull/940
- [ ] `src/utils/user_data.rs`
- [ ] `src/backend/egl/mod.rs`
- [ ] `src/backend/egl/display.rs`
- [ ] `src/backend/egl/ffi.rs`
- [ ] `src/backend/egl/device.rs`
- [ ] `src/backend/egl/context.rs`
- [ ] `src/backend/renderer/gles2/mod.rs`

## Help
If you want to help, simply create a draft first to avoid multiple people fix the same file!

Clone the repo:
```bash
# assuming with ssh
git clone git@github.com:Smithay/smithay.git

# create your branch
git switch -c fix_756_<filename>.rs
```
now go to one of the files above and add
```rust
#![forbid(unsafe_op_in_unsafe_fn)]
```
to the given file to see the issues only in the given files and write a message here, so I can edit this post.

PR #948, opened 2023-03-18
adding unsafe block

TornaxO7:
Needed for https://github.com/Smithay/smithay/issues/946


src/backend/egl/context.rs
```diff
@@ -56,8 +56,8 @@ impl EGLContext {
         assert!(!config_id.is_null(), "EGL configuration id pointer is null");
         assert!(!context.is_null(), "EGLContext pointer is null");
 
-        let display = EGLDisplay::from_raw(display, config_id)?;
-        let pixel_format = display.get_pixel_format(config_id)?;
+        let display = unsafe { EGLDisplay::from_raw(display, config_id) }?;
+        let pixel_format = unsafe { display.get_pixel_format(config_id) }?;
```
Drakulix, line 60: Is `get_pixel_format` declared unsafe? If it is we should take a look if we can lift that requirement, that function just returns an enum after all.

TornaxO7:
> I feel like the EGL module in particular was written quite some time ago and although it was updated quite a bit since then, it still throws the `unsafe` keyword around whenever it sees a pointer. (So everywhere.)
> 
> There is probably opportunity here to make some functions safe instead. But looking at #954 it is questionable whenever that is worth it.
> 
> @i509VCB any opinion?

Sorry for the late reply!
So should I close this issue?

i509VCB:
> I feel like the EGL module in particular was written quite some time ago and although it was updated quite a bit since then, it still throws the `unsafe` keyword around whenever it sees a pointer. (So everywhere.)
> 
> There is probably opportunity here to make some functions safe instead. But looking at #954 it is questionable whenever that is worth it.
> 
> @i509VCB any opinion?

Yeah I'd agree that we may want to defer this until we concretely know what to do with moving to glutin or not.

> should I close this issue?

Given drakulix is quite deep into 10-bit color stuff and that might involve going into EGL, it is quite possible you'll get huge conflicts in git. If we don't go to glutin we can discuss restarting this pull request for EGL.

TornaxO7:
Okeydokey, I'll close this then!

PR #955, opened 2023-03-18
Fix CropRenderElement nesting

chrisduerr:
Previously the `CropRenderElement` would not take the existing element's cropping into account. This can create problems when nesting cropped render elements which can easily happen with viewporter for example.

The approach taken by this patch is to calculate the new cropped source rectangle and then only return the intersection to the original source rectangle. This allows `CropRenderElement` to crop stuff off without ever increasing the source rectangle and reverting existing cropping.

---

Just wanted to point out that this fix almost seems to simple to work just like this. So if someone can think of other potential edgecases that might break this, please let me know. This was only very briefly tested with viewporter (though it did work).

### DIFF
--- a/src/backend/renderer/element/utils/elements.rs
+++ b/src/backend/renderer/element/utils/elements.rs
@@ -154,13 +154,14 @@ impl<E: Element> CropRenderElement<E> {
             // Ok, for the src we need to know how much we cropped from the element geometry
             // and then bring that rectangle into buffer space. For this we have to first
             // apply the element transform and then scale it to buffer space.
-            let src = element_relative_intersection.to_f64().to_logical(1.0).to_buffer(
+            let mut src = element_relative_intersection.to_f64().to_logical(1.0).to_buffer(
                 physical_to_buffer_scale,
                 transform,
                 &element_geometry.size.to_f64().to_logical(1.0),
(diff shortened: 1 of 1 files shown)


cmeissl:
I don't think this is the appropriate fix (at least not without some additional logic), this will crop the size without considering the element src loc. I think it should be enough to offset the `src` by the `element_src` loc. Something like `src.loc += element_src.loc;` before the intersection.

chrisduerr:
> this will crop the size without considering the element src loc

What do you mean by this? You're intersecting two rectangles, so both locs are taken into account, right?

Or do you mean that they should be relative to each other? I personally think that would make things unnecessarily difficult to reason about. Especially since the `CropRenderElement` is not relative but in physical coordinates.

> Something like src.loc += element_src.loc; before the intersection.

I don't think this would work, as you would have to know the previous crop elements sizes to calculate which size you need to pass in to the higher-level cropped element?

chrisduerr:
> After looking into this again it should be enough to add only add src.loc += element_src.loc;.

This makes sense, considering that I noticed the issue only because the offset from the top-left was incorrect. The size wasn't actually wrong.

I've tested it and verified that just setting the loc does work for me.

PR #956, opened 2023-03-19
Handle creation shader function

TornaxO7:
This fix is needed for https://github.com/Smithay/smithay/issues/946


Drakulix (approved):
Nitpick about naming, but good idea! :)

PR #957, opened 2023-03-20
adding error handling for `gl.CreateProgram` in `gles2/mod.rs`.

TornaxO7:
According to the [doc](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glCreateProgram.xhtml):
> This function returns 0 if an error occurs creating the program object.

### DIFF
--- a/src/backend/renderer/gles2/mod.rs
+++ b/src/backend/renderer/gles2/mod.rs
@@ -494,6 +494,9 @@ pub enum Gles2Error {
         /// Uniform type that was declared when compiling
         declared: UniformType,
     },
+    /// Indicates that an error occured during the `gl.CreateProgram` function.
+    #[error("An error occured while trying to create a program object")]
+    CreateProgram,
 }
 
 impl From<Gles2Error> for SwapBuffersError {
@@ -514,6 +517,7 @@ impl From<Gles2Error> for SwapBuffersError {
             | x @ Gles2Error::BufferAccessError(_)
             | x @ Gles2Error::MappingError
             | x @ Gles2Error::UnexpectedSize
+            | x @ Gles2Error::CreateProgram
             | x @ Gles2Error::BlitError
             | x @ Gles2Error::CreateShaderObject
             | x @ Gles2Error::UniformTypeMismatch { .. }
@@ -537,6 +541,7 @@ impl From<Gles2Error> for SwapBuffersError {
             | x @ Gles2Error::UnexpectedSize
             | x @ Gles2Error::BlitError
             | x @ Gles2Error::CreateShaderObject
+            | x @ Gles2Error::CreateProgram
             | x @ Gles2Error::UniformTypeMismatch { .. }
             | x @ Gles2Error::UnknownUniform(_)
             | x @ Gles2Error::BindBufferEGLError(_) => SwapBuffersError::TemporaryFailure(Box::new(x)),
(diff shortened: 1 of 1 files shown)


src/backend/renderer/gles2/mod.rs
```diff
@@ -537,6 +541,7 @@ impl From<Gles2Error> for SwapBuffersError {
             | x @ Gles2Error::UnexpectedSize
             | x @ Gles2Error::BlitError
             | x @ Gles2Error::CreateShaderObject
+            | x @ Gles2Error::CreateProgram
```
TornaxO7, line 544: Please note, that I don't know if this error should be put here or to `SwapBuffersError::ContextLost`.

src/backend/renderer/gles2/mod.rs
```diff
@@ -514,6 +517,7 @@ impl From<Gles2Error> for SwapBuffersError {
             | x @ Gles2Error::BufferAccessError(_)
             | x @ Gles2Error::MappingError
             | x @ Gles2Error::UnexpectedSize
+            | x @ Gles2Error::CreateProgram
```
TornaxO7, line 520: Please note, that I don't know if this error should be put here or to `SwapBuffersError::ContextLost`.

Drakulix: This one will fail initializing, which is not temporary. So context lost please. :)

TornaxO7: okeydokey!

TornaxO7: done!

src/backend/renderer/gles2/mod.rs
TornaxO7: moved to `ContextLost`

src/backend/renderer/gles2/mod.rs
```diff
@@ -622,6 +627,10 @@ unsafe fn link_program(
     let vert = compile_shader(gl, ffi::VERTEX_SHADER, vert_src)?;
     let frag = compile_shader(gl, ffi::FRAGMENT_SHADER, frag_src)?;
     let program = gl.CreateProgram();
```
i509VCB, line 629: Unrelated to this pull request, but it creating the program fails, we end up leaking the vert and frag shaders. This is something we need to try to deal with in the gles renderer.

i509VCB (approved):

src/backend/renderer/gles2/mod.rs
TornaxO7: could you explain me, how to free the shaders then I can add the free-management for it.

i509VCB: so when you use the try operator (`?`), it will immediately return `Err` if there is a failure.

What needs to be done here is that if creating `frag` fails you need to delete the `vert` shader and then return `Err`.

Same with `program`.

TornaxO7: May I ask what I'd need to do in order to "delete" a variable?

i509VCB: I think you need to use `glDeleteShader` for the shader objects.

TornaxO7: okeydokey

TornaxO7: Should be done now.

i509VCB:
Looks fine, conflict needs to be resolved.

TornaxO7:
ok damn, a lot of code has changed now and I don't really know which parts of the code needs to have the check here.

PR #959, opened 2023-03-21
10-bit prerequisites 

Drakulix:
A couple of commits from #935, that don't yet enable 10-bit support (as that PR requires some more rework and splitting the GL ES renderer), but that can already be reviewed and merged without causing any harm.


src/wayland/shm/mod.rs
```diff
@@ -154,6 +156,13 @@ impl ShmState {
     pub fn global(&self) -> GlobalId {
         self.shm.clone()
     }
+
+    /// Updates the list of formats advertised by the global.
+    ///
+    /// This will only affect new binds to the wl_shm global.
+    pub fn add_formats(&mut self, formats: impl IntoIterator<Item = wl_shm::Format>) {
```
i509VCB, line 163: I guess this begs the question, do we need a way to remove formats from future binds?

src/backend/drm/gbm.rs
```diff
@@ -0,0 +1,336 @@
...
+/// Possible errors for attaching a [`framebuffer::Handle`] with [`framebuffer_from_bo`]
+#[derive(Debug, Error)]
+#[error("failed to add a framebuffer")]
+pub struct DrmError {
+    /// Error message associated to the access error
+    pub errmsg: &'static str,
+    /// Device on which the error was generated
+    pub dev: Option<PathBuf>,
+    /// Underlying device error
+    pub source: drm::SystemError,
```
i509VCB, line 176: I assume thiserror has an attribute to forward Error::source to this field?

src/backend/drm/surface/gbm.rs
```diff
@@ -180,17 +177,26 @@ where
...
         };
 
-        let fb = match attach_framebuffer(&drm, &buffer) {
+        let fb = match framebuffer_from_bo(drm.device_fd(), &buffer, true) {
             Ok(fb) => fb,
-            Err(err) => return Err((swapchain.allocator, err)),
+            Err(err) => {
+                return Err((
+                    swapchain.allocator,
+                    Error::DrmError(DrmError::Access {
```
i509VCB, line 185: This appears at least twice, maybe a From implementation could make things neater.

src/backend/drm/surface/gbm.rs
```diff
@@ -56,6 +44,7 @@ where
     pub fn new(
         drm: DrmSurface,
         mut allocator: A,
+        color_formats: &[Fourcc],
```
cmeissl, line 47: Maybe this could benefit from a comment in the functions docs describing that the formats are
tested in the order they appear.

src/backend/drm/surface/gbm.rs
```diff
@@ -102,18 +91,31 @@ where
             Err(err) => return Err((allocator, err.into())),
         };
 
-        if !plane_formats.iter().any(|fmt| fmt.code == code) {
+        let opaque_code = get_opaque(code).unwrap_or(code);
+        if !plane_formats
+            .iter()
+            .any(|fmt| fmt.code == code || fmt.code == opaque_code)
+        {
```
cmeissl, line 98: Do we always want to prefer a opaque format even if a later format could directly match?

src/backend/drm/compositor/mod.rs
```diff
@@ -1155,6 +1150,7 @@ where
         planes: Option<Planes>,
         mut allocator: A,
         framebuffer_exporter: F,
+        color_formats: &[DrmFourcc],
```
cmeissl, line 1153: Same here, probably worth to add a description in the docs

anvil/src/udev.rs
```diff
@@ -95,6 +95,15 @@ use smithay::{
...
 use tracing::{debug, error, info, trace, warn};
 
+// we cannot simply pick the first supported format of the intersection of *all* formats, because:
+// - we do not want something like Abgr4444, which looses color information, if something better is available.
+// - some formats might perform terribly.
+// - we might need some work-arounds, if one supports modifiers, but the other does not.
+// - we can't handle formats with small alpha channels yet in the renderer like `Abgr2101010`.
+//
+// So lets just pick `ARGB8888`/`ABGR8888` for now, they are widely supported.
+const SUPPORTED_FORMATS: &[Fourcc] = &[Fourcc::Abgr8888, Fourcc::Argb8888];
```
cmeissl, line 105: Does this intentionally not include 10bit variants? Winit has been changed to prefer 10bit.

cmeissl (commented):
Looks, good. Just some small nits and questions.

src/wayland/shm/mod.rs
Drakulix: Given that we would still need to support the old formats for already bound clients, I just see this function as a way to lazily initialize the list (until a renderer is created) not to further alter the list later.

src/backend/drm/surface/gbm.rs
Drakulix: Well given users might want to try 10-bit formats first and later 8-bit formats, I would assume so. If the opaque_fallback is allowed, it shouldn't make a difference functionally. That said we need to add a check for underlays before allowing this, that is still missing.

anvil/src/udev.rs
Drakulix: Oh, then I need to drop the winit changes. 10-bit isn't ready yet in this PR.

src/backend/drm/surface/gbm.rs
Drakulix: Done

src/backend/drm/gbm.rs
Drakulix: Done

src/backend/drm/surface/gbm.rs
```diff
@@ -430,6 +446,33 @@ impl Drop for FbHandle {
     }
 }
 
+struct BufferWrapper<'a, B>(&'a B);
```
i509VCB: Do we need two of these? Having a duplicate type is begging for behavior to desync.

Drakulix: There aren't two, later commits do extract and unify the code.

src/backend/drm/compositor/gbm.rs
```diff
@@ -237,9 +256,37 @@ impl<'a, T: 'static> PlanarBuffer for BufferObjectInternal<'a, T> {
     }
 }
 
+struct BufferWrapper<'a, B>(&'a B);
```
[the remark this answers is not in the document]
Drakulix: done

src/backend/drm/surface/gbm.rs
cmeissl: Yeah, seems correct. imo we can can just disable underlays completely for now if we select a opaque format for the primary plane during initialization in `DrmCompositor::new`. I want to change the plane selection algorithm anyway.

Drakulix: Should I do this in this PR or do you write the updated plane selection code soon anyway?

cmeissl: Given that the issue already existed prior to this PR I don't think it has to be solved here. I have no ETA for the plane selection, so I would suggest to create an Issue out of this review comment and I will take care of it. Either by disabling the underlay planes for opaque types or in a new plane assignment algorithm.

cmeissl (approved):

PR #960, opened 2023-03-23
fallback to gbm bo_map if bo_write is unavailable

cmeissl:


test_gbm_bo_write.c
i509VCB, line 1: We already have two of these files, I'd think we would want to move these to a folder

cmeissl: yeah, I had the same feeling. will do this if @Drakulix has no objections against the check in general

Cargo.toml
```diff
@@ -81,6 +81,7 @@ backend_drm = ["drm", "drm-ffi"]
 backend_gbm = ["gbm", "cc", "pkg-config"]
 backend_gbm_has_fd_for_plane = []
 backend_gbm_has_create_with_modifiers2 = []
+backend_gbm_has_bo_write = []
```
bwidawsk, line 84: One other thought I've had for these autodetected is an explicit disable. That makes testing breakages easier. TBH though, I don't know of a good way to do that.

bwidawsk (approved):

bwidawsk:
Here is the compositor implementation which I'm currently using:
https://gitlab.freedesktop.org/bwidawsk/sudbury/-/blob/cros/src/drm_kms/minigbm.rs

I'd definitely prefer to have this in Smithay itself, but this PR is no longer required for functioning.

Thanks @cmeissl for whipping it up so fast.

PR #962, opened 2023-03-24
gles2: Always rebinding egl_images

Drakulix:
External images on nvidia gpus seem to suffer from synchronization issues without this change. OES_EGL_image_external also seems to suggest, that this is a good idea.


cmeissl (approved):
I re-read the spec and also think this should be fine.
wlroots seems to skip it for external (`// External changes are immediately made visible by the GL implementation`), but weston seems to always call `glEGLImageTargetTexture2DOES`.

PR #963, opened 2023-03-24
Fix inaccurate renderer nameing

### LINKED ISSUE #895: Rendering functions should document expected `&[RenderElement]` z-order
Current documentation for `DamageTrackedRenderer` states the following:

> Render this output

During rendering I would usually expect back to front element order, however this function expects front to back order.
While users might still run into and notice this issue by doing it wrong the first time, there should be documentation reassuring them what the correct element order is meant to be.

CC @cmeissl

cmeissl:
The naming of `DamageTrackedRenderer` implies that it is a renderer, which it is not. Fix this inaccuracy by renaming:

- `DamageTrackedRenderer` -> `OutputDamageTracker`
- `DamageTracker` -> `DamageBag`
- `DamageTrackerSnapshot` -> `DamageSnapshot`

This also documents the expected z-order of elements.

fixes #895

### DIFF
--- a/src/backend/renderer/damage.rs
+++ b/src/backend/renderer/damage.rs
@@ -326,7 +326,9 @@ impl OutputDamageTracker {
         &self.mode
     }
 
-    /// Render this output
+    /// Render this output with the provided [`Renderer`]
+    ///
+    /// - `elements` for this output in front-to-back order
     #[instrument(level = "trace", parent = &self.span, skip(renderer, elements))]
     pub fn render_output<E, R>(
         &mut self,
@@ -458,6 +460,8 @@ impl OutputDamageTracker {
     }
 
     /// Damage this output and return the damage without actually rendering the difference
+    ///
+    /// - `elements` for this output in front-to-back order
     #[instrument(level = "trace", parent = &self.span, skip(elements))]
     pub fn damage_output<E>(
         &mut self,
--- a/src/backend/renderer/damage.rs
+++ b/src/backend/renderer/damage.rs
@@ -1,16 +1,24 @@
-//! Helper for effective damage tracked rendering
+//! Helper for effective output damage tracking
 //!
 //! # Why use this implementation
 //!
-//! The [`DamageTrackedRenderer`] in combination with the [`RenderElement`] trait
+//! The [`OutputDamageTracker`] in combination with the [`RenderElement`] trait
 //! can help you to reduce resource consumption by tracking what elements have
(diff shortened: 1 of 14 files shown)


src/backend/renderer/damage.rs
```diff
@@ -254,17 +262,17 @@ impl TryInto<(Size<i32, Physical>, Scale<f64>, Transform)> for DamageTrackedRend
...
+    mode: OutputDamageTrackerMode,
     last_state: RendererState,
     span: tracing::Span,
 }
 
-/// Errors thrown by [`DamageTrackedRenderer::render_output`]
+/// Errors thrown by [`OutputDamageTracker::render_output`]
 #[derive(thiserror::Error)]
-pub enum DamageTrackedRendererError<R: Renderer> {
+pub enum OutputDamageTrackerError<R: Renderer> {
```
Drakulix, line 275: Actually, while we are at it (sorry for bringing this up only just now)... I would vastly prefer to just name this `Error` and solve conflicts by renaming the import in other modules.

This is not C, we don't have to prefix everything with the module-name or similar.

Drakulix (commented):
Looks good! Really liking the new names much better. Just a small nitpick...

src/backend/renderer/damage.rs
i509VCB: I'd agree with this, but then I ask that question about the rest of smithay. Should Gles2Renderer be gles2::Renderer or do we want the explicit name. Some of the reexports add a prefix (gbm::Device exported as GbmDevice)

cmeissl: I am fine with changing it, but we should stay consistent. So if we change it here we should also it everywhere else. Not sure if gradually, starting with this one, or changing all at once is better. But I lean towards making that change gradually. This could potentially affect a big surface of the API. No one likes to review such big PRs and it might take a bit so the potential for rebase conflicts is pretty high.

Drakulix: Right and at least for error types I think we are semi-consistent in the sense, that we prefer shorter names.
Also I don't think for main structs like the `Gles2Renderer` or the `OutputDamageTracker` renaming really makes sense. For the Gles2Renderer it is particularly problematic given the `Renderer`-trait just one module above.

cmeissl: renamed to `Error`

Drakulix (approved):

PR #964, opened 2023-03-25
solid color rendering

cmeissl:


Drakulix (approved):
Looks good! I assume support for the SinglePixelBuffer protocol will then come in [the separate PR](https://github.com/Smithay/smithay/pull/719).

PR #966, opened 2023-03-26
wayland: support xdg_wmbase v4

### LINKED ISSUE #544: Support for xdg shell version 4
Version 4 of `xdg-shell` has been released and adds [xdg_toplevel::configure_bounds](https://wayland.app/protocols/xdg-shell#xdg_toplevel:event:configure_bounds)

> ### Recommended window geometry bounds
>
> The configure_bounds event may be sent prior to a xdg_toplevel.configure event to communicate the bounds a window geometry size is recommended to constrain to.
>
>The passed width and height are in surface coordinate space. If width and height are 0, it means bounds is unknown and equivalent to as if no configure_bounds event was ever sent for this surface.
>
>The bounds can for example correspond to the size of a monitor excluding any panels or other shell components, so that a surface isn't created in a way that it cannot fit.
>
>The bounds may change at any point, and in such a case, a new xdg_toplevel.configure_bounds will be sent, followed by xdg_toplevel.configure and xdg_surface.configure.

cmeissl:
this adds support for configure_bounds which was introduced in v4

fixes #544

### DIFF
--- a/src/wayland/shell/xdg/handlers/surface/toplevel.rs
+++ b/src/wayland/shell/xdg/handlers/surface/toplevel.rs
@@ -191,7 +191,11 @@ where
     })
 }
 
-pub fn send_toplevel_configure(resource: &xdg_toplevel::XdgToplevel, configure: ToplevelConfigure) {
+pub fn send_toplevel_configure(
+    resource: &xdg_toplevel::XdgToplevel,
+    configure: ToplevelConfigure,
+    send_bounds: bool,
+) {
     let data = resource.data::<XdgShellSurfaceUserData>().unwrap();
     let (width, height) = configure.state.size.unwrap_or_default().into();
     // convert the Vec<State> (which is really a Vec<u32>) into Vec<u8>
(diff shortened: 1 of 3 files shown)


src/wayland/shell/xdg/handlers/surface/toplevel.rs
```diff
@@ -206,6 +210,12 @@ pub fn send_toplevel_configure(resource: &xdg_toplevel::XdgToplevel, configure:
     };
     let serial = configure.serial;
 
+    // send bounds if requested
+    if send_bounds && resource.version() >= xdg_toplevel::EVT_CONFIGURE_BOUNDS_SINCE {
```
Drakulix, line 214: Do we need an extra `send_bounds`, if `None` (or width=0 and height=0) mean the same thing?

Drakulix (commented):
Looks good!

src/wayland/shell/xdg/handlers/surface/toplevel.rs
cmeissl: I added this to prevent excessive `configure_bounds` events. The spec says it must be sent initially and whenever it changes. Without that we would always send the bounds, I am not sure how a client will react on that, maybe it is not an issue at all.

Drakulix: But the state is double buffered, no? Couldn't we just send the event whenever it changes?

Drakulix: Oh now I see, the `pub fn send_toplevel_configure` was confusing me, I thought this was part of the public api. In this case `send_configure` does internally exactly what I would expect. Consider this solved.

Drakulix (approved):

PR #967, opened 2023-03-26
wayland: support xdg_wmbase v5

cmeissl:
this adds support for wm capabilities  which were introduced in v5

### DIFF
--- a/src/wayland/shell/xdg/handlers/surface/toplevel.rs
+++ b/src/wayland/shell/xdg/handlers/surface/toplevel.rs
@@ -205,6 +205,7 @@ pub fn send_toplevel_configure(
     resource: &xdg_toplevel::XdgToplevel,
     configure: ToplevelConfigure,
     send_bounds: bool,
+    send_capabilities: bool,
 ) {
     let data = resource.data::<XdgShellSurfaceUserData>().unwrap();
     let (width, height) = configure.state.size.unwrap_or_default().into();
@@ -224,6 +225,24 @@ pub fn send_toplevel_configure(
     if send_bounds && resource.version() >= xdg_toplevel::EVT_CONFIGURE_BOUNDS_SINCE {
         let bounds = configure.state.bounds.unwrap_or_default();
         resource.configure_bounds(bounds.w, bounds.h);
+    }
+
+    // send the capabilities if requested
+    if send_capabilities && resource.version() >= xdg_toplevel::EVT_WM_CAPABILITIES_SINCE {
+        let mut capabilities = configure
+            .state
+            .capabilities
+            .capabilities()
+            .copied()
+            .collect::<Vec<_>>();
+        let capabilities = {
+            let ptr = capabilities.as_mut_ptr();
+            let len = capabilities.len();
+            let cap = capabilities.capacity();
+            ::std::mem::forget(capabilities);
+            unsafe { Vec::from_raw_parts(ptr as *mut u8, len * 4, cap * 4) }
+        };
+        resource.wm_capabilities(capabilities);
     }
 
     // Send the toplevel configure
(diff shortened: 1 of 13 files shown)


src/wayland/shell/xdg/handlers/surface/toplevel.rs
```diff
@@ -97,26 +97,36 @@ where
                 });
             }
             xdg_toplevel::Request::SetMaximized => {
-                let handle = make_toplevel_handle(toplevel);
-                XdgShellHandler::maximize_request(state, handle);
+                if toplevel_has_capability(toplevel, xdg_toplevel::WmCapabilities::Maximize) {
```
Drakulix, line 100: So we pre-filter the requests being passed to downstream by the capabilities it sets?
Does something in the protocol demand, that we don't handle these events?

Or is this to keep <v5-clients causing *unexpected* events? Because I would say let downstream handle that instead of making it impossible.

cmeissl: The protocol defines that the requests should be ignored when the capability is absent. Though this could result in a race if the capabilities change during runtime. Like on creation fullscreen is supported, but later removed.

src/wayland/shell/xdg/handlers/surface/toplevel.rs
```diff
@@ -205,6 +205,7 @@ where
 pub fn send_toplevel_configure(
     resource: &xdg_toplevel::XdgToplevel,
     configure: ToplevelConfigure,
+    send_capabilities: bool,
```
Drakulix, line 208: Why do we need this bool, can't we just send capability whenever they have changed?

cmeissl: This is what the `bool` represents, same as with `send_bounds`. `send_toplevel_configure` is an internal function called from within `XdgToplevel::send_configure`. We could move this calls to `XdgToplevel`, but so far I kept everything directly defined in the `xdg_toplevel` protocol in a single function and only extensions like `xdg_decoration` are called outside of `send_toplevel_configure`.

cmeissl:
So, I think this should be now complete from the protocol implementation itself. Anvil currently makes no use of it.
There is one last thing that may need some changes. `XdgToplevel::configure` currently tests for changes and ignores the call if nothing changed. IIRC @kchibisov mentioned that this could be problematic for XWayland clients in case the request is ignored and no configure is sent. Though I would prefer to track this in a separate issue as this could already be an issue.

kchibisov:
>IIRC @kchibisov mentioned that this could be problematic for XWayland clients in case the request is ignored and no configure is sent

This defined in the specification. You must configure back after `set_fullscreen` and `set_maximized`. It's not just `XWayland`, because you violate the spec of not doing so.

cmeissl:
> > IIRC @kchibisov mentioned that this could be problematic for XWayland clients in case the request is ignored and no configure is sent
> 
> This defined in the specification. You must configure back after `set_fullscreen` and `set_maximized`. It's not just `XWayland`, because you violate the spec of not doing so.

Would `[xdg_surface::configure](https://wayland.app/protocols/xdg-shell#xdg_surface:event:configure)` be sufficient or is `[xdg_toplevel::configure](https://wayland.app/protocols/xdg-shell#xdg_toplevel:event:configure)` required to be sent?

kchibisov:
You need a top-level one, because it demands a state. At least that's what I understand from the `xdg-shell` protocol.

cmeissl:
> You need a top-level one, because it demands a state. At least that's what I understand from the `xdg-shell` protocol.

Indeed, re-reading that paragraph sounds like it needs the state. Hm, so preventing an unchanged state to send a configure is an issue. I see two options, either we add a force option or we just leave the decision to downstream. I find the force way ugly and sending another configure shouldn't do any harm unless the client throws away everything on each configure. Though that only needs the toplevel and surface configure, bounds/capabilities/decorations can still be de-duplicated I believe.

cmeissl:
Ok, so this now contains two commits. The first one changing `send_configure` to always generate a configure event and adding a `send_pending_configure` function to get the previous behavior. The second one actually implementing `WmCapabilities`. I tried to verify this with `foot` which already supports `WmCapababilites` and `epiphany` which doesn't.
`foot` successfully hides capabilities not announced while with epihany the requests are just ignored and nothing seems to break.

@kchibisov Can you please take a quick look if this is correct now from the protocol point of view?

Drakulix (approved):
LGTM

PR #974, opened 2023-04-02
drm-extras: Don't put null bytes in manufacturer string

PolyMeilex:
WlOutput can't use manufacturer name that contains \u{0} char

### DIFF
--- a/smithay-drm-extras/src/edid.rs
+++ b/smithay-drm-extras/src/edid.rs
@@ -66,7 +66,12 @@ fn get_manufacturer_name(edid: &edid_rs::EDID) -> String {
 
     hwdata::pnp_id_to_name(&code)
         .map(|name| name.to_string())
-        .unwrap_or_else(|| code.into_iter().collect())
+        .unwrap_or_else(|| {
+            code.into_iter()
+                .map(|v| (v as u8).to_string())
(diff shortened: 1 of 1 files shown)


smithay-drm-extras/src/edid.rs
```diff
@@ -66,7 +66,12 @@ fn get_manufacturer_name(edid: &edid_rs::EDID) -> String {
 
     hwdata::pnp_id_to_name(&code)
         .map(|name| name.to_string())
-        .unwrap_or_else(|| code.into_iter().collect())
+        .unwrap_or_else(|| {
+            code.into_iter()
+                .map(|v| (v as u8).to_string())
```
i509VCB, line 71: This seems to be creating a String per character which seems wasteful.

Also why are we getting the character, converting it to the number of the codepoint and then writing out the number?

PolyMeilex: This is meant to output human readable string, and some of manufacturer codes contain characters that are not human readable like 0x0 (or any non readable char in range of 0-255), so this as a fallback to quarrying the name of the manufacturer from hwdata basically outputs them as numbers instead eg. "12_0_9" which is an upgrade over just outputting "Unknown", or string that contains unreadable characters or invalid cstring because of null bytes.

PR #975, opened 2023-04-02
Adding link to calloop crate

TornaxO7:
The [docs] about [`calloop`] arent' correctly linked a the moment.

[docs]: https://smithay.github.io/smithay/smithay/index.html#the-event-loop-and-state-handling
[`calloop`]: https://crates.io/crates/calloop

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,6 +44,8 @@
 //! ### Logging
 //!
 //! TODO: docs about tracing
+//!
+//! [`calloop`]: https://crates.io/crates/calloop
 
 pub mod backend;
 #[cfg(feature = "desktop")]


elinorbgr:
Actually, I think this is an artifact of our local run, in the docs.rs version of the docs ( https://docs.rs/smithay/latest/smithay/ ), the link is properly linked to the documentation of calloop.

So I think it's instead a matter of figuring out why our rustdoc CI setup does not do the link properly, even though we also build calloop's docs.

TornaxO7:
> Actually, I think this is an artifact of our local run, in the docs.rs version of the docs ( https://docs.rs/smithay/latest/smithay/ ), the link is properly linked to the documentation of calloop.

Yes you're right.

> So I think it's instead a matter of figuring out why our rustdoc CI setup does not do the link properly, even though we also build calloop's docs.

Sounds reasonable.

PR #979, opened 2023-04-03
drm: fix element switching direct scan-out <> composition

cmeissl:
When en element switches between direct scan-out and composition we will dynamically add or remove fake elements to communicate the opaque regions of overlay planes and realize hole punching in case of underlays. Re-using the id of the elements will miss with the damage state of the internal output damage tracker. Instead we now use an artifical Id per plane for the fake elements.

cc @chrisduerr

### DIFF
--- a/src/backend/drm/compositor/elements.rs
+++ b/src/backend/drm/compositor/elements.rs
@@ -23,13 +23,13 @@ pub struct HolepunchRenderElement {
 }
 
 impl HolepunchRenderElement {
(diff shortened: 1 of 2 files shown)


chrisduerr:
This does **not** fix the issue for me.

chrisduerr:
The latest revision **does** fix the issue for me.

Drakulix (approved):
LGTM

PR #982, opened 2023-04-05
feature: Color management

Drakulix:
TODO:

- [x] color module for abstracting over different implementations
  - [x] color module based on LittleCMS
    - [ ] add lcms-plugin to optimize pipelines and create matrix-based color mappings where possible instead of always creating 3D-LUT based ones
    - [x] Upstream
      - [x] ~https://github.com/kornelski/rust-lcms2/pull/9~ Not my PR, but I guess https://github.com/kornelski/rust-lcms2/commit/0f7ec7c4fc614aa0ac0610b720eff9b221ed0e69 fixes this as well
      - [x] https://github.com/kornelski/rust-lcms2/pull/10
      - [x] https://github.com/kornelski/rust-lcms2/pull/11
  - [x] no-op color module, behaving exactly as before
     - [x] Pure identity variant
     - [ ] Add variant supporting sRGB->Linear->sRGB for at least gamma correct blending
  - [ ] *possibly* experiment with [rcms](https://github.com/cpsdqs/rcms) as an alternative to LittleCMS. (Likely would need some upstream contributions though.)
- [ ] Renderer:
  - [x] Adjust renderer-trait to take output profile on frame creation
  - [x] Adjust frame-trait so that `draw`-calls take an input profile
  - [x] gles2: Compute `ColorTransform` (heavily cached) for every draw call and upload LUTs to shader
  - [x] multigpu: Update for new api
  - [x] glow: Update for new api
  - [x] compositor
    - [x] Simple: Disable composition if transformation isn't identity
    - [ ] Hard: Offload pre/post-curves to `GAMMA_LUT`/`DEGAMMA_LUT` properties and mappings to `CTM` property, if all necessary properties are available and data is compatible
- [ ] Build test case comparing results rendering with color-correction and using lcms directly to convert buffers in software to verify our implementation does the right thing.
- [x] smallvil
  - [x] Support sRGB profile
- [x] anvil:
  - [x] Support sRGB profile
  - [x] Allow output profile to be generated from EDID
  - [x] Allow custom output profiles to be set

Future PRs:

- Allow easy change of white-point for implementing features similar to `flux`/`Night Mode`/`redshift`
- HDR:
  - Make it possible to express HDR-metadata in output-profiles and set `HDR_OUTPUT_METADATA` property accordingly
  - Make it possible for buffers to be uploaded with color-profiles other than sRGB
  - Direct scanout HDR-buffers by setting HDR-metadata
  - Tone-map non matching HDR-buffers


jibsta210:
hey @Drakulix — daily driving cosmic on a panther lake OLED that advertises full HDR caps (P3 + BT.2020 + ST 2084 + 525 nit peak / 0.0005 nit OLED blacks) and watching it sit in sdr mode every day is what brought me here.

read through the 15 commits in this PR and the discussion in pop-os/cosmic-comp#3 / #1384. a few honest questions before any code:

1. is the architectural approach in this PR (lcms + GLES color pipeline + render_elements color api) still the direction youd take today, or would you redesign? three years is a long time and id rather not rebase a stale design.

2. if the approach still holds — would you accept a rebase + cleanup PR from a contributor? master is 1369 commits ahead so the rebase wont be small, happy to take it on if there's a path to land, but only if you or another smithay maintainer has bandwidth to engage on review.

3. what was the original gating block back in 2023 — your time, missing pieces in the PR itself, or upstream protocol churn? knowing what stopped it would help me not run into the same wall.

no expectations either way. happy to redirect if hdr/color mgmt isn't the right angle right now, just figured id check before disappearing for a month rebasing 5k lines of code.

PandorasFox:
similarly, i have an hdr primary monitor (LG ULTRAGEAR+) - i have some experience engineering color management/HDR pipelining woes from a prior job; I've also been curious about the general state/design and what contributions would be welcome. i've been getting back into photography/editing hdr RAWs and have a bit of a vested interest in this at this point.

jibsta210:
> similarly, i have an hdr primary monitor (LG ULTRAGEAR+) - i have some experience engineering color management/HDR pipelining woes from a prior job; I've also been curious about the general state/design and what contributions would be welcome. i've been getting back into photography/editing hdr RAWs and have a bit of a vested interest in this at this point.

so i actually got HDR working on cosmic now - i just need to tune the colors... I will post a build on my GIT soon and maybe you can assist with that tuning if you have experience.

HigherOrderLogic:
> 1. is the architectural approach in this PR (lcms + GLES color pipeline + render_elements color api) still the direction youd take today, or would you redesign? three years is a long time and id rather not rebase a stale design.

I'm not really involved in either Smithay or Cosmic's development, but fwiw, I believe the Cosmic milestone mentioned migrating to Vulkan renderer before actually implementing color management/HDR stuff.

jibsta210:
@PandorasFox — said I'd post the build, here it is:

- **cosmic-comp fork** (HDR pipeline + live tuning): https://github.com/jibsta210/cosmic-comp/tree/feat/hdr-experiment
- **smithay fork** (atomic-DRM HDR state staging): https://github.com/jibsta210/smithay/tree/feat/hdr-experiment
- **GUI for slider-tuning the shader live**: https://github.com/jibsta210/cosmic-hdr-tuner

What works:
- Connector goes into HDR mode: `Colorspace=DCI-P3_RGB_D65` (or `BT2020_RGB`) + `HDR_OUTPUT_METADATA` blob (PQ + display primaries + 525-nit max / 0.0005-nit min mastering luminance) + `max_bpc=10`.
- 10-bit `Abgr2101010` swapchain end-to-end, atomic commits accept and stay accepted.
- Postprocess shader runs on every output frame: sRGB → linear → Rec.709→target gamut → ref-white scale → inverse PQ EOTF → 10-bit FB. (Math links: BT.2087 Annex 1 for the Rec.709→BT.2020 matrix, ST 2084 / BT.2100 for PQ, BT.2408 for ref-white concept.)

Where I'm stuck and would love a sanity check:
- **Visually washed / undersaturated** at every combination I've tried (BT.2020 tag + BT.2020 matrix + 250 nits, DCI-P3 tag + DCI-P3 matrix + 250 nits, gamut strength 0%/100%). Brightness is fine when ref_white is in range, but colors look less saturated than the same content in SDR.
- Panel: Tandem OLED on Dell XPS 16 (DA16260). EDID:
  ```
  Primaries:  R(0.6889, 0.3100), G(0.2400, 0.7209), B(0.1360, 0.0449)
  White:      D65 (0.3130, 0.3291)
  Native max: 400 cd/m² full-coverage, 525 cd/m² 10% rect
  Native min: 0.0005 cd/m²
  Min SDR:    20 cd/m² full-coverage
  Display Interface Features Data Block:
    Supported color space and EOTF standard combination 1: DCI-P3, BT.2020/SMPTE ST 2084
  ```
  Native primaries are *between* DCI-P3 and BT.2020 — closer to DCI-P3 in R/G, closer to BT.2020 in blue.

Specific questions where domain experience would help:
1. For a panel that reports `(DCI-P3, BT.2020/SMPTE ST 2084)` in its DisplayID Display Interface Features block, is the canonical HDR10 signaling path (`Colorspace=BT2020_RGB` + BT.2020-tagged metadata blob) right, or do these panels actually want `DCI-P3_RGB_D65` + DCI-P3 primaries because their native is P3?
2. Is sRGB → linear → matrix → PQ the correct order, or should the matrix run in PQ-encoded space?
3. Should I be feeding the shader the EDID's reported native primaries directly (instead of standard BT.2020/DCI-P3 constants), so the matrix targets the actual panel rather than the nominal container?
4. ABL on this panel kicks in around 400-nit full-coverage. Anything to do compositor-side, or is that purely panel firmware?

More generally — if you (or anyone reading) have config suggestions, kernel parameters, EDID quirks, or just "you're missing X obvious thing" hints that could cause the wash-out, throw them at me. Could be something basic I'm overlooking.

The shader is at `src/backend/render/shaders/offscreen.frag` (`color_mode == 5.0` branch). The Rust glue that builds the metadata blob + Colorspace prop is in `src/backend/kms/drm_helpers.rs::HdrColorContainer`.

Honest disclosure: a lot of this iteration was paired with Claude — I drove the architectural calls, the AI helped move fast through the GLSL/atomic-DRM details. Code is GPL-3 in case that matters for your appetite to look. Happy to land any of this in smithay/cosmic-comp upstream once shape and math are right.

jibsta210:
circling back — did a fresh implementation of `wp_color_management_v1` v2 in a smithay fork branch (`jibsta210/smithay@feat/hdr-experiment`), since the protocol's evolved a lot since 2023 and i wanted HDR working on my own panel. just doing this for fun — if any of it's useful for finishing this PR (or you'd merge a different shape entirely), lmk. otherwise no worries.

ai co-authored.

Drakulix:
Hey, thanks for your interest.

>     1. is the architectural approach in this PR (lcms + GLES color pipeline + render_elements color api) still the direction youd take today, or would you redesign? three years is a long time and id rather not rebase a stale design.

Partially, but those have just been initial thoughts. I am pretty sure we'll need to rely on something like `LittleLCMS` to handle ICC files eventually, though not necessarily for all transformations.

I am not sure I'll bother implementing HDR support for the GL renderer though. I lot of code has been cleaned up and the initial shadow buffer rendering approach has been removed from the renderer again for maintainability. I am currently working on a vulkan renderer for smithay, which will probably what informs the Color/HDR-api once I have some time to experiment on that and afterwards, I'll figure out if I can/want/should backport that to the GLES renderer as well. It might very well be that the maintainance burden isn't worth while as they aren't many gpus/targets, which can output HDR, but don't support vulkan.

How exactly the render-elements will need to be redesigned is still very much open and exploration is needed. I don't want smithay to force compositors into too ridgid of a specific color pipeline. Additionally the api needs to allow the `DrmCompositor` to eventually take the color-informations to offload the necessary transformations to the new drm color pipeline api.

>     2. if the approach still holds — would you accept a rebase + cleanup PR from a contributor? master is 1369 commits ahead so the rebase wont be small, happy to take it on if there's a path to land, but only if you or another smithay maintainer has bandwidth to engage on review.

I am not sure that is very worthwhile. Simply having a handler for the protocol isn't the difficult part and can be done quickly by any one. Without any code actually using it, I don't want to have this code in upstream smithay.

>     3. what was the original gating block back in 2023 — your time, missing pieces in the PR itself, or upstream protocol churn? knowing what stopped it would help me not run into the same wall.

Upstream took it's sweet time and cosmic-comp had different priorities to get 1.0 out of the door. After the vulkan renderer, HDR and overall color management is very much the first thing on my todo list though.

Side-node: Due to the complexity of the topic and it being pretty much in an api exploration stage, I would be hesitant to take on the efford on working on an external contribution for this feature. Especially if AI-generated, since we really need to get the performance characteristics and color-correct output right, while designing for long-term maintanance.

PR #983, opened 2023-04-05
replacing `dev_id: dev_t` with DrmNode

TornaxO7:
Since the `DrmNode` basically acts as a `dev_id`, why not using it instead of the `dev_t` type?

There are some breaking changes sadly (like return-type of a function differs), so I'm not sure if it's really "worth it".
What do you guys think?

### DIFF
--- a/src/backend/drm/device/mod.rs
+++ b/src/backend/drm/device/mod.rs
@@ -8,7 +8,6 @@ use std::time::{Duration, SystemTime};
 use calloop::{EventSource, Interest, Poll, PostAction, Readiness, Token, TokenFactory};
 use drm::control::{connector, crtc, plane, Device as ControlDevice, Event, Mode, ResourceHandles};
 use drm::{ClientCapability, Device as BasicDevice, DriverCapability};
-use nix::libc::dev_t;
 
 pub(super) mod atomic;
 mod fd;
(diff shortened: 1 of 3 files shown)


i509VCB (approved):
Seems fine to me.

Breaking changes and smithay are somewhat intertwined so a breaking change this small is no big deal.

There is the issue of a DrmDevice by invariant has to have a primary node, but for api consistency it makes sense to move to `DrmNode`.

src/backend/drm/device/mod.rs
```diff
@@ -372,7 +375,7 @@ impl DrmDevice {
     }
 
     /// Returns the device_id of the underlying drm node
-    pub fn device_id(&self) -> dev_t {
+    pub fn device_id(&self) -> DrmNode {
```
PolyMeilex, line 378: We now have `DrmDevice::device_id` that returns `DrmNode` and says in its docs `Returns the device_id of the underlying drm node`, and then `DrmNode` has `dev_id()` that returns `dev_t` and also says in its docs `Returns the device_id of the underlying DRM node.`. 

This is kinda confusing, I believe that we should distinguish between the two, eg. leave `device_id()` as it was and add `device_node()` getter instead, with appropriate docs that it returns a DrmNode, not the device_id

TornaxO7: Good point! I'll fix that.

TornaxO7: hm... I'd even go further and remove `device_id` and replace it with `device_node` because you can easily retrieve the ID from the node afterwards.

PolyMeilex: I don't mind

TornaxO7: done!

PR #984, opened 2023-04-06
Fix typo in documentation

chrisduerr:


chrisduerr:
I'm not sure if I mentioned this before, but it might be nice for all these `self -> Self` functions to also have `#[must_use]` annotations.

Drakulix (approved):

PR #988, opened 2023-04-11
Single pixel buffer

cmeissl:


src/backend/renderer/element/surface.rs
```diff
@@ -266,33 +270,51 @@ where
...
 
         let view = state.view()?;
         let buffer = state.buffer()?;
+        let buffer_type = crate::backend::renderer::buffer_type(buffer)?;
         let id = Id::from_wayland_resource(surface);
 
-        let texture = state.texture::<R>(renderer.id())?;
+        if matches!(buffer_type, crate::backend::renderer::BufferType::SinglePixel) {
+            let spb = crate::wayland::single_pixel_buffer::get_single_pixel_buffer(buffer).ok()?;
+            let rgba = spb.rgba8888();
```
Drakulix, line 306: So we truncate it to rgba8888, just to convert it to floats afterwards..

Can we add a `rgba16f` variant please?

Drakulix (approved):
Good start. I really like the re-use of the TextureRenderElement.

src/backend/renderer/element/surface.rs
cmeissl: yeah, that is kinda pointless. any reason why to use half float? After reading the spec (and looking at what swaybg sends) the values range from 0 - u32::MAX. So dividing by `u32::MAX as f32` (or just `0xffffffff`) should give us `f32` per component?

bwidawsk (approved):
LGTM, and I'm currently using this in my compositor.

bwidawsk:
Is there anything needed to get this merged?

src/wayland/single_pixel_buffer/handlers.rs
```diff
@@ -0,0 +1,82 @@
...
+                id: buffer,
+                r,
+                g,
+                b,
+                a,
+            } => {
+                data_init.init(buffer, SinglePixelBufferUserData { r, g, b, a });
+            }
+            wp_single_pixel_buffer_manager_v1::Request::Destroy => {}
+            _ => todo!(),
```
chrisduerr, line 56: `todo!()` shouldn't be used in this fashion, you should probably use `unimplemented!()` instead.

chrisduerr (changes requested):
+1 to compositors implementing this protocol using this PR:
https://github.com/catacombing/catacomb/commit/595eb4d9f5f4636e5d8783463b76bfb44fb84d5c

I think it would be nice to have this merged upstream soon since it requires some changes to Smithay methods (buffer_dimensions/buffer_type) and without these wrappers need to be used downstream.

PR #995, opened 2023-04-17
vulkan: add `get_supported_formats` method

zackartz:
Adds a method `get_supported_formats` to `PhysicalDevice` that returns a Vec of supported `vk::Format`.


zackartz:
moving to i509VCB's fork

PR #998, opened 2023-04-18
Fix texture leak

cmeissl:
viewporter created a ref-cycle which prevented the surface data to be cleaned up properly which resulted in "leaking" textures


Drakulix (approved):
LGTM

PR #1001, opened 2023-04-20
decouple RenderFrameResult from DrmCompositor

Drakulix:


src/backend/drm/compositor/mod.rs
```diff
@@ -642,48 +642,60 @@ type RenderFrameErrorType<A, F, R> = RenderFrameError<
     R,
 >;
 
+#[derive(Debug)]
+/// Defines the element for the primary plane in cases where a new buffer was composited
```
cmeissl, line 646: This is not necessarily true, it may also reference a previous buffer if there was no damage on the primary plane.

cmeissl (approved):
Just a small nit

PR #1002, opened 2023-04-21
drm: do not submit empty frames

cmeissl:
if a frame resulted in no damage do not try to submit the frame through drm. this would result in an error on legacy and atomic does not guarantee that a vblank will be triggered.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -52,7 +52,7 @@
 //! #     drm::{DrmDevice, DrmDeviceFd},
 //! #     renderer::{
 //! #       element::surface::WaylandSurfaceRenderElement,
-//! #       gles::{GlesRenderbuffer, GlesRenderer},
+//! #       gles::{GlesTexture, GlesRenderer},
 //! #     },
 //! # };
 //! # use drm_fourcc::{DrmFormat, DrmFourcc, DrmModifier};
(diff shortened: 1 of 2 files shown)


src/backend/drm/compositor/mod.rs
```diff
@@ -2003,15 +2014,25 @@ where
...
     /// when a vblank event is received, that denotes successful scan-out of the frame.
     /// Otherwise the underlying swapchain will eventually run out of buffers.
     ///
     /// `user_data` can be used to attach some data to a specific buffer and later retrieved with [`DrmCompositor::frame_submitted`]
     pub fn queue_frame(&mut self, user_data: U) -> FrameResult<(), A, F> {
-        self.queued_frame = self.next_frame.take().map(|state| {
+        self.queued_frame = self.next_frame.take().and_then(|state| {
+            if state.is_empty() {
+                return None;
+            }
```
Drakulix, line 2034: Didn't we want to make this return an error, so downstream can explicitly handle, that it needs to re-schedule?

cmeissl: Initially yes, but we already ignore the frame if we have no next buffer (render_frame has not been called) which was also not documented. So I thought that might be viable alternative. But I am not biased in either direction, I am also fine with making it an error. But in that case we should also discuss making no next frame an error.

Drakulix: I am for consistency and explicit-ness(?). So feel free to add both error variants. :)

cmeissl: Will do! But I just realized that this could lead to an unexpected situation for downstream when switching from GbmBufferedSurface to DrmCompositor. The next frame logic was taken over from next buffer, so iirc queue_buffer also just ignores the call if there is no next buffer. Do we want to change that too?

Drakulix: I think we should for consistency.

Drakulix (approved):

PR #1004, opened 2023-04-28
Add `zwp_idle_inhibit` protocol

chrisduerr:


src/wayland/idle_inhibit/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+}
+
+/// Handler trait for idle-inhibit.
+pub trait IdleInhibitHandler {
+    /// Enable idle inhibition for the output of the provided surface.
+    fn inhibit(&mut self, surface: WlSurface);
+
+    /// Stop inhibition for the provided surface.
+    fn uninhibit(&mut self, surface: WlSurface);
```
Drakulix, line 128: Could you add a comment describing, that this function is only called, if the client explicitly destroys the protocol object? Or maybe in simpler terms, that it is not called, if the surface is destroyed, even if it had an active inhibitor.

Drakulix (commented):
LGTM, but CI doesn't seem to be happy with your doc-comment example.

src/wayland/idle_inhibit/inhibitor.rs
```diff
@@ -0,0 +1,43 @@
+//! idle-inhibit inhibitor.
+
+use _idle_inhibit::zwp_idle_inhibitor_v1::{Request, ZwpIdleInhibitorV1};
+use wayland_protocols::wp::idle_inhibit::zv1::server as _idle_inhibit;
+use wayland_server::protocol::wl_surface::WlSurface;
+use wayland_server::{Client, DataInit, Dispatch, DisplayHandle};
+
+use crate::wayland::idle_inhibit::{IdleInhibitHandler, IdleInhibitManagerState};
+
+/// State of the zwp_idle_inhibitor_v1 Global
```
PolyMeilex, line 10: `zwp_idle_inhibitor_v1` is not a global, this is idle_inhibitor's user data.

Drakulix (approved):
LGTM! Thanks for protocol implementation :)

PR #1009, opened 2023-05-03
wayland: Surface Transactions

Drakulix:
Best reviewed commit-by-commit:

**[compositor: Use the TransactionQueue](https://github.com/Smithay/smithay/commit/71f145b59fbc71598cd8c18375eb2c5da7b0e5ee)**

Changes the internal handling of the compositor module to use the TransactionQueue. Should result in identical behavior, doesn't expose anything new to downstream.

**[compositor: Make it possible to add Blockers](https://github.com/Smithay/smithay/commit/6946f4b5e29f5e181eab73a80b215d673877b23c)**

Adds new public api to the compositor module to allow downstream to add Blockers to a surface, which will be considered on the next commit.

Couple of open questions for this commit:
- [x] ~~The `CompositorClientState::blocker_cleared`-method takes a lot of really generic arguments. It works for the upcoming dmabuf-commit, but might be problematic for other use-cases. Could we require less? Should we possibly move the responsibility to re-run the commit-handler to downstream and just take a DisplayHandle?~~ Necessary to call commit and less of an issue with the changes below.
- [x] ~~If we keep the `commit`-callback here, do we maybe want to skip the original commit-callback in the first place? We can easily bubble up the return value of `apply_ready` to conditionally skip the call.~~ `commit` call has been moved into `Transaction::apply`
- [x] ~~If there are pending blockers, post-commit-hooks will not observe any changed state. Luckily smithay doesn't use them for anything, does any compositor? Do we consider this expected behaviour? Do we want to run them once the transaction is done instead? Can we maybe just drop them?~~ `post_commit`-hooks as well

**[buffer: Add buffer_attached callback](https://github.com/Smithay/smithay/commit/91ffd61aa41ac69b2764b38ef4e381e1ead3c77e)**

Adds a new `buffer_attached`-callback to the `BufferHandler`. Prerequisite for the next commit

**[WIP: anvil: Wait for dmabufs to become ready](https://github.com/Smithay/smithay/commit/b50e3a3fcedd7d63a859ce610595a792717d0a28)**

Very rudimentary test of the above functionality.

Possible todos:
- [x] The code should be moved into a different module
- [x] Maybe we can add some implementations/abstractions in smithay to make this easier, specifically for the dmabuf case?
- [x] We should `poll` new buffers before adding a blocker for them. We can skip a complete event loop cycle, if the buffers are already ready, which is not unlikely.

### DIFF
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -9,13 +9,16 @@ use smithay::{
         WindowSurfaceType,
     },
     output::Output,
-    reexports::wayland_server::protocol::{wl_buffer::WlBuffer, wl_output, wl_surface::WlSurface},
+    reexports::wayland_server::{
+        protocol::{wl_buffer::WlBuffer, wl_output, wl_surface::WlSurface},
+        Client,
+    },
     utils::{Logical, Point, Rectangle, Size},
     wayland::{
         buffer::BufferHandler,
         compositor::{
-            get_parent, is_sync_subsurface, with_states, with_surface_tree_upward, CompositorHandler,
-            CompositorState, TraversalAction,
+            get_parent, is_sync_subsurface, with_states, with_surface_tree_upward, CompositorClientState,
+            CompositorHandler, CompositorState, TraversalAction,
         },
         shell::{
             wlr_layer::{
@@ -25,11 +28,15 @@ use smithay::{
             xdg::{XdgPopupSurfaceData, XdgToplevelSurfaceData},
         },
     },
+    xwayland::XWaylandClientData,
 };
 
-use crate::state::{AnvilState, Backend};
 #[cfg(feature = "xwayland")]
 use crate::CalloopData;
+use crate::{
+    state::{AnvilState, Backend},
+    ClientState,
+};
 
 mod element;
 mod grabs;
@@ -88,6 +95,19 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
     fn compositor_state(&mut self) -> &mut CompositorState {
         &mut self.compositor_state
     }
+    fn client_compositor_state<'a>(&self, client: &'a Client) -> &'a CompositorClientState {
+        if let Some(state) = client.get_data::<XWaylandClientData>() {
+            state
+                .user_data()
+                .insert_if_missing(CompositorClientState::default);
+            return state.user_data().get::<CompositorClientState>().unwrap();
+        }
+        if let Some(state) = client.get_data::<ClientState>() {
+            return &state.compositor_state;
+        }
+        panic!("Unknown client data type")
+    }
+
     fn commit(&mut self, surface: &WlSurface) {
         #[cfg(feature = "xwayland")]
         X11Wm::commit_hook::<CalloopData<BackendData>>(surface);
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -97,10 +97,7 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
     }
     fn client_compositor_state<'a>(&self, client: &'a Client) -> &'a CompositorClientState {
         if let Some(state) = client.get_data::<XWaylandClientData>() {
-            state
-                .user_data()
-                .insert_if_missing(CompositorClientState::default);
-            return state.user_data().get::<CompositorClientState>().unwrap();
+            return &state.compositor_state;
         }
         if let Some(state) = client.get_data::<ClientState>() {
             return &state.compositor_state;
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -1,7 +1,7 @@
 use std::cell::RefCell;
 
 #[cfg(feature = "xwayland")]
-use smithay::xwayland::X11Wm;
+use smithay::xwayland::{X11Wm, XWaylandClientData};
 use smithay::{
     backend::renderer::utils::on_commit_buffer_handler,
     desktop::{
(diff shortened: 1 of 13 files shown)


src/wayland/compositor/mod.rs
```diff
@@ -453,6 +458,21 @@ pub struct CompositorClientState {
...
 }
 
+impl CompositorClientState {
+    pub fn blocker_cleared<D: CompositorHandler>(
+        &self,
+        surface: &WlSurface,
+        state: &mut D,
+        dh: &DisplayHandle,
+    ) {
+        if let Some(queue) = self.queue.lock().unwrap().as_mut() {
```
cmeissl, line 468: Is it possible that commit again tries to access the queue? That could lead to a deadlock, which could be prevented if the lock is dropped before invoking commit.

src/wayland/compositor/mod.rs
```diff
@@ -425,6 +426,10 @@ pub fn add_destruction_hook(surface: &WlSurface, hook: fn(&SurfaceData)) {
     PrivateSurfaceData::add_destruction_hook(surface, hook)
 }
 
+pub fn add_blocker(surface: &WlSurface, blocker: impl Blocker + Send + 'static) {
```
elinorbgr, line 429: We'll need documentation here before this PR is merged.

src/wayland/compositor/mod.rs
```diff
@@ -453,6 +458,21 @@ pub struct CompositorClientState {
     queue: Mutex<Option<TransactionQueue>>,
 }
 
+impl CompositorClientState {
+    pub fn blocker_cleared<D: CompositorHandler>(
```
elinorbgr, line 462: We'll need documentation here before this PR is merged.

src/wayland/compositor/mod.rs
elinorbgr: :+1: There is no reason to hold the lock while `commit()` is invoked, and I suggest we stick to a "hold lock as shortly as needed" policy.

src/wayland/compositor/mod.rs
```diff
@@ -453,6 +458,21 @@ pub struct CompositorClientState {
...
+        &self,
+        surface: &WlSurface,
+        state: &mut D,
+        dh: &DisplayHandle,
+    ) {
+        if let Some(queue) = self.queue.lock().unwrap().as_mut() {
+            if queue.apply_ready(dh) {
+                CompositorHandler::commit(state, surface)
```
elinorbgr, line 470: I think we probably should update the documentation of `CompositorHandle::commit()` to reflect more precisely when I may be invoked.

src/wayland/buffer/mod.rs
```diff
@@ -16,8 +16,15 @@ use wayland_server::protocol::wl_buffer;
 /// [`DelegateDispatch`](wayland_server::DelegateDispatch) implementations to notify the compositor when a
 /// buffer is destroyed.
 pub trait BufferHandler {
+    /// Called when a buffer is attached to a surface.
```
elinorbgr, line 19: I think we could make this comment more detailed, maybe adding something like

> Note that at this point the buffer has not yet been committed to the surface, and it is possible that it will never be. This callback is meant to handle preparatory work to handle the buffer (such as tracking DMA fences), but should not be used as an indication that the contents of that buffer will be displayed for sure, nor about when they will be displayed.

elinorbgr:
Ok so, thinking more about it, regarding the commit callback and your questions on the second commit, a general answer:

Given how the code is right now, it's entirely possible that the invocation of `apply_ready` in `blocker_cleared` might apply transactions that are unrelated to the surface passed as argument, and in a future-proof view, applying a transaction can involve multiple surfaces, and ideally all of them should have their `commit()` callback invoked.

So, I think maybe we should check if it'd be possible to have those callbacks (and the post_commit hooks) invoked directly in `Transaction::apply` ? I believe it would solve the whole problem, if it is possible.

src/wayland/compositor/mod.rs
Drakulix: done

Drakulix:
> So, I think maybe we should check if it'd be possible to have those callbacks (and the post_commit hooks) invoked directly in Transaction::apply ? I believe it would solve the whole problem, if it is possible.

Seems like a good solution.
This is ready for another review round. :)

src/backend/allocator/dmabuf.rs
```diff
@@ -299,3 +319,195 @@ where
             .and_then(|b| AsDmabuf::export(&b).map_err(|err| AnyError(err.into())))
     }
 }
+
+/// [`crate::wayland::compositor::Blocker`] implementation for an accompaning [`DmabufSource`]
```
cmeissl, line 323:
```suggestion
/// [`crate::wayland::compositor::Blocker`] implementation for an accompanying [`DmabufSource`]
```

anvil/src/shell/mod.rs
```diff
@@ -80,14 +86,24 @@ impl FullscreenSurface {
...
     fn compositor_state(&mut self) -> &mut CompositorState {
         &mut self.compositor_state
     }
+    fn client_compositor_state<'a>(&self, client: &'a Client) -> &'a CompositorClientState {
+        #[cfg(feature = "xwayland")]
+        if let Some(state) = client.get_data::<XWaylandClientData>() {
+            state
+                .user_data()
+                .insert_if_missing(CompositorClientState::default);
+            return state.user_data().get::<CompositorClientState>().unwrap();
```
cmeissl, line 99: Could that be directly added to `XWaylandClientData`? `launch` initializes `XWaylandClientData` where it would be possible to directly add `CompositorClientState` as a field instead of using `user_data`.

anvil/src/shell/buffer.rs
```diff
@@ -0,0 +1,35 @@
...
+                    data.state
+                        .client_compositor_state(&client)
+                        .blocker_cleared(&mut data.state, &data.display.handle());
+                    Ok(())
+                });
+                if res.is_ok() {
+                    add_blocker(surface, blocker);
+                }
+            }
+        }
```
cmeissl, line 33: How does that handle clients that attach multiple buffers during a single commit? I guess that would add blockers for buffers that are no longer in use?

What would happen if a client attaches and commits buffers faster then the render loop in the compositor? Like for example `glmar2-es2-wayland` which does not wait for frame callbacks. Wouldn't that also insert blockers for buffers that are potentially already released. Could that stall updates because the buffer is already in use by the client? If the client destroys that buffers, which can happen in case of glmark, it might not be an issue, but this is not guaranteed. iirc this highly depends on how `mesa` handles the buffers.

cmeissl (commented):
Overall I really like the blocker API, I just have a few questions that I left as comments.

Drakulix:
Now based on https://github.com/Smithay/smithay/pull/1014 to make use of a pre-commit hook instead of buffer_attached for adding the blocker.

cmeissl (commented):
The whole `anvil/src/shell/buffer.rs` module looks like a left over from moving the block to the pre-commit handler.

Drakulix:
> The whole `anvil/src/shell/buffer.rs` module looks like a left over from moving the block to the pre-commit handler.

Removed.

cmeissl (approved):
LGTM

PR #1012, opened 2023-05-04
Fix input_method panic

chrisduerr:
This fixes a regression introduced in 881dfbf, which caused input method to panic on the `get_input_method` request.


chrisduerr:
This patch basically just matches what is done in `text_input` already:

https://github.com/rano-oss/smithay/blob/48d0dd680941fad6d12b6fbcee7280a0d7468897/src/wayland/text_input/mod.rs#L131-L132

chrisduerr:
@rano-oss

Drakulix (approved):

PR #1013, opened 2023-05-04
wayland: Remove Arc<Mutex<Inner>> from xdg shell state

PolyMeilex:
This used to be needed because `destroyed` handler did not have access to compositor state, but that limitation is long gone.


Drakulix (approved):
LGTM

PR #1014, opened 2023-05-04
Pass state to surface destruct hook

PolyMeilex:


src/wayland/compositor/mod.rs
```diff
@@ -418,7 +420,24 @@ pub fn add_post_commit_hook(surface: &WlSurface, hook: fn(&DisplayHandle, &WlSur
...
-pub fn add_destruction_hook(surface: &WlSurface, hook: fn(&SurfaceData)) {
+///
+/// D generic is the compositor state, same as used in `CompositorState::new<D>()`
+pub fn add_destruction_hook<D, F>(surface: &WlSurface, hook: F)
+where
+    F: Fn(&mut D, &SurfaceData) + Send + 'static,
+    D: 'static,
+{
+    let user_state_type = surface.data::<SurfaceUserData>().unwrap().user_state_type;
+    assert_eq!(
```
i509VCB, line 431: I'm wondering if that should be a debug assert, but downcast_mut already tests the typeids so we wouldn't be eliminating the runtime checks.

i509VCB, line 431: Maybe we should mention the type name here? I recall some function in std gives you the name.

i509VCB (approved):
Is this worth adding to pre and post commit hooks as well?

src/wayland/compositor/mod.rs
PolyMeilex: good call

src/wayland/compositor/tree.rs
```diff
@@ -7,11 +7,15 @@ use super::{
...
 };
 use std::{
+    any::Any,
     fmt,
-    sync::{atomic::Ordering, Mutex},
+    sync::{atomic::Ordering, Arc, Mutex},
 };
 use wayland_server::{backend::ObjectId, protocol::wl_surface::WlSurface, DisplayHandle, Resource};
 
+type CommitHook = dyn Fn(&mut dyn Any, &DisplayHandle, &WlSurface) + Send + Sync;
```
i509VCB, line 16: Is there a reason why commit hook needs Sync but destroy does not?

Drakulix: Because the hook is wrapped in an Arc, so we can clone the whole vector and release the lock.

Because Arcs are references, this means that technically multiple references could exist across multiple threads (even though they don't in this case), which is why `Arc` only implements `Send` if `T: Send + Sync`.
I don't think this additional bound is particularly limiting or hurtful, because it just affects captured variables, but we could work around that with a newtype and an unsafe `Send`-impl. But I don't really want to do that in case somebody in the future touches the code and breaks thread-safety accidentally.

i509VCB (approved):

PR #1017, opened 2023-05-05
wayland: Return serial on configure

Drakulix:
Downstream might want to track configures send and check if they have already been acked. This is quite difficult with the current api, because downstream has to compare the changes it made to the configure struct passed to `LayerShellHandler::ack_configure` and `XdgShellHandler::ack_configure` respectively.

This PR changes `send_configure` calls to return the serial used for the configure, so downstream can simply compare the serial in `ack_configure` to be sure state was acknoledged.


cmeissl (approved):
LGTM

PR #1019, opened 2023-05-07
Add ext-session-lock protocol support

chrisduerr:
This adds support for the ext-session-lock protocol, which allows
clients to lock the Wayland session and present surfaces of their choice
as lockscreens.

---

This PR does not include an Anvil implementation yet. At this point it's basically just a 1:1 copy of what I required to make this protocol work in my compositor, so before trying to implement this for Anvil I'd like some feedback on the approach taken here. I think it should serve well as a general-purpose implementation, but I'm open to suggestions.

### DIFF
--- a/src/wayland/idle_inhibit/mod.rs
+++ b/src/wayland/idle_inhibit/mod.rs
@@ -137,16 +137,16 @@ pub trait IdleInhibitHandler {
 #[macro_export]
 macro_rules! delegate_idle_inhibit {
     ($(@<$( $lt:tt $( : $clt:tt $(+ $dlt:tt )* )? ),+>)? $ty: ty) => {
-        smithay::reexports::wayland_server::delegate_global_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
-            smithay::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibit_manager_v1::ZwpIdleInhibitManagerV1: ()
+        $crate::reexports::wayland_server::delegate_global_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
+            $crate::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibit_manager_v1::ZwpIdleInhibitManagerV1: ()
         ] => $crate::wayland::idle_inhibit::IdleInhibitManagerState);
 
-        smithay::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
-            smithay::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibit_manager_v1::ZwpIdleInhibitManagerV1: ()
+        $crate::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
+            $crate::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibit_manager_v1::ZwpIdleInhibitManagerV1: ()
         ] => $crate::wayland::idle_inhibit::IdleInhibitManagerState);
 
-        smithay::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
-            smithay::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibitor_v1::ZwpIdleInhibitorV1: $crate::wayland::idle_inhibit::inhibitor::IdleInhibitorState
+        $crate::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
+            $crate::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibitor_v1::ZwpIdleInhibitorV1: $crate::wayland::idle_inhibit::inhibitor::IdleInhibitorState
         ] => $crate::wayland::idle_inhibit::IdleInhibitManagerState);
     };
 }
--- a/src/wayland/session_lock/lock.rs
+++ b/src/wayland/session_lock/lock.rs
@@ -49,7 +49,9 @@ where
         match request {
             Request::GetLockSurface { id, surface, output } => {
                 // Assign surface a role and ensure it never had one before.
-                if compositor::give_role(&surface, LOCK_SURFACE_ROLE).is_err() {
+                if compositor::get_role(&surface).is_some()
+                    || compositor::give_role(&surface, LOCK_SURFACE_ROLE).is_err()
+                {
                     lock.post_error(Error::Role, "Surface already has a role.");
                     return;
                 }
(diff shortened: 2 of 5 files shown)


src/wayland/idle_inhibit/mod.rs
```diff
@@ -137,16 +137,16 @@ pub trait IdleInhibitHandler {
...
-        smithay::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
-            smithay::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibit_manager_v1::ZwpIdleInhibitManagerV1: ()
+        $crate::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
+            $crate::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibit_manager_v1::ZwpIdleInhibitManagerV1: ()
         ] => $crate::wayland::idle_inhibit::IdleInhibitManagerState);
 
-        smithay::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
-            smithay::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibitor_v1::ZwpIdleInhibitorV1: $crate::wayland::idle_inhibit::inhibitor::IdleInhibitorState
+        $crate::reexports::wayland_server::delegate_dispatch!($(@< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $ty: [
+            $crate::reexports::wayland_protocols::wp::idle_inhibit::zv1::server::zwp_idle_inhibitor_v1::ZwpIdleInhibitorV1: $crate::wayland::idle_inhibit::inhibitor::IdleInhibitorState
```
chrisduerr, line 149: I noticed that I explicitly referred to smithay in the imports here, which I believe to be incorrect. So I decided to sneak the fix into this PR.

If this PR is not fit for merging anytime soon or you'd rather see it in a different PR in general, I can open one just for these changes (or you can if you want).

Drakulix:
> This PR does not include an Anvil implementation yet. At this point it's basically just a 1:1 copy of what I required to make this protocol work in my compositor, so before trying to implement this for Anvil I'd like some feedback on the approach taken here. I think it should serve well as a general-purpose implementation, but I'm open to suggestions.

I think the overall approach should likely work well, but an example implementation for anvil will actually be necessary for a thorough review. As I understand anvil needs to store the `SessionLockState`, track the client setting up everything and then calling `lock`. That "tracking if the client is ready to signal the successful lock" feels like it can be quite complex for a given compositor, as it involves tracking outputs, switching render-paths and doing all of that atomically so nothing is leaked.
Maybe I am overestimating this and the current API is entirely satisfactory, but if tracking this turns out to be unnecessary complex, maybe we want to provide some more helpers in this module.

The module itself looks absolutely fine to me as is. I just don't think it allows to fully asses what an implementation using it might look like, which is obviously important for review.

chrisduerr:
> That "tracking if the client is ready to signal the successful lock" feels like it can be quite complex for a given compositor, as it involves tracking outputs, switching render-paths and doing all of that atomically so nothing is leaked.
Maybe I am overestimating this and the current API is entirely satisfactory, but if tracking this turns out to be unnecessary complex, maybe we want to provide some more helpers in this module.

I have no clue how that would work in Anvil, I just redraw my output and then call lock, that doesn't seem too difficult.

Surely Anvil could just blank out each display in a single render call and then call the lock method? That shouldn't be too difficult I think?

I probably won't get around to an Anvil implementation soon, considering that I don't care about Anvil at all. But maybe I can spare some time this weekend.

Drakulix:
> Surely Anvil could just blank out each display in a single render call and then call the lock method? That shouldn't be too difficult I think?

I would hope, it is as simple as just putting a flag in `AnvilState` (or potentially an `Option<LockStatus>`) and a new if-condition inside the `render_output`-call to use the lock-surfaces instead of rendering the whole desktop. Hotplugging outputs, etc, might complicate things a little.

src/wayland/session_lock/lock.rs
```diff
@@ -0,0 +1,124 @@
...
+                let lock_state = state.lock_state();
+                if lock_state.locked_outputs.contains(&output) {
+                    lock.post_error(Error::DuplicateOutput, "Output is already locked.");
+                    return;
+                }
+                lock_state.locked_outputs.push(output.clone());
+
+                // Ensure surface has no existing buffers attached.
+                let has_buffer = compositor::with_states(&surface, |states| {
+                    let attributes = states.cached_state.current::<SurfaceAttributes>();
```
cmeissl, line 67: iirc `current` would only return a buffer after it has been committed, but if I understood the spec correctly it would also be an error to attach a buffer. A attached buffer would be found in `pending`.

The spec states:
> Providing a wl_surface which already has a role or already has a buffer attached or committed is a protocol error, as is attaching/committing a buffer before the first ext_session_lock_surface_v1.configure event.

So maybe that should also be checked in the pre_commit handler, like testing for attached and committed buffer if the intial configure has not been sent.

src/wayland/session_lock/lock.rs
```diff
@@ -0,0 +1,124 @@
...
+        lock: &ExtSessionLockV1,
+        request: Request,
+        data: &SessionLockState,
+        _display: &DisplayHandle,
+        data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            Request::GetLockSurface { id, surface, output } => {
+                // Assign surface a role and ensure it never had one before.
+                if compositor::give_role(&surface, LOCK_SURFACE_ROLE).is_err() {
```
cmeissl, line 52: Does the protocol allow the surface to be re-used with the same role? `give_role` does not return an error in this case, only if the role is different.

src/wayland/session_lock/lock.rs
```diff
@@ -0,0 +1,124 @@
...
+
+                // Initialize surface data.
+                compositor::with_states(&surface, |states| {
+                    states
+                        .data_map
+                        .insert_if_missing_threadsafe(|| Mutex::new(LockSurfaceAttributes::default()))
+                });
+
+                // Add pre-commit hook for updating surface state.
+                compositor::add_pre_commit_hook(&surface, |_dh, surface| {
```
cmeissl, line 88: If re-use is allowed this should test if it is the first time a lock surface has been requested and only add a hook once. One way would be to return the `bool` from `states.data_map.insert_if_missing...`.

src/wayland/session_lock/lock.rs
```diff
@@ -0,0 +1,124 @@
...
+                });
+                if has_buffer {
+                    lock.post_error(Error::AlreadyConstructed, "Surface has a buffer attached.");
+                    return;
+                }
+
+                let data = ExtLockSurfaceUserData {
+                    surface: surface.clone(),
+                };
+                let shell_surface = data_init.init(id, data);
```
cmeissl, line 78: I guess `shell_surface` is some kind of copy-paste thing and should be named `lock_surface`?

src/wayland/session_lock/lock.rs
chrisduerr: > if the intial configure has not been sent.

But there is no initial configure?

src/wayland/session_lock/lock.rs
chrisduerr: Pretty sure the protocol **really** wants the surface to never be reused from anything.

src/wayland/session_lock/lock.rs
chrisduerr: Probably.

chrisduerr:
I've fixed all the mentioned issues in the protocol implementation but I don't really have an interest in implementing this protocol for Anvil considering that I'm unfamiliar with both input handling and rendering.

So I'll just leave this protocol to someone who actually thinks Anvil is a worthwhile project to put their time into.

PR #1020, opened 2023-05-08
input: Add `KeyboardHandle::set_xkb_config` method

ids1024:
This allows changing the keymap used the keyboard without removing and re-adding the `wl_keyboard`. This seems to be the best way to do dynamic xkb configuration, so settings can be changed without restarting the compositor.

Like `wlr_keyboard_set_keymap`, this creates a new `xkb_state`, updates it with pressed keys, and sends modifiers to clients.

I think this should be correct but it needs testing, so leaving as a draft for now.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -1,7 +1,7 @@
 //! Keyboard-related types for smithay's input abstraction
 
 use crate::backend::input::KeyState;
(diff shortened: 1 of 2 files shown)


ids1024:
With some testing, it seems to be working as expected.

Drakulix (approved):
LGTM

PR #1021, opened 2023-05-09
wayland.idle_notify: Initial implementation

PolyMeilex:
Draft: it would be good to do anvil impl for this one, and do some proper testing.

### DIFF
--- a/src/wayland/idle_notify/mod.rs
+++ b/src/wayland/idle_notify/mod.rs
@@ -1,0 +1,345 @@
+//! This interface allows clients to monitor user idle status.
+//!
+//! ```
+//! # extern crate wayland_server;
+//! # #[macro_use] extern crate smithay;
+//! use smithay::delegate_idle_notify;
+//! use smithay::wayland::idle_notify::{IdleNotifierState, IdleNotifierHandler};
+//! # use smithay::input::{Seat, SeatHandler, SeatState, pointer::CursorImageStatus};
+//! # use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
+//!
+//! struct State { idle_notifier: IdleNotifierState<Self> }
+//! # let mut event_loop = smithay::reexports::calloop::EventLoop::<State>::try_new().unwrap();
+//! # let mut display = wayland_server::Display::<State>::new().unwrap();
+//! // Create the primary_selection state
+//! let idle_notifier = IdleNotifierState::<State>::new(
+//!     &display.handle(),
+//!     event_loop.handle(),
+//! );
+//!
+//! let state = State { idle_notifier };
+//!
+//! // implement the necessary traits
+//! # impl SeatHandler for State {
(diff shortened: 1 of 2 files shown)


src/wayland/idle_notify/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        D: 'static,
+    {
+        let global = display.create_global::<D, ExtIdleNotifierV1, _>(1, ());
+        Self {
+            global,
+            notifications: HashMap::new(),
+            loop_handle,
+            is_inhibited: false,
+        }
+    }
```
chrisduerr, line 126: Anvil uses a different state for implementing states and for the event loop data, making this very inconvenient.

I'd assume that this is why you didn't bother with the Anvil implementation?

PolyMeilex: > I'd assume that this is why you didn't bother with the Anvil implementation?

No, I did not consider that, but that's exactly why I wanted to do anvil impl before I drafting, to catch stuff like that. 

Sounds like this can be fixed with additional generic, will try to play around with that.

src/wayland/idle_notify/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    /// Is idle state inhibited, eg. by wp idle inhibit protocol
+    pub fn is_inhibited(&mut self) -> bool {
+        self.is_inhibited
+    }
+
+    /// Should be called whenever activity occurs on a seat, eg. mouse/keyboard input
+    pub fn notify_activity(&mut self, seat: &WlSeat) {
```
YaLTeR, line 159: Shouldn't this accept a `Seat`? Is it even possible to get all WlSeats out of a Seat?

src/wayland/idle_notify/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    fn notifications(&self) -> impl Iterator<Item = &ExtIdleNotificationV1> {
+        self.notifications
+            .values()
+            .into_iter()
+            .flat_map(|seat| seat.values())
+    }
+
+    fn reinsert_timer(&self, notification: &ExtIdleNotificationV1) {
```
YaLTeR, line 186: I'm thinking that maybe this logic with idle timers and inhibiting could be extracted into utils? Because if the compositor wants to do something upon idle too, subject to inhibiting, it'll have to duplicate the exact same logic essentially.

chrisduerr (commented):
I just used this protocol to implement idle-notify in my compositor:
https://github.com/catacombing/catacomb/blob/master/src/protocols/idle_notify/mod.rs

Some things I've noticed:
 - Minor API changes, simple enough to fix (see my module)
 - WlSeat seems impossible to access, I just needed things to work so I decided to just use the same map for all seats (might be an okay MVP for single-seat usage)
 - The loop_handle is annoying with Anvil, but wasn't a problem for my compositor

PR #1022, opened 2023-05-10
Make it easier to render windows with alpha

Drakulix:
This PR tries to make it easier to render windows in a semi-transparent state, e.g. for fade-in/fade-out animations.

Best reviewed commit-by-commit:

## https://github.com/Smithay/smithay/pull/1022/commits/f36fe3a2d34f49237d83c0da54fed03be533195e

Stores an alpha value inside the `WaylandSurfaceRenderElement` and makes it possible to specify that on construction.
This is necessary to reflect the state inside `Element::opaque_regions` and in-line with all other element-implementations, that also provide a way to customize alpha values on element creation.

## https://github.com/Smithay/smithay/pull/1022/commits/c4443bf842a3c68236342523271ff10c93bb2127

Adds an `alpha`-argument to the `AsRenderElement::render_elements`-method. Given alpha values needs to be specified on creation (see previous commit), this can't be part of `RenderElements::draw`.
That fact makes it impossible to create elements with custom alpha values, when using abstractions like the `Space`, that handle element creation for you. No wrapper element or similar is able to adjust the stored alpha value later.
As such adding this argument to `AsRenderElements` is a logical solution.


cmeissl (approved):
LGTM

PR #1024, opened 2023-05-14
xwayland: several fixes for window stacking

nicolasavru:
1. Since we're reparenting, we need to reorder the parent windows, not the client window. Reordering occurs relative to siblings and the client windows have no siblings under their respective parent windows. The exception is override_redirect windows, which don't get reparented.

2. When adding windows to client_list and client_list_stacking, ignore MapNotifys for the parent windows. If we don't, we end up with duplicates in the lists, one entry from the client window and one from the parent window.

3. Dedupe the code of update_stacking_order_downwards and update_stacking_order_upwards.

4. Attempt to make the docstrings of update_stacking_order_downwards and update_stacking_order_upwards cleare. In particular, try to be consistent with "abpve" and "below" to match X11's terminology. "before", "after", "in front of", and "behind" can be confusing, especially since client_list_stacking is ordered bottom -> top.

5. The implementations of update_stacking_order_downwards and update_stacking_order_upwards were seemingly swapped.

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -178,6 +178,27 @@ pub enum Reorder {
     Bottom,
 }
 
+enum StackingDirection {
+    Downwards,
+    Upwards,
+}
+
+impl StackingDirection {
+    fn pos_comparator(&self, pos: usize, last_pos: usize) -> bool {
+        match self {
+            Self::Downwards => last_pos > pos,
+            Self::Upwards => last_pos < pos,
+        }
+    }
+
+    fn stack_mode(&self) -> StackMode {
+        match self {
+            Self::Downwards => StackMode::ABOVE,
+            Self::Upwards => StackMode::BELOW,
+        }
+    }
+}
+
 /// Handler trait for X11Wm interactions
 pub trait XwmHandler {
     /// [`X11Wm`] getter for a given ID.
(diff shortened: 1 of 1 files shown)


nicolasavru:
I'm still not confident in the update_stacking_order_downwards/upwards logic, so I'd especially appreciate a verification that I understood the intent behind those algorithms correctly.

Drakulix (approved):
Thanks! Very exciting to see somebody else pick up the Xwayland mess. After all a lot of this code was added very recently and by me, learning X-stuff on the go. So I certainly expect some more subtle (or not so subtle :laughing: ) bugs in there.

The logic looks fine to me, it certainly captures the intent (though, that says nothing about, if it is actually a good design..)!

Has this fixed any issues you were having, or were you able to verify somehow, that these changes do in fact result in a more reliable stacking behaviour?

Drakulix:
(Also CI seems to be unhappy with your PR. :grimacing: )

nicolasavru:
Just clippy and formatting... fixed.

The raise_window change did fix the issue I was having: raise_window just did nothing before this change and now seems to work correctly. I'm not actually using update_stacking_order_downwards/upwards in my application, I tested them by using them to repeatedly reverse the order of all my X11 windows and observing the results with `xwininfo -root -tree` and `xprop -root` (with DISPLAY set to my application, of course). That's how I noticed the duplicate entries in _NET_CLIENT_LIST[_STACKING].

I'm learning the X stuff as I go also. :)

PR #1027, opened 2023-05-17
portability: Fallback for `SealedFile`, and remove `Boottime`

ids1024:
`memfd_create` and sealing only exist on Linux and on recent versions of FreeBSD. For NetBSD, OpenBSD, Illumos, Darwin, etc. it is necessary to fallback to something like `shm_open`.

Using `shm_open` twice to get a writable and read-only fd seems to match what wlroots does (in `allocate_shm_file_pair`). They don't seem to use `memfd_create` and sealing at all.

`CLOCK_BOOTTIME` and `CLOCK_UPTIME` don't seem to exist on platforms like NetBSD. Getting system uptime is possible, but involves `sysctl` and not a `clockid_t`. OpenBSD seems to have both `CLOCK_BOOTTIME` and `CLOCK_UPTIME`. But they are different (`CLOCK_UPTIME` excludes time spent suspended.) This isn't used anywhere. So unless it's expected to be important for something, may be simplest to remove it?

### DIFF
--- a/src/utils/clock.rs
+++ b/src/utils/clock.rs
@@ -15,26 +15,6 @@ impl ClockSource for Monotonic {
 
 impl NonNegativeClockSource for Monotonic {}
 
-/// Clock based on boottime
-#[derive(Debug)]
-pub struct Boottime;
-
-#[cfg(target_os = "linux")]
-impl ClockSource for Boottime {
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
LGTM

PR #1030, opened 2023-05-19
Lockless input filters

bwidawsk:
My goal is to be able to acquire the KbdInternal lock within filters. Holding locks during callbacks lead to deadlocks, therefore I opted to try to remove the lock rather than trying to pass along the state I cared about.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -218,6 +218,7 @@ where
 }
 
 /// Handle to the underlying keycode to allow for different conversions
+#[derive(Clone)]
 pub struct KeysymHandle<'a> {
     keycode: u32,
     keymap: &'a xkb::Keymap,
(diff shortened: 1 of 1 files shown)


Drakulix:
I am afraid (given this is potentially a hot-path), that cloning the entire keyboard state/keymap for every pressed key will negatively impact performance.
What issue is this trying to solve, why do you need to access the KeyboardHandle during the filter? Can't we just give a reference to the guard to the filter closure instead?

bwidawsk:
At least currently, the clone seems cheap, a couple of literals and some pointers. The current problem for me is to be able to acquire the current focus. Passing the guard to the filter closure would also work for me. Both of these solutions make it clear to the caller that the closure will be holding a lock, so that is acceptable as well.

@cmeissl Any thoughts since we discussed this privately?

bwidawsk:
Back to this now and I realize that guard actually holds `KbdInternal` which seems like something we don't want to pass back to the compositor. @Drakulix thoughts?

bwidawsk:
> > Back to this now and I realize that guard actually holds `KbdInternal` which seems like something we don't want to pass back to the compositor. @Drakulix thoughts?
> 
> Right, but you could wrap it in an `KeyboardInnerHandle`, no?

Not sure how to reconcile this properly. KeysymHandle wants a immutable borrow of the guard, so I can't instantiate the inner without a clone somewhere.

PR #1031, opened 2023-05-20
allow downstream to override vertex shaders

HackedOS:


Drakulix:
Tbh I don't really see a use-case for this functionality.

PR #1032, opened 2023-05-21
Add buffer age reset function

chrisduerr:
The PPP's driver seems to run into non-recoverable framebuffer errors when the `reset_buffers` function is called and the `InternalSlot` is reset while the `InternalSlot::buffer` is kept.

This patch adds a new function which only resets each slot's buffer age rather than resetting the entire slot, allowing a force-redraw on the PPP without triggering the framebuffer error.

---

This was the only solution me and @cmeissl could come up with, but he suggested to either go for something like this or to just always reset the entire slot in `reset_buffers` without keeping the buffer around. Both of those fix the issue on the PPP based on my testing.

Leaving the decision on which solution to go with to @Drakulix (unless you have an even better suggestion).

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,0 @@
-[toolchain]
-channel = "1.65.0"
--- a/src/backend/allocator/swapchain.rs
+++ b/src/backend/allocator/swapchain.rs
@@ -233,7 +233,7 @@ where
         self.slots = Default::default();
     }
 
-    /// Remove all internally cached buffers to e.g. reset age values
+    /// Remove all internally cached buffers.
     pub fn reset_buffers(&mut self) {
         for slot in &mut self.slots {
             if let Some(internal_slot) = Arc::get_mut(slot) {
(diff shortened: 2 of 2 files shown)


Drakulix:
> This was the only solution me and @cmeissl could come up with, but he suggested to either go for something like this or to just always reset the entire slot in reset_buffers without keeping the buffer around. Both of those fix the issue on the PPP based on my testing.

Interesting, resetting currently resets the user_data as well, which might leak a framebuffer handle? Maybe the drm-driver is not too happy with multiple framebuffer handles referring to the same buffer?

In any way, I think **both** is the correct solution here. Arguably resetting the buffers should actually cause a new allocation, but for performance considerations a `reset_age` call also seems like a good idea.
So could you add a commit that also resets the whole `slot` on `reset_buffers`?

Drakulix (approved):
LGTM.

@cmeissl Does anything come to mind, why this might be a bad idea?

cmeissl:
> LGTM.
> 
> @cmeissl Does anything come to mind, why this might be a bad idea?

No, this is fine imo.
I just want to add a few things.
Imo the issue is not leaking a fb handle, we clean them up on drop. The opposite seems to happen, destroying the fb seems to somehow mess up the GEM object. It looks like the object is freed despite beeing still referenced by the buffer. This makes the next call to addFB2 fail. So as long as we either add a fb for a newly allocated buffer or keep both, the buffer and fb, around, things seem to work fine. While I believe this is a driver bug, this PR can work around it and does no harm otherwise. Actually I think it is the right thing to either keep everything or reset everything together.

PR #1038, opened 2023-05-25
anvil: Set text input focus, without requring keyboard grab

ids1024:
Also adds a `TextInputSeat` trait, like `InputMethodSeat`.

Draft, because I'm not entirely sure how this is supposed to work. I needed to make this change to get fcitx5 to work with Anvil. Smithay is already calling `set_focus` in response to `GrabKeyboard` events from the IME. But fcitx5 doesn't seem to call `grab_keyboard` until it receives `activate`, which is only sent when the text input client calls `zwp_text_input_v3::enable`, but clients only do that once they get focus through `zwp_text_input_v3::enter`...

Comparing wlroots/sway, the keyboard grab doesn't seem to be involved in setting text input focus. So maybe this is the more correct way to do so? If so, should the `set_focus` call be removed?

Presumably without this change some other IME may work, and that's what this was being tested with? Though I'd expect fcitx5 to be the most prominent user of the `input-method-unstable-v2` protocol. (IBus doesn't support v2 yet: https://github.com/ibus/ibus/pull/2256.)

For reference, I tested by configuring pinyin input in `fcitx5-configtool`, and running these commands in separate terminals to watch what was going on in the protocol:

```
WAYLAND_DISPLAY=wayland-2 WAYLAND_DEBUG=1 fcitx5 |& grep input_method
GTK_IM_MODULE=wayland WAYLAND_DISPLAY=wayland-2 WAYLAND_DEBUG=1 gedit |& grep text_input
```

CC @rano-oss

### DIFF
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -63,7 +63,7 @@ use smithay::{
         shm::{ShmHandler, ShmState},
         socket::ListeningSocketSource,
         tablet_manager::TabletSeatTrait,
-        text_input::TextInputManagerState,
+        text_input::{TextInputManagerState, TextInputSeat},
         viewporter::ViewporterState,
         virtual_keyboard::VirtualKeyboardManagerState,
         xdg_activation::{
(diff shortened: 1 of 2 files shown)


rano-oss:
I agree with your assessment of the problem. If fcitx does it this we way should too. The set focus in the keyboard grab should then be removed, I am not sure but I think we still have to "pass through" the focus, but that should be simple to test. 
Also, yes, fcitx is the most prominent user of v2 and all the input methods v2 was tested with should work anyway (I only tested with standalone ones).

rano-oss:
You also need to set the popup surface role to none, if we do that internally in the set focus call in some way, we can simplify the set_focus calls to just take a surface, which would be nice :)

PR #1039, opened 2023-05-25
layer: Fix `LayerMap` `zone` to account for output transform

ids1024:
This fixes the positions of layer shell surfaces with rotated displays on cosmic-comp, and fixes the position of tiled windows (which are based on the exclusive zone).


Drakulix (approved):
LGTM

PR #1041, opened 2023-05-26
Updates for next `calloop` version, using `BorrowedFd`

ids1024:
Supersedes https://github.com/Smithay/smithay/pull/839, the other parts of which should be merged now.

This is blocked on release of `calloop` and `udev`.

The same awkwardness exists as in the previous PR around `display.backend().poll_fd()`. That returns a `BorrowedFd<'_>` that we can't simply insert into an event loop.

This is also a bit awkward with `child_stdout` in `xserver.rs` (I guess we want to remove the event source without closing that?)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -354,11 +354,8 @@ jobs:
           restore-keys: ${{ runner.os }}-cargo-registry-
 
       - name: Rust toolchain
-        # nightly changed no_coverage to coverage(off) which breaks the build
-        # see: https://github.com/rust-lang/rust/pull/114656
-        uses: dtolnay/rust-toolchain@master
+        uses: dtolnay/rust-toolchain@nightly
         with:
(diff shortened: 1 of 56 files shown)


ids1024:
Calloop is released now, but this is still blocked on a `udev` release, and some solution to the couple awkward things marked `// XXX` here.

ids1024:
> This is also a bit awkward with child_stdout in xserver.rs (I guess we want to remove the event source without closing that?)

Actually, that part seems fine. XWayland applications work without issue after changing that to work the natural way.

ids1024:
Marking ready for review. This isn't ready to be *merged*, since it depends on https://github.com/Smithay/wayland-rs/pull/635, and release of that and `udev`. But if those changes are satisfactory, this should be suitable to merge after those things, and updating to point to released versions.

elinorbgr (approved):

PR #1042, opened 2023-06-01
anvil: Show new window on output with cursor, not first output

ids1024:
Behavior with only one output should be unchanged. This seems fairly important for usability with multiple outputs. Especially with dialog windows.

A full-featured compositor may have much fancier logic, but for Anvil this seems better at least.

### DIFF
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -313,12 +313,21 @@ fn ensure_initial_configure(surface: &WlSurface, space: &Space<WindowElement>, p
     };
 }
 
-fn place_new_window(space: &mut Space<WindowElement>, window: &WindowElement, activate: bool) {
-    // place the window at a random location on the primary output
+fn place_new_window(
+    space: &mut Space<WindowElement>,
+    pointer_location: Point<f64, Logical>,
+    window: &WindowElement,
+    activate: bool,
+) {
+    // place the window at a random location on same output as pointer
     // or if there is not output in a [0;800]x[0;800] square
     use rand::distributions::{Distribution, Uniform};
 
-    let output = space.outputs().next().cloned();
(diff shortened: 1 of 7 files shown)


bwidawsk (approved):
Looks fine to me, but if position is random, maybe the output should be also?

anvil/src/shell/mod.rs
```diff
@@ -313,12 +313,27 @@ fn ensure_initial_configure(surface: &WlSurface, space: &Space<WindowElement>, p
...
+    activate: bool,
+) {
+    // place the window at a random location on same output as pointer
     // or if there is not output in a [0;800]x[0;800] square
     use rand::distributions::{Distribution, Uniform};
 
-    let output = space.outputs().next().cloned();
+    let output = space
+        .outputs()
+        .find(|o| {
```
bwidawsk, line 328: I think you can use `space::output_under` instead.

anvil/src/shell/mod.rs
```diff
@@ -313,12 +313,21 @@ fn ensure_initial_configure(surface: &WlSurface, space: &Space<WindowElement>, p
...
+    activate: bool,
+) {
+    // place the window at a random location on same output as pointer
     // or if there is not output in a [0;800]x[0;800] square
     use rand::distributions::{Distribution, Uniform};
 
-    let output = space.outputs().next().cloned();
+    let output = space
+        .output_under(pointer_location)
+        .next()
```
bwidawsk, line 328: I hadn't realized `output_under` returned an Iterator... I wouldn't have mentioned the change had I known. Still LGTM though.

bwidawsk (approved):
Still LGTM

Drakulix (approved):
LGTM

PR #1043, opened 2023-06-01
input-method: Call `compositor::give_role` on popup

ids1024:
And post a protocol error if the surface already has a role. Unfortunately while the protocol says to raise a protocol error in this case, it doesn't actually define an error enum...


rano-oss:
LGTM

Drakulix (approved):

PR #1046, opened 2023-06-06
Add get_or_insert abstractions for surface data

chrisduerr:
This adds two new public methods to the `UserDataMap` stored for surfaces, aimed at making developers' lives a little easier.

The `get_or_insert` and `get_or_insert_threadsafe` methods allow getting guaranteed access to the user data without any `unwrap` in the user's code.

The `insert` and `insert_threadsafe` methods have been added as an internal abstraction to ensure the code stays consistent. I've kept them private assuming that the `if_missing` checks are necessary because the underlying storage has no deduplication mechanisms.

### DIFF
--- a/src/utils/user_data.rs
+++ b/src/utils/user_data.rs
@@ -132,7 +132,7 @@ impl UserDataMap {
         match self.get() {
             Some(data) => data,
             None => {
-                // Innsert the new node.
+                // Insert the new node.
                 self.insert(init);
 
                 // Return it again immediately.
@@ -146,7 +146,7 @@ impl UserDataMap {
         match self.get() {
             Some(data) => data,
             None => {
-                // Innsert the new node.
+                // Insert the new node.
                 self.insert_threadsafe(init);
 
                 // Return it again immediately.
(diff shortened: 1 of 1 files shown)


src/utils/user_data.rs
```diff
@@ -127,6 +127,34 @@ impl UserDataMap {
...
+            }
+        }
+    }
+
+    /// Access the user data of a given type, initializing it if required.
+    pub fn get_or_insert_threadsafe<T: Send + Sync + 'static, F: FnOnce() -> T>(&self, init: F) -> &T {
+        match self.get() {
+            Some(data) => data,
+            None => {
+                // Innsert the new node.
```
i509VCB, line 149:
```suggestion
                // Insert the new node.
```

src/utils/user_data.rs
```diff
@@ -127,6 +127,34 @@ impl UserDataMap {
         None
     }
 
+    /// Access the user data of a given type, initializing it if required.
+    pub fn get_or_insert<T: 'static, F: FnOnce() -> T>(&self, init: F) -> &T {
+        match self.get() {
+            Some(data) => data,
+            None => {
+                // Innsert the new node.
```
i509VCB, line 135:
```suggestion
                // Insert the new node.
```

i509VCB (approved):

PR #1049, opened 2023-06-16
Shm: Fix client_lies_about_buffer_size

bwidawsk:
This fixes a basic WLCS test that checks that width is less than stride (stride is in bytes)


Drakulix (commented):
Actual fix is fine, but we should not have another format table (and also probably get rid of the `alpha` table via the same way).

Drakulix (approved):
LGTM

PR #1050, opened 2023-06-18
Feature/egl fence

cmeissl:
This PR tries to get rid of `glFinish` in most cases.
It is doing so by introduction a `SyncPoint` that is returned from `Frame::finish` which can be used to wait
for the rendering to fully realize.
Additional the `SyncPoint` can transport a fence, that in case it represents a native fence, be forwarded to kms
to let the driver wait for us.

### DIFF
--- a/anvil/src/render.rs
+++ b/anvil/src/render.rs
@@ -1,13 +1,13 @@
 use smithay::{
     backend::renderer::{
-        damage::{Error as OutputDamageTrackerError, OutputDamageTracker},
+        damage::{Error as OutputDamageTrackerError, OutputDamageTracker, RenderOutputResult},
         element::{
             surface::WaylandSurfaceRenderElement,
             utils::{
                 ConstrainAlign, ConstrainScaleBehavior, CropRenderElement, RelocateRenderElement,
                 RescaleRenderElement,
             },
-            AsRenderElements, RenderElement, RenderElementStates, Wrap,
+            AsRenderElements, RenderElement, Wrap,
         },
         ImportAll, ImportMem, Renderer,
     },
@@ -15,7 +15,7 @@ use smithay::{
         constrain_space_element, ConstrainBehavior, ConstrainReference, Space, SpaceRenderElements,
     },
     output::Output,
-    utils::{Physical, Point, Rectangle, Size},
+    utils::{Point, Rectangle, Size},
 };
 
 #[cfg(feature = "debug")]
@@ -202,7 +202,7 @@ pub fn render_output<R>(
     damage_tracker: &mut OutputDamageTracker,
     age: usize,
     show_window_preview: bool,
-) -> Result<(Option<Vec<Rectangle<i32, Physical>>>, RenderElementStates), OutputDamageTrackerError<R>>
+) -> Result<RenderOutputResult, OutputDamageTrackerError<R>>
 where
     R: Renderer + ImportAll + ImportMem,
     R::TextureId: Clone + 'static,
--- a/anvil/src/winit.rs
+++ b/anvil/src/winit.rs
@@ -359,9 +359,9 @@ pub fn run_winit() {
             });
 
             match render_res {
-                Ok((damage, states)) => {
-                    let has_rendered = damage.is_some();
-                    if let Some(damage) = damage {
+                Ok(render_output_result) => {
+                    let has_rendered = render_output_result.damage.is_some();
+                    if let Some(damage) = render_output_result.damage {
                         if let Err(err) = backend.submit(Some(&*damage)) {
                             warn!("Failed to submit buffer: {}", err);
                         }
@@ -382,11 +382,11 @@ pub fn run_winit() {
 
                     // Send frame events so that client start drawing their next frame
                     let time = state.clock.now();
-                    post_repaint(&output, &states, &state.space, None, time);
+                    post_repaint(&output, &render_output_result.states, &state.space, None, time);
 
                     if has_rendered {
                         let mut output_presentation_feedback =
-                            take_presentation_feedback(&output, &state.space, &states);
+                            take_presentation_feedback(&output, &state.space, &render_output_result.states);
                         output_presentation_feedback.presented(
                             time,
                             output
--- a/anvil/src/x11.rs
+++ b/anvil/src/x11.rs
@@ -404,7 +404,7 @@ pub fn run_x11() {
             );
 
             match render_res {
-                Ok((damage, states)) => {
+                Ok(render_output_result) => {
                     trace!("Finished rendering");
                     if let Err(err) = backend_data.surface.submit() {
                         backend_data.surface.reset_buffers();
@@ -414,7 +414,7 @@ pub fn run_x11() {
                     };
 
                     #[cfg(feature = "debug")]
-                    if damage.is_some() {
+                    if render_output_result.damage.is_some() {
                         if let Some(renderdoc) = state.renderdoc.as_mut() {
                             renderdoc.end_frame_capture(
                                 state.backend_data.renderer.egl_context().get_context_handle(),
@@ -430,11 +430,11 @@ pub fn run_x11() {
 
                     // Send frame events so that client start drawing their next frame
                     let time = state.clock.now();
-                    post_repaint(&output, &states, &state.space, None, time);
+                    post_repaint(&output, &render_output_result.states, &state.space, None, time);
 
-                    if damage.is_some() {
+                    if render_output_result.damage.is_some() {
                         let mut output_presentation_feedback =
-                            take_presentation_feedback(&output, &state.space, &states);
+                            take_presentation_feedback(&output, &state.space, &render_output_result.states);
                         output_presentation_feedback.presented(
                             time,
                             output
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -123,7 +123,7 @@
 use std::{
     cell::RefCell,
     collections::{HashMap, HashSet},
-    os::unix::io::AsFd,
+    os::unix::io::{AsFd, OwnedFd},
     rc::Rc,
     sync::{Arc, Mutex},
 };
@@ -143,7 +143,7 @@ use crate::{
             gbm::{GbmAllocator, GbmDevice},
             Allocator, Buffer, Slot, Swapchain,
         },
-        drm::{DrmError, PlaneDamageClips},
+        drm::{plane_has_property, DrmError, PlaneDamageClips},
         renderer::{
             buffer_y_inverted,
             damage::{Error as OutputDamageTrackerError, OutputDamageTracker, OutputNoMode},
@@ -151,6 +151,7 @@ use crate::{
                 Element, Id, RenderElement, RenderElementPresentationState, RenderElementState,
                 RenderElementStates, RenderingReason, UnderlyingStorage,
             },
+            sync::SyncPoint,
             utils::{CommitCounter, DamageBag, DamageSnapshot},
             Bind, Blit, DebugFlags, ExportMem, Frame as RendererFrame, Offscreen, Renderer, Texture,
         },
@@ -331,6 +332,7 @@ struct PlaneConfig<B> {
     pub alpha: f32,
     pub buffer: Owned<B>,
     pub plane_claim: PlaneClaim,
+    pub sync: Option<(SyncPoint, Option<Arc<OwnedFd>>)>,
 }
 
 impl<B> Clone for PlaneConfig<B> {
@@ -343,6 +345,7 @@ impl<B> Clone for PlaneConfig<B> {
             alpha: self.alpha,
             buffer: self.buffer.clone(),
             plane_claim: self.plane_claim.clone(),
+            sync: self.sync.clone(),
         }
     }
 }
@@ -481,6 +484,7 @@ impl<B: AsRef<framebuffer::Handle>> FrameState<B> {
     fn test_state(
         &mut self,
         surface: &DrmSurface,
+        supports_fencing: bool,
         plane: plane::Handle,
         state: PlaneState<B>,
         allow_modeset: bool,
@@ -492,24 +496,7 @@ impl<B: AsRef<framebuffer::Handle>> FrameState<B> {
         let backup = current_config.clone();
         *current_config = state;
 
-        let res = surface.test_state(
-            self.planes
-                .iter()
-                // Filter out any skipped planes
-                .filter(|(_, state)| !state.skip)
-                .map(|(handle, state)| super::surface::PlaneState {
-                    handle: *handle,
-                    config: state.config.as_ref().map(|config| super::PlaneConfig {
-                        src: config.src,
-                        dst: config.dst,
-                        transform: config.transform,
-                        alpha: config.alpha,
-                        damage_clips: config.damage_clips.as_ref().map(|d| d.blob()),
-                        fb: *config.buffer.as_ref(),
-                    }),
-                }),
-            allow_modeset,
-        );
+        let res = surface.test_state(self.build_planes(supports_fencing, true), allow_modeset);
 
         if res.is_err() {
             // test failed, restore previous state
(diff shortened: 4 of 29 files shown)


src/backend/renderer/gles/mod.rs
```diff
@@ -2363,14 +2363,39 @@ impl<'frame> Frame for GlesFrame<'frame> {
     fn finish(mut self) -> Result<SyncPoint, Self::Error> {
         self.finish_internal()
     }
+
+    fn wait(&mut self, sync: &SyncPoint) {
+        // if the sync point holds a EGLFence we can try
+        // to directly insert it in our context
+        if let Some(fence) = sync.get::<EGLFence>() {
```
Drakulix, line 2370: We need to compare here, that this is the same `EGLDisplay` or does this rely on `fence.wait()` to fail in that case?

Drakulix: After reading though it again.. I think we need to compare EGLDisplayHandles here.

src/backend/renderer/multigpu/mod.rs
```diff
@@ -1135,6 +1133,7 @@ where
                     frame
                         .clear([0.0, 0.0, 0.0, 0.0], &damage)
                         .map_err(Error::Target)?;
+                    frame.wait(&sync);
```
Drakulix, line 1136: We don't actually have to wait in this case. If this is a dmabuf imported by multiple device, sychronization is guaranteed to be handled internally.

Skipping this is important for the nvidia driver, which will wait correctly for the dmabuf, but can't import the native-fence.

src/backend/renderer/multigpu/mod.rs
```diff
@@ -1186,6 +1185,8 @@ where
                     copy_rects = Vec::from([Rectangle::from_loc_and_size((0, 0), buffer_size)]);
                 }
 
+                sync.wait();
```
Drakulix, line 1188: On the other hand we should be able to do a `frame.wait(&sync)` here instead, no?
(Although OpenGL will also synchronize this correctly internally.)

Drakulix (commented):
Looks good for a first draft.

src/backend/renderer/gles/mod.rs
cmeissl: I was (and I am still) not sure about that. `eglWaitSync` defines
> If display does not match the EGLDisplay passed to eglCreateSync when sync was created, the behavior is undefined. 

which should be fine as this always uses the display it was created from. I would expect it to fail if the current bound context does not support the provided fence.
Can you point me to the paragraph where it demands that the displays has to match the current bound one?

If we need that check I would prefer to extend the `wait` function to receive the current display and check that internally.

Drakulix: > which should be fine as this always uses the display it was created from

Yes it is fine, but it is not helpful, if you give the fence to a different renderer (e.g. inside the MultiRenderer).
Because it won't synchronize on the fence, but just insert a server-side wait for the original context:

> The server only blocks execution of commands for the specific context on which eglWaitSync was issued; other contexts implemented by the same server are not affected..

So in those cases, when we call `Renderer::wait` we need to either return an error or try to import the file descriptor from the other context to do the right thing. Otherwise we are not blocking in our current display/context on anything.

Drakulix: Also:

>  If no context is current for the currently bound client API (i.e., eglGetCurrentContext returns EGL_NO_CONTEXT), an EGL_BAD_MATCH error is generated. 

So we would want to call `make_current` before.

cmeissl: Okay, I changed `EGLFence::wait` to take a display and return an error in case the display is not equal to the display used for creating the fence. `GlesRenderer::wait` now also calls `eglMakeCurrent` before trying to use the fence.

src/backend/renderer/multigpu/mod.rs
cmeissl: right, removed the call

src/backend/renderer/multigpu/mod.rs
cmeissl: I added this to make sure the frame has finished before calling copy_framebuffer. So I don't think waiting in the target frame would work. In case of gl this is probably not needed, yeah.

anvil/src/render.rs
```diff
@@ -202,7 +203,13 @@ pub fn render_output<'a, R>(
...
     age: usize,
     show_window_preview: bool,
-) -> Result<(Option<Vec<Rectangle<i32, Physical>>>, RenderElementStates), OutputDamageTrackerError<R>>
+) -> Result<
+    (
+        Option<(SyncPoint, Vec<Rectangle<i32, Physical>>)>,
+        RenderElementStates,
+    ),
+    OutputDamageTrackerError<R>,
+>
```
Drakulix, line 212: This type is getting a bit unwieldy (I am actually a little surprised clippy is not warning on this tbh). Maybe we can put the `Result::Ok` part in a new struct?

anvil/src/winit.rs
```diff
@@ -361,7 +361,9 @@ pub fn run_winit() {
             match render_res {
                 Ok((damage, states)) => {
                     let has_rendered = damage.is_some();
-                    if let Some(damage) = damage {
+                    if let Some((sync, damage)) = damage {
+                        // wait for the rendering to finish
+                        sync.wait();
```
Drakulix, line 366: I wonder if this will not just cause additional delay (as a proper compositor would block on the submitted dmabuf anyway). For true explicit sync we would send the fence to the compositor, but I think we can continue to rely on implicit sync here and just drop the sync point?

anvil/src/x11.rs
```diff
@@ -406,33 +406,41 @@ pub fn run_x11() {
...
-                        warn!("Failed to submit buffer: {}. Retrying", err);
-                    } else {
-                        state.backend_data.render = false;
-                    };
 
-                    #[cfg(feature = "debug")]
-                    if damage.is_some() {
+                    let rendered = if let Some((sync, _)) = damage {
+                        // wait for the rendering to finish
+                        sync.wait();
```
Drakulix, line 412: see winit

src/backend/egl/fence.rs
```diff
@@ -0,0 +1,156 @@
+//! TODO: Module docs
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, OwnedFd};
+use std::sync::Arc;
+use std::time::Duration;
+
+use crate::backend::egl::{ffi, wrap_egl_call, EGLDisplay, Error};
+
+#[derive(Debug)]
+struct InnerEGLFence {
+    display: EGLDisplay,
```
Drakulix, line 10: EGLDisplay stores quite a bit of stuff, like DrmFormats and extensions. I would expect this type to be rather small. Can we instead store an `Arc<EGLDisplayHandle>` (maybe even a `Weak`) and a `supports_from_native` bool or something?
Also please don't evaluate extensions for every fence creation, but create this bool once when initializing the `EGLDisplay`. This can't change for the lifetime of an `EGLDisplay` anyway.

src/backend/renderer/gles/mod.rs
```diff
@@ -2434,24 +2467,21 @@ impl<'frame> GlesFrame<'frame> {
             }
         }
 
+        // if we support egl fences we should use it
+        if self.renderer.extensions.iter().any(|s| s == "GL_OES_EGL_sync") {
```
Drakulix, line 2471: Given this is potentially a hot path, maybe also make this a bool or at least a capability, so we don't have to do string comparisions, each time we finish a frame.

src/backend/renderer/multigpu/mod.rs
Drakulix: Given we only have the GLRenderer and the API currently expects implicit sync, I would not bother explicitly waiting here. I expect a discussion on copy_framebuffer and other apis to come up, when implementing the vulkan renderer.

Also I didn't meant `target_frame.wait`, but `render.renderer().wait`, I didn't realize this was an API of `Frame`.

src/backend/renderer/mod.rs
```diff
@@ -216,6 +218,12 @@ pub trait Frame {
     /// Output transformation that is applied to this frame
     fn transformation(&self) -> Transform;
 
+    /// Wait for the [`SyncPoint`] to be signaled
+    fn wait(&mut self, sync: &sync::SyncPoint) -> Result<(), Self::Error> {
```
Drakulix, line 222: Why is this an api of `Frame` and not `Renderer` (or both?)

Drakulix (commented):
Couple of more remarks. I really like how this is shaping up.

anvil/src/render.rs
cmeissl: It does warn, but it was hidden behind the `glow` build error ^^ I already changed it to use a `OutputRenderResult`

anvil/src/winit.rs
cmeissl: removed the explicit wait

anvil/src/x11.rs
cmeissl: removed the explicit wait

src/backend/renderer/gles/mod.rs
cmeissl: yeah, added a capability for fencing

src/backend/renderer/multigpu/mod.rs
cmeissl: removed the explicit wait here

src/backend/renderer/mod.rs
cmeissl: The intend behind this was that it only makes sense if you are currently rendering something to insert a server wait. I can add the same function to `Renderer` if you want?

cmeissl: added `wait` to the `Renderer` trait

src/backend/renderer/sync/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use downcast_rs::{impl_downcast, Downcast};
+
+#[cfg(feature = "backend_egl")]
+mod egl;
+
+/// A fence that will be signaled in finite time
+pub trait Fence: std::fmt::Debug + Send + Sync + Downcast {
+    /// Queries the state of the fence
+    fn is_signaled(&self) -> bool;
```
i509VCB, line 12: I wonder if we could avoid needing to clone a bunch of state for the vulkan renderer by making checks for signalling be done on the renderer/device. SyncPoint however makes that a little more complicated to do.

Essentially my opinion is that a fence implementation should be a sort of handle into the renderer/device.

i509VCB: Thinking about the implementation, I would probably be able to make this work with a `Weak<ash::Device>` and an underlying vulkan `Fence` (with a drop of sort to ensure the underlying fence doesn't go away while a fence live) (pre-exporting ahead of time). So that I don't need to clone a whole lot of the renderer.

This would also mean that it would still be Send + Sync

src/backend/drm/compositor/mod.rs
```diff
@@ -1006,14 +1073,15 @@ where
...
-        if let Some((dmabuf, geometry)) = primary_dmabuf {
+        if let Some((sync, dmabuf, geometry)) = primary_dmabuf {
             let blit_damage = damage
                 .iter()
                 .filter_map(|d| d.intersection(geometry))
                 .collect::<Vec<_>>();
 
             trace!("blitting frame with damage: {:#?}", blit_damage);
 
+            sync.wait();
```
Drakulix, line 1084: Sorry I am only seeing this now, but I am a little worried again, that this might cause unnecessary blocks in the client, but for the screencopy use-case.

It totally makes sense for the `commit` and `page_flip` functions, but here we have a renderer. Can't and shouldn't we do `renderer.wait(&sync)` here?

src/backend/renderer/glow.rs
```diff
@@ -270,17 +270,21 @@ impl<'frame> Frame for GlowFrame<'frame> {
...
     }
 
-    fn finish(mut self) -> Result<(), Self::Error> {
+    fn finish(mut self) -> Result<sync::SyncPoint, Self::Error> {
         self.finish_internal()
     }
+
+    fn wait(&mut self, sync: &sync::SyncPoint) -> Result<(), Self::Error> {
+        self.frame.as_mut().unwrap().wait(sync)
+    }
```
Drakulix, line 279: `GlowRenderer` also need a `wait`-implementation!

src/backend/renderer/multigpu/mod.rs
```diff
@@ -1519,9 +1524,13 @@ where
...
     }
 
-    fn finish(mut self) -> Result<(), Self::Error> {
+    fn finish(mut self) -> Result<sync::SyncPoint, Self::Error> {
         self.finish_internal()
     }
+
+    fn wait(&mut self, sync: &sync::SyncPoint) -> Result<(), Self::Error> {
+        self.frame.as_mut().unwrap().wait(sync).map_err(Error::Render)
+    }
```
Drakulix, line 1533: `MultiRenderer` also needs a `wait`-implementation!

Drakulix (commented):
Almost good to go. :sweat_smile: 
I think these are the last things to address.

src/backend/renderer/glow.rs
cmeissl: oh, good catch!

src/backend/renderer/mod.rs
```diff
@@ -226,7 +234,7 @@ pub trait Frame {
     /// Leaking the frame instead will leak resources and can cause any of the previous effects.
     /// Leaking might make the renderer return Errors and force it's recreation.
     /// Leaking may not cause otherwise undefined behavior and program execution will always continue normally.
-    fn finish(self) -> Result<(), Self::Error>;
+    fn finish(self) -> Result<sync::SyncPoint, Self::Error>;
```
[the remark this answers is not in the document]
cmeissl: The function is already `must_use` as it returns a `Result`, we could either add a message better describing why it must be used or we could maybe put `must_use` on the `SyncPoint`

src/backend/renderer/glow.rs
cmeissl: done

src/backend/renderer/multigpu/mod.rs
cmeissl: done

src/backend/drm/compositor/mod.rs
cmeissl: yeah, I replaced it with a potential server wait and also changed the function to return a `SyncPoint` instead of waiting for the rendering of overlay elements.

Drakulix (approved):
LGTM now! Thanks for working on this :)

src/backend/drm/compositor/mod.rs
```diff
@@ -519,42 +523,90 @@ impl<B: AsRef<framebuffer::Handle>> FrameState<B> {
...
+                        // Try to extract a native fence out of the supplied sync point if any
+                        // If the sync point has no native fence or the surface does not support
+                        // fencing force a wait
+                        if let Some((sync, fence)) = config.sync.as_mut() {
+                            if supports_fencing && fence.is_none() {
+                                *fence = sync.native().map(Arc::new);
+                            }
+
+                            if fence.is_none() {
+                                sync.wait();
```
i509VCB, line 549: Do we want to wait between each map or accumulate each fence that needs to be waited on and wait in a single spot?

src/backend/drm/compositor/mod.rs
```diff
@@ -519,42 +523,90 @@ impl<B: AsRef<framebuffer::Handle>> FrameState<B> {
...
```
i509VCB, line 594: Similar question to commit

src/backend/drm/surface/gbm.rs
```diff
@@ -23,10 +25,16 @@ pub struct GbmBufferedSurface<A: Allocator<Buffer = BufferObject<()>> + 'static,
     current_fb: Slot<BufferObject<()>>,
     pending_fb: Option<(Slot<BufferObject<()>>, U)>,
     #[allow(clippy::type_complexity)]
-    queued_fb: Option<(Slot<BufferObject<()>>, Option<Vec<Rectangle<i32, Physical>>>, U)>,
+    queued_fb: Option<(
```
i509VCB, line 28: (bo, sync, damage, user data)?

src/backend/renderer/mod.rs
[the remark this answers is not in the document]
i509VCB: Ah yeah I really meant `SyncPoint`, but `#[must_use]` can't really apply to traits? I guess in that case put it on the implementations?

src/backend/renderer/sync/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// Blocks the current thread until the fence is signaled.
+    fn wait(&self);
+
+    /// Returns whether this fence has a native handle
+    fn is_native(&self) -> bool;
+
+    /// Get the native fence fd if available
+    fn native(&self) -> Option<OwnedFd>;
```
i509VCB, line 21: What are the expected semantics this function should consider?

I see a few things here:
1. Every call to `Fence::native` must return a new fd. This seems to imply that implementations should export the sync fd when this function is called.
2. One possible implementation strategy I could think of is to export the fd before returning the sync point and just dup the fd. However `dup`ing the fd returns an io::Result.

Another question: could we return a `BorrowedFd` here and be fine? I am not sure if the fd needs to be valid for the lifetime of the `IN_FENCES_PTR` or whatever it is called or could you just drop the fence immediately after commiting/page flip?

src/backend/drm/compositor/mod.rs
cmeissl: I changed that part a bit so that it now also applies to `test_state`. There is a new `build_planes` function that tries to extract the native fence and wait if that fails (if we are not only testing the state).

src/backend/drm/surface/gbm.rs
cmeissl: yeah, that starts to get ugly. replaced with an internal struct.

src/backend/renderer/mod.rs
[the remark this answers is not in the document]
cmeissl: `SyncPoint` is a struct, so we can put `#[must_use]` on it -> done
(Already revealed some missing waits in the `buffer_test`)

src/backend/renderer/sync/mod.rs
cmeissl: Yes, 1. is correct, currently it expects to always return a new fd as indicated by returning an `OwnedFd` (of course this is some kind of implementation detail as the `Fence` could also dup the fd). For the `Fence` trait this is correct imo as both, EGLFence and VkFence, export a new fd and transfer ownership.
Personally I don't think 2. would be a good fit, we do not need a native fd for all use-cases. If you just want to insert the fence in another renderer (as a server wait) or blocking wait we do not necessarily need the fd at all. In case we want to import a fd into a renderer we again have to transfer ownership and I personally don't want to introduce another error case by `dup` here.

`IN_FENCE_FD` already works with `BorrowedFd` in the drm abstraction. In case of linux the native fence (the fd) is expected to point to a `sync_file`. On a test commit kms only checks if the provided fd points to a valid `sync_file`, on commit it will import the sync_file internally. But I opted to cache the `OwnedFd` in the `DrmCompositor` so that we do not have to re-export the fd on each atomic test.

src/backend/egl/fence.rs
```diff
@@ -0,0 +1,163 @@
...
+pub struct EGLFence(Arc<InnerEGLFence>);
+
+impl EGLFence {
+    /// Returns whether the display supports creating fences from native fences
+    pub fn supports_from_native(display: &EGLDisplay) -> bool {
+        display.supports_native_fences
+    }
+
+    /// Create a fence from a native fence
+    pub fn from_native(display: &EGLDisplay, native: OwnedFd) -> Result<Self, Error> {
```
i509VCB, line 38: Would it be more accurate to call this importing instead? Vulkan uses the term import for this type of native sync fd to fence

src/backend/egl/fence.rs
```diff
@@ -0,0 +1,163 @@
...
+        ];
+        let handle = wrap_egl_call(|| unsafe {
+            ffi::egl::CreateSync(
+                **display_handle,
+                ffi::egl::SYNC_NATIVE_FENCE_ANDROID,
+                attributes.as_ptr(),
+            )
+        })
+        .map_err(Error::CreationFailed)?;
+        // SAFETY: Ownership has been transferred to egl
```
i509VCB, line 61: Because EGL took ownership of the fd, it would probably be more clear to use mem::forget

src/backend/renderer/gles/mod.rs
```diff
@@ -3054,8 +3097,13 @@ impl<'frame> GlesFrame<'frame> {
 
 impl<'frame> Drop for GlesFrame<'frame> {
     fn drop(&mut self) {
-        if let Err(err) = self.finish_internal() {
-            warn!("Ignored error finishing GlesFrame on drop: {}", err);
+        match self.finish_internal() {
+            Ok(sync) => {
+                sync.wait();
```
i509VCB, line 3102: Is this is going to still cause GlesFrame to block even the SyncPoint is used in KMS?

src/backend/renderer/sync/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+impl_downcast!(Fence);
+
+/// A sync point the will be signaled in finite time
+#[derive(Debug, Clone)]
+#[must_use]
```
i509VCB, line 27: We can put a message here if useful that the compiler will add to diagnostic's:
```rust
#[must_use = "message here"]
```

i509VCB (approved):

src/backend/egl/fence.rs
cmeissl: In case of an error we want to make sure we close the fd, but yes, I can restructure that a bit to use `into_raw_fd` and close the fd in case of an error.

src/backend/renderer/sync/mod.rs
cmeissl: added a message

src/backend/egl/fence.rs
cmeissl: tbh I have no preference, both is fine for me. @Drakulix any preference?

PR #1054, opened 2023-06-27
Allow customizing GlesRenderer capabilities

### LINKED ISSUE #1023: 10-bit color support drastically reduced performance
I've tested the time it takes to call `render` in my compositor and it went from ~1ms while starting up Firefox with a fixed website to ~10ms.

The bisect identified that one of the following commits is at fault:

```
594fb497ef37a5b7f7dc36cbbca77e8f3a65af65
c5b9baebb571fc2799b807a93062cc5c39f5e910
06389b115127b90c5607d2f8add97ec6c00247a0
99f605a51661100b75c4259ddcb1a8e9c7ae312c
57ceb42c72544ae5a1bfb6d7ba254fc73c62ab10
7196c3e0ae43c49a2163e8063703189d535e2551
318738bb84cbe82da3effd04a28cbee091542b7e
```

These don't build, so I could not bisect any further, but it looks like https://github.com/Smithay/smithay/pull/935 is likely at fault.

chrisduerr:
This patch separates the retrieval of the available driver capabilities from `GlesRenderer::new`, allowing users to define their own capability list.

The driver's defaults can be easily retrieved with the new `Capability::supported_capabilities` function.

The `None` value is allowed to mimic the current behavior to ensure the complexity of the API isn't increased unnecessarily for people not interested in this feature.

Closes #1023.

---

I haven't added any of these changes to the CHANGELOG.md because that file seems to be in complete disarray anyway. If you want a changelog entry, please let me know where and what.


src/backend/renderer/gles/mod.rs
```diff
@@ -366,6 +368,106 @@ pub enum Capability {
     ColorTransformations,
 }
 
+impl Capability {
+    /// Find all capabilities supported by the driver.
+    pub fn supported_capabilities() -> Vec<Capability> {
```
i509VCB, line 373: You need to specify a context must be current and possibly return Err if no context is current. The other option is to borrow the context.

src/backend/renderer/gles/mod.rs
```diff
@@ -501,7 +605,10 @@ impl GlesRenderer {
     /// - The renderer might use two-pass rendering internally to facilitate color space transformations.
     ///   As such it reserves any stencil buffer for internal use and makes no guarantee about previous framebuffer
     ///   contents being accessible during the lifetime of a `GlesFrame`.
-    pub unsafe fn new(context: EGLContext) -> Result<GlesRenderer, GlesError> {
+    pub unsafe fn new<C>(context: EGLContext, capabilities: C) -> Result<GlesRenderer, GlesError>
+    where
+        C: Into<Option<Vec<Capability>>>,
```
i509VCB, line 610: I'd actually prefer the developer be explicit about what features are desired, if you want everything the driver supports then just pass the result of get_capabilities into this function.

src/backend/renderer/multigpu/egl.rs
```diff
@@ -67,7 +68,11 @@ impl<R: From<GlesRenderer> + Renderer<Error = GlesError>> GraphicsApi for EglGle
     type Device = EglGlesDevice<R>;
     type Error = Error;
 
-    fn enumerate(&self, list: &mut Vec<Self::Device>) -> Result<(), Self::Error> {
+    fn enumerate(
+        &self,
+        list: &mut Vec<Self::Device>,
+        capabilities: &Option<Vec<Capability>>,
```
i509VCB, line 74: Could it be desirable to use different capabilities on different devices?

src/backend/renderer/gles/mod.rs
```diff
@@ -366,6 +368,106 @@ pub enum Capability {
     ColorTransformations,
 }
 
+impl Capability {
```
i509VCB, line 371: I think this would be better located on the gles renderer type.

src/backend/renderer/gles/mod.rs
```diff
@@ -366,6 +368,106 @@ pub enum Capability {
...
+        // Get available extensions.
+        let extensions = match GlExtensions::new(&gl) {
+            Ok(extensions) => extensions,
+            Err(_) => return Vec::new(),
+        };
+
+        let mut capabilities = Vec::with_capacity(4);
+
+        // Get maximum supported OpenGL version.
+        let gl_version = version::GlVersion::try_from(&gl).unwrap_or(version::GLES_2_0);
```
i509VCB, line 386: What if the compositor wants to request a specific GL version?

src/backend/renderer/gles/mod.rs
chrisduerr: Why? To me that just seems like an unnecessary burden on most of the consumers.

src/backend/renderer/multigpu/egl.rs
chrisduerr: Maybe, but this was about as complex as I was interested in making the multigpu changes.

I can also remove them for now and have multigpu support added by someone that makes use of that feature.

src/backend/renderer/gles/mod.rs
chrisduerr: That isn't possible with the current GlesRenderer either, right?

i509VCB: Yeah I guess that is an issue we can address later given that we don't allow that at the moment either

src/backend/renderer/multigpu/mod.rs
```diff
@@ -50,6 +50,7 @@ use std::{
     rc::Rc,
 };
 
+use crate::backend::renderer::gles::Capability;
```
i509VCB, line 53: This causes the multigpu code to depend on types in gles. Capabilities probably could be put to a later time here since we probably want more broad capabilities that are renderer generic.

chrisduerr:
Closing in favor of https://github.com/Smithay/smithay/pull/1055.

PR #1056, opened 2023-06-27
Some small perf improvements

cmeissl:
Some things I noticed during profiling


src/wayland/dmabuf/mod.rs
```diff
@@ -422,9 +422,23 @@ impl DmabufFeedbackBuilder {
...
+    fn eq(&self, other: &Self) -> bool {
+        // Note: The dmabuf feedback can not change, it is initialized
+        // with the DmabufFeedbackBuilder, so if the arc ptr equal we
+        // can short-circuit the expensive equality test saving us
+        // a few cpu cycles
+        if Arc::ptr_eq(&self.0, &other.0) {
+            return true;
+        }
+
+        self.0 == other.0
```
bwidawsk, line 438: Wouldn't the inverse apply? ie, instead of this, why not just `false`

src/backend/renderer/gles/mod.rs
```diff
@@ -2570,6 +2569,7 @@ impl<'frame> GlesFrame<'frame> {
 
         // as a last option we force finish, this is unlikely to happen
         unsafe {
+            self.renderer.gl.Flush();
```
bwidawsk, line 2572: I don't think flush buys anything here.

bwidawsk (approved):

src/backend/renderer/gles/mod.rs
cmeissl: Yeah, right, glFinish alone should be enough.

src/wayland/dmabuf/mod.rs
cmeissl: No, not necessarily. The feedback could still hold the same formats and tranches, in which case we do not want to send it to the client. Sending the same feedback could result in unnecessary re-allocations in the client.

src/backend/renderer/gles/mod.rs
cmeissl: removed the unnecessary flush

Drakulix (approved):

PR #1057, opened 2023-06-29
Perf/profiling

cmeissl:
Some changes I have been using locally for profiling anvil/smithay, mainly with `tracy`.

Example:

Screenshot from 2023-06-29 11-53-02

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ x11rb = { version = "0.11.1", optional = true }
 xkbcommon = { version = "0.5.0", features = ["wayland"]}
 scan_fmt = { version = "0.2.3", default-features = false }
 encoding = { version = "0.2.33", optional = true }
+profiling = "1.0"
 
 [dev-dependencies]
 clap = { version = "4", features = ["derive"] }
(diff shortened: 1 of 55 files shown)


Drakulix (commented):
Awesome. I meant to do this at some point, highly appreciated.

cmeissl:
> Awesome. I meant to do this at some point, highly appreciated.

Cool, I was not sure there is interest in it as we already have `tracing` and this adds another dependency. Though if no feature is enabled the dependency should be a no-op.

PolyMeilex (approved):
This sounds awesome!
Just off curiosity, how does this differ from something like `tracing_tracy`?

cmeissl:
> This sounds awesome! Just off curiosity, how does this differ from something like `tracing_tracy`?

With this the use of `tracing_tracy` is still possible. Initially I used `tracing_tracy`, but the result were a bit off. For getting meaningful data you would have to enable the trace log level which is pretty heavy in smithay as we log some quite big messages. So using `profiling` with `tracy_client` is way more lightweight and it also leaves the decision to downstream which framework it wants to use.

Drakulix (approved):

PR #1058, opened 2023-06-29
fix: offscreen texture has_alpha determination

gsrushton:
GlesTexture::from_raw takes an `opaque` parameter. When creating an offscreen texture this parameter was set to `has_alpha` which would imply transparency, I.E. the opposite of opaque.


Drakulix (approved):
Nice catch! Thanks :)

PR #1059, opened 2023-06-29
Refining fence api for nvidia driver

Drakulix:
First commit makes sure the rendered contents are visible to the other gpu before reading out the buffer.

An alternative might be to re-import the dmabuf instead, but waiting is just as good. For non-nvidia systems the code should just import the fence to the other gpu directly, causing a server-wait just like a potential re-import.

Second commit moves responsibility to synchronize kms-submission of the `DrmCompositor` to downstream. This allows compositors to skip waiting on the SyncPoint for the PrimaryPlaneElement for drivers that correctly synchronize on the submitted dmabufs.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -496,7 +496,7 @@ impl<B: AsRef<framebuffer::Handle>> FrameState<B> {
         let backup = current_config.clone();
         *current_config = state;
 
-        let res = surface.test_state(self.build_planes(supports_fencing, true), allow_modeset);
+        let res = surface.test_state(self.build_planes(supports_fencing), allow_modeset);
 
         if res.is_err() {
             // test failed, restore previous state
@@ -512,7 +512,7 @@ impl<B: AsRef<framebuffer::Handle>> FrameState<B> {
         supports_fencing: bool,
         event: bool,
     ) -> Result<(), crate::backend::drm::error::Error> {
-        surface.commit(self.build_planes(supports_fencing, false), event)
+        surface.commit(self.build_planes(supports_fencing), event)
     }
 
     fn page_flip(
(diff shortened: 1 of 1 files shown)


src/backend/drm/compositor/mod.rs
```diff
@@ -733,12 +724,13 @@ pub struct RenderFrameResult<'a, B: Buffer, F: AsRef<framebuffer::Handle>, E> {
 }
 
 impl<'a, B: Buffer, F: AsRef<framebuffer::Handle>, E> RenderFrameResult<'a, B, F, E> {
-    /// Returns whether the frame would block the current thread when queued
-    pub fn would_block(&self) -> bool {
+    /// Returns a sync point, if synchronization with kms submission can't be guaranteed through the available apis.
+    pub fn needs_sync(&self) -> Option<SyncPoint> {
```
cmeissl, line 728: I am not sure if we really want to return a sync point here. The reason for `would_block` only returning a `bool` was to prepare for multiple sync points in the future when we want to expose one per plane element too. So the idea was to have a quick way to identify if sync is needed with a bool and leave it to downstream to figure out how to wait. If I would have had more time I thought about adding a helper that can be used to wait or insert it into calloop transparently (taking all sync points into consideration in the future).

Drakulix: It should be possible to merge fences, so I am not sure, if we need to make this api unnecessarily complicated. Though if the api can't import a fence, we can't block on the fences of a dmabuf in combination of a render-syncpoint, as we can't get them all into the same api... Though maybe we could handle this internally in the SyncPoint and allow it to contain multiple fences in the future?

Drakulix: But I am also happy with just returning a bool.

cmeissl: imo a simple bool should be enough until we have to think about merging fences. It could still be handy when we find a proper api for multiple fences as a quick check.

PR #1061, opened 2023-07-01
Fix unnecessary damage

cmeissl:
This PR tries to reduce damage on the primary plane.

The `OutputDamageTracker` now automatically damages areas no longer covered by opaque regions.
This enables us to return to only report opaque regions from overlay plane elements in the `DrmCompositor`
and no longer forward damage of the original element.

~~The last commit changes the `Element` trait to return an `Option` for opaque regions.  This can save us from some allocations for non opaque elements. They were already tracked as on Option in most elements anyway.
The commit is fully optional, so I can easily drop it if we do not want to change the api.~~


cmeissl:
@chrisduerr As this PR changes the opaque region handling (again) can you please check that it does not regress your initial opaque region issue?

chrisduerr:
LGTM

Drakulix (approved):

PR #1062, opened 2023-07-02
egl: Only query eglGetError() when an error happened

linkmauve:
Most EGL calls return a value which indicates whether the call succeeded or not, so avoid checking for errors when none happened.

I went for specialisation since checking for `!= EGL_FALSE` gets redundant pretty quickly, same for `!.is_null()`.  There are a few functions returning unexpected errors though, in our case `eglClientWaitSync()` and `eglDupNativeFenceFDANDROID()`, returning respectively `EGL_FALSE` even though it is a `EGLint`, and `-1`.

This has been tested with the [Catacomb](https://github.com/catacombing/catacomb) compositor, it indeed reduces the amount of `eglGetError()` calls to exactly zero according to [apitrace](https://apitrace.github.io/).

### DIFF
--- a/src/backend/egl/error.rs
+++ b/src/backend/egl/error.rs
@@ -152,8 +152,32 @@ impl EGLError {
     }
 }
 
-/// Wraps a raw egl call and returns error codes from `eglGetError`, if it fails.
-pub fn wrap_egl_call<R, F: FnOnce() -> R>(call: F) -> Result<R, EGLError> {
+/// Wraps a raw egl call and returns error codes from `eglGetError()`, only if the result of the
+/// call is different from the `err` value.
+pub fn wrap_egl_call<R: PartialEq, F: FnOnce() -> R>(call: F, err: R) -> Result<R, EGLError> {
     let res = call();
-    EGLError::from_last_call().map(|()| res)
+    if res != err {
+        Ok(res)
+    } else {
+        EGLError::from_last_call().map(|()| res)
+    }
+}
+
+/// Wraps a raw egl call and returns error codes from `eglGetError()`, only if the pointer returned
+/// is null.
+pub fn wrap_egl_call_ptr<R, F: FnOnce() -> *const R>(call: F) -> Result<*const R, EGLError> {
+    let res = call();
+    if !res.is_null() {
+        Ok(res)
+    } else {
+        EGLError::from_last_call().map(|()| res)
+    }
+}
+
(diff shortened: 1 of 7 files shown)


src/backend/egl/error.rs
```diff
@@ -153,7 +153,28 @@ impl EGLError {
 }
 
 /// Wraps a raw egl call and returns error codes from `eglGetError`, if it fails.
-pub fn wrap_egl_call<R, F: FnOnce() -> R>(call: F) -> Result<R, EGLError> {
+pub fn wrap_egl_call<R: PartialEq, F: FnOnce() -> R>(call: F, err: R) -> Result<R, EGLError> {
```
cmeissl, line 156: Can you please extend the doc comment with a description about the `err` paremeter and what
it is exactly used for.

src/backend/egl/error.rs
```diff
@@ -153,7 +153,28 @@ impl EGLError {
...
-    EGLError::from_last_call().map(|()| res)
+    if res != err {
+        Ok(res)
+    } else {
+        EGLError::from_last_call().map(|()| res)
+    }
+}
+
+/// Wraps a raw egl call and returns error codes from `eglGetError`, if it fails.
+pub fn wrap_egl_call_ptr<R, F: FnOnce() -> *const R>(call: F) -> Result<*const R, EGLError> {
```
cmeissl, line 166: Same here, maybe just a short description that the pointer returned from
the wrapped call is checked against `null`

src/backend/egl/error.rs
```diff
@@ -153,7 +153,28 @@ impl EGLError {
...
+    let res = call();
+    if !res.is_null() {
+        Ok(res)
+    } else {
+        EGLError::from_last_call().map(|()| res)
+    }
+}
+
+/// Wraps a raw egl call and returns error codes from `eglGetError`, if it fails.
+pub fn wrap_egl_call_bool<F: FnOnce() -> ffi::egl::types::EGLBoolean>(
```
cmeissl, line 176: While pretty obvious here, we should also define the error semantics here.

cmeissl:
Thanks for working on this! In general I am fine with the proposed changes, just some small nits on the docs.

linkmauve:
Fixed the documentation, I made a mental memo to change it at first, then completely forgot about it, sorry!

Drakulix (approved):
LGTM, just waiting for CI to agree

PR #1064, opened 2023-07-03
cleanup texture cache

cmeissl:


src/backend/drm/compositor/mod.rs
```diff
@@ -1952,6 +1952,13 @@ where
 
             match render_res {
                 Ok(render_output_result) => {
+                    if render_output_result.damage.is_none() {
+                        // if we receive no damage we can assume no rendering took place
+                        // and we should trigger a cleanup of the renderer texture cache
+                        // to prevent holding textures longer then necessary
+                        renderer.cleanup_texture_cache();
```
Drakulix, line 1959: So that means, we kinda code the `DrmCompositor` around the assumption, that rendering will implicitly cleanup the texture cache? I kinda don't like that tbh, but I don't have a much better solution at hand.

Drakulix: This needs to be at least very well documented on the `cleanup_texture_cache` method.

Drakulix (commented):
The first commit is certainly fine this way (apart from docs). I don't love the second, but I am not sure, if calling `cleanup_texture_cache` all the time is a better solution. It probably not negligible in terms of performance. Can we maybe do a more explicit "no rendering happened"-state than using `damage.is_none` as a proxy for that information.

src/backend/drm/compositor/mod.rs
cmeissl: Yeah, I don't like that either. But we are currently doing implicit cleanup which is probably OK. Removing that and relying on explicit cleanup with a separate function is worse imo. I am not sure there is an easy way to make the explicit cleanup less costly by only invoking it when really needed. But without it we potentially keep a lot of memory alive unnecessarily. Doing it the way I implemented will at least make sure we do not cleanup twice. I can extend the result of the output damage tracker to explicit tell us if rendering took place. But with the current api I don't see a way on how to encode the cleanup requirements (including the implicit cleanup) other then documenting it in the renderer trait docs. Would that be enough? Clarifying the implicit cleanup in the docs and adding a renderer bool to the result?

Drakulix: > Clarifying the implicit cleanup in the docs and adding a renderer bool to the result?

Yeah I would be fine with it as an immediate measure. Maybe when the vulkan renderer finally materializes we can take a step back and evaluate that approach again.

Drakulix: Okay, so while I am all for merging this, can we at least add a note of implicit cleanup happening on `Frame::drop` to the `Frame`-trait and `Renderer::cleanup_texture_cache` docs?

cmeissl: Okay, I added some comment to the trait and the cleanup function.

Drakulix (approved):

src/backend/renderer/multigpu/mod.rs
```diff
@@ -1224,6 +1240,10 @@ where
                         )
                         .map_err(Error::Target)?;
                     let sync = frame.finish().map_err(Error::Target)?;
+                    render
```
YaLTeR, line 1243: Just to clarify, why does this call `cleanup_texture_cache()` explicitly, even though the new docs say that `Frame::finish()` does it automatically?

cmeissl: The `frame` here is from the `target`, so the call to `cleanup_texture_cache` should clean-up the other renderer in this case.

YaLTeR: I see, thanks

PR #1068, opened 2023-07-11
desktop: Return whenever re-arrangement of LayerMap caused changes

Drakulix:


src/desktop/wayland/layer.rs
```diff
@@ -409,6 +412,7 @@ impl LayerMap {
                 let size_changed = layer.0.surface.with_pending_state(|state| {
                     state.size.replace(size).map(|old| old != size).unwrap_or(true)
                 });
+                changed = changed && size_changed;
```
PolyMeilex, line 415: So, if the only way for `changed` to become true is location change, this mean that if only size/s changed `changed` will never become true?

In different words: if location did not change this will always evaluate to `false && size_changed = false`

Is that right? I get an impression that perhaps this was meant to be an OR?

PR #1069, opened 2023-07-12
drm/atomic: Use new mode when testing

Drakulix:
When using the `DrmSurface::use_mode` function, we are currently testing the plane config with the old mode previously set, which will fail on most (if not all) drivers.

### DIFF
--- a/src/backend/drm/surface/atomic.rs
+++ b/src/backend/drm/surface/atomic.rs
@@ -442,10 +442,10 @@ impl AtomicDrmSurface {
             [&PlaneState {
                 handle: self.plane,
                 config: Some(PlaneConfig {
-                    src: Rectangle::from_loc_and_size(Point::default(), pending.mode.size()).to_f64(),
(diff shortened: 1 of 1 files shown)


Vixea:
Huh, enlighten me but to me pending seems like it would well be pending or in process of becoming the current mode not the previous mode. Why is it called pending?

Drakulix:
> Huh, enlighten me but to me pending seems like it would well be pending or in process of becoming the current mode not the previous mode. Why is it called pending?

The pending-state is used, when commiting to apply any changes, that have been accumulated. This method is one of those, that change the pending state.
As such we do actually set `pending.mode` later in the same function: https://github.com/Smithay/smithay/pull/1069/files#diff-6bcab28ce0de67371f28dcd4a7e52a15270bd50b473e6403eeff1eafc4397a6bR472

But before we do that, we test, that the new mode would actually be accepted. And to do that, we have to use the new mode and not whatever was previously stored in the pending state.

Tl;dr This method is updating said pending mode, so it can't rely on that.

cmeissl (approved):
LGTM

PR #1071, opened 2023-07-12
Add back manual `Debug` impls removed in f628336b

ids1024:
These are still needed to implemented `Debug` without a `D: Debug` bound.

If there were *one* manual debug impl like this I'd add a comment noting that it is the same as the derived implementation but with a different bound, and link to https://github.com/rust-lang/rust/issues/26925. But adding that comment everywhere would be verbose...

This still slightly simplifies many of these implementations, since they used to require `Debug` bounds for `::KeyboardFocus`, `::PointerFocus` etc. When removing those I switched to deriving `Debug` since that *seemed* to work.

### DIFF
--- a/src/desktop/wayland/popup/grab.rs
+++ b/src/desktop/wayland/popup/grab.rs
@@ -1,4 +1,7 @@
-use std::sync::{Arc, Mutex};
+use std::{
+    fmt,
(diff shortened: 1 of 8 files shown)


Drakulix (approved):
Thanks!

PR #1075, opened 2023-07-14
Add static output mode support for DrmCompositor

chrisduerr:
This patch adds new functions to the `DrmCompositor` to allow users to control when its output mode should be updated, rather than using the Output to update the mode automatically.

---

This *seems* to work with my compositor and fixes my issues. Though no guarantees it doesn't mess with damage tracking somehow.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -123,6 +123,7 @@
 use std::{
     cell::RefCell,
     collections::{HashMap, HashSet},
+    fmt::Debug,
     os::unix::io::{AsFd, OwnedFd},
     rc::Rc,
     sync::{Arc, Mutex},
@@ -157,7 +158,7 @@ use crate::{
         },
         SwapBuffersError,
     },
-    output::{Output, OutputModeSource, OutputNoMode},
+    output::{OutputModeSource, OutputNoMode},
     utils::{Buffer as BufferCoords, DevPath, Physical, Point, Rectangle, Scale, Size, Transform},
 };
 
@@ -1243,9 +1244,13 @@ where
     <F as ExportFramebuffer<A::Buffer>>::Error: std::error::Error + Send + Sync,
     G: AsFd + Clone,
 {
-    /// Initialize a new [`DrmCompositor`]
+    /// Initialize a new [`DrmCompositor`].
     ///
-    /// - `output` is used to determine the current mode, scale and transform
+    /// The [`OutputModeSource`] can be created from an [`&Output`], which will automatically track
+    /// the output's mode changes. An [`OutputModeSource::Static`] variant should only be used when
+    /// manually updating modes using [`DrmCompositor::set_output_mode_source`].
+    ///
+    /// - `output_mode_source` is used to determine the current mode, scale and transform
     /// - `surface` for the compositor to use
     /// - `planes` defines which planes the compositor is allowed to use for direct scan-out.
     ///           `None` will result in the compositor to use all planes as specified by [`DrmSurface::planes`]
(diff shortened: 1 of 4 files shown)


src/backend/drm/compositor/mod.rs
```diff
@@ -1263,6 +1263,40 @@ where
...
+    /// Initialize a new [`DrmCompositor`] with custom output mode source.
+    ///
+    /// This method should only be used when trying to create a [`DrmCompositor`] that manually
+    /// updates its output mode through [`DrmCompositor::set_output_mode_source`]. If you want the
+    /// mode changes to be handled automatically, use [`DrmCompositor::new`] instead.
+    ///
+    /// See also: [`DrmCompositor::new`]
+    #[allow(clippy::too_many_arguments)]
+    #[instrument(skip(allocator, framebuffer_exporter))]
+    pub fn new_with_output_mode_source(
```
chrisduerr, line 1296: This method could in theory be made private with a second method that just takes a size/scale/transform and handles this opaquely.

However I've found that for my compositor just implementing `From<MyOutput> for OutputModeSource` made this rather convenient and it reduces the number of parameters (which are already excessive), so I think just having this public should be perfectly fine (and maybe even better than the alternative).

src/output.rs
```diff
@@ -440,3 +440,51 @@ impl PartialEq<Output> for WeakOutput {
         self.upgrade().map(|o| &o == other).unwrap_or(false)
     }
 }
+
+/// Source for determining output mode information.
+#[derive(PartialEq, Clone, Debug)]
+pub enum OutputModeSource {
```
chrisduerr, line 446: Moved all this to `output`, I think that's a much better place for it anyway (and the file is smaller too).

src/output.rs
```diff
@@ -440,3 +440,51 @@ impl PartialEq<Output> for WeakOutput {
...
+        /// Size of the static output
+        size: Size<i32, Physical>,
+        /// Scale of the static output
+        scale: utils::Scale<f64>,
+        /// Transform of the static output
+        transform: Transform,
+    },
+}
+
+impl TryFrom<&OutputModeSource> for (Size<i32, Physical>, utils::Scale<f64>, Transform) {
```
chrisduerr, line 460: Let me know if my rust-fu betrayed me, but I don't think there's a way to automatically implement this for both `&OutputModeSource` and `OutputModeSource`?

src/output.rs
```diff
@@ -440,3 +440,51 @@ impl PartialEq<Output> for WeakOutput {
...
+            OutputModeSource::Static {
+                size,
+                scale,
+                transform,
+            } => Ok((*size, *scale, *transform)),
+        }
+    }
+}
+
+impl TryFrom<OutputModeSource> for (Size<i32, Physical>, utils::Scale<f64>, Transform) {
```
chrisduerr, line 479: Switched this to `TryFrom`, because it's better and I don't know why it wasn't `TryFrom` to begin with.

src/output.rs
i509VCB: `Copy` would probably workaround the ref impl, but OutputModeSource can't because of the `Output`

chrisduerr:
CI failure seems unrelated?

src/backend/drm/compositor/mod.rs
Drakulix: Maybe we can even simplify this into one `new`-function taking a `impl Into<OutputModeSource>` and implement that for `Output`?

src/backend/renderer/damage.rs
```diff
@@ -360,14 +320,27 @@ impl OutputDamageTracker {
...
+    /// sources. For known modes use [`OutputDamageTracker::new`] or
+    /// [`OutputDamageTracker::from_output`] instead.
+    pub fn from_mode_source(output_mode_source: OutputModeSource) -> Self {
+        Self {
+            mode: output_mode_source,
+            last_state: Default::default(),
+            span: info_span!("render_damage"),
+        }
+    }
+
```
Drakulix, line 341: Same about these constructors. This is not part of the released api and even that I wouldn't care too much about stability here. Lets just do one constructor for `impl Into<OutputModeSource>` and implement that for `Output` and a static tuple (or introduce another small completely public struct for static size, scale and transform).

Drakulix (commented):
Liking this a lot. I think we can even simplify this a little further (see comments)

src/backend/renderer/damage.rs
chrisduerr: I don't think that really makes things easier. We already have "a small tuple", it's the `OutputModeSource` variant.

It's just going to get incredibly complicated and difficult for people to figure out which types are allowed here by relying on an overwhelming amount of from/into impls.

Another method isn't that difficult, but it can make reading the API much, much simpler.

Drakulix (approved):

PR #1076, opened 2023-07-14
layer_map: Don't ignore dead surfaces

Drakulix:
This check makes it impossible to figure out which `LayerSurface` belonged to a `WlrLayerSurface` passed to `WlrLayerShellHandler::layer_destroyed`.

This is completely unnecessary, as all the state should still be available, when this handler is triggered. This is especially useful to figure out what output a now dead layer-surface was previously shown on to re-render said output.


PolyMeilex (approved):
Yeah, this sounds good to me, it was probably needlessly inherited from pre 0.30 days anyway.

PR #1077, opened 2023-07-14
utils/clock: Rename duration_since

bwidawsk:
In order to avoid colliding with the logical definition of function std::time::duration_since, rename utils' opposite implementation to be more fitting.

For more details:
https://doc.rust-lang.org/std/time/struct.SystemTime.html#method.duration_since

### DIFF
--- a/src/utils/clock.rs
+++ b/src/utils/clock.rs
@@ -69,9 +69,9 @@ pub struct Time<Kind> {
 }
 
 impl<Kind> Time<Kind> {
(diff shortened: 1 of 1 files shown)


bwidawsk:
Should I rename this to `elapsed`?

Vixea:
Yeah imo sounds better

Drakulix (approved):
LGTM, definitely less confusing

PR #1082, opened 2023-07-19
egl: Support ANGLE EGL platform on winit Wayland

ids1024:
For this to work I also had to modify `LIB` to point to the `libEGL.so` from ANGLE. And set `VK_ICD_FILENAMES`, because naturally it wasn't working with Nvidia drivers, but was with Intel.

This uses Vulkan. It seems ANGLE only provides the Wayland platform with Vulkan.

This then runs, but errors with "Depth/stencil buffer format combination not allowed for blit." It seems it needs to use `DEPTH24_STENCIL8`. Presumably that's what the default framebuffer is using?

### DIFF
--- a/src/backend/egl/ffi.rs
+++ b/src/backend/egl/ffi.rs
@@ -356,4 +356,9 @@ pub mod egl {
     // Accepted in the <attribute> parameter of eglQueryWaylandBufferWL:
     pub const EGL_TEXTURE_FORMAT: i32 = 0x3080;
     pub const WAYLAND_Y_INVERTED_WL: i32 = 0x31DB;
+
+    pub const PLATFORM_ANGLE_ANGLE: u32 = 0x3202;
+    pub const PLATFORM_ANGLE_TYPE_ANGLE: i32 = 0x3203;
+    pub const PLATFORM_ANGLE_NATIVE_PLATFORM_TYPE_ANGLE: i32 = 0x348F;
+    pub const PLATFORM_ANGLE_TYPE_VULKAN_ANGLE: i32 = 0x3450;
 }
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
LGTM

src/backend/egl/ffi.rs
```diff
@@ -356,4 +356,9 @@ pub mod egl {
     // Accepted in the <attribute> parameter of eglQueryWaylandBufferWL:
     pub const EGL_TEXTURE_FORMAT: i32 = 0x3080;
     pub const WAYLAND_Y_INVERTED_WL: i32 = 0x31DB;
+
+    pub const PLATFORM_ANGLE_ANGLE: u32 = 0x3202;
+    pub const PLATFORM_ANGLE_TYPE_ANGLE: i32 = 0x3203;
+    pub const PLATFORM_ANGLE_NATIVE_PLATFORM_TYPE_ANGLE: i32 = 0x348F;
+    pub const PLATFORM_ANGLE_TYPE_VULKAN_ANGLE: i32 = 0x3450;
```
Drakulix, line 363: Of course ANGLE doesn't seem to have an extension to define these?

i509VCB: gl-generator is effectively abandoned so these might not be in the registry. It's a non trivial task to update gl-generator either

ids1024: I tried to update this in gl-generator (https://github.com/brendanzab/gl-rs/pull/544), but the master branch seems to run into other issues in Smithay currently. So just defining here seems to be the best option currently.

heavyrain266: Have you tried some newer generators like the one used by glow?

i509VCB: phosphorus only works with gl.xml, egl.xml support had a pull request but that was abandoned

Drakulix: Well, certainly unfortunate, but I guess this works.

PR #1084, opened 2023-07-19
gles: Support more formats for GlesTexture

bwidawsk:
Currently the only texture format that can be sampled from as an offscreen texture is RGBA8 persuant to the GLES2 spec. Since BGRA is already required, and since buffers that are allocated for textures end up being BGRA, allow this format as well.


Drakulix (approved):
https://registry.khronos.org/OpenGL/extensions/EXT/EXT_texture_format_BGRA8888.txt says, that the format is to be accepted by `TexImage2D`. So this is at least certainly a valid backing format, though I am not sure if some implementations might reject creating a FRAMEBUFFER from such a texture, causing `Bind<GlesTexture>::bind` to fail in this case. But that is also valid in our implementation, so... no reason not to merge this.

PR #1086, opened 2023-07-21
Feature/kms optimizations

cmeissl:


src/backend/drm/surface/mod.rs
```diff
@@ -391,125 +386,7 @@ impl DrmSurface {
...
-        }
-
-        trace!(
-            "Supported scan-out formats for plane ({:?}): {:?}",
-            plane,
-            formats
-        );
-
-        Ok(formats)
+        plane_formats(self, plane)
```
Drakulix, line 389: Isn't it a bit weird, that we do enumerate this for `PlaneInfo`, but the `DrmSurface` still have to re-query this every time this function is called?

src/backend/drm/compositor/mod.rs
```diff
@@ -2727,35 +3059,73 @@ where
...
+        };
+        let is_compatible = previous_state
+            .plane_state(plane_info.handle)
+            .map(|state| {
+                state
+                    .config
+                    .as_ref()
+                    .map(|other| {
+                        // Note: We do not use the plane config `is_compatible` test
+                        // here as we exclude the destition location from the test
```
Drakulix, line 3086:
```suggestion
                        // here as we exclude the destination location from the test
```

Drakulix (commented):
Looks mostly good to me

src/backend/drm/compositor/mod.rs
```diff
@@ -1671,19 +1680,9 @@ where
...
-                &self.surface,
-                self.supports_fencing,
-                self.planes.primary.handle,
-                primary_plane_state,
-                self.surface.commit_pending(),
-            )
-            .map_err(FrameError::DrmError)?;
+        // unconditionally set the primary plane state
+        // if this would fail the test we are screwed anyway
+        next_frame_state.set_state(self.planes.primary.handle, primary_plane_state);
```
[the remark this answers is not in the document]
cmeissl: Yeah, this only sets the "initial" frame state to a swapchain slot. The only real difference is that this now no longer fails in case the device has been deactivated (tty switch), so upstream would notice that a bit later. But this is no issue imo as this can happen anytime anyway.

src/backend/drm/surface/mod.rs
cmeissl: `PlaneInfo` uses the same function under the hood, but neither are cached in the `DrmSurface`.  We can get rid of the `DrmSurface::supported_formats` call in the `DrmCompositor`, but we might want to keep the function for `GbmBufferedSurface`. Do you suggest to remove the `supported_formats` completely? Or to cache the PlaneInfo in the `DrmSurface` and return the formats from the cached info?

cmeissl: Ah, we already call `planes` in `DrmDevice::create_surface`, so we could just put that once in the `DrmSurface`?

Drakulix: > Ah, we already call planes in DrmDevice::create_surface, so we could just put that once in the DrmSurface?

Right, that is what I was thinking.

bwidawsk:
FYI this fixes a direct scanout bug when clear color is black

bwidawsk:
I'd like to see this move along due to the bug fix. Anything I can do to help?

src/backend/drm/surface/mod.rs
cmeissl: Just pushed a commit with these changes, if you agree with them I would incorporate them correctly in the history and re-base the branch.

cmeissl:
> 

Sorry for taking so long, I continued to work on this and hope we can get it merged in the next couple of days.

src/backend/drm/surface/mod.rs
Drakulix: Yep, looks good to me.

Drakulix (approved):

Issue #1087, opened 2023-07-21
anvil: Tablet events move cursor but don't send mouse events


ids1024:
I'm not entirely sure how tablet events should be handled, and how this interacts with applications using `tablet-unstable-v2`.

But it seems unexpected that moving the cursor using the tablet moves the same cursor as the mouse does, but the application gets no mouse events and believes the cursor has not moved.

PR #1088, opened 2023-07-21
anvil: Don't store `pointer_location`, use from `PointerHandle`

ids1024:
If the pointer location is tracked within Smithay, it seems bad for to have a separate copy of it. Assuming we never want them to get out of sync.

Defining a method for `pointer_location()` is awkward since we can't split borrows in method calls, so the call has to be moved earlier, in some uses. But the code is slightly verbose to duplicate everywhere it's used. Maybe just store a copy of `PointerHandle` in the state, to avoid having to get it from the seat and unwrap?

More importantly, there are a few cases marked `XXX` where pointer events were not generated.
(https://github.com/Smithay/smithay/issues/1087 is one of them). Do we want to just generate pointer motion (but not relative motion) events in all of these cases?

### DIFF
--- a/anvil/src/shell/x11.rs
+++ b/anvil/src/shell/x11.rs
@@ -52,7 +52,12 @@ impl<BackendData: Backend> XwmHandler for CalloopData<BackendData> {
     fn map_window_request(&mut self, _xwm: XwmId, window: X11Surface) {
         window.set_mapped(true).unwrap();
         let window = WindowElement::X11(window);
-        place_new_window(&mut self.state.space, self.state.pointer_location, &window, true);
+        place_new_window(
+            &mut self.state.space,
+            self.state.pointer.current_location(),
+            &window,
+            true,
+        );
         let bbox = self.state.space.element_bbox(&window).unwrap();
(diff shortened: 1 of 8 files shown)


anvil/src/shell/x11.rs
```diff
@@ -206,9 +211,7 @@ impl<BackendData: Backend> XwmHandler for CalloopData<BackendData> {
     }
 
     fn resize_request(&mut self, _xwm: XwmId, window: X11Surface, _button: u32, edges: X11ResizeEdge) {
-        let seat = &self.state.seat; // luckily anvil only supports one seat anyway...
```
Drakulix, line 209: This comment is still very much true, if we store a `PointerHandle` in the state.

Drakulix (commented):
Doesn't seem to be terrible idea and would certainly be showing off better, that the current location shouldn't be a state global.

ids1024:
I was a bit confused by the wlcs test failure, but that was due to `surface_under` using the previous pointer location. Fixed by making it take `pos` as an argument.

bwidawsk:
LGTM

Drakulix (approved):

PR #1090, opened 2023-07-25
POC of smithay-derive

PolyMeilex:
TODO:
[ ] Support all modules
[ ] Add old macro style as well, not only derive, and remove old definitions
[ ] If we add non-derive style macro, crate should probably be renamed to `smithay-macros` or `smithay-proc`
[ ] Use in anvil and smallvil

### DIFF
--- a/src/wayland/module_registry.rs
+++ b/src/wayland/module_registry.rs
@@ -1,0 +1,109 @@
+use proc_macro2::TokenStream as TokenStream2;
+use std::collections::HashMap;
+
+#[derive(Debug)]
+pub struct ModuleRegistry {
+    modules: HashMap<String, ModuleDescriptor>,
+}
+
+impl Default for ModuleRegistry {
+    fn default() -> Self {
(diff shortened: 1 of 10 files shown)


ids1024:
So, this should also eliminate the need to write/copy long and arcane `macro_rules! delegate_*` declarations? That will be nice. That is not a highlight of dealing with code in smithay and mithay-client-toolkit.

`smithay-macros` may be a better name if we add other kinds of proc macros to it later. But it doesn't matter that much.

src/wayland/module_registry.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+impl ModuleRegistry {
+    pub fn new() -> Self {
+        let mut registry = Self {
+            modules: HashMap::new(),
+        };
+
+        registry.regiter(super::compositor::descriptor());
```
Drakulix, line 21: `regiter`?

PR #1093, opened 2023-08-05
XWayland keyboard grab protocol

ids1024:


src/wayland/xwayland_keyboard_grab.rs
```diff
@@ -0,0 +1,176 @@
...
+        &mut self,
+        data: &mut D,
+        handle: &mut KeyboardInnerHandle<'_, D>,
+        keycode: u32,
+        state: KeyState,
+        modifiers: Option<keyboard::ModifiersState>,
+        serial: Serial,
+        time: u32,
+    ) {
+        // TODO set_focus?
```
Drakulix, line 44: Yeah I am not sure, if the KeyboardHandler does this when setting the grab already, but we definitely want to set focus to `None` at the start of the grab.

Drakulix: Though this would send a wl_keyboard.leave event, I am not sure, if that is what Xwayland expects? Or if it wants the surface to get focus?

Drakulix: From the protocol "forcing the keyboard focus for the given seat upon the given surface.",
I guess we want to call `set_focus` upon `self.surface` here.

This means we might need a requirement for `From<WlSurface>` for `<D as SeatHandler>::KeyboardFocus`. I wonder if that is too generic and we want to introduce a wrapper-type like `XwaylandGrabSurface(pub WlSurface)` here and specifically require a `From`-implementation for that type. This would allow compositors to special case this particular Focus-variant.

Drakulix (commented):
Looks good!

src/wayland/xwayland_keyboard_grab.rs
ids1024: If we define the helper trait as `trait XWaylandKeyboardGrabHandler: SeatHandler`, we can require a method returning `Self::KeyboardFocus`, which might be cleaner than using `From` with a special wrapper type if it's only used here.

I guess for either Anvil or Cosmic this `WlSurface` needs to first be converted to the corresponding `X11Surface`.

Drakulix: Right the X11Surface would be cleaner to handle input properly.

I have no clear vision of how you would want to model this, but feel free to go ahead with this approach. It sounds sensible and I would like to see how it would look in practice.

src/wayland/xwayland_keyboard_grab.rs
```diff
@@ -0,0 +1,197 @@
...
+pub trait XWaylandKeyboardGrabHandler: SeatHandler {
+    /// XWayland client has requested a keyboard grab for `surface` on `seat`
+    fn grab(
+        &mut self,
+        surface: wl_surface::WlSurface,
+        seat: wl_seat::WlSeat,
+        grab: XWaylandKeyboardGrab<Self>,
+    );
+
+    fn keyboard_focus_for_xsurface(&self, surface: &wl_surface::WlSurface) -> Self::KeyboardFocus;
```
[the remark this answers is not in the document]
ids1024: We could have it return an `Option` then log an error if it's `None`. Or if it fails because XWayland is misbehaving, a protocol error would make more sense than crashing the compositor. Though the protocol doesn't define any protocol errors. It doesn't technically even mention that the `wl_surface` should be associated with an X window and mapped. But if it's only used by XWayland, we only need to deal with how XWayland uses the protocol anyway.

Theoretically some sort of race may be possible without `xwayland_shell_v1`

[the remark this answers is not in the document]
Drakulix: Right so `None` and ignoring the grab (which is valid behaviour for a compositor according to the protocol anyway) seems like a good middle-ground. This allows compositor to replace `unwrap` with `?` at least easily, should the surface not be mappable to an `X11Surface`.

ids1024:
Would it be better for Smithay to call `set_grab` here, instead of requiring a `grab` handler method? That's what is done for input-method. Is there any reason a compositor would want to do something other than just setting the grab like the Anvil implementation here?

Or it could provide a default implementation for the `grab` method, at least.

Drakulix:
> Would it be better for Smithay to call `set_grab` here, instead of requiring a `grab` handler method? That's what is done for input-method. Is there any reason a compositor would want to do something other than just setting the grab like the Anvil implementation here?
> 
> Or it could provide a default implementation for the `grab` method, at least.

Default implementation is fine, but given the protocol explicitly does not guarantee the grab to be set and suggest compositors might even show indicators for keys being captured, I say this should be compositor policy.

Drakulix (approved):
LGTM

PR #1095, opened 2023-08-09
Minor smallvil fixes

YaLTeR:
Trying to make a compositor and these are some things I bumped into.


Drakulix (approved):
Looks all good! Thanks for putting in the time

PR #1100, opened 2023-08-15
Smallvil fixes 2

YaLTeR:
This time: fix clipboard and add basic popup support.

### DIFF
--- a/smallvil/src/handlers/compositor.rs
+++ b/smallvil/src/handlers/compositor.rs
@@ -38,7 +38,7 @@ impl CompositorHandler for Smallvil {
             }
         };
 
-        xdg_shell::handle_commit(&self.space, surface);
+        xdg_shell::handle_commit(&mut self.popups, &self.space, surface);
         resize_grab::handle_commit(&mut self.space, surface);
     }
 }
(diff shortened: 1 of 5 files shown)


PolyMeilex (approved):
Looks great 👍

smallvil/src/handlers/compositor.rs
```diff
@@ -40,6 +43,25 @@ impl CompositorHandler for Smallvil {
...
+                    .lock()
+                    .unwrap()
+                    .initial_configure_sent
+            });
+            if !initial_configure_sent {
+                // NOTE: This should never fail as the initial configure is always
+                // allowed.
+                popup.send_configure().expect("initial configure failed");
+            }
+        }
```
PolyMeilex, line 64: Could we move this to xdg specific commit handler? Aka `xdg_shell::handle_commit` from line 44
My idea was to have everything xdg related in a single xdg file, where both commit and xdg handler logic lives

PolyMeilex (changes requested):

smallvil/src/handlers/compositor.rs
YaLTeR: Done.

PolyMeilex (approved):
Thanks! Popups support in Smallvil was long overdue, so it's great to see it.

PR #1106, opened 2023-08-28
Security Context protocol

ids1024:
The compositor should not expose the `wp_security_context_manager_v1` global to clients connected to a security context. Otherwise, I guess this would be up to the compositor to set its own restrictions.

This implementation is currently incomplete.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -443,7 +443,7 @@ jobs:
       - name: Build Documentation
         env: 
           RUSTDOCFLAGS: --cfg=docsrs
-        run: cargo doc --no-deps --features "test_all_features" -p smithay -p calloop:0.10.6 -p drm -p gbm -p input -p nix:0.26.2 -p udev -p slog -p wayland-server -p wayland-backend -p wayland-protocols:0.30.1 -p winit -p x11rb
+        run: cargo doc --no-deps --features "test_all_features" -p smithay -p calloop:0.10.6 -p drm -p gbm -p input -p nix:0.26.4 -p udev -p slog -p wayland-server -p wayland-backend -p wayland-protocols:0.30.1 -p winit -p x11rb
         
       - name: Setup index
         run: cp ./doc_index.html ./target/doc/index.html
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ tempfile = { version = "3.0", optional = true }
 thiserror = "1.0.25"
 udev = { version = "0.7", optional = true }
 wayland-egl = { version = "0.30.0", optional = true }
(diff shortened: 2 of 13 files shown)


ids1024:
Rebased on top of https://github.com/Smithay/smithay/pull/1108, so CI can pass. This seems to be working correctly now, testing Anvil with a flatpak containing the wayland-rs `list_globals` example, I can confirm the security context global appear when run in a flatpak when `flatpak` is built from master with support for this.

Drakulix (approved):
Looks good to me in general, though I wonder, if we should force `can_view` to just return false for any client, that already has a security context?

While I am all for flexibility, I am not sure, for what scenario you could want a faulty implementation.

A filter for more restrictive filtering could still be provided.

ids1024:
>  I wonder, if we should force can_view to just return false for any client, that already has a security context?

I think so, but I couldn't think of a good way to do that. The `ClientData` associated with the client is managed by the compositor, not Smithay, so I don't know if Smithay has a good way to test if `Client` has a security context... except by having a method to query that implemented by the compositor.

If Smithay doesn't use the security context for anything except excluding clients with a security context from using this protocol, the filter function already handles that so it's a bit redundant to add anything else.

Drakulix:
> > I wonder, if we should force can_view to just return false for any client, that already has a security context?
> 
> I think so, but I couldn't think of a good way to do that. The `ClientData` associated with the client is managed by the compositor, not Smithay, so I don't know if Smithay has a good way to test if `Client` has a security context... except by having a method to query that implemented by the compositor.

We could add a trait requirement on `ClientData` to the implementation, so that we could query the security context. Though I get how that might be a bit annoying.

> If Smithay doesn't use the security context for anything except excluding clients with a security context from using this protocol, the filter function already handles that so it's a bit redundant to add anything else.

Smithay itself wouldn't do anything else, but I think it is a good idea to allow downstream to add additional filters.

ids1024:
I don't think there's anywhere we could a dd a trait bound for the `ClientData` since that uses dynamic dispatch, and `insert_client` is called by the compositor. And yeah, that would be a bit annoying and add some complexity if it were possible.

Drakulix:
Yeah, I remember [creative workarounds](https://smithay.github.io/smithay/smithay/wayland/compositor/trait.CompositorHandler.html#tymethod.client_compositor_state) to this I wrote in the past. They wouldn't really add any more safety over this approach.

ids1024:
Yeah, I thought about something like that. Though you couldn't even call `client_compositor_state` in a `can_view` function since it is only passed the `client` and `global_data`, not the compositor's state.

And if it's only going to be used to test `is_some`, that's basically just a slightly more complicated filter function.

Drakulix:
Right. I am gonna merge this as is, once the CI-PR has been sorted out.

PR #1108, opened 2023-08-30
Fix clippy lints on Rust 1.72

ids1024:
Should fix CI build.

I'm not sure about the use of `Arc` in `MemoryRenderBuffer`. It seems `MemoryRenderBufferInner` isn't `Send` because `Box<dyn Any>>` isn't. `Send` can be added there if a `Send` bound is added on `TextureId`. But then `GlesTexture` isn't `Send` so that presumably isn't desired in general.

### DIFF
--- a/src/backend/vulkan/mod.rs
+++ b/src/backend/vulkan/mod.rs
@@ -335,6 +335,7 @@ impl Instance {
         info!("Created new instance");
         info!("Enabled instance extensions: {:?}", inner.enabled_extensions);
 
+        #[allow(clippy::arc_with_non_send_sync)]
         Ok(Instance(Arc::new(inner)))
     }
 
@@ -383,10 +384,6 @@ impl Instance {
         &self.0.instance
     }
 }
-
-// SAFETY: Destruction is externally synchronized (using an internal Arc).
-unsafe impl Send for Instance {}
-unsafe impl Sync for Instance {}
 
 /// A Vulkan physical device.
 ///
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -443,7 +443,7 @@ jobs:
       - name: Build Documentation
         env: 
           RUSTDOCFLAGS: --cfg=docsrs
-        run: cargo doc --no-deps --features "test_all_features" -p smithay -p calloop:0.10.6 -p drm -p gbm -p input -p nix:0.26.2 -p udev -p slog -p wayland-server -p wayland-backend -p wayland-protocols:0.30.1 -p winit -p x11rb
+        run: cargo doc --no-deps --features "test_all_features" -p smithay -p calloop:0.10.6 -p drm -p gbm -p input -p nix:0.26.4 -p udev -p slog -p wayland-server -p wayland-backend -p wayland-protocols:0.30.1 -p winit -p x11rb
         
       - name: Setup index
         run: cp ./doc_index.html ./target/doc/index.html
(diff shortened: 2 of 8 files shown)


Drakulix:
> It seems MemoryRenderBufferInner isn't Send because Box<dyn Any>> isn't. Send can be added there if a Send bound is added on TextureId. But then GlesTexture isn't Send so that presumably isn't desired in general.

Good point, given this is a trait object (and has to be), we also can't implement `Send` if the texture is `Send`, so it looks like anything dealing with textures in this way is currently not `Send` anyway. So we could just use an `Rc` here. @cmeissl what do you think of this lint? I guess it would make sense to convert it?

cmeissl:
> > It seems MemoryRenderBufferInner isn't Send because Box> isn't. Send can be added there if a Send bound is added on TextureId. But then GlesTexture isn't Send so that presumably isn't desired in general.
> 
> Good point, given this is a trait object (and has to be), we also can't implement `Send` if the texture is `Send`, so it looks like anything dealing with textures in this way is currently not `Send` anyway. So we could just use an `Rc` here. @cmeissl what do you think of this lint? I guess it would make sense to convert it?

Okay, so the only reason I see why we need the inner buffer (and be able to clone it) is to access the data in the render element. I think we can get rid of that completely by removing the memory buffer render element and (re-)using the texture render element. IIRC we did something similar in the wayland render element.

But replacing the `Arc` with `Rc` will also work for now and is fine by me.
(I have to continue working on the single pixel buffer PR, this already replaces the wayland render element with a texture render element. So I could extend that and apply the same to the other render elements?)

src/backend/vulkan/mod.rs
```diff
@@ -335,6 +335,7 @@ impl Instance {
         info!("Created new instance");
         info!("Enabled instance extensions: {:?}", inner.enabled_extensions);
 
+        #[allow(clippy::arc_with_non_send_sync)]
```
i509VCB, line 338: I think this raises a question, do we want to mark the private inner types as send and Sync or add a bunch of allow attributes for these types of cases?

ids1024: I think in this case it would be reasonable to implement `Send` and `Sync` for the inner type. For the use in `EGLBufferReader::new` we can't do that since it's a `*mut wl_display` that it reference counts. (Though that could possibly be done in a different way.)

ids1024:
```rust
 // SAFETY (Host Synchronization): InstanceInner is always stored in an Arc, therefore destruction is
 // synchronized (since the inner value of an Arc is always dropped on a single thread).
```

```rust
// SAFETY: Destruction is externally synchronized (using an internal Arc).
unsafe impl Send for Instance {}
unsafe impl Sync for Instance {}
```

I think this is safe, though it isn't really related to the fact an `Arc` is used. Without `Arc` you'd still only be able to destroy the object once, from one thread.

i509VCB: Dropping an InstanceInner would be safe since it assumes it owns the underlying ash::Instance. Plus if it's Send and Sync the call to vkDestroyInstance is still externally synchronized (by the borrow checker). And ash makes that function unsafe, so you'd need to know the semantics of InstanceInner to not cause undefined behavior.

I think we can mark InstanceInner as Send + Sync and then remove the manual Send + Sync impls on Instance.

ids1024:
Okay, changed to ` Rc<RefCell<MemoryRenderBufferInner>>` if it's not going to be `Send` or `Sync` anyway, and updated `InstanceInner` to be `Send + Sync`.

Drakulix (approved):
Looks good to me, lets get CI going again.

PR #1110, opened 2023-09-01
input: Only forward modifiers in KeyboardInnerHandle::set_focus

### LINKED ISSUE #1107: KeyboardInnerHandle::set_focus transfers pressed keys to newly focused client
In my compositor, I have a global keyboard shortcut, `logo + tab`, that cycles the focused window.

When using this shortcut, the newly focused window appears to repeat the keys I had pressed. This also happens with other keyboard shortcuts that change windows such as `logo + 1..9`.

When focusing a window, I call `KeyboardHandle::set_focus`, which calls `KeyboardInnerHandle::set_focus`, [which appears to transfer all pressed keys and modifiers to the new client.](https://github.com/Smithay/smithay/blob/master/src/input/keyboard/mod.rs#L750-L764)

Passing `Vec::new()` for keys fixes my issue. [I don't see a similar global keyboard shortcut in Anvil](https://github.com/Smithay/smithay/blob/master/anvil/src/input_handler.rs#L960-L978), so maybe this was missed? I don't think this is intended behavior.

luveti:
This should fix #1107.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -26,23 +26,6 @@ pub use modifiers_state::ModifiersState;
 mod xkb_config;
 pub use xkb_config::XkbConfig;
 
-const MODIFIER_KEYSYMS: [Keysym; 14] = [
-    xkb::KEY_Shift_L,
-    xkb::KEY_Shift_R,
-    xkb::KEY_Control_L,
-    xkb::KEY_Control_R,
-    xkb::KEY_Caps_Lock,
-    xkb::KEY_Shift_Lock,
-    xkb::KEY_Meta_L,
-    xkb::KEY_Meta_R,
-    xkb::KEY_Alt_L,
-    xkb::KEY_Alt_R,
-    xkb::KEY_Super_L,
-    xkb::KEY_Super_R,
-    xkb::KEY_Hyper_L,
-    xkb::KEY_Hyper_R,
-];
-
 /// Trait representing object that can receive keyboard interactions
 pub trait KeyboardTarget<D>: IsAlive + PartialEq + Clone + fmt::Debug + Send
 where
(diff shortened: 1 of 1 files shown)


src/input/keyboard/mod.rs
```diff
@@ -751,14 +768,16 @@ impl<'a, D: SeatHandler + 'static> KeyboardInnerHandle<'a, D> {
...
+                    .filter_map(|keycode| {
+                        MODIFIER_KEYSYMS.contains(keycode).then(|| {
+                            KeysymHandle {
+                                // Offset the keycode by 8, as the evdev XKB rules reflect X's
+                                // broken keycode system, which starts at 8.
+                                keycode: keycode + 8,
+                                state: &self.inner.state,
+                                keymap: &self.inner.keymap,
+                            }
+                        })
```
Drakulix, line 780: Have you tested this? I am not sure we actually need to forward any pressed keys. The `focus.modifiers` event should be enough to update modifiers on the client-side, I believe.

luveti: I have! I've been using it with just passing a `Vec::new()` to the `focus.enter` call and it has been working fine that way as well.

I can update my PR to just do that instead if you'd prefer.

Drakulix: Yes please, then we don't need that list of modifier keysyms, which I would prefer. :)

luveti: Agreed! I searched through xkb for a function that would tell me if a Keysym is a modifier but didn't find anything during a quick search.

Drakulix (approved):
Looks good! Thanks

cmeissl (approved):
LGTM

ids1024:
How does this compare to behavior in other compositors? `wl_keyboard::enter` takes an array of pressed keys, so the previously implementation of this *looks* like how the Wayland protocol is intended to implement.

Drakulix:
> How does this compare to behavior in other compositors? `wl_keyboard::enter` takes an array of pressed keys, so the previously implementation of this _looks_ like how the Wayland protocol is intended to implement.

Hmm sway does seem to indeed send the pressed keys: https://github.com/swaywm/sway/blob/89f85312687b7fd1777c936e169e6400cee0a4b4/sway/input/seat.c#L164-L166

Drakulix:
So I was wondering why I don't observe this behaviour in cosmic-comp and it seems to come do to, at which point you trigger the shortcut. cosmic-comp triggers them on key-release, which means the key is removed from `pressed_keys` before the focus changes. There is obviously value in allowing this to work on key-pressed as well, so if we don't want to always send an empty set of keys, we need to at least honour the compositors which to intercept a key and consequentially remove the key from pressed_keys again.

luveti:
@ids1024 @Drakulix I observed that XFCE (X11) will maintain pressed keys when switching windows. So I've implemented the [second option](https://github.com/Smithay/smithay/issues/1107#issuecomment-1698957006) and it appears to work well.

Drakulix (approved):
Now this is only a bugfix, not forwarding any intercepted key presses. I don't think this is in any way controversial.

Should anyone want to discuss over semantics of key presses on focus swaps, they can open a new issue. This doesn't change behaviour, so lets move on with this PR.

Thanks @luveti for working on this! :)

PR #1112, opened 2023-09-01
xwm: Add XwmHandler::cursor_changed

luveti:
Related issue is #1113.

An example of how this can be used is:
```rs
fn cursor_changed(&mut self, _xwm: XwmId, serial: u32) {
    self.active_cursor = Some(serial);
    if self.cursors.contains_key(&serial) {
        return;
    }
    let (width, height, xhot, yhot, data) =
        self.xwm.as_ref().unwrap().get_cursor_image().unwrap();
    let data: Vec<u8> = data.iter().flat_map(|val| val.to_le_bytes()).collect();
    let texture = self
        .renderer
        .import_memory(
            data.as_slice(),
            Fourcc::Argb8888,
            (width as i32, height as i32).into(),
            false,
        )
        .unwrap();
    self.cursors
        .insert(serial, (xhot as f64, yhot as f64, texture));
}
```

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -88,7 +88,7 @@ use x11rb::{
     protocol::{
         composite::{ConnectionExt as _, Redirect},
         render::{ConnectionExt as _, CreatePictureAux, PictureWrapper},
-        xfixes::{ConnectionExt as _, SelectionEventMask},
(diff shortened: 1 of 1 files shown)


Drakulix:
And why do you need this? Xwayland correctly sets a wl_surface for any set cursor and forwards the image that way.

luveti:
I'm closing this. There is a reply in the related issue https://github.com/Smithay/smithay/issues/1113#issuecomment-1703211475.

PR #1114, opened 2023-09-02
Touchpad gestures

YaLTeR:
Implements zwp_pointer_gestures_v1 and the entire Smithay support for it.

The Wayland gesture objects store whether they are currently active (i.e. between begin() and end()) on a given `WlSurface`. When receiving gesture update or end event, this field is checked, and if it doesn't match, then the gesture is cancelled. This way, gesture objects bound mid-gesture, or something equally weird happening, won't receive update() with no begin().

You could do something smarter, e.g. mutter will synthesize a gesture begin event if you move the pointer onto a surface mid-gesture, but doing this sounds quite more complex (need to track ongoing gestures on the pointer), and I don't think this usecase is very realistic (for this to happen you have to do something along move the pointer with a mouse while also making a gesture on the touchpad).

Anyway, here's a demo on my compositor, but it works in Anvil too, even on Xwayland windows.

https://github.com/Smithay/smithay/assets/1794388/29dae2c4-2dc9-48e3-b592-a357618b9f17

Here you can see the pinch zoom/rotate gesture visually in gtk4-demo, then the swipe gesture only in `WAYLAND_DEBUG` on the right, then the hold gesture by stopping the kinetic scrolling by putting a finger on the touchpad.

By default an Anvil build won't have the hold gestures, to enable them build with the `smithay/libinput_1_19` feature.

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -50,12 +50,17 @@ use crate::state::Backend;
 use smithay::{
     backend::{
         input::{
-            Device, DeviceCapability, PointerMotionEvent, ProximityState, TabletToolButtonEvent,
+            Device, DeviceCapability, GestureBeginEvent, GestureEndEvent, GesturePinchUpdateEvent as _,
+            GestureSwipeUpdateEvent as _, PointerMotionEvent, ProximityState, TabletToolButtonEvent,
             TabletToolEvent, TabletToolProximityEvent, TabletToolTipEvent, TabletToolTipState,
         },
         session::Session,
     },
-    input::pointer::RelativeMotionEvent,
+    input::pointer::{
+        GestureHoldBeginEvent, GestureHoldEndEvent, GesturePinchBeginEvent, GesturePinchEndEvent,
+        GesturePinchUpdateEvent, GestureSwipeBeginEvent, GestureSwipeEndEvent, GestureSwipeUpdateEvent,
+        RelativeMotionEvent,
+    },
     wayland::{
         seat::WaylandFocus,
         tablet_manager::{TabletDescriptor, TabletSeatTrait},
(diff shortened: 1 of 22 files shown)


anvil/src/input_handler.rs
```diff
@@ -927,6 +939,108 @@ impl AnvilState<UdevData> {
...
+        );
+    }
+
+    fn on_gesture_pinch_update<B: InputBackend>(&mut self, evt: B::GesturePinchUpdateEvent) {
+        let pointer = self.pointer.clone();
+        pointer.gesture_pinch_update(
+            self,
+            &GesturePinchUpdateEvent {
+                time: evt.time_msec(),
+                delta: smithay::backend::input::GesturePinchUpdateEvent::delta(&evt),
```
YaLTeR, line 998: This is a bit ugly due to the name collision, any suggestions?

Drakulix: Just rename them on import? :/
Either the libinput ones to `LibinputGesturePinchUpdateEvent` or smithay's to `SmithayGesturePinchUpdateEvent`? Maybe you have ideas for better names. :sweat_smile:

Drakulix (approved):
Thanks for doing the work!

Lots and lots of plumbing code, but nothing problematic. I didn't even realize libinput just gave us gesture events (too long since I wrote the bindings...), so thanks for wiring that up!

I'll wait for a resolution on the name-conflict and possibly merge the clippy PR first so we get working CI, but otherwise this looks good for merge to me.

anvil/src/input_handler.rs
YaLTeR: Oh, right. This is just for trait methods, they don't even have to be named. `use X as _` is sufficient.

YaLTeR:
Oh right, Smallvil.

Drakulix:
> Oh right, Smallvil.

I think it is fine to leave pointer gestures out of smallvil (cc @PolyMeilex).
So just stubs with comments should be enough to fix that.

YaLTeR:
Oh, and I forgot to write docs

YaLTeR:
finally

PR #1120, opened 2023-09-05
Input method popup

rano-oss:
Continuation of #1066. 

- [x] Improved code for entering and leaving surfaces
- [x] Removed race condition when deactivating/activating input methods moving between surfaces
- [x] Removed bug where IME clients could send multiple commits before text input client is ready (the IME client should wait for a done signal from the text input).
- [x] Fixed bug where popup surface would have incorrect initial placement
- [x] Todos (not specifically sure what the ones remaining are about, I might need some help here)
- [x] Surface placement bug, see the image (I need help understanding how to debug/fix this)
image


src/desktop/wayland/popup/mod.rs
```diff
@@ -64,23 +71,31 @@ impl PopupKind {
...
+                        .unwrap()
+                        .lock()
+                        .unwrap()
+                        .current
+                        .geometry
+                })
+                .loc
+            }
+            // TODO
+            PopupKind::InputMethod(ref t) => t.rectangle().loc.to_logical(1), // XXX
```
i509VCB, line 97: I'd probably question the to_logical here with the scale of 1

wlcs_anvil/src/main_loop.rs
```diff
@@ -95,23 +95,6 @@ pub fn run(channel: Channel<WlcsEvent>) {
...
-            let position = Point::from((
-                rectangle.loc.x + rectangle.size.w,
-                rectangle.loc.y + rectangle.size.h,
-            ));
-            input_method.with_surface(|surface| {
-                elements.extend(AsRenderElements::<DummyRenderer>::render_elements(
-                    &smithay::desktop::space::SurfaceTree::from_surface(surface),
-                    &mut renderer,
-                    position.to_physical_precise_round(scale),
-                    scale,
```
i509VCB, line 110: The scale was considered here. I wonder if that's why the size is wrong now?

rano-oss: Thanks, that gives me some ideas about where to start debugging this.

src/desktop/wayland/popup/mod.rs
ids1024: Yeah, this was a placeholder in my original PR, hence *both* `TODO` and `XXX` comments. :laughing:

I'm not sure exactly how it should be handled. It will need to be tested at a couple DPI scales.

rano-oss: Should I get the scale from the compositor state using a trait?

ids1024:
Even with the `parent_geometry` logic here on the latest commit, in Anvil I'm not seeing IMEs positioned relative to the parent window.

Other than popup positioning, it seems to be working pretty well now switching between a couple windows using the IME.

src/desktop/wayland/popup/mod.rs
Drakulix: The scale depends on the output the popup is placed upon. Since you don't know the placement, you need to let the compositor provide it during rendering.

ids1024:
Ah, so I guess `gometry()` should return the equivalent of https://wayland.app/protocols/xdg-shell#xdg_surface:request:set_window_geometry, which here should be a position of `(0, 0)` and the size of the attached buffer, I guess? (There's no other size communicated for the popup.)

While location should be relative to the parent window, and the position of the parent window *should* already be handled in `AsRenderElements` without changes here...

ids1024:
Oh, it looks like I forget `GTK_IM_MODULE=wayland` is needed to test this with GTK...

With that set, closing the IME popup and re-opening it doesn't seem to work as expected.

src/desktop/wayland/popup/mod.rs
rano-oss: I added a suggestion to solve this now, not sure if I can do it this way, but maybe it can be this simple?

rano-oss:
@ids1024 I believe this pr does what it is intended to do, I will continue to test more with fcitx5 and see if I can iron out the bugs. Changes might come in this pr or in a new one.

rano-oss:
image

Seems to work fine for me, and alacritty only uses text_input_v3 so it definitely works for that. Just some odd behavior where text will move if not committed, but I assume that is a fcitx5 problem.

ids1024:
Okay, that seems to match what I see using fcitx5 from git with alacritty. It seems older versions have some issues here.

With gedit (using `GTK_IM_MODULE=wayland`), popup positioning seems to sometimes be correct? But it's not consistent. Not sure exactly what's going on.

ids1024:
> I also did a println! sanity test to make sure fcitx runs through the protocol, and it does

For reference, I'd run `fcitx5` with `WAYLAND_DEBUG=1`. Then it prints every wayland message it sends or receives from the server.

src/wayland/input_method/input_method_handle.rs
```diff
@@ -69,39 +70,15 @@ impl InputMethodHandle {
...
-        let inner = self.inner.lock().unwrap();
-        let popup = inner.popup.inner.lock().unwrap();
-        if popup.surface_role.is_some() {
-            if let Some(surface) = &popup.surface {
-                if surface.alive() {
-                    f(surface);
-                }
-            }
-        }
+    pub(crate) fn set_text_input_rectangle(&self, x: i32, y: i32, width: i32, height: i32) {
```
Drakulix, line 73: Could we at some point in the chain (this function or `set_rectangle` not sure where the best point is) convert this into a `Rectangle` type please?

Drakulix (commented):
Code-wise definitely looks a lot cleaner than the old implementation.

src/wayland/input_method/input_method_handle.rs
rano-oss: I did in the next point of the chain, but I have now moved it forward, also trimmed some code on the way so all is good :rocket:

ids1024:
Functionally, this mostly works with a recent build of fcitx5, at least, which is certainly an improvement over the existing version.

It's hard to say if the issues with gedit are issues here, in GTK, in fcitx5, or in the design of the protocol itself.

So that's worth more investigation, but not really a blocker to merging this, I'd say.

rano-oss:
Just realized I had a function I didn't use, I guess that might have something to do with the movement of the square

Drakulix (approved):
Alright, lets go ahead with this then.

PR #1122, opened 2023-09-05
allow to specify a kind for an element

cmeissl:
this can give the backend a hint about how to handle the element.
for now the only supported use-case is to not accidentally assign non cursor content to the cursor plane

In the future this could be extended to reflect `wp_content_type`.

Only a Draft to discuss alternatives.

### Scan-out priority

One alternative would be to introduce some kind of `ScanoutPriority` which could be used to limit the planes
an element can or should be placed on. But this is way more complex, not only for implementation, but also from
the design perspective. Maybe the two things could be considered orthogonal? The `ScanoutPriority` could look roughly like this:

```rs
pub enum ScanoutPriority {
  Undesired,
  Undefined,
  Priorize(usize)
}
```

### Seperate paramter

Another alternative I considered is to explicitly pass the cursor element(s) to `DrmCompositor::render_frame`.
But this not only breaks the current API, it also turned out to create a mess with multiple backends.

### DIFF
--- a/anvil/src/drawing.rs
+++ b/anvil/src/drawing.rs
@@ -5,7 +5,7 @@ use smithay::{
         element::{
             surface::WaylandSurfaceRenderElement,
             texture::{TextureBuffer, TextureRenderElement},
-            AsRenderElements,
+            AsRenderElements, Kind,
         },
         ImportAll, Renderer, Texture,
     },
@@ -93,6 +93,7 @@ where
                             None,
                             None,
                             None,
+                            Kind::Cursor,
                         ))
                         .into(),
                     ]
@@ -103,7 +104,12 @@ where
             CursorImageStatus::Surface(surface) => {
                 let elements: Vec<PointerRenderElement<R>> =
                     smithay::backend::renderer::element::surface::render_elements_from_surface_tree(
-                        renderer, surface, location, scale, alpha,
+                        renderer,
+                        surface,
+                        location,
+                        scale,
+                        alpha,
+                        Kind::Cursor,
                     );
                 elements.into_iter().map(E::from).collect()
             }
--- a/src/backend/renderer/element/mod.rs
+++ b/src/backend/renderer/element/mod.rs
@@ -317,6 +317,22 @@ impl RenderElementStates {
     }
 }
 
+/// Defines the kind of an [`Element`]
+///
+/// This can give the backend a hint about how to handle the element
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Default)]
+pub enum Kind {
+    /// The element represents a cursor
+    ///
+    /// An element representing a cursor is expected to change infrequently.
+    /// Not marking a cursor element as `Cursor` may result in lower performance and increased power usage.
+    /// In contrast, marking elements that change frequently as `Cursor` can degrade performance significantly.
+    Cursor,
+    /// The element kind is unspecified
+    #[default]
+    Unspecified,
+}
+
 /// A single element
 pub trait Element {
     /// Get the unique id of this element
@@ -355,6 +371,10 @@ pub trait Element {
     /// already encoded alpha in it's underlying representation.
     fn alpha(&self) -> f32 {
         1.0
+    }
+    /// Returns the [`Kind`] for this element
+    fn kind(&self) -> Kind {
+        Kind::default()
     }
 }
 
@@ -435,6 +455,10 @@ where
 
     fn alpha(&self) -> f32 {
         (*self).alpha()
+    }
+
+    fn kind(&self) -> Kind {
+        (*self).kind()
     }
 }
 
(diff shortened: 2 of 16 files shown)


Drakulix (commented):
Oh I like that idea, that also means we could preferably put wp_content_type `video` surfaces on overlay/underlay planes in the future, which should benefit from that the most.

I think this solution is strictly better than the last proposed one.
Scanout priority might also be useful, but can absolutely be done later and additionally.

E.g. I imagine that `game` surfaces could  be preferred on the primary plane and everything else gets composed on an overlay plane, if possible. Priorities could make the algorithm consider other planes for compositing. But I think that approach would still be orthogonal. 

You simply don't want content that is often changing on the cursor plane, which could still happen (e.g. if the cursor is hidden) with *just* priorities. I think we need something *like* `kind` anyway.

src/backend/renderer/element/mod.rs
```diff
@@ -317,6 +317,19 @@ impl RenderElementStates {
     }
 }
 
+/// Defines the kind of an [`Element`]
+///
+/// This can give the backend a hint about how
+/// to handle the element
```
YaLTeR, line 323: It would be good to document here what exactly this affects (cursor plane assignment) and why it matters (animating elements will cause performance problems). Because right now you might assume that it's a good idea to set the DnD surface's kind to cursor, but the DnD surface can very well animate. For example, in one of my apps the DnD surface can show a playing video preview:

https://github.com/Smithay/smithay/assets/1794388/f77298b9-f4a4-4bcd-bdbf-be6865c384d2

Also libadwaita's tab DnD surface will animate its size occasionally.

anvil/src/drawing.rs
```diff
@@ -103,7 +104,12 @@ where
             CursorImageStatus::Surface(surface) => {
                 let elements: Vec<PointerRenderElement<R>> =
                     smithay::backend::renderer::element::surface::render_elements_from_surface_tree(
-                        renderer, surface, location, scale, alpha,
+                        renderer,
+                        surface,
+                        location,
+                        scale,
+                        alpha,
+                        Some(Kind::Cursor),
```
YaLTeR, line 112: I wonder if maybe it's a good idea to set kind separately so it's not spammed into every invocation of these functions. Since it's something you set on individual elements generally.

src/backend/renderer/element/surface.rs
```diff
@@ -235,6 +235,7 @@ pub fn render_elements_from_surface_tree<R, E>(
     location: impl Into<Point<i32, Physical>>,
     scale: impl Into<Scale<f64>>,
     alpha: f32,
+    kind: Option<Kind>,
```
YaLTeR, line 238: Why Option<> here?

YaLTeR:
Comments aside, this seems to solve my issue with undesirable surfaces getting cursor planed :+1:

src/backend/renderer/element/mod.rs
cmeissl: Good point, I will clarify what this `Kind` is intended for. I want to keep it generically and not depend on the `DrmCompositor` internals, so would something like:
> An element representing a cursor is expected to change infrequently. Failing to mark cursor elements can result in sub-optimal performance. In contrast, marking elements that change frequently as `Cursor` can degrade performance significantly. `

src/backend/renderer/element/surface.rs
cmeissl: The idea behind making this an `Option` was in preparation of `wp_content_type`. On `None` it could map the content type defined by the client to a `Kind`. So providing `Some` would override the client defined content type.
But that might not be desired. Not sure what the best semantics are for this parameter, we could make it a fallback I guess. So if the surface does not specify a content type it would fallback to the provided `Kind`.

anvil/src/drawing.rs
cmeissl: Currently no render element has a public setter, everything is done a construction. Setting it after construction would require a mutable reference, not sure if this would make a good API. It would work to create a "overload" function, something like `render_element_from_surface_tree_with_kind`, but tbh I am not sure if this would be an improvement.

src/backend/renderer/element/mod.rs
YaLTeR: Sounds good, perhaps only reword this a little bit:

> Failing to mark cursor elements can result in sub-optimal performance.

Maybe

> Not marking a cursor element as `Cursor` may result in lower performance and increased power usage.

src/backend/renderer/element/surface.rs
cmeissl: @YaLTeR I changed it to not use an `Option` for now. When integrating `wp_content_type` we can decide how to handle that. We could still test for `Kind::Unspecified` and in this case use whatever content type the client defined for the surface if any.

src/backend/renderer/element/mod.rs

Drakulix (approved):

PR #1123, opened 2023-09-06
Take Wayland buffer offset into account

YaLTeR:
It was ignored until now, breaking offset DnD surfaces for example.

I'm not sure if it interacts correctly with viewporter and input region; the protocol is not entirely clear on this. But I'm also not sure after reading the protocol that the existing code handles it correctly. So far nothing seems to have broken.

A good test for this fix is GTK tests/testdndresize; alternatively, dragging images and videos out of [Loupe](https://flathub.org/apps/org.gnome.Loupe) or [Identity](https://flathub.org/apps/org.gnome.gitlab.YaLTeR.Identity).

### DIFF
--- a/src/backend/renderer/utils/wayland.rs
+++ b/src/backend/renderer/utils/wayland.rs
@@ -126,7 +126,8 @@ impl RendererSurfaceState {
                     .buffer_dimensions
                     .unwrap()
                     .to_logical(self.buffer_scale, self.buffer_transform);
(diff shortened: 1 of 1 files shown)


Drakulix (approved):

cmeissl:
This seems to have broken cursor input for at least `weston-terminal` in anvil. Entering and exiting `weston-terminal` will each time further offset the cursor making interaction with the window impossible at some time.

PR #1128, opened 2023-09-10
Perf/various optimizations

cmeissl:
WIP: Some experiments for getting the render time further down.

Based on #1122 

Quick compare with 2 outputs, one empty the other with glmark:
Screenshot from 2023-09-14 20-58-09

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ xkbcommon = { version = "0.5.0", features = ["wayland"]}
 scan_fmt = { version = "0.2.3", default-features = false }
 encoding = { version = "0.2.33", optional = true }
 profiling = "1.0"
+smallvec = "1.11"
 
 [dev-dependencies]
 clap = { version = "4", features = ["derive"] }
--- a/anvil/Cargo.toml
+++ b/anvil/Cargo.toml
@@ -59,4 +59,5 @@ x11 = ["smithay/backend_x11", "x11rb", "smithay/renderer_gl", "smithay/backend_v
 xwayland = ["smithay/xwayland", "x11rb", "smithay/x11rb_event_source", "xcursor"]
 profile-with-puffin = ["profiling/profile-with-puffin", "puffin_http"]
 profile-with-tracy = ["profiling/profile-with-tracy"]
+profile-with-tracy-mem = ["profile-with-tracy"]
 renderer_sync = []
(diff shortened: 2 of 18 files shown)


Drakulix (commented):
I wish this was split into more commits, because I feel like we are getting into muddy waters here.

A couple of things are obviously a good idea.
That includes making the damage snapshot copy-on-write (already neatly separated) or adding `has_dmabuf_format` (not so separated).

Others need more testing (preferrably in different compositors with different workloads) to convince me. E.g. cosmic-comp probably has a lot more small RenderElements, than anvil for various UI elements.
- Using `Vec`s in places of `Map`s. I can see how some of these only have a very small amount of elements, which might make this more efficient.
- Using Fxhash instead of std' HashMaps. I haven't found any good characteristics on when fxhash might be faster and why this applies here.

I also would like to point out, that this depends on the rust-version used, as stdlib collections are still frequently optimized. So there might be reason to use these optimizations, if you are stuck with older versions, and less reason with newer ones. Profiling runs should thus also be tagged with the rust-version used to build.

cmeissl:
> 

Sure, as always I will split it into mutliple commits. I want to be able to profile each change independently.
I expect that after splitting we can merge some obvious optimizations directly, while others may lead to nothing and may be dropped. 

Candidates for merge are:
- DamageSnasphot COW
- has_dmabuf_format
- rectangle subtract many (after more testing)
- render_output_with (this can save quite some time, bind/unbind is rather expensive)

Possible micro-optimizations:
- replacing a few hash maps with vec (only where we can assume a small set, which is true for the planes)

Lower priority:
- The FxHash stuff. This definitely needs a bigger test-set. But we use HashMaps/Sets in quite a few places, so it is something I definitely want to explore.

cmeissl:
Okay, so everything should be nicely split up in smallish commits now.
I also removed the alternative hashing for now.

Drakulix (approved):
Every single optimization seems very sensible to me and *should* be an obvious improvement.

Lets wait for a few test runs, but I am pretty certain we can merge this as is.

YaLTeR:
Screenshot from 2023-09-15 22-10-51

Yellow = this PR, red = master.

Not an exact benchmark here; I opened a few alacritties and weston-presentation-shm, then switched to Firefox and scrolled that around a bit.

cmeissl:
Okay, also did another trace:

image

So imo this is good to get merged

PR #1129, opened 2023-09-11
shm/pool: Introduce a thread for dropping

YaLTeR:
Visual demonstration of Pool drops stalling the main thread:

Screenshot from 2023-09-11 08-06-31

Real stalled vblank events before this PR:

image

Real stalled vblank events after this PR:

image

(yes, there's only one, and it's unrelated)

### DIFF
--- a/src/wayland/shm/pool.rs
+++ b/src/wayland/shm/pool.rs
@@ -5,7 +5,11 @@ use std::{
     num::NonZeroUsize,
     os::unix::io::{AsRawFd, OwnedFd, RawFd},
     ptr,
-    sync::{Once, RwLock},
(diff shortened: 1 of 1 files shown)


YaLTeR:
Uhh, so it seems that Sender got Sync only in Rust 1.72.0. What should I do here?

i509VCB:
> Uhh, so it seems that Sender got Sync only in Rust 1.72.0. What should I do here?

SyncSender?

Drakulix (approved):

ids1024:
Perhaps io-uring would be good for this and many other things; but then we'd still want a fallback for when that isn't available...

PR #1130, opened 2023-09-11
xdg-shell: Update xdg shell to version 6

PolyMeilex:


Drakulix (approved):
LGTM, but clippy seems to disagree

PR #1132, opened 2023-09-13
drm leasing

Drakulix:
TODO:

- [x] Do a drm-rs release
- [x] Make the handler withdraw/re-annouce successfully leased/released connectors internally, so downstream doesn't need to track and call `withdraw_connector` and `add_connector` respectively.
- [x] Add suspend/resume to temporarily revoke all leases/advertised connectors while session is inactive / master was lost
- [x] Implement in anvil
- [x] Actually test the damn thing

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -52,7 +52,7 @@ use smithay::{
         vulkan::{version::Version, Instance, PhysicalDevice},
         SwapBuffersError,
     },
-    delegate_dmabuf,
+    delegate_dmabuf, delegate_drm_lease,
     desktop::{
         space::{Space, SurfaceTree},
         utils::OutputPresentationFeedback,
@@ -67,8 +67,8 @@ use smithay::{
         },
         drm::{
             self,
-            control::{connector, crtc, ModeTypeFlags},
-            Device,
+            control::{connector, crtc, Device, ModeTypeFlags},
+            Device as _,
         },
         input::Libinput,
         nix::fcntl::OFlag,
@@ -83,6 +83,9 @@ use smithay::{
         compositor,
         dmabuf::{
             DmabufFeedback, DmabufFeedbackBuilder, DmabufGlobal, DmabufHandler, DmabufState, ImportError,
+        },
+        drm_lease::{
+            DrmLease, DrmLeaseBuilder, DrmLeaseHandler, DrmLeaseRequest, DrmLeaseState, LeaseRejected,
         },
         input_method::{InputMethodHandle, InputMethodSeat},
     },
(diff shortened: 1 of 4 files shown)


Drakulix:
Ready for review, just waiting for a drm-rs release. I tested this extensively and tried to cleanup the code as much as possible.

src/backend/allocator/dmabuf.rs
```diff
@@ -397,11 +397,11 @@ impl DmabufSource {
         ];
         for (idx, handle) in dmabuf.handles().enumerate() {
             // SAFETY: This is stored together with the Dmabuf holding the owned file descriptors
-            let fd = handle.as_raw_fd();
+            let fd = handle.as_fd();
             if matches!(
                 poll::poll(
                     &mut [poll::PollFd::new(
-                        fd,
+                        &fd,
```
cmeissl, line 404: `PollFd::new` takes a reference of something implementing `AsFd` and `dmabuf.handles()` return an iterator over `BorrowedFd`. So I guess we could directly pass `handle` here?

src/wayland/drm_lease/mod.rs
```diff
@@ -0,0 +1,966 @@
+//! DRM Lease protocol
+//!
+//! This module provides helpers to handle the wp_drm_lease_v1 protocol,
+//! which allows clients to leese DRM resources from the compositor.
```
cmeissl, line 4:
```suggestion
//! which allows clients to lease DRM resources from the compositor.
```

src/wayland/drm_lease/mod.rs
```diff
@@ -0,0 +1,966 @@
...
+//! This is in particular useful for VR applications, that would like to take over
+//! a directly attached VR display from the DRM device otherwise controlled by the compositor.
+//! In theory however any sorts of DRM resources can be lend out to less privileged clients via this protocol.
+//!
+//! ## How to use
+//!
+//! To setup the drm_lease global, you will need to first provide the `DrmNode` you want to lease resources from.
+//! You can usually get that from your [`DrmDevice`](crate::backend::drm::DrmDevice). Once the global is up,
+//! you can advertise connectors as available through [`DrmLeaseState::add_connector`].
+//! Should a connector become unaviailable or is used by the compositor, you may remove it again using [`DrmLeaseState::withdraw_connector`].
```
cmeissl, line 15:
```suggestion
//! Should a connector become unavailable or is used by the compositor, you may remove it again using [`DrmLeaseState::withdraw_connector`].
```

anvil/src/udev.rs
```diff
@@ -508,6 +518,60 @@ pub fn run_udev() {
...
+            .as_mut()
+            .unwrap()
+    }
+
+    fn lease_request(
+        &mut self,
+        node: DrmNode,
+        request: DrmLeaseRequest,
+    ) -> Result<DrmLeaseBuilder, LeaseRejected> {
+        let backend = self.backend_data.backends.get(&node).unwrap();
```
cmeissl, line 537: Could this result in a race when the device gets removed after the client issues a lease request but before the server processing the request?

cmeissl (approved):
:+1: Just a theoretical question about the use in anvil

anvil/src/udev.rs
Drakulix: Yeah probably. I guess I need to change the `unwrap` into a `return LeaseRejected` for the error case. Nice catch.

PR #1133, opened 2023-09-15
Feature/egl context priority

cmeissl:
This PR adds support for egl context priority using the [EGL_IMG_context_priority](https://registry.khronos.org/EGL/extensions/IMG/EGL_IMG_context_priority.txt) egl extension.

It also changes the multigpu code to request a high priority context by default.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -354,8 +354,11 @@ jobs:
           restore-keys: ${{ runner.os }}-cargo-registry-
 
       - name: Rust toolchain
-        uses: dtolnay/rust-toolchain@nightly
+        # nightly changed no_coverage to coverage(off) which breaks the build
+        # see: https://github.com/rust-lang/rust/pull/114656
+        uses: dtolnay/rust-toolchain@master
         with:
+          toolchain: nightly-2023-06-17
           components: llvm-tools-preview
 
       - name: grcov cache
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -34,7 +34,7 @@ use smithay::{
             compositor::DrmCompositor, CreateDrmNodeError, DrmDevice, DrmDeviceFd, DrmError, DrmEvent,
             DrmEventMetadata, DrmNode, DrmSurface, GbmBufferedSurface, NodeType,
         },
-        egl::{self, EGLDevice, EGLDisplay},
+        egl::{self, context::ContextPriority, EGLDevice, EGLDisplay},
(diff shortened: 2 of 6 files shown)


Drakulix:
Can we make the second commit add the priority as constructor arguments for the `GbmGlesBackend` and `EglGlesBackend` instead?

cmeissl:
> Can we make the second commit add the priority as constructor arguments for the `GbmGlesBackend` and `EglGlesBackend` instead?

Yes, though we only have a default impl and `with_factory`. I can add a new function taking the priority, but what should the default do? An alternative would be to not request any priority by default and use the factory for this. I considered that first, but that forces downstream to use the factory which can be easily forgotten.

Drakulix:
> > Can we make the second commit add the priority as constructor arguments for the `GbmGlesBackend` and `EglGlesBackend` instead?
> 
> Yes, though we only have a default impl and `with_factory`. I can add a new function taking the priority, but what should the default do? An alternative would be to not request any priority by default and use the factory for this. I considered that first, but that forces downstream to use the factory which can be easily forgotten.

I guess we have two options:
- Make `Default` do what it currently does. A context with "medium" or "default" priority. Additionally add a `with_priority`-constructor. The factory-constructor would remain unchanged.
- Drop `Default` imo and force downstream to always specify a priority or at least explicitly pass `None` for "default priority" (if not using a factory).

I am happy with both options.

i509VCB:
I believe vulkan has an extension similar to this, so we can elect to specify a priority from the multigpu code, although it's an optional extension: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_global_priority.html

cmeissl:
> > > Can we make the second commit add the priority as constructor arguments for the `GbmGlesBackend` and `EglGlesBackend` instead?
> > 
> > 
> > Yes, though we only have a default impl and `with_factory`. I can add a new function taking the priority, but what should the default do? An alternative would be to not request any priority by default and use the factory for this. I considered that first, but that forces downstream to use the factory which can be easily forgotten.
> 
> I guess we have two options:
> 
>     * Make `Default` do what it currently does. A context with "medium" or "default" priority. Additionally add a `with_priority`-constructor. The factory-constructor would remain unchanged.
> 
>     * Drop `Default` and force downstream to always specify a priority or at least explicitly pass `None` for "default priority" (if not using a factory).
> 
> 
> I am happy with both options.

I opted for the first option, which really seems more sensible as `Medium` is defined as the default by the spec (and it does not break the API)

Drakulix (approved):

PR #1134, opened 2023-09-17
wip

cmeissl:
PR for showcasing tracy gpu profiling

## TODO
- [ ] make the whole thing optional (maybe abstract gpu profiling and only depend on tracy in anvil?)
- [ ] wait for release of profiling (https://github.com/aclysma/profiling/pull/56)
- [ ] cpu and gpu timing seems to be off (see https://github.com/wolfpld/tracy/issues/17). maybe calibration can help here a bit?

Example:
image


ids1024:
> maybe abstract gpu profiling and only depend on tracy in anvil?

Are there any other profilers one might use with similar GPU profiling instrumentation? If so, a trait abstracting them makes sense. If not, just having an optional tracy dependency might make sense.

Would it be reasonable to have a trait method of `Renderer` that calls `profiler.scope`? (And could also be implemented by a Vulkan renderer, when that exists.) I guess handling `span_location!` and also being agnostic to different profilers is awkward here.

src/backend/renderer/gles/profiler.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        let mut bits = 0;
+        unsafe { gl.GetQueryivEXT(ffi::TIMESTAMP_EXT, ffi::QUERY_COUNTER_BITS_EXT, &mut bits) };
+
+        let mut gpu_timestamp = 0;
+        unsafe { gl.GetInteger64v(ffi::TIMESTAMP_EXT, &mut gpu_timestamp) };
+
+        let context = client
+            .new_gpu_context(
+                Some("GlesRenderer"),
```
ids1024, line 36: When the compositor is used with multiple GPUs, this should probably have different names. Maybe the drm node, though vendor and driver can also be queried through EGL.

src/backend/renderer/gles/profiler.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+struct ExitedGpuTracepoint {
+    span: GpuSpan,
+}
+
+impl Drop for EnteredGpuTracepoint {
+    fn drop(&mut self) {
+        debug_assert!(self.span.is_none());
```
ids1024, line 184: Doing things this way clashes with the use of `?` for error handling; it will panic if `?` prevents it from getting to the `exit` call.

Not sure what to do about this.

YaLTeR:
This can be closed now

PR #1136, opened 2023-09-19
WIP Use euclid crate for geometry types

ids1024:
This does not yet compile.

It seems one difference between the Euclid types and the ones Smithay currently has is that `euclid`'s points cannot be added, and there's a seperate vector type. That is probably better for making sure code dealing with coordinates is correct. Though that will require some changes.


anvil/src/drawing.rs
```diff
@@ -169,7 +169,7 @@ where
         } else {
             3
         };
-        Rectangle::from_loc_and_size((0, 0), (24 * digits, 35)).to_f64()
+        Rectangle::new((0, 0), (24 * digits, 35)).to_f64()
```
i509VCB, line 172: There is actually a `rect` method in euclid to make this even shorter

ids1024: Right. I guess the cases with `origin` `(0, 0)` should just be equivalent to `Rect::from_size`.

ids1024:
Another thing `euclid` doesn't have is functions like `Size::to_point`. But it's probably best to express things in ways other than a cast like that.

Not sure how best to make use of things like `Euclid::Translation2D` and `Euclid::Scale` for conversion between coordinate spaces.

With or without this change, `Logical` should be split into absolute coordinates, space-relative, surface/renderelement-relative.

ids1024:
Another difference is that apparently casts in `euclid` like `to_i32` are panicking if out of range. Unlike what Smithay has, which seems to just be using `as i32`. That's something to be cautious about, though casts like that are probably already problematic if there's any real situation where they can be out of range.

Probably as long as signed types are used, this would only be an issue if a client sent some extreme values.

PR #1140, opened 2023-09-24
Update winit to 0.29.2 version

### LINKED ISSUE #146: winit key repeating
In the winit backend key repeating is enabled, whereas when using libinput no such feature exists.

winit backend should have key repeating disabled to be consistent (or atleast an option to disable it).

kchibisov:
Make better winit integration by providing a calloop's WaylandSource
for it. The winit backend now doesn't use vsync and utilizes frame
callbacks, when the user cooperates and asks for `redraw_requested`.

Fixes #146.

### DIFF
--- a/anvil/src/winit.rs
+++ b/anvil/src/winit.rs
@@ -10,7 +10,7 @@ use smithay::backend::renderer::ImportEgl;
 #[cfg(feature = "debug")]
 use smithay::{
     backend::{allocator::Fourcc, renderer::ImportMem},
-    reexports::winit::platform::wayland::WindowExtWayland,
+    reexports::winit::raw_window_handle::{HasWindowHandle, RawWindowHandle},
 };
 
 use smithay::{
@@ -33,6 +33,7 @@ use smithay::{
         calloop::EventLoop,
         wayland_protocols::wp::presentation_time::server::wp_presentation_feedback,
         wayland_server::{protocol::wl_surface, Display},
+        winit::platform::pump_events::PumpStatus,
     },
     utils::{IsAlive, Scale, Transform},
     wayland::{
@@ -97,7 +98,7 @@ pub fn run_winit() {
             return;
         }
     };
-    let size = backend.window_size().physical_size;
+    let size = backend.window_size();
 
     let mode = Mode {
         size,
@@ -211,27 +212,26 @@ pub fn run_winit() {
     let mut pointer_element = PointerElement::<GlesTexture>::default();
 
     while state.running.load(Ordering::SeqCst) {
-        if winit
-            .dispatch_new_events(|event| match event {
-                WinitEvent::Resized { size, .. } => {
-                    // We only have one output
-                    let output = state.space.outputs().next().unwrap().clone();
-                    state.space.map_output(&output, (0, 0));
-                    let mode = Mode {
-                        size,
-                        refresh: 60_000,
-                    };
-                    output.change_current_state(Some(mode), None, None, None);
-                    output.set_preferred(mode);
-                    crate::shell::fixup_positions(&mut state.space, state.pointer.current_location());
-                }
-                WinitEvent::Input(event) => {
-                    state.process_input_event_windowed(&display_handle, event, OUTPUT_NAME)
-                }
-                _ => (),
-            })
-            .is_err()
-        {
+        let status = winit.dispatch_new_events(|event| match event {
+            WinitEvent::Resized { size, .. } => {
+                // We only have one output
+                let output = state.space.outputs().next().unwrap().clone();
+                state.space.map_output(&output, (0, 0));
+                let mode = Mode {
+                    size,
+                    refresh: 60_000,
+                };
+                output.change_current_state(Some(mode), None, None, None);
+                output.set_preferred(mode);
+                crate::shell::fixup_positions(&mut state.space, state.pointer.current_location());
+            }
+            WinitEvent::Input(event) => {
+                state.process_input_event_windowed(&display_handle, event, OUTPUT_NAME)
+            }
+            _ => (),
+        });
+
+        if let PumpStatus::Exit(_) = status {
             state.running.store(false, Ordering::SeqCst);
             break;
         }
@@ -294,8 +294,15 @@ pub fn run_winit() {
                         backend.renderer().egl_context().get_context_handle(),
                         backend
                             .window()
-                            .wayland_surface()
-                            .unwrap_or_else(std::ptr::null_mut),
+                            .window_handle()
+                            .map(|handle| {
+                                if let RawWindowHandle::Wayland(handle) = handle.as_raw() {
+                                    handle.surface.as_ptr()
+                                } else {
+                                    std::ptr::null_mut()
+                                }
+                            })
+                            .unwrap_or_else(|_| std::ptr::null_mut()),
                     );
                 }
                 let age = if *full_redraw > 0 {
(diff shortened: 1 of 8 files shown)


smallvil/src/winit.rs
```diff
@@ -24,10 +22,15 @@ pub fn init_winit(
     let display_handle = &mut data.display_handle;
     let state = &mut data.state;
 
-    let (mut backend, mut winit) = winit::init()?;
+    let (mut backend, winit_source) = winit::WinitGraphicsBackend::new(GlAttributes {
+        version: (3, 0),
```
i509VCB, line 26: Off topic for this pull request, this seems to have always been 3.0, but the winit backend should be possible even with gles 2.0?

@Drakulix?

examples/minimal.rs
```diff
@@ -161,36 +167,41 @@ pub fn run_winit() -> Result<(), Box<dyn std::error::Error>> {
...
+                WinitEvent::Input(event) => match event {
+                    InputEvent::Keyboard { event } => {
+                        keyboard.input::<(), _>(
+                            &mut state,
+                            event.key_code(),
+                            event.state(),
+                            0.into(),
+                            0,
+                            |_, _, _| {
+                                //
```
i509VCB, line 182: empty comment?

anvil/src/winit.rs
```diff
@@ -212,8 +217,8 @@ pub fn run_winit() {
     let mut pointer_element = PointerElement::<GlesTexture>::default();
 
     while state.running.load(Ordering::SeqCst) {
-        if winit
-            .dispatch_new_events(|event| match event {
+        for event in winit.get_events() {
```
i509VCB, line 220: I think we should move anvil over the an event source to match the udev and x11 backends. Plus the end goal is to have an event source from winit.

examples/minimal.rs
kchibisov: It was there before.

anvil/src/winit.rs
kchibisov: You're free to do anvil yourself.

src/backend/winit/mod.rs
```diff
@@ -308,18 +209,24 @@ where
...
-        // So the order here is important.
-        self.renderer.bind(self.egl.clone())?;
-
-        // Were we told to resize?
-        if let Some(size) = self.resize_notification.take() {
-            self.egl.resize(size.w, size.h, 0, 0);
+        // NOTE we must resize before making the current context current, otherwise the back
+        // buffer will be latched. Some nvidia drivers may not like it, but a lot of wayland
+        // software does the order that way due to mesa latching back buffer on each
+        // `make_current`.
```
Drakulix, line 222: Urgh, so this order is explicitly required for correct behavior in mesa and the other for nvidia? That's stupid.

In that case we need a test for the driver here, because this order crashes the client on nvidia.

kchibisov: Are you sure that it still crashes on nvidia that way? We use the order like that in alacritty for a while and it's not crashing. I think they changed it a while ago to not crash though.

Drakulix: It might be fixed now, but I definitely remember debugging this a whole afternoon. What anvil does differently to most wayland clients is, that it will unbind surface/context again, since we might have multiple. So I am not sure, if that same bug is triggered by most other clients anyway, though I believe you with alacritty.

I will re-test it with this branch and an up-to-date driver, before we touch the code. I would love to be able to keep this order.

src/backend/winit/mod.rs
```diff
@@ -372,234 +281,390 @@ where
...
+                        device: WinitVirtualDevice,
+                    }));
+                }
+                Event::UserEvent(event) => match event {
+                    WinitUserEvent::Shutdown => window_target.exit(),
+                },
+                Event::WindowEvent { event, .. } => match event {
+                    WindowEvent::Resized(size) => {
+                        trace!("Resizing window to {size:?}");
+                        current_size = size;
```
YaLTeR, line 476: If the window is resized without a Resized event (i.e. with request_inner_size()) then this is not updated, messing up the input.

kchibisov: I guess we could just wrap the way you set the `inner_size`, otherwise we'd have to go back to calling `inner_size` quite frequently, which is not that fast.

kchibisov: Alternatively, we could simply wrap the way you get the `&WinitWindow` and provide it via some callback and then will just ensure the size/state of the window after each call to the `&WinitWindow`.

src/backend/winit/mod.rs
kchibisov: @Drakulix were you able to verify whether nvidia works. We're doing basically what smithay does in alacritty without issues. Though, we don't unbind completely.

kchibisov:
I decided to use old API that was already here, but I added a way to integrate into calloop without any thread.

src/backend/winit/mod.rs
```diff
@@ -174,118 +112,129 @@ where
     let _guard = span.enter();
     info!("Initializing a winit backend");
 
-    let events_loop = EventLoop::new();
-    let winit_window = Arc::new(builder.build(&events_loop).map_err(Error::InitFailed)?);
+    let event_loop = EventLoopBuilder::new().build().expect("TODO");
```
YaLTeR, line 115: Seems like a left-over expect().

src/backend/winit/mod.rs
```diff
@@ -174,118 +112,129 @@ where
...
                     )
-                    .map_err(EGLError::CreationFailed)?
-                },
-                true,
-            )
-        } else {
-            unreachable!("No backends for winit other then Wayland and X11 are supported")
+                }
+            }
+            _ => todo!(),
```
YaLTeR, line 170: Needs to be unreachable!(), presumably.

src/backend/winit/mod.rs
```diff
@@ -308,18 +263,24 @@ where
         &mut self.renderer
     }
 
-    /// Bind the underlying window to the underlying renderer
+    /// Bind the underlying window to the underlying renderer.
+    ///
+    /// The `size` argument is used to resize the underlying renderer. When `None` no
```
YaLTeR, line 268: I guess this is left over from the previous iteration.

src/backend/winit/mod.rs
```diff
@@ -401,205 +374,282 @@ impl WinitEventLoop {
...
-                    device: WinitVirtualDevice,
-                }));
-                self.initialized = true;
-            }
-
-            self.events_loop
-                .run_return(move |event, _target, control_flow| match event {
-                    Event::RedrawEventsCleared => {
-                        *control_flow = ControlFlow::Exit;
+        let event_loop = unsafe { self.event_loop.get_mut() };
```
YaLTeR, line 381: Might need a SAFETY comment.

src/backend/winit/mod.rs
Drakulix: I don't have a system running nvidia+x11 anymore, but under wayland at least this seems to be running the same on all my gpus. I call that good enough. If there actually was an issue here, they fixed it or it is x11 exclusive (which is probably broken anyway, because untested).

Drakulix:
Just testing this I can't get anvil to repaint without sending it input events. Works as expected with the x11 interface.

Drakulix (approved):
LGTM. Thanks a bunch

PR #1142, opened 2023-09-25
Use safe IO types in a couple more places

ids1024:
This still leaves `XWaylandEvent::Ready` using a `RawFd`, which is awkward. Not sure exactly what the logic is there, and that field isn't used in anvil or cosmic-comp, but maybe ref-counting could be used? And `RawFd`s are used in `Session`/`LibSeatSession`, but it's not clear how to update that either.


Drakulix (approved):
LGTM

PR #1144, opened 2023-09-26
Make DrmEventMetadata Copy

YaLTeR:
Any reason it wasn't?


Drakulix (approved):
Don't think there is any.

PR #1145, opened 2023-09-26
geometry: Generic Debug implementation

Drakulix:
Allows downstream to implement their own coordinate spaces without sacrificing `fmt::Debug` for it.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -439,36 +439,10 @@ impl<Kind> Point<f64, Kind> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Drakulix:
> LGTM, we just need to keep in mind that the exact return value of `type_name::<S>()` is unspecified, and currently apparently possibly includes the whole module path, not just the identifier name of the type.

Yes, it might look a little ugly in some cases, but I still think this is an overall improvement.

If we want to avoid that, we could make the name an associated const of a new `CoordinateSpace`-trait?

PR #1146, opened 2023-09-26
shm/pool: set trace log level for InnerPool::new instrumentation

nicolasavru:


Drakulix (approved):
Yeah, that is probably a good idea, given the expected frequency this is called in.

PR #1147, opened 2023-09-27
Replace encoding dep with encoding_rs.

nicolasavru:
encoding is unmaintained: https://rustsec.org/advisories/RUSTSEC-2021-0153

### DIFF
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -11,7 +11,7 @@ use crate::{
     },
     utils::{user_data::UserDataMap, IsAlive, Logical, Rectangle, Serial, Size},
 };
-use encoding::{DecoderTrap, Encoding};
+use encoding_rs::WINDOWS_1252;
 use std::{
     collections::HashSet,
     sync::{Arc, Mutex, Weak},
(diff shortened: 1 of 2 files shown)


src/xwayland/xwm/surface.rs
```diff
@@ -553,14 +553,8 @@ impl X11Surface {
...
             Ok(Some(wm_class)) => (
-                encoding::all::ISO_8859_1
-                    .decode(wm_class.class(), DecoderTrap::Replace)
-                    .ok()
-                    .unwrap_or_default(),
-                encoding::all::ISO_8859_1
-                    .decode(wm_class.instance(), DecoderTrap::Replace)
-                    .ok()
-                    .unwrap_or_default(),
+                WINDOWS_1252.decode(wm_class.class()).0.to_string(),
```
i509VCB, line 556: Windows-1252 is an extension of ISO-8859-1 which would technically allow more valid character values? I'm not sure if this is correct?

Although before that discussion happens I can't figure out what encoding the X11 spec or ICCCM wants for WM_CLASS. The ICCCM says it is a `STRING` which doesn't mean seem to mention any encoding. Where as other properties like WM_WINDOW_ROLE explicitly mention ISO-8859-1

@Drakulix might know more about this?

Drakulix: STRING is sadly completely undefined and as a result we don't know when it comes to old apis, as far as I am aware. If we were strict we could probably reject everything that isn't ASCII here, but since most modern apis do define `ISO-8859-1` as the encoding, it makes sense to accept that for everything `STRING`.

Drakulix: > I'm not sure if this is correct?

I am quite certain it isn't, if I don't see a X11 spec proving otherwise. I have never seen this encoding being mentioned anywhere, while working on the Xwayland code.

nicolasavru: https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#target_atoms has STRING as Latin-1, which should be ISO-8859-1. Unless I'm missing something though, we're only reading the data from X, not writing to it, so I think this should be fine.

nicolasavru: The reason I used windows-1252is because encoding_rs doesn't offer a separate iso-8859-1 decoder. If there is concern about X giving smithay bad data, then I think we'd want to use the windows-1251 decoding and then manually filter out the character ranges we don't want.

Drakulix: Yeah I am actually not sure, if there is any concern here. Obviously this data is client controlled, but I am not sure, what a bad client could achieve here with the wider character encoding.

Drakulix: I think this is fine. I assume you did a quick test with these changes?

Drakulix (approved):

PR #1150, opened 2023-09-28
Simplify the CI a bit

elinorbgr:
A few changes to hopefully simplify the CI and reduce its running time:

- Use `cargo check` instead of `cargo test` when there is no test to run anyway
- Use `cargo hack` to check features individually instead of creating a huge matrix of jobs
  - Should save on runtime overall for no longer having a huge list of containers to run and being rate-limited by github
  - Makes the CI result less clogged to read
  - This will automatically pick up new features if / when they are added
- Stop downgrading `log`, this was done to workaround an issue with slog but we don't use it anymore so that should no longer be needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,33 +143,54 @@ jobs:
       - name: Check
         run: cargo check --features "test_all_features"
 
+  smithay-check-features:
+    needs:
+      - format
+      - clippy-check
+
+    env:
+      RUSTFLAGS: "-D warnings"
+
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+
+      - name: Get date for registry cache
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+      - name: Cargo registry cache
+        uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


elinorbgr:
Doesn't look to affect much the runtime in the end, kinda surprised by that tbh. But it's still worth doing it imo, for the other advantages.

elinorbgr:
With some tweaking or the job dependency graph we get a nice speedup though, 10 minutes instead of 14 :)

The main bottleneck now seems to be the msrv build, for some reason it insists in re-downloading the crates.io index instead of using the cached one. I suspect this is because the index we are caching is built using the [sparse protocol](https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html), but our msrv (1.65) is too old to use it...

elinorbgr:
(the stuck "anvil-tests" in the CI summary is a glitch from github, this job does not actually exist, the CI is green)

Drakulix:
Is any step still at least compiling our examples? I am not sure from just reading the CI file.

elinorbgr:
There is still a job that runs `cargo test --features "test_all_features"` on smithay ("smithay-tests"), so this should compile the examples alongside

Drakulix (approved):

PR #1151, opened 2023-09-30
wayland: add wlr-data-control protocol

kchibisov:
wlr_data_control is a protocol used to implement clipboard managers or
access clipboard without creating a window. The implementation of it
ties to regular selections, thus the selection handling was unified
to reduce the maintainance burden.

The present selection modules, like `walyand/primary_selection` and
`wayland/data_device` moved into the new `wayland/selection` module.
Keeping their original implementation, where it was possible.

The new selection module uses the common structure of `seat_data`,
`device`, `source`, and `mod.rs` used in the said above modules, however
it uses the hand-rolled dynamic dispatch with the `selection_dispatch!`
macro. The offers and selection replies are all handled together,
so the code is unique in the most cases.

As a side effect of the selection update and merging the handling of
the primary and clipboard selection into the single `SelectionHandler`
trait (users could still just use one of them, it's not a must to have
both), the xwayland was changed to use some types from the
`wayland/selection` module.

--

I haven't tested XWayland, but the protocol is implemented in anvil, and it should be dead simple to add into any compositor.

The implementation was tested with `wl-clipboard` and https://github.com/YaLTeR/niri/pull/27


anvil/src/state.rs
```diff
@@ -218,38 +193,50 @@ delegate_data_device!(@<BackendData: Backend + 'static> AnvilState<BackendData>)
...
-            if let Some(source) = source {
-                if let Ok(Err(err)) = with_primary_source_metadata(&source, |metadata| {
-                    xwm.new_selection(SelectionType::Primary, Some(metadata.mime_types.clone()))
-                }) {
-                    warn!(?err, "Failed to set Xwayland primary selection");
-                }
-            } else if let Err(err) = xwm.new_selection(SelectionType::Primary, None) {
-                warn!(?err, "Failed to clear Xwayland primary selection");
+            if let Err(err) = xwm.new_selection(ty, source.map(|source| source.mime_types())) {
+                warn!(?err, "Failed to set Xwayland primary selection");
```
Drakulix, line 203: Nitpick

```suggestion
                warn!(?err, ?ty, "Failed to set Xwayland selection");
```

src/wayland/selection/wlr_data_control/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+#[derive(Debug)]
+pub struct DataControlState {
+    manager_global: GlobalId,
+}
+
+impl DataControlState {
+    /// Register new [ZwlrDataControlManagerV1] global.
+    ///
+    /// Passing `primary_selection` will enable support for primary selection as well.
+    pub fn new<D>(display: &DisplayHandle, primary_selection: Option<&PrimarySelectionState>) -> Self
```
Drakulix, line 78: Can we please add a variant here, that allows setting a `filter` for which clients can bind to this global?

Drakulix (commented):
While I really like the simplicity of the `data_control` implementation, there is no way to filter what `wlr_data_device`-clients see, right?

Would it maybe make sense to add a method to the `DataControlHandler` to make it possible for downstream to filter selections send? (Default implementation could be "allow all" and it seems like `send_selection` would be an easy point to hook into.)

I don't have any particular use-cases in mind for this, but the current implementation feels like all or nothing. Either I want the protocol with all of the security problems it brings, or I don't advertise it at all.

Overall though, this is a great PR with so many good improvements. Thanks a bunch!

src/wayland/selection/wlr_data_control/mod.rs
kchibisov: Done.

Drakulix (approved):

PR #1153, opened 2023-10-01
Make output_update() public

YaLTeR:
It's useful for compositors to manage output overlap for non-Window surfaces, e.g. pointer and DnD surfaces.

I put it in `desktop/wayland/utils` as there's nothing Space-specific about it.

Example use scenario: https://github.com/YaLTeR/niri/commit/01906b190318a5e12feeb0dfa20a602a8561809f


Drakulix (approved):
LGTM. Thanks!

PR #1154, opened 2023-10-02
wayland: add cursor_shape_v1 support

kchibisov:
--

The protocol is dead simple and requiring a full rendering impl sounds overkill for it, given
that compositors can opt-in into this protocol, when they can handle the cursor themes.

Tested that it only reaches the compositor with https://github.com/YaLTeR/niri/pull/32 , but the actual rendering is not done, and only default cursor is properly rendered.

Testing client https://github.com/Smithay/client-toolkit/blob/master/examples/themed_window.rs (press `Shift + n` to cycle icons).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ appendlist = "1.4"
 ash = { version = "0.37.1", optional = true }
 bitflags = "2.2.1"
 calloop = "0.12.2"
+cursor-icon = "1.0.0"
 cgmath = "0.18.0"
 downcast-rs = "1.2.0"
 drm-fourcc = "^2.2.0"
--- a/anvil/src/drawing.rs
+++ b/anvil/src/drawing.rs
@@ -35,7 +35,7 @@ impl<T: Texture> Default for PointerElement<T> {
     fn default() -> Self {
         Self {
             texture: Default::default(),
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -30,6 +30,7 @@ appendlist = "1.4"
 ash = { version = "0.37.1", optional = true }
 bitflags = "2.2.1"
 calloop = "0.12.2"
+cursor-icon = "1.0.0"
```
kchibisov, line 33: This is a winit crate used mostly in clients now, it's used by sctk for its themed cursor as well.

Drakulix: The crate notes in it's documentation, that it aims to be cross-platform. Is there reason to assume it might not include all the cursor variants the wayland protocol does at some point to keep it cross-platform? In that case I would assume it would be better to define our own enum.

Drakulix (approved):
Looks good to me.

Cargo.toml
kchibisov: @Drakulix the crate targets w3c specification, which wayland also targets. So it should be fine. The reason for it as well, is that you can call `name()` on icon to load themes, etc.

Drakulix: Great, then it is a good fit.

PR #1155, opened 2023-10-03
Take a global filter function in `InputMethodManagerState::new`

ids1024:
I still wonder if there's a cleaner way to deal with global visibility, but don't have any suggestions at the moment.

KDE exposes the input method protocol only to the process it starts as an IME. This is definitely a protocol that one may want to restrict.


Drakulix (approved):

ids1024:
Updated to fix failing doctest.

PR #1158, opened 2023-10-03
Avoid unnecessary uses of `into_raw_fd` and `borrow_raw`

ids1024:
The use in `virtual_keyboard_handle.rs` additionally was incorrectly using `as_raw_fd` where it should use `into_raw_fd`. Which is unsound but does the right thing unless another thread happens to open an fd at the same time.

### DIFF
--- a/src/backend/x11/mod.rs
+++ b/src/backend/x11/mod.rs
@@ -99,7 +99,7 @@ use nix::{
 use std::{
     collections::HashMap,
     io,
-    os::unix::io::{AsRawFd, BorrowedFd, FromRawFd, OwnedFd},
+    os::unix::io::{AsRawFd, FromRawFd, OwnedFd},
     sync::{
         atomic::{AtomicU32, Ordering},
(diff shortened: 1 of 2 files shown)


src/backend/libinput/mod.rs
```diff
@@ -614,9 +614,7 @@ impl<S: Session> libinput::LibinputInterface for LibinputSessionInterface<S> {
         Ok(unsafe { OwnedFd::from_raw_fd(fd) })
     }
 
-    fn close_restricted(&mut self, fd: OwnedFd) {
-        let _ = self.0.close(fd.into_raw_fd());
-    }
+    fn close_restricted(&mut self, _fd: OwnedFd) {}
```
Drakulix, line 617: This is unfortunately not correct, session-close does more then just call `close` on the fd. We need to convert back here.

ids1024: Ah right. I misread that as just a call to nix's `close`.

PR #1160, opened 2023-10-09
Start using wlcs-rs for wlcs_anvil

bwidawsk:
This is the first part of multiple changes to utilize wlcs-rs crate and generic renderer infrastructure. The purpose is to be able to have multiple compositors share WLCS testing code where appropriate. See #904 (to be updated) for details.


bwidawsk:
Using #904 instead

PR #1161, opened 2023-10-10
Take `OwnedFd` instead of `RawFd` is `Session` methods

ids1024:
Using `OwnedFd` here should be safe. If `close` finds a `RawFd` that matches, either it is a file descriptor for that device, or this is some fd not opened through this API, and will close a device for which the fd was already freed.

Alternately, `Session` could have an associated type, defined as something like `libseat::Device`, and take that instead of `OwnedFd`. But implementing `libinput::LibinputInterface` requires just taking `OwnedFd`s, so this would mean simply moving the mapping from fds to devices/ids to `LibinputSessionInterface` instead.

I guess in https://github.com/PolyMeilex/libseat-rs/pull/4 I was assuming `libseat_close_device` would close the fd. As is, it isn't possible to use that API in libseat and actually close the fd without unsafe code. Maybe `Device` isn't necessary and it could just take the id to close the device. If the fd outlives the device being closed, I guess operations on it may fail? But it won't invalidate the fd, at least. So it's not a "safety" issue.

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -3,7 +3,6 @@ use std::ffi::OsString;
 use std::{
     collections::{hash_map::HashMap, HashSet},
     convert::TryInto,
-    os::unix::io::FromRawFd,
     path::Path,
     sync::{atomic::Ordering, Mutex},
     time::{Duration, Instant},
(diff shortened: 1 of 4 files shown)


ids1024:
This was merged as part of https://github.com/Smithay/smithay/pull/1162.

PR #1162, opened 2023-10-11
Use Rustix instead of nix

ids1024:
This still has a dependency on `nix`, for the `Errno` type used by errors from `drm`. May be best to avoid exposing either crate in a public API where possible...

The `sigaction` code is commented out since rustix doesn't support that. It seems to be problematic to do through direct system calls (as rustix does by default on linux), rather than through libc. Maybe that could just be done through calls to `libc`, but it would be good if Rustix had a solution here.

Also needs to be tested on FreeBSD.


ids1024:
Updated the `sigaction` code to call `libc`, and tested this on FreeBSD. I seem to be having an issue there, but not related to this PR.

This should be in good shape now.

Drakulix (approved):
LGTM

PR #1164, opened 2023-10-12
fix/im-activate-logic

### LINKED ISSUE #1163: Virtual keyboard not tracking focus loss
The virtual keyboard should get closed when focus is removed from input elements, however the virtual keyboard is only closed when a window is quit entirely.

Could not bisect because most commits in Smithay's history don't work.

rano-oss:
closes #1163


Drakulix:
@chrisduerr Please confirm this fixes #1163

PR #1165, opened 2023-10-12
xwm: Send synthetic configure event for reparented windows

Drakulix:
Apparently clients don't like getting positions relative to their parent, if we reparent them? Anyway this works.

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -94,11 +94,11 @@ use x11rb::{
         render::{ConnectionExt as _, CreatePictureAux, PictureWrapper},
         xfixes::{ConnectionExt as _, SelectionEventMask},
         xproto::{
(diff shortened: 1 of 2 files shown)


src/xwayland/xwm/mod.rs
```diff
@@ -2401,3 +2401,31 @@ fn send_selection_notify_resp(
...
+) -> Result<(), ConnectionError> {
+    conn.send_event(
+        false,
+        *win,
+        EventMask::STRUCTURE_NOTIFY,
+        ConfigureNotifyEvent {
+            response_type: CONFIGURE_NOTIFY_EVENT,
+            sequence: 0,
+            event: *win,
+            window: *win,
```
PolyMeilex, line 2419: No idea how X works, is assigning window id to event field intentional?

Drakulix: Yes. https://gitlab.freedesktop.org/wayland/weston/-/blob/main/xwayland/window-manager.c#L725-726

PolyMeilex (approved):

PR #1166, opened 2023-10-13
Implement ext-session-lock

YaLTeR:
Picking up from #1019 with minimal changes, all in subsequent commits. cc @chrisduerr 

No anvil impl, but I've got a niri impl: https://github.com/YaLTeR/niri/compare/main...session-lock

Preliminary testing with swaylock and https://github.com/Smithay/client-toolkit/pull/419 seems to work, gonna give this more testing over the next few days.

### DIFF
--- a/src/wayland/session_lock/lock.rs
+++ b/src/wayland/session_lock/lock.rs
@@ -7,7 +7,7 @@ use crate::wayland::compositor;
 use crate::wayland::compositor::SurfaceAttributes;
 use _session_lock::ext_session_lock_surface_v1::ExtSessionLockSurfaceV1;
 use _session_lock::ext_session_lock_v1::{Error, ExtSessionLockV1, Request};
-use wayland_protocols::ext::session_lock::v1::server as _session_lock;
+use wayland_protocols::ext::session_lock::v1::server::{self as _session_lock, ext_session_lock_surface_v1};
 use wayland_server::{Client, DataInit, Dispatch, DisplayHandle, Resource};
 
 use crate::wayland::session_lock::surface::{ExtLockSurfaceUserData, LockSurface, LockSurfaceAttributes};
@@ -83,9 +83,9 @@ where
 
                 // Initialize surface data.
                 compositor::with_states(&surface, |states| {
-                    states
-                        .data_map
-                        .insert_if_missing_threadsafe(|| Mutex::new(LockSurfaceAttributes::default()))
+                    states.data_map.insert_if_missing_threadsafe(|| {
+                        Mutex::new(LockSurfaceAttributes::new(lock_surface.clone()))
+                    })
                 });
 
                 // Add pre-commit hook for updating surface state.
@@ -94,9 +94,15 @@ where
                         let attributes = states.data_map.get::<Mutex<LockSurfaceAttributes>>();
                         let mut attributes = attributes.unwrap().lock().unwrap();
 
-                        if let Some(state) = attributes.last_acked {
-                            attributes.current = state;
-                        }
+                        let Some(state) = attributes.last_acked else {
+                            attributes.surface.post_error(
+                                ext_session_lock_surface_v1::Error::CommitBeforeFirstAck,
+                                "Committed before the first ack_configure.",
+                            );
+                            return;
+                        };
+
+                        attributes.current = state;
                     });
                 });
 
--- a/src/wayland/session_lock/lock.rs
+++ b/src/wayland/session_lock/lock.rs
@@ -3,8 +3,10 @@
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
-use crate::wayland::compositor;
+use crate::backend::renderer::buffer_dimensions;
+use crate::utils::Size;
 use crate::wayland::compositor::SurfaceAttributes;
+use crate::wayland::compositor::{self, BufferAssignment};
 use _session_lock::ext_session_lock_surface_v1::ExtSessionLockSurfaceV1;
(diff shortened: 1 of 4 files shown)


src/wayland/session_lock/lock.rs
```diff
@@ -0,0 +1,134 @@
...
+        request: Request,
+        data: &SessionLockState,
+        _display: &DisplayHandle,
+        data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            Request::GetLockSurface { id, surface, output } => {
+                // Assign surface a role and ensure it never had one before.
+                if compositor::get_role(&surface).is_some()
+                    || compositor::give_role(&surface, LOCK_SURFACE_ROLE).is_err()
```
Drakulix, line 53: Even though the protocol says "If the surface already has a role", there is reason to allow clients reusing wl_surfaces, that previously have had the `LOCK_SURFACE_ROLE`.

Other protocols have the same wording, still clients expect to be able to re-use wl_surfaces for XdgSurfaces or LayerSurfaces. So imo the first condition here should be removed.

Drakulix (commented):
Looks good to me on first glance, but waiting with merging for that "more testing". Give me a ping once this is ready for a secon review please. :)

src/wayland/session_lock/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// State of the [`ExtSessionLockManagerV1`] Global.
+#[derive(Debug)]
+pub struct SessionLockManagerState {
+    pub(crate) locked_outputs: Vec<WlOutput>,
+}
+
+impl SessionLockManagerState {
+    /// Create new [`ExtSessionLockManagerV1`] global.
+    pub fn new<D>(display: &DisplayHandle) -> Self
```
ids1024, line 74: This should probably take a `filter` callback like other potentially sensitive protocols.

src/wayland/session_lock/lock.rs
```diff
@@ -0,0 +1,142 @@
...
+        lock: &ExtSessionLockV1,
+        request: Request,
+        data: &SessionLockState,
+        _display: &DisplayHandle,
+        data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            Request::GetLockSurface { id, surface, output } => {
+                // Assign surface a role and ensure it never had one before.
+                if compositor::give_role(&surface, LOCK_SURFACE_ROLE).is_err() {
```
chrisduerr, line 52: Why did you relax the role requirements here?

According to the protocol:
> Providing a wl_surface which already has a role or already has a buffer attached or committed is a protocol error

chrisduerr (commented):
Changes from my patch look fine. Just one small question.

src/wayland/session_lock/lock.rs
Drakulix: Same is said in xdg-shell and layer-shell and still clients expect to re-use surfaces previously assigned the same rule, which is also accepted by multiple server-side implementations of said protocols. Doing so differently hear would be inconsistent, which is why I requested said change. I don't think it has any problematic implications.

ids1024:
Running several times on Sway, it seems to be working reliably with the example using a `round_trip` instead of a `flush`.

src/wayland/session_lock/lock.rs
chrisduerr: Ah fair enough, just missed your review because it was collapsed.

src/wayland/session_lock/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+use wayland_protocols::ext::session_lock::v1::server as _session_lock;
+use wayland_server::protocol::wl_output::WlOutput;
+use wayland_server::protocol::wl_surface::WlSurface;
+use wayland_server::{Client, DataInit, Dispatch, DisplayHandle, GlobalDispatch, New};
+
+use crate::wayland::session_lock::lock::SessionLockState;
+use crate::wayland::session_lock::surface::{LockSurface, LockSurfaceConfigure};
+
+pub mod lock;
+pub mod surface;
```
ids1024, line 62: I think for other protocols in Smithay, modules like this are private, and their types are re-exported.

src/wayland/session_lock/lock.rs
```diff
@@ -0,0 +1,142 @@
...
+
+                // Add pre-commit hook for updating surface state.
+                compositor::add_pre_commit_hook::<D, _>(&surface, |_state, _dh, surface| {
+                    compositor::with_states(surface, |states| {
+                        let attributes = states.data_map.get::<Mutex<LockSurfaceAttributes>>();
+                        let mut attributes = attributes.unwrap().lock().unwrap();
+
+                        let Some(state) = attributes.last_acked else {
+                            attributes.surface.post_error(
+                                ext_session_lock_surface_v1::Error::CommitBeforeFirstAck,
```
ids1024, line 107: The protocol also defines `null_buffer` and `dimensions_mismatch` protocol errors. It doesn't seem to mention them, but presumably we should have a check here that a null buffer is committed and it's dimensions match what was given to `configure`.

YaLTeR: Done.

For the NULL buffer I'm checking for `BufferAssignment::Removed`, which as far as I understand doesn't cover committing with no buffer in the first place, but I'm not sure how to easily check for that.

For dimensions I think this is right? I tested with scale 1 and 2 and dynamically resizing the output, seems to work.

YaLTeR:
Haven't hit any issues so far.

Drakulix (approved):
LGTM

PR #1169, opened 2023-10-17
Fix IM and virtual keyboard handling

kchibisov:
The series is split into 2 commits, the first one addresses the virtual keyboard issues I've faced and the second tries to solve IM issues, which were left on the table. Each commit has a description on what it tried to fix.

--

This is by far the ugliest thing I ever wrote, but to make it nice I'd need to delete all the input handling in smithay, which I don't have time to do.

### DIFF
--- a/src/wayland/input_method/input_method_handle.rs
+++ b/src/wayland/input_method/input_method_handle.rs
@@ -9,31 +9,43 @@ use wayland_protocols_misc::zwp_input_method_v2::server::{
     zwp_input_method_v2::{self, ZwpInputMethodV2},
     zwp_input_popup_surface_v2::ZwpInputPopupSurfaceV2,
 };
-use wayland_server::backend::ClientId;
+use wayland_server::{backend::ClientId, protocol::wl_surface::WlSurface};
 use wayland_server::{
     protocol::wl_keyboard::KeymapFormat, Client, DataInit, Dispatch, DisplayHandle, Resource,
 };
 
 use crate::{
     input::{keyboard::KeyboardHandle, SeatHandler},
-    utils::{alive_tracker::AliveTracker, Rectangle, SERIAL_COUNTER},
+    utils::{alive_tracker::AliveTracker, Logical, Rectangle, SERIAL_COUNTER},
     wayland::{compositor, seat::WaylandFocus, text_input::TextInputHandle},
 };
 
 use super::{
     input_method_keyboard_grab::InputMethodKeyboardGrab,
-    input_method_popup_surface::{PopupHandle, PopupSurface},
+    input_method_popup_surface::{PopupHandle, PopupParent, PopupSurface},
     InputMethodHandler, InputMethodKeyboardUserData, InputMethodManagerState,
-    InputMethodPopupSurfaceUserData,
+    InputMethodPopupSurfaceUserData, INPUT_POPUP_SURFACE_ROLE,
 };
-
-const INPUT_POPUP_SURFACE_ROLE: &str = "zwp_input_popup_surface_v2";
 
 #[derive(Default, Debug)]
 pub(crate) struct InputMethod {
-    pub instance: Option<ZwpInputMethodV2>,
+    pub instance: Option<Instance>,
     pub popup_handle: PopupHandle,
     pub keyboard_grab: InputMethodKeyboardGrab,
+}
+
+#[derive(Debug)]
+pub(crate) struct Instance {
+    pub object: ZwpInputMethodV2,
+    pub serial: u32,
+}
+
+impl Instance {
+    /// Send the done incrementing the serial.
+    pub(crate) fn done(&mut self) {
+        self.object.done();
+        self.serial += 1;
+    }
 }
 
 /// Handle to an input method instance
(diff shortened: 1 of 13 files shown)


src/wayland/input_method/input_method_handle.rs
```diff
@@ -77,19 +84,16 @@ impl InputMethodHandle {
...
 }
 
 /// User data of ZwpInputMethodV2 object
 pub struct InputMethodUserData<D: SeatHandler> {
     pub(super) handle: InputMethodHandle,
     pub(crate) text_input_handle: TextInputHandle,
     pub(crate) keyboard_handle: KeyboardHandle<D>,
+    pub(crate) popup_geometry_callback: fn(&D, &WlSurface) -> Rectangle<i32, Logical>,
+    pub(crate) new_popup: fn(&mut D, PopupSurface),
+    pub(crate) dismiss_popup: fn(&mut D, PopupSurface),
```
kchibisov, line 96: Yes. Otherwise you force IME into the users.

kchibisov:
The other question is that the `parent_geometry` doesn't really work, at least the way it's used.

The issue is that parent geometry could change without client losing focus and the popup needs a way for it to be updated. The most reliable solution for popup to actually ask geometry during space renderer each time from the compositor, so it ensures that everything is up to date.

How something like that should be approached? I'm pretty sure it's not the first time something like that happens and some other elements may also need to query for the parent geometry all the time.

kchibisov:
Converting to draft since it needs more tuning to actually work, and probably will have more to come.

I'd still like to know what to do with `geometry`.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -169,54 +165,38 @@ where
...
-    let new_keymap = match new_keymap {
-        Some(keymap) => keymap,
-        None => {
+    let context = xkb::Context::new(xkb::CONTEXT_NO_FLAGS);
+    let new_keymap = match unsafe {
+        xkb::Keymap::new_from_fd(
+            &context,
+            // SAFETY: We still need to close the FD afterwards, so using `as_raw` to map
+            // the fd.
+            fd.as_raw_fd(),
```
Drakulix, line 186: xkbcommon 0.7 (which we just updated to) does take an `OwnedFd` here. Please rebase and remove the unsafe.

Drakulix (approved):
Looks good to me.

https://github.com/Smithay/smithay/pull/1156 was just merged and causes a slight conflict.

rano-oss:
Sorry I am a bit late to the party, it can probably be merged the way it is for now I guess?
Placement of popup surface is incorrect as it should be the compositor that decides where to place the surface.
Text input only communicates placement of the text cursor. Technically my solution is also incorrect, I just made it correct in 90% of cases, but I guess we I should start looking into how to handle collision, see picture:
image
There is also a weird bug sometimes were no text is written and keys pressed go for menu items, not sure why yet, maybe a modifier is stuck?
The last thing I found is that when I open and close an ime a couple of times, the keyboard spits out rubbish.
I haven't had the time to chase down why this happens

rano-oss:
Do we really need to have an implementation of dismiss_popup? Can we not just call "PopupManager::dismiss_popup" in the ime code?

kchibisov:
>Placement of popup surface is incorrect as it should be the compositor that decides where to place the surface.
Text input only communicates placement of the text cursor. Technically my solution is also incorrect, I just made it correct in 90% of cases, but I guess we I should start looking into how to handle collision, see picture:

I expose all the state needed to handle collisions, there's a `protected_region`, which downstream should deal with.

>There is also a weird bug sometimes were no text is written and keys pressed go for menu items, not sure why yet, maybe a modifier is stuck?

It could be a `commit` logic discarding, basically or how we count commits. There could be a modifier stuck, but it would mean that the smithay input stuff is bugged. Also, IME works during that you can like re-trigger it just fine...

>The last thing I found is that when I open and close an ime a couple of times, the keyboard spits out rubbish.
I haven't had the time to chase down why this happens

Given that we just pass things around it could be your IME doing so for whatever reason. I don't have that issue with my IME, for example. This is also has nothing to do with the way things work on this PR in general, since it just uses the smithay infra for that and what I did is basically moved the whole logic to bring up/down IME into the activate stuff.

kchibisov:
>Do we really need to have an implementation of dismiss_popup? Can we not just call "PopupManager::dismiss_popup" in the ime code?

Yes, because `PopupManager` is not mandatory and could be excluded.

rano-oss:
I used fcitx5, which IME are you using?

rano-oss:
> > Do we really need to have an implementation of dismiss_popup? Can we not just call "PopupManager::dismiss_popup" in the ime code?
> 
> Yes, because `PopupManager` is not mandatory and could be excluded.

but would it not be sufficient to check if we have a popup, then just call it?

kchibisov:
>but would it not be sufficient to check if we have a popup, then just call it?

Call what? We don't have `PopupManager` in a `wayland` module, it's a `desktop` module which could be excluded. Users also may not even use the `PopupManager`.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
kchibisov: I can't remove the `unsafe` though. The function is still `unsafe`.

rano-oss:
> > Placement of popup surface is incorrect as it should be the compositor that decides where to place the surface.
> > Text input only communicates placement of the text cursor. Technically my solution is also incorrect, I just made it correct in 90% of cases, but I guess we I should start looking into how to handle collision, see picture:
> 
> I expose all the state needed to handle collisions, there's a `protected_region`, which downstream should deal with.
> 
Could we at least make a correct example in anvil then? Like just the 90% correct one or something?
> > There is also a weird bug sometimes were no text is written and keys pressed go for menu items, not sure why yet, maybe a modifier is stuck?
> 
> It could be a `commit` logic discarding, basically or how we count commits. There could be a modifier stuck, but it would mean that the smithay input stuff is bugged. Also, IME works during that you can like re-trigger it just fine...

Well, I guess I should just create an issue for this then

rano-oss:
> > but would it not be sufficient to check if we have a popup, then just call it?
> 
> Call what? We don't have `PopupManager` in a `wayland` module, it's a `desktop` module which could be excluded. Users also may not even use the `PopupManager`.

Alright, in that case, could we just not check if the popup manager is used? Like we do not really have an alternative for it right now anyway?

kchibisov:
>Could we at least make a correct example in anvil then? Like just the 90% correct one or something?

I don't want to do that myself, but feel free. Only cosmic can probably handle popup re constaints logic correctly.

>Alright, in that case, could we just not check if the popup manager is used? Like we do not really have an alternative for it right now anyway?

I just tell downstream that popup should be dissmissed, how it's done is up to the downstream. Anvil uses `PopupManager` for example, catacomb doesn't use it. I don't know what you really want from me here, sorry. So propose a patch.

kchibisov:
>I used fcitx5, which IME are you using?

https://github.com/tadeokondrak/anthywl

rano-oss:
Alright I will try and see if I can figure something out on my own then. Anyway great work :slightly_smiling_face: , focus seems to work a lot better now at least.

PR #1170, opened 2023-10-18
utils/clock: Use `rustix`, without `unsafe`; infallible

ids1024:
Rustix's `lock_gettime` does not return an error, and notes it is infallible, while `clock_gettime_dynamic` can handle some additional clocks that may result in an error.

I don't think the clock helper in Smithay needs to handle any of those, so returning a `Result` in `new` is unneeded.

This also uses an associated const for `ClockSource::ID`, so the clock id doesn't need to be stored a runtime when it's already expressed in the type. Const generics would be suitable, but don't work with enums yet.


Drakulix:
@ids1024 Could you rebase this?

Issue #1148, opened 2023-09-27
Every `wl_pointer::button` event is followed immediately by a `wl_pointer::motion`


ids1024:
This behavior seems undesirable and incorrect, even if clients normally aren't impacted by absolute motion to the same position.  In particular it *is* an issue with pointer locking, since `wl_pointer::motion` events shouldn't be sent at all.

`gdb` shows `unset_grab` is responsible here. It seems the issue is that `PointerGrab::button` for `ClickGrab` is calling `unset_grab`, which then sends a `motion` event.

Should a grab really be used here? I guess a `motion` event may be necessary with other types of grabs, but not with this one, since it doesn't block motion events. But it's not clear how to do this. Perhaps wlroots could be worth looking at, since they have somewhat similar mechanisms...

With https://github.com/Smithay/smithay/issues/1126 this `motion` also ends up in the same frame as the `button` event. Which maybe is correct if it is sent; or maybe it's implicit in the protocol that `motion` and `button` are separate event types that need to be in different frames? Though more generally, if `set_grab` and `unset_grab` send `motion` events, those should be followed by a `frame`.

Drakulix:
> Should a grab really be used here? I guess a motion event may be necessary with other types of grabs, but not with this one, since it doesn't block motion events.

I think the grab has it's values here. But I would challenge that `set_grab` and `unset_grab` send events. Why is this current necessary and could we shift this responsibility to downstream?

ids1024:
It seems in Sway this is handled with `seatop_down`, which restores `seatop_default` when the last button is release. This is seperate from wlroots pointer grabs.

Anyway, there are several problems here:
* Synthesizing calls to `motion` anywhere lower in the stack than where pointer constraints are handled is broken with pointer locking, since it generates motion events that should be suppressed.
  - So calling it here is wrong if we handle that in the compositor's input event handler.
* If this event is sent, it *probably* shoudn't be part of the frame of the button event? So it would be desirable for the vent to be sent after that is over and that `frame` has been sent.
* A `motion` isn't necessary or desirable on ungrab for this particular grab, but may be necessary for others.

Drakulix:
> A motion isn't necessary or desirable on ungrab for this particular grab

Right, so we shouldn't send it.

The necessity should thus be expressed by the Grab-trait or preferrably the grab would somehow emit this event itself, if needed.

ids1024:
[...]

PR #1173, opened 2023-10-18
Add `restore_focus` argument to `PointerInnerHandle::unset_grab`

ids1024:
Used in `ClickGrab` to prevent `motion` events from occurring with every `button` event. Otherwise, behavior should be unchanged. This matches the argument taken by `KeyboardInnerHandle::unset_grab`.
    
This seems like the simplest solution. It would also be possible to add a method to the `PointerGrab` trait indicating if focus should be restored, but that's complicated since `unset_grab` can't access the
grab when it's `Borrowed`, so it would have to add a bool to `GrabStatus::Borrowed`, etc.
    
This still doesn't send a `frame`, but since this takes a serial and a time, it probably will be sent along with other pointer events, and hopefully part of a `frame`. The Wayland spec isn't all that specific
about when things can/should be part of a `frame`...
    
Calling `motion` is also incorrect with pointer constraints, but grabs other than `ClickGrab` generally shouldn't exist while a constraint is active. It would be good to enforce that some way.

The first commit here is a change I had as part of an attempt at a different approach, which turned out more complicated; but this part seemed a bit neater.

Fixes https://github.com/Smithay/smithay/issues/1148.

### DIFF
--- a/anvil/src/shell/grabs.rs
+++ b/anvil/src/shell/grabs.rs
@@ -64,7 +64,7 @@ impl<BackendData: Backend> PointerGrab<AnvilState<BackendData>> for MoveSurfaceG
         handle.button(data, event);
         if handle.current_pressed().is_empty() {
             // No more buttons are pressed, release the grab.
-            handle.unset_grab(data, event.serial, event.time);
+            handle.unset_grab(data, event.serial, event.time, true);
         }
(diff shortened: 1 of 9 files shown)


Drakulix (approved):
LGTM

PR #1176, opened 2023-10-24
xwm: Fix reordering

Drakulix:
This was unfortunately reversed in https://github.com/Smithay/smithay/pull/1024, which wasn't obvious due to the refactoring that PR was doing.


PolyMeilex (approved):
LGTM

PR #1177, opened 2023-10-25
wayland/compositor: add wl_compositor@v6 constructor

kchibisov:
`wl_compositor@v6` requires special handling in downstream compositors, so smithay can't provide it unconditionally, thus allow setting a version for `wl_compositor` when creating a global.

--

This is probably not really a good solution, Alternative could be to have a separate constructor which builds _at least_ v6.

### DIFF
--- a/src/wayland/compositor/mod.rs
+++ b/src/wayland/compositor/mod.rs
@@ -110,14 +110,16 @@ mod handlers;
 mod transaction;
 mod tree;
 
+use std::cell::RefCell;
 use std::{any::Any, sync::Mutex};
 
 pub use self::cache::{Cacheable, MultiCache};
 pub use self::handlers::{RegionUserData, SubsurfaceCachedState, SubsurfaceUserData, SurfaceUserData};
 use self::transaction::TransactionQueue;
 pub use self::transaction::{Blocker, BlockerState};
-use self::tree::PrivateSurfaceData;
 pub use self::tree::{AlreadyHasRole, TraversalAction};
+use self::tree::{PrivateSurfaceData, SuggestedSurfaceState};
+use crate::utils::Transform;
 use crate::utils::{user_data::UserDataMap, Buffer, Logical, Point, Rectangle};
 use wayland_server::backend::GlobalId;
 use wayland_server::protocol::wl_compositor::WlCompositor;
@@ -399,6 +401,32 @@ where
     PrivateSurfaceData::with_states(surface, f)
 }
 
+/// Send the `scale` and `transform` preferences for the given surface when it supports them.
+///
+/// The new state is only send when it differs from the already cached one on the calling thread.
+pub fn send_surface_state(surface: &WlSurface, data: &SurfaceData, scale: i32, transform: Transform) {
+    if surface.version() < 6 {
+        return;
+    }
+
+    // NOTE we insert default for checks below to work properly.
+    let mut storage = data
+        .data_map
+        .get_or_insert(|| RefCell::new(SuggestedSurfaceState::default()))
+        .borrow_mut();
+
+    if storage.scale != scale {
+        surface.preferred_buffer_scale(scale);
+        storage.scale = scale;
+    }
+
+    let transform = transform.into();
+    if storage.transform != transform {
+        surface.preferred_buffer_transform(transform);
+        storage.transform = transform;
+    }
+}
+
 /// Retrieve the metadata associated with a `wl_region`
 ///
 /// If the region is not managed by the `CompositorGlobal` that provided this token, this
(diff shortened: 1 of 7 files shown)


kchibisov:
I'll fix CI after ack on approach, since change in it is a rewrite of the whole thing.

YaLTeR:
Hm, which special handling is *required*?

kchibisov:
>Hm, which special handling is required?

Downstream has to actually send the events for all the surfaces, it's not as automated as with `wl_output` and downstream could decide to use different scale factors for multiple windows on the same output.

YaLTeR:
> Downstream has to actually send the events for all the surfaces

Do you mean preferred_buffer_scale/transform? I couldn't find where it says that the compositor should send them?

YaLTeR:
Hm, I don't see where it says that in the protocol? I'd assume that if a surface never gets a preferred buffer scale, it should keep using wl_output scale?

kchibisov:
Should be ready to go.

src/wayland/compositor/mod.rs
```diff
@@ -564,18 +564,43 @@ impl CompositorClientState {
...
+    /// The same as [`new`], but binds at least version 6 of [`wl_compositor`].
+    ///
+    /// This means that for clients to scale and apply transformation with
+    /// non-default values [`wl_surface::preferred_buffer_scale`] and
+    /// [`wl_surface::preferred_buffer_transform`] must be sent.
+    ///
+    /// [`new`]: Self::new
+    /// [`wl_compositor`]: wayland_server::protocol::wl_compositor
+    /// [`wl_surface::preferred_buffer_scale`]: WlSurface::preferred_buffer_scale
+    /// [`wl_surface::preferred_buffer_transform`]: WlSurface::preferred_buffer_transform
```
Drakulix, line 591: No sorry, we are not asking downstream to send raw events. Downstream would have to check the version of global the client has bound to do this correctly, etc. This needs a function similar to `set_preferred_scale` for the fractional scale protocol.

kchibisov:
I'm not sure I like the way it's because downstream can't call `send_surface_state` during the `with_surfaces`, since the data can only be mutated. Maybe this cache should all go into the `SurfaceData` instead of `PrivateSurfaceData` but that would mean that you need to get one cache first and then access the other one with the interior mutability?

The same issue is if you want to call basically any `compositor/mod.rs` api during the `with_surfaces`, but I guess you can avoid doing so for some, it's just most of them are just getters, but in this particular case you'd need an API which can actually set things...

kchibisov:
A particular place I don't like with the given approach is https://github.com/kchibisov/niri/blob/37692b264f93e6018d0af04d7459b538e51618ae/src/layout.rs#L1958 , but it's not like we want the method on the `SurfaceData` or require `SurfaceData` before calling this function, since it could be annoying for downstream.

kchibisov:
Added a separate commit for `new_subsuface`, so downstream can send a scale right away when parent is known and they can't present parent across the outputs.

Drakulix:
> A particular place I don't like with the given approach is https://github.com/kchibisov/niri/blob/37692b264f93e6018d0af04d7459b538e51618ae/src/layout.rs#L1958 , but it's not like we want the method on the `SurfaceData` or require `SurfaceData` before calling this function, since it could be annoying for downstream.

Honestly I don't know. We require the `SurfaceData` in other places and it has been fine for downstream?
I don't have a particular strong preference here, but taking `SurfaceData` instead of the surface for `send_surface_state` would be more consistent with other apis.

>  Added a separate commit for new_subsuface, so downstream can send a scale right away when parent is known and they can't present parent across the outputs.

That seems like a good addition. :+1: 

Little bike-shed: I am not a huge fan of how we are overloading the word `state` through the whole crate.
Any other ideas for `send_surface_state`? It's a little long, but maybe even something like `send_surface_scale_transform` or `send_preferred_scale_transform`? (Or split the two up? Though that might cause some unnecessary overhead..)

kchibisov:
>Honestly I don't know. We require the SurfaceData in other places and it has been fine for downstream?
I don't have a particular strong preference here, but taking SurfaceData instead of the surface for send_surface_state would be more consistent with other apis.

Yeah, but it's just other APIs don't mutate... I'll see how it looks though.

>send_preferred_scale_transform

Ok, but once something new is added it'll go `scale_transform_subpixel`? The state is inline with what it actually is, since `xdg-shell` also uses the combination `surface state` https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml?ref_type=heads#L439 .

Drakulix:
> Ok, but once something new is added it'll go `scale_transform_subpixel`? The state is inline with what it actually is, since `xdg-shell` also uses the combination `surface state` https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml?ref_type=heads#L439 .

That is somewhat assuring. I guess as long as we don't have a better idea, lets keep it like this for now.

Drakulix (approved):
LGTM.

PR #1178, opened 2023-10-27
Use pixman for compositing pointer

bwidawsk:
This needs to be in draft until pixman-rs is more stable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ scan_fmt = { version = "0.2.3", default-features = false }
 encoding_rs = { version = "0.8.33", optional = true }
 profiling = "1.0"
 smallvec = "1.11"
+pixman = { git = "https://github.com/cmeissl/pixman-rs.git", version = "0.1.0", features = ["drm-fourcc"] }
 
 [dev-dependencies]
 clap = { version = "4", features = ["derive"] }
--- a/anvil/src/drawing.rs
+++ b/anvil/src/drawing.rs
@@ -1,7 +1,5 @@
 #![allow(clippy::too_many_arguments)]
 
-
-
 use smithay::{
     backend::renderer::{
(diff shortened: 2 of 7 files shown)


i509VCB:
Is the rationale here that the gpu upload for cursors which likely can't use a dmabuf fb is slower than copying into a dumb buffer?

cmeissl:
> Is the rationale here that the gpu upload for cursors which likely can't use a dmabuf fb is slower than copying into a dumb buffer?

The issue with the current code is not the upload, we need that as a fallback anyway for drawing the cursor on the primary plane.

The issue we have is that we can not use the gles renderer to draw the cursor on the cursor plane. This is the reason why we currently draw to an offscreen buffer and then copy the offscreen pixels to a gbm buffer for scan-out on the cursor plane. But this is pretty costly, especially downloading the pixels from the gpu. 
A software renderer would solve that by directly drawing to the gbm buffer.

For some cases it would be possible to directly copy the cursor pixels to the buffer, but that would at least exclude scaling and output transformations.

Cargo.toml
```diff
@@ -68,6 +68,7 @@ scan_fmt = { version = "0.2.3", default-features = false }
 encoding_rs = { version = "0.8.33", optional = true }
 profiling = "1.0"
 smallvec = "1.11"
+pixman = { git = "https://github.com/cmeissl/pixman-rs.git", version = "0.1.0", features = ["drm-fourcc"] }
```
ids1024, line 71: If the dependency is only used in `backend/drm/compositor`, it should probably be an optional dependency that `backend_drm` requires.

bwidawsk:
I think this PR is going to get closed in favor of using the pixman renderer with special hooks to avoid a copy. I'll keep it open until a new PR supersedes it.

PR #1183, opened 2023-10-27
Accept `F` instead of `&mut F` in `with_surfaces_surface_tree` again

ids1024:
https://github.com/Smithay/smithay/pull/1177 changed this. But it isn't actually needed for the other changes there, since there's a blanket implementation of `FnMut` for `&mut` references to types implementing `FnMut`.

So this is more general, and helpful for simply passing a closure to the function, etc.


kchibisov:
@ids1024 as long as it works it should be fine. The goal was to remove the `Copy` marker, since you should be able to pass `&mut` data with the closure, like `&mut Vec` to collect all the surfaces inside the `with_surfaces` API.

Drakulix (approved):

PR #1188, opened 2023-10-30
 Add API to manipulate XKB layouts

kchibisov:
While this is not ideal it does the job to implement basic layouts
functionality like switching layout on compositor and propagating the
change to client, resolving latin character for Keysym handles and
set/restore layouts.

--

Showcase could be found here https://github.com/YaLTeR/niri/pull/65

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -11,9 +11,11 @@ use std::{
 use thiserror::Error;
 use tracing::{debug, error, info, info_span, instrument, trace};
 
+use xkbcommon::xkb::ffi::XKB_STATE_LAYOUT_EFFECTIVE;
 pub use xkbcommon::xkb::{self, keysyms, Keycode, Keysym};
 
 use super::{Seat, SeatHandler};
+use private::XkbTypesHandle;
 
 #[cfg(feature = "wayland_frontend")]
 mod keymap_file;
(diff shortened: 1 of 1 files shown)


src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+
+    /// Get the human readablle name for the layout.
+    fn layout_name(&self, layout: Layout) -> &str {
+        self.keymap().layout_get_name(layout.0)
+    }
+
+    /// Get the number of layouts in the keymap.
+    fn num_layouts(&self) -> usize {
+        self.keymap().num_layouts() as usize
+    }
```
kchibisov, line 391: Not sure how useful this is.

src/input/keyboard/mod.rs
```diff
@@ -838,3 +1020,15 @@ impl<D: SeatHandler + 'static> KeyboardGrab<D> for DefaultGrab {
...
+mod private {
+    use super::xkb;
+    // Glue trait to avoid duplicated code.
+    pub trait XkbTypesHandle {
+        /// Get the reference to the xkb keymap.
+        fn keymap(&self) -> &xkb::Keymap;
+
+        /// Get the reference to the xkb state.
+        fn state(&self) -> &xkb::State;
+    }
```
kchibisov, line 1033: Trick to not expose `xkbcommon` crate.

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+                    .key_char()
+                    .map(|key| key.is_ascii_alphabetic())
+                    .unwrap_or(false)
+                {
+                    return Some(*keysym);
+                }
+            }
+        }
+
+        Some(base_sym)
```
kchibisov, line 290: I decided to go with something like that for now, it sort of works, even though not ideal, but it should be enough to not trigger bindings you don't want to trigger, like due to different `Enter` locations. Not sure how it's relevant though.

There's no helper to check for latin, but `key_char` logic sort of does that already.

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+            layout.0,
+        );
+
+        if state != 0 {
+            self.mods_state.update_with(self.state);
+            *self.mods_changed = true;
+        }
+    }
+
+    /// Switches layout forward cycling when cyclying when it reaches the end.
```
YaLTeR, line 337:
```suggestion
    /// Switches layout forward cycling when it reaches the end.
```

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+    state: &'a mut xkb::State,
+    keymap: &'a mut xkb::Keymap,
+    mods_state: &'a mut ModifiersState,
+    mods_changed: &'a mut bool,
+}
+
+impl<'a> XkbContext<'a> {
+    /// Set layout of the keyboard to the given index.
+    ///
+    /// Returns `true` when the layout got applied.
```
YaLTeR, line 320: Nowhere to be seen it seems

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+        }
+    }
+
+    /// Switches layout forward cycling when cyclying when it reaches the end.
+    pub fn cycle_next_layout(&mut self) {
+        let next_layout = (self.active_layout().0 + 1) % self.keymap.num_layouts();
+        self.set_layout(Layout(next_layout));
+    }
+
+    /// Switches layout backward cycling when cyclying when it reaches the start.
```
YaLTeR, line 343:
```suggestion
    /// Switches layout backward cycling when it reaches the start.
```

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+        (0..self.keymap().num_layouts())
+            .find(|&idx| {
+                self.state()
+                    .layout_index_is_active(idx, XKB_STATE_LAYOUT_EFFECTIVE)
+            })
+            .map(Layout)
+            .unwrap_or_default()
+    }
+
+    /// Get the human readablle name for the layout.
```
YaLTeR, line 383:
```suggestion
    /// Get the human readable name for the layout.
```

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+pub trait XkbContextHandler: XkbTypesHandle {
+    /// Get the active layout of the keyboard.
+    fn active_layout(&self) -> Layout {
+        (0..self.keymap().num_layouts())
+            .find(|&idx| {
+                self.state()
+                    .layout_index_is_active(idx, XKB_STATE_LAYOUT_EFFECTIVE)
+            })
+            .map(Layout)
+            .unwrap_or_default()
```
YaLTeR, line 380: Does `unwrap_or_default()` make sense here?

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+        self.keymap().layout_get_name(layout.0)
+    }
+
+    /// Get the number of layouts in the keymap.
+    fn num_layouts(&self) -> usize {
+        self.keymap().num_layouts() as usize
+    }
+
+    /// Iterate over layouts present in the keymap.
+    fn layouts(&self) -> Box<dyn Iterator<Item = Layout>> {
```
YaLTeR, line 394: `impl Iterator<Item = Layout> + '_`?

src/input/keyboard/mod.rs
YaLTeR: Isn't it already exposed?

> `pub use xkbcommon::xkb::{self,`

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+        };
+
+        // Try to look other layouts and find the one with ascii character.
+        for layout in self.layouts() {
+            if layout == effective_layout {
+                continue;
+            }
+
+            for keysym in self.raw_syms_for_key_in_layout(self.keycode, layout) {
+                // NOTE: Only check for ascii alphabetic since the rest of the punctuation should
```
YaLTeR, line 278: Doesn't Dvorak put punctuation at the top-left? Also Colemak puts <kbd>;</kbd> on <kbd>P</kbd>. And Russian has <kbd>;</kbd> where QWERTY has <kbd>/</kbd>, etc.

For a more specific example, I have default binds to <kbd>,</kbd> and <kbd>.</kbd> in niri, in Russian that's <kbd>б</kbd> and <kbd>ю</kbd>.

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+            return Some(base_sym);
+        };
+
+        // Try to look other layouts and find the one with ascii character.
+        for layout in self.layouts() {
+            if layout == effective_layout {
+                continue;
+            }
+
+            for keysym in self.raw_syms_for_key_in_layout(self.keycode, layout) {
```
YaLTeR, line 277: If we check only `first()` for `base_sym` then maybe this should also do that? I.e. change to `if let Some(keysym) = <...>.first()`

src/input/keyboard/mod.rs
```diff
@@ -249,12 +251,166 @@ impl<'a> KeysymHandle<'a> {
...
+    ///
+    /// This method is handy to implement layout agnostic bindings. Keep in mind that
+    /// it could be not-ideal to use just this function, since some layouts utilize non-standard
+    /// shift levels and you should look into [`Self::modified_sym`] first.
+    ///
+    /// The `None` is returned when the underlying keycode doesn't produce a valid keysym.
+    pub fn raw_latin_sym_or_raw_current_sym(&self) -> Option<Keysym> {
+        let effective_layout = Layout(self.state.key_get_layout(self.keycode));
+        // NOTE: There's always a keysym in the current layout given that we have modified_sym.
+        let base_sym = *self.raw_syms().first()?;
```
YaLTeR, line 264: I see xkb has `key_get_one_sym` for when you only want a single keysym rather than multiple (it returns `NoSymbol` in case there would be multiple), but I don't see a corresponding `by_level` function. Maybe it would make sense to replicate this here by only proceeding when `raw_syms()` has exactly one element. But also it's probably fine as is?

src/input/keyboard/mod.rs
kchibisov: yes, default is base layout.

src/input/keyboard/mod.rs
kchibisov: not stable yet.

src/input/keyboard/mod.rs
kchibisov: Does it really matter to resolve punctuation here? The thing is that you don't resolve from punctuation key a latin layout, because it's sort of a punctuation. I don't mind having to check for punctuation as well here.

src/input/keyboard/mod.rs
kchibisov: it's fine as is, you need api for `layout + level`, the thing I use is basically the only one available.

src/input/keyboard/mod.rs
kchibisov: Hm, but where you can build it then and how it's being used?

src/input/keyboard/mod.rs
YaLTeR: wdym? has been for a while

YaLTeR: Ah, right, this is a trait.

src/input/keyboard/mod.rs
YaLTeR: I'm thinking of the keys you might reasonably want to bind like [, ], ,, ., ;, ', that become letters in Russian for instance

kchibisov: Ah, the 'ю' and 'б' things, probably should add then...

src/input/keyboard/mod.rs
kchibisov: The are used in `mods`, so I sort of exposed it.

YaLTeR (approved):

Drakulix (approved):

PR #1190, opened 2023-10-31
security-context: Expose client_id

Drakulix:
Expose the client_id, that created the security context. This can be used to inherit properties like e.g. privileged protocols to new sockets created by this protocol.


ids1024:
A minor detail, but would it be better to call it something like `creator_client_id`? Or maybe something else.

If I didn't know how security contexts work, I might expect `client_id` to be the id of the client running in the context, and not the sandboxing engine that created the context.

Drakulix:
> A minor detail, but would it be better to call it something like creator_client_id?

Fixed :)

PolyMeilex (approved):

PR #1195, opened 2023-11-01
Update `glow` to 0.14

ids1024:


ids1024:
Hm, looking at the clippy error, it seems a change, in particular https://github.com/grovesNL/glow/commit/d62451c7, make `glow::Context` no longer `Send`/`Sync` by adding a raw pointer to it, and not explicitly implementing those traits.

Show `glow` be implementing those wrappers, or should we wrap it to make it `Sync`? Assuming we need the renderer to be `Send`/`Sync`.

Drakulix:
> Hm, looking at the clippy error, it seems a change, in particular [grovesNL/glow@d62451c7](https://github.com/grovesNL/glow/commit/d62451c7), make `glow::Context` no longer `Send`/`Sync` by adding a raw pointer to it, and not explicitly implementing those traits.
> 
> Show `glow` be implementing those wrappers, or should we wrap it to make it `Sync`? Assuming we need the renderer to be `Send`/`Sync`.

It is kinda right for glow to not explicitly implement those, given that this depends on the platform and even on unix this is not clear cut. As long as the context is only made current on one thread at a time, you can move it around, but otherwise stuff breaks.

So I guess it would be more correct for our wrapper to implement Send/Sync and make sure to uphold that invariant.

ids1024:
Yeah, so using it with multiple threads can be unsafe. But `glow` is all unsafe, and it's up to the caller to know what safety constraints apply to their platform and use of glow.

So I guess it's a matter of taste whether or not such a type should be `Send` or `Sync`.

SuperSonicHub1:
I'm trying to upgrade the egui and egui-glow dependencies in [`smithay-egui`](https://github.com/Smithay/smithay-egui) to `0.28` and was blocked by this PR still having not gone through. I'd like to help; what needs to be done to merge this PR?

ids1024:
I guess we're still waiting for glow 0.14 to fix the `Send + Sync` bound? It was changed but not released yet: https://github.com/grovesNL/glow/pull/267

ids1024:
Updated to 0.14, and updated https://github.com/pop-os/cosmic-comp/pull/885.

Drakulix (approved):

PR #1198, opened 2023-11-04
Ensure that damage is non-overlapping

kchibisov:
The previous logic was blindly merging and doing only a forward pass, however this won't work when after merging 2 rectangles already processed one could end up in overlap.

This is an example of old renderer doing so:

```
[ 697101.942] wl_surface@26.damage_buffer(66, 350, 36, 48)
[ 697101.956] wl_surface@26.damage_buffer(126, 446, 828, 168)
[ 697101.965] wl_surface@26.damage_buffer(90, 638, 744, 528)
[ 697101.974] wl_surface@26.damage_buffer(162, 1262, 420, 72)
[ 697101.984] wl_surface@26.damage_buffer(114, 1358, 312, 48)
[ 697101.994] wl_surface@26.commit()

damage to be rendered: [
    Rectangle<smithay::utils::geometry::Physical> {
        x: 158,
        y: 478,
        width: 828,
        height: 168,
    },
    Rectangle<smithay::utils::geometry::Physical> {
        x: 194,
        y: 1294,
        width: 420,
        height: 72,
    },
    Rectangle<smithay::utils::geometry::Physical> {
        x: 146,
        y: 1390,
        width: 312,
        height: 48,
    },
    Rectangle<smithay::utils::geometry::Physical> {
        x: 32,
        y: 382,
        width: 834,
        height: 1152,
    },
]
```

The new algorithm trying splits damage bounding box into non-overlapping regions containing damage, and then running tile-based damage shaping, where for each tile we try to find total intersection + union of damage rectangles.

--

~I still should add tests for common patterns, but the main idea of implementation should be good.~

### DIFF
--- a/src/backend/renderer/damage/shaper.rs
+++ b/src/backend/renderer/damage/shaper.rs
@@ -1,0 +1,409 @@
+use std::mem;
+
+use crate::utils::{Physical, Rectangle, Size};
+
+/// The recommended minimum tile side.
+const DEFAULT_MIN_TILE_SIDE: i32 = 16;
+
+/// The maximum ratio of the largest damage rectangle to the current damage bbox.
+const MAX_DAMAGE_TO_DAMAGE_BBOX_RATIO: f32 = 0.9;
+
+/// The gap between tiles damage to merge tiles into one.
+const TILE_DAMAGE_GAP_FRACTION: i32 = 4;
+
+/// State of the damage shaping.
+#[derive(Debug, Default)]
+pub struct DamageShaper<const MIN_TILE_SIDE: i32 = DEFAULT_MIN_TILE_SIDE> {
+    /// Cache for tiles to avoid re-allocation.
+    tiles_cache: Vec<Tile>,
+    /// The damage accumulated during shaping.
+    out_damage: Vec<Rectangle<i32, Physical>>,
+}
+
+impl<const MIN_TILE_SIDE: i32> DamageShaper<MIN_TILE_SIDE> {
+    /// Shape damage rectangles.
+    #[profiling::function]
+    pub fn shape_damage(&mut self, in_damage: &mut Vec<Rectangle<i32, Physical>>) {
+        self.out_damage.clear();
+        self.tiles_cache.clear();
+
+        // Call the implementation without direction.
+        self.shape_damage_impl(in_damage, None, false);
+
+        // The shaped damage is inside of `out_damage`, so swap it with `in_damage` since
+        // it's irrelevant.
+        mem::swap(&mut self.out_damage, in_damage);
+    }
+
+    // A divide and conquer hybrid damage shaping algorithm.
+    //
+    // The key idea is to split rectangles by non-overlapping segments on dominating axis (largest
+    // side of the damage bounding box).
+    //
+    // When damage is overlapping through the entire `in_damage` span we shape damage using the
+    // _tile_ based shaper, where the damage bounding box is split into tiles and damage is being
+    // computed for each tile individually by walking all the current damage rectangles.
+    #[inline(never)]
+    fn shape_damage_impl(
+        &mut self,
+        in_damage: &mut [Rectangle<i32, Physical>],
+        last_direction: Option<DamageSplitAxis>,
+        invert_direction: bool,
+    ) {
+        // Optimize small damage input.
+        if in_damage.is_empty() {
+            return;
+        } else if in_damage.len() == 1 {
+            self.out_damage.push(in_damage[0]);
+            return;
+        }
+
+        // Compute the effective damage bounding box and the maximum damaged area rectangle.
+
+        let (x_min, y_min, x_max, y_max, max_damage_area) =
+            in_damage
+                .iter()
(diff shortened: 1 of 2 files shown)


kchibisov:
This should be ready for review.

src/backend/renderer/damage/shaper.rs
```diff
@@ -0,0 +1,407 @@
...
+            return;
+        } else if in_damage.len() == 1 {
+            self.out_damage.push(in_damage[0]);
+            return;
+        }
+
+        // Compute the effective damage bounding box and the maximum damaged area rectangle.
+
+        let (x_min, y_min, x_max, y_max, max_damage_area) =
+            in_damage.iter().fold((i32::MAX, i32::MAX, 0, 0, 0), |acc, rect| {
```
YaLTeR, line 64: Are input rects' locations guaranteed to be > 0? My monitor rn is located at negative coordinates.

src/backend/renderer/damage/shaper.rs
```diff
@@ -0,0 +1,407 @@
...
+        }
+
+        // The coordinate where the first rectangle ends and where the potential overlap
+        // will end.
+        let mut overlap_end = match direction {
+            DamageSplitAxis::X => {
+                if Some(direction) != last_direction {
+                    in_damage.sort_unstable_by(|lhs, rhs| {
+                        // Sort ascending by X and then descending by width, so the first
+                        // rectangle will overlap the most.
```
YaLTeR, line 120: But it it's [0, 1], [1, 10], [2, 9] according to this sort then the first rectangle isn't the one that overlaps the most?

If this is intended then comments here should be clarified as to what's going on.

src/backend/renderer/damage/shaper.rs
```diff
@@ -0,0 +1,407 @@
...
+                        lhs.loc.y.cmp(&rhs.loc.y).then(rhs.size.h.cmp(&lhs.size.h))
+                    });
+                }
+                in_damage[0].loc.y + in_damage[0].size.h
+            }
+        };
+
+        // The start of overlap.
+        let mut overlap_start_idx = 0;
+        for idx in overlap_start_idx + 1..in_damage.len() {
```
YaLTeR, line 140:
```suggestion
        for idx in 1..in_damage.len() {
```
It's 0, might as well.

src/backend/renderer/damage/shaper.rs
```diff
@@ -0,0 +1,407 @@
...
+                // Advance the overlap.
+                overlap_start_idx = idx;
+                overlap_end = rect_end;
+            } else {
+                overlap_end = overlap_end.max(rect_end);
+            }
+        }
+
+        // When rectangle covers the entire bounding box and we've tried different direction of
+        // splitting perform the .
```
YaLTeR, line 160: You accidentally a word.

src/backend/renderer/damage/shaper.rs
```diff
@@ -0,0 +1,407 @@
...
+    fn shape_damage_tiled(
+        &mut self,
+        in_damage: &[Rectangle<i32, Physical>],
+        bbox: Rectangle<i32, Physical>,
+        tile_size: Size<i32, Physical>,
+    ) {
+        self.tiles_cache.clear();
+        let tile_gap: Size<i32, Physical> = From::from((
+            tile_size.h / TILE_DAMAGE_GAP_FRACTION,
+            tile_size.w / TILE_DAMAGE_GAP_FRACTION,
```
YaLTeR, line 191: Is it intended that h and w are swapped here?

YaLTeR (commented):
If I was writing this I'd absolutely add some [proptest](https://lib.rs/crates/proptest) tests. Have it generate a random Vec of Rectangles of input, run the damage shaper, then verify that:
1. All input rectangles are covered with damage.
2. There's no overlap in the output.

src/backend/renderer/damage/shaper.rs
kchibisov: _ascending by X_ **then** descending by _width_ everything you showed is correct.

src/backend/renderer/damage/shaper.rs
kchibisov: nope.

kchibisov:
>If I was writing this I'd absolutely add some [proptest](https://lib.rs/crates/proptest) tests. Have it generate a random Vec of Rectangles of input, run the damage shaper, then verify that:

Such algorithms could generally be proved by induction, you have a routine to shape tiles, and it's being checked, and you have a top-level driver that just uses sort and find gap inside of it, so all you need to test is sort of tested already.

src/backend/renderer/damage/shaper.rs
kchibisov: yeah, but it's more clear that way.

kchibisov:
>All input rectangles are covered with damage.

this thing is harder than you think, because it's basically requires the same algorithm we're using, but for opaque regions.

YaLTeR (approved):

cmeissl (approved):
LGTM
Thanks!

PR #1200, opened 2023-11-04
Implement `Eq` and `Hash` for `PopupSurface`

### LINKED ISSUE #1199: Implement `Hash` and `Eq` for `PopupSurface`
It would be helpful if `PopupSurface` could implement `Hash` and `Eq` so that it could be used in hash maps - compositors may wish to store custom state related to a particular popup in a hash map.

Antikyth:
Does what it says on the tin. This allows `PopupSurface` to be used in `HashMap`s, should compositors wish to associate custom state with a popup.

Closes #1199.

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -1478,6 +1478,15 @@ impl std::cmp::PartialEq for PopupSurface {
     }
 }
 
+impl std::cmp::Eq for PopupSurface {}
+
+impl std::hash::Hash for PopupSurface {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // self.alive().hash(state);
+        self.wl_surface.hash(state);
+    }
+}
+
 impl PopupSurface {
     /// Is the toplevel surface referred by this handle still alive?
     pub fn alive(&self) -> bool {


src/wayland/shell/xdg/mod.rs
```diff
@@ -1478,6 +1478,15 @@ impl std::cmp::PartialEq for PopupSurface {
     }
 }
 
+impl std::cmp::Eq for PopupSurface {}
+
+impl std::hash::Hash for PopupSurface {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // self.alive().hash(state);
```
kchibisov, line 1485:
```suggestion
```

Antikyth: This matches above where it is commented out in the `PartialEq` implementation. Since `Hash` has to be kept in sync with the `PartialEq` implementation, either the commented out part should be entirely removed, or this should remain in case it is un-commented-out.

kchibisov: having dead commented out code looks like an artifact, so it must be removed.

Drakulix:
So the commented out `alive` check is actually the smoking gun, telling us, why this implementation is not possible.

Wayland objects (which includes the `WlSurface` and `XdgPopup` types the `PopupSurface` is composed of) don't implement `Eq`, because wayland-rs doesn't guarantee that dead objects compare as equal, even if they refer to the same dead object.

Which violates the invariant of `Eq`, which disallows `Hash`. So these implementations are unsafe and you simply can't construct a `HashMap` using this type as a key, sorry.

kchibisov:
>So the commented out alive check is actually the smoking gun, telling us, why this implementation is not possible.
It used to be there, because we didn't want any comparison with dead popups to succeed and it's commented out, because it is actually not needed.

Commenting out code is not a good way to document anything. You can always write something and explain why something doesn't work. You can also add doc comments to trait impl methods explaining that.

Like commented out code should never be the case in nearly any language, it's like the worst.

Drakulix:
> >So the commented out alive check is actually the smoking gun, telling us, why this implementation is not possible.
> 
> It used to be there, because we didn't want any comparison with dead popups to succeed and it's commented out, because it is actually not needed.
> 
> 
> 
> Commenting out code is not a good way to document anything. You can always write something and explain why something doesn't work. You can also add doc comments to trait impl methods explaining that.
> 
> 
> 
> Like commented out code should never be the case in nearly any language, it's like the worst.

I don't disagree. I also don't know how it happened, it certainly wasn't meant to be documentation.

It is just what promoted me to lookup the wayland-client PartialEq implementation.

Drakulix:
@Antikyth first of all thank you for the contribution. Can you maybe explain your use-case for this a bit more? I am trying to think of a good alternative to suggest, given that this code won't work as is, and I think would help knowing what kind of data you are storing.

PR #1201, opened 2023-11-04
Start global serial counter at 1

roscale:
Qt apps don't acknowledge the xdg_surface configure event when the serial number is 0, therefore, no toplevel appears.

This bug never happens on most compositors because the serial number 0 is used for other things, and by the time a configure event comes, the number is already incremented.

The minimal example suffers from this bug because it's so minimal, and unfortunately serial number 0 is used for the first configure event.

### DIFF
--- a/src/utils/serial.rs
+++ b/src/utils/serial.rs
@@ -71,7 +71,9 @@ pub struct SerialCounter {
 impl SerialCounter {
     /// Retrieve the next serial from the counter
     pub fn next_serial(&self) -> Serial {
-        let _ = self.serial.compare_exchange(0, 1, Ordering::AcqRel, Ordering::SeqCst);
+        let _ = self
+            .serial
+            .compare_exchange(0, 1, Ordering::AcqRel, Ordering::SeqCst);
         Serial(self.serial.fetch_add(1, Ordering::AcqRel))
     }
 }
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Wow nice catch! The counter can wrap-around though, so I guess we still have problem in here, although it's unlikely to cause any issues?

i509VCB (approved):
Checking qt, this is probably a bug. I'd probably give the Qt devs a note about this https://code.qt.io/cgit/qt/qtwayland.git/tree/src/plugins/shellintegration/xdg-shell/qwaylandxdgshell.cpp#n385

I'm curious if this affects any other toolkits.

roscale:
@Drakulix 

It's extremely unlikely, but I can fix it.
However, I've never used atomic variables before and I don't know exactly what I should write to keep it correct.

Maybe this, or maybe just an if else with load but I'm not sure. And I'm also not sure about the memory orderings in compare_exchange.

```rust
    pub fn next_serial(&self) -> Serial {
        let _ = self.serial.compare_exchange(0, 1, Ordering::AcqRel, Ordering::SeqCst);
        Serial(self.serial.fetch_add(1, Ordering::AcqRel))
    }
```

Drakulix:
No if else, the fix needs to be one operation. I believe the compare_exchange would be correct. Can you update the PR with that?

Drakulix:
Thanks again!

PR #1203, opened 2023-11-06
xdg-activation: Add client_id to token data

Drakulix:
Makes it easier to special case certain (e.g. privileged) clients, as surface and seat (which could otherwise be used to get the client) are optional.

### DIFF
--- a/src/wayland/xdg_activation/dispatch.rs
+++ b/src/wayland/xdg_activation/dispatch.rs
@@ -87,7 +87,7 @@ where
 {
     fn request(
         state: &mut D,
-        _: &Client,
+        client: &Client,
         token: &xdg_activation_token_v1::XdgActivationTokenV1,
         request: xdg_activation_token_v1::Request,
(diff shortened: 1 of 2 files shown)


src/wayland/xdg_activation/dispatch.rs
```diff
@@ -87,7 +87,7 @@ where
 {
     fn request(
         state: &mut D,
-        _: &Client,
+        client: &Client,
```
i509VCB, line 90: I'd wonder would the `ClientId` be preferable or pass a whole `Client`? You can get the id from a `Client` but also do lookup on the `Client` you get.

Drakulix: The original client could disconnect before the token is used by another client. I didn't want to force the object to keep on living. In that sense it serves as an `Option<Client>`.

PolyMeilex (approved):

PR #1205, opened 2023-11-07
implement Hash for Serial

i509VCB:


Drakulix (commented):
Can't this be derived?

PR #1211, opened 2023-11-07
Improve backend/x11/mod docs

SpyrosRoum:
The documentation claims that X11Backend::new returns both X11Backend and X11Surface, but there is not X11Surface returned


Drakulix (approved):
Nice catch! Thanks :)

PR #1213, opened 2023-11-08
Use clippy to disallow e/print/ln macros

### LINKED ISSUE #1034: Configure clippy::disallowed_macros
We should forbid use of print/println/eprint/eprintln since we use tracing for logging.

We can configure clippy::disallowed_macros to do that.

SpyrosRoum:
closes #1034

### DIFF
--- a/anvil/src/main.rs
+++ b/anvil/src/main.rs
@@ -53,11 +53,14 @@ fn main() {
             tracing::error!("Unknown backend: {}", other);
         }
         None => {
-            println!("USAGE: anvil --backend");
-            println!();
-            println!("Possible backends are:");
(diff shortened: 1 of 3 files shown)


SpyrosRoum:
Didn't realize clippy runs with the top level config for anvil and smallvil.

I will switch `print` to tracing logs for those as part of this PR.

SpyrosRoum:
Ended up allowing print for usage messages in smallvil, I think it makes more sense to print help messages like that.

However if you think I should change them to tracing logs let me know

Drakulix (approved):
LGTM! Thanks. :)

PR #1218, opened 2023-11-09
keyboard: allow setting keymap directly

nicolasavru:
Also make the Layout.0 public to allow constructing Layouts from u32s.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -403,7 +403,7 @@ impl<'a> XkbContextHandler for KeysymHandle<'a> {
 ///
 /// The layout may become invalid after calling [`KeyboardHandle::set_xkb_config`]
 #[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Layout(xkb::LayoutIndex);
+pub struct Layout(pub xkb::LayoutIndex);
 
 /// Result for key input filtering (see [`KeyboardHandle::input`])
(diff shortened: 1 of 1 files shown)


Drakulix:
To be clear: I don't oppose this change.
But what is the use-case for it? It would be easier to review and judge if this should be part of the public api, if I would knew what exactly you do with it.

nicolasavru:
A passthrough compositor. In the wayland client part of my code, I'm getting the keymap as a string/file descriptor and the layout index from update_modifiers (https://github.com/Smithay/client-toolkit/pull/430 for that), and need to pass that information on to clients running against my compositor. So I have the full keymap file and can construct a Keymap, but I don't have the RMLVO string which is what compositors usually construct a keymap from.

kchibisov (approved):

Drakulix (approved):

ids1024:
https://github.com/Smithay/smithay/pull/750 has something similar. But I noticed the API is potentially unsound since `xkb::Keymap` is reference-counted, in a way that isn't thread safe. (See the unsafe `Send` impl in `src/input/keyboard/mod.rs`).

nicolasavru:
I can modify this api to accept a String instead and construct the keymap inside set_keymap with Keymap::new_from_string (that's how I'm constructing the Keymap outside the function currently). I'll send a pr for that in a few min.

nicolasavru:
#1221 for that change. Thanks for catching that.

PR #1219, opened 2023-11-09
input: Fix relative motion docs

bwidawsk:


src/input/pointer/mod.rs
```diff
@@ -239,9 +239,9 @@ impl<D: SeatHandler + 'static> PointerHandle<D> {
 
     /// Notify about relative pointer motion
     ///
-    /// This will internally send the appropriate button event to the client
-    /// objects matching with the currently focused surface, if the client uses
-    /// the relative pointer protocol.
+    /// The new location is relative to the current location in surface (focus) local coordinate
+    /// space. This will internally take care of notifying the appropriate client objects
+    /// of enter/motion/leave events.
```
Drakulix, line 244: I don't think relative_motion will cause `enter` or `leave` events and it will for certain not cause `motion` events.

Drakulix (changes requested):

Issue #1202, opened 2023-11-06
Abstract `smithay::desktop::Window` over XWayland surfaces


saihaze:
According to the comments in `smithay::desktop`, this is marked as "TODO".
Breaking API changes will be necessary. An example is `Window::toplevel()`, which will return an `Option` instead, as the window may be an X11 one.
I will do the job myself.

PR #1220, opened 2023-11-10
Make `smithay::desktop::Window` support XWayland windows

saihaze:
As mentioned in https://github.com/Smithay/smithay/issues/1202, `Window` is supposed to abstract over both Xdg toplevels and X11 surfaces. So now I have done it.
There is a breaking API change - `Window::toplevel()` now returns an `Option`. And the `new()` function will be deprecated, replaced by `new_wayland_window()` and `new_x11_window()`.
Fixes are applied to Smallvil and Anvil to make them compile and work. But to take full advantage of this PR, Anvil may need major refactor in the future.

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -108,7 +108,7 @@ impl<BackendData: Backend> AnvilState<BackendData> {
                 for element in self.space.elements() {
                     #[allow(irrefutable_let_patterns)]
                     if let WindowElement::Wayland(window) = element {
-                        let toplevel = window.toplevel();
+                        let toplevel = window.toplevel().unwrap();
                         let mode_changed = toplevel.with_pending_state(|state| {
                             if let Some(current_mode) = state.decoration_mode {
                                 let new_mode =
--- a/anvil/src/shell/grabs.rs
+++ b/anvil/src/shell/grabs.rs
@@ -311,7 +311,7 @@ impl<BackendData: Backend> PointerGrab<AnvilState<BackendData>> for ResizeSurfac
 
         match &self.window {
             WindowElement::Wayland(w) => {
-                let xdg = w.toplevel();
+                let xdg = w.toplevel().unwrap();
                 xdg.with_pending_state(|state| {
                     state.states.set(xdg_toplevel::State::Resizing);
                     state.size = Some(self.last_window_size);
@@ -355,7 +355,7 @@ impl<BackendData: Backend> PointerGrab<AnvilState<BackendData>> for ResizeSurfac
 
             match &self.window {
                 WindowElement::Wayland(w) => {
-                    let xdg = w.toplevel();
+                    let xdg = w.toplevel().unwrap();
                     xdg.with_pending_state(|state| {
                         state.states.unset(xdg_toplevel::State::Resizing);
                         state.size = Some(self.last_window_size);
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -242,7 +242,7 @@ fn ensure_initial_configure(surface: &WlSurface, space: &Space<WindowElement>, p
                     .initial_configure_sent
             });
             if !initial_configure_sent {
-                toplevel.toplevel().send_configure();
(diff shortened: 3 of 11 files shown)


colinmarc:
In my compositor, I have a similar abstraction, but I delay the creation of the wrapping object for an X11 window (and the mapping of the window) until the surface has been matched (until `wl_surface()` returns `Some`). 

I'm not sure how robust that is, but it seems to work, and I suspect there are significant benefits to holding the invariant that a mapped window always has a surface. I see that this just adds `unwrap` everywhere in Anvil for now, but my experience was that it's actually quite hard to do much with a window that doesn't have a surface anyway, so you end up pushing that "delay until the surface is matched" code into a bunch of places on the compositor side.

saihaze:
> 

> In my compositor, I have a similar abstraction, but I delay the creation of the wrapping object for an X11 window (and the mapping of the window) until the surface has been matched (until `wl_surface()` returns `Some`).
> 
> I'm not sure how robust that is (and tbc, I'm new to all this), but it seems to work, and I suspect there are significant benefits to holding the invariant that a mapped window always has a surface. I see that this just adds `unwrap` everywhere in Anvil for now, but my experience was that it's actually quite hard to do much with a window that doesn't have a surface anyway, so you end up pushing that "delay until the surface is matched" code into a bunch of places on the compositor side.

The original design of `toplevel()` assumes that the window must be a Wayland one. But an X11 window will never have a `ToplevelSurface`. That's why an `Option` is returned instead now.

colinmarc:
> The original design of `toplevel()` assumes that the window must be a Wayland one. But an X11 window will never have a `ToplevelSurface`. That's why an `Option` is returned instead now.

Right - to clarify, I'm talking about the `wl_surface` part, and the fact that X11Windows may or may not have a `wl_surface` yet when they get mapped, because of a race. I was suggesting a way to get around the following problematic pattern:

```rust
match tl {
    WindowSurface::Wayland((s, _)) => {
        // do something with surface
    }
    #[cfg(feature = "xwayland")]
    WindowSurface::X11(s) => {
        if let Some(surface) = &s.wl_surface() {
            // do something with surface. if it hasn't been matched with the X11Window yet, we might skip some important state updates.
        }
    }
}
```

But this is just a drive-by suggestion from someone who's still learning this stack, feel free to ignore me.

Drakulix:
> because of a race. 

It's also not because of a race. X11 clients can unmap an X11 window again, which will cause Xwayland to remove the wl_surface again. And it can also be mapped again, which will cause it to get another wl_surface!

This is part of the reason why I feel like `Window` abstracts very poorly over X11Window.

> Fixes are applied to Smallvil and Anvil to make them compile and work. But to take full advantage of this PR, Anvil may need major refactor in the future.

Yeah for anvil this large refactor would need to be part of this PR. If anvil isn't even remotely doing what is best practice using smithay, we might as well delete it.

Also at least the geometry code is painfully wrong, the two function have the same name, but the result haves widely different meanings and X11surface's geometry also needs to be updated to work correctly.

Additionally I see no way to access the X11Surface type from a window.

saihaze:
OK. Seems that this PR is a bad design. I will try something new.

Drakulix:
> OK. Seems that this PR is a bad design. I will try something new.

I don't think the PR is in general a bad idea. I just feel like the `Window` type needs many more changes for this to be usable.

saihaze:
> I don't think the PR is in general a bad idea. I just feel like the Window type needs many more changes for this to be usable.
Maybe I can introduce a new type for this? Name it like `GenericWindow`. It will be an Anvil `WindowElement` without server-side decorations. Then I will rewrite some parts of Anvil to take advantage of it.

Drakulix:
> > I don't think the PR is in general a bad idea. I just feel like the Window type needs many more changes for this to be usable.
> 
> Maybe I can introduce a new type for this? Name it like `GenericWindow`. It will be an Anvil `WindowElement` without server-side decorations. Then I will rewrite some parts of Anvil to take advantage of it.

Feel free to start with that, if that feels easier. Though I would like to eventually unify this into a single `Window` type.

PR #1221, opened 2023-11-10
keyboard: accept String instead of Keymap when setting keymap

nicolasavru:
Fixes soundness issue introduced in #1218.


Drakulix:
@ids1024 You pointed out the original soundness issue, would you mind taking a look at this?

ids1024 (approved):
This looks good to me. I'd agree that a `String` in the public API is the best option to do this safely currently. That's what I ended up doing in https://github.com/Smithay/client-toolkit/pull/299, though I didn't get around to updating my Smithay PR for that.

Drakulix (approved):

PR #1223, opened 2023-11-10
Change `PointerHandle.motion` and friends to take relative coords

colinmarc:
The existing API is helpful in that it makes it harder to mess up the math, but it's not flexible enough to support having different windows with different scaling factors. That's relevant if, for example, you want to force a scale factor of 1 for Xwayland windows.

The existing API takes `Option<(<D as SeatHandler>::PointerFocus, Point<i32, Logical>)>`, with the point being the surface origin, rather than the relative coordinates of the point on that surface. Then, internally, it converts into relative coordinates to generate the wayland event. A user familiar with wayland but not reading the documentation closely might think that the point should be the relative coords, not the surface origin (this happened to me).

Additionally, if you have different scale factors for different windows, it's impossible (or very difficult) to get the calculation right, since smithay is assuming that the coordinate space for the surface is the same as the global compositor space.

### DIFF
--- a/src/desktop/wayland/popup/grab.rs
+++ b/src/desktop/wayland/popup/grab.rs
@@ -20,7 +20,7 @@ use crate::{
         },
         SeatHandler,
     },
-    utils::{DeadResource, IsAlive, Logical, Point, Serial},
+    utils::{DeadResource, IsAlive, RelativePoint, Serial},
     wayland::{compositor::get_role, seat::WaylandFocus, shell::xdg::XDG_POPUP_ROLE},
 };
 
@@ -303,7 +303,10 @@ where
                 // We set the focus to root as this will make
                 // sure the grab will stay alive until the
                 // toplevel is destroyed or the grab is unset
-                focus: Some((root.into(), (0, 0).into())),
+                focus: Some(RelativePoint {
+                    surface: root.into(),
+                    loc: (0f64, 0f64).into(),
+                }),
                 location: (0f64, 0f64).into(),
             },
         }
@@ -547,7 +550,7 @@ where
         &mut self,
         data: &mut D,
         handle: &mut PointerInnerHandle<'_, D>,
-        focus: Option<(<D as SeatHandler>::PointerFocus, Point<i32, Logical>)>,
+        focus: Option<RelativePoint<D::PointerFocus>>,
         event: &MotionEvent,
     ) {
         if self.popup_grab.has_ended() {
@@ -565,7 +568,7 @@ where
                     .current_grab()
                     .as_ref()
                     .and_then(|f2| f2.wl_surface())
-                    .map(|s| f1.0.same_client_as(&s.id()))
+                    .map(|s| f1.surface.same_client_as(&s.id()))
             })
             .unwrap_or(false)
         {
@@ -579,7 +582,7 @@ where
         &mut self,
         data: &mut D,
         handle: &mut PointerInnerHandle<'_, D>,
-        focus: Option<(<D as SeatHandler>::PointerFocus, Point<i32, Logical>)>,
+        focus: Option<RelativePoint<D::PointerFocus>>,
         event: &RelativeMotionEvent,
     ) {
         handle.relative_motion(data, focus, event);
@@ -605,7 +608,7 @@ where
                 .and_then(|f| {
                     self.popup_grab
                         .current_grab()
-                        .and_then(|f2| f.0.wl_surface().map(|s| f2.same_client_as(&s.id())))
+                        .and_then(|f2| f.surface.wl_surface().map(|s| f2.same_client_as(&s.id())))
                 })
                 .unwrap_or(false)
         {
--- a/src/input/pointer/mod.rs
+++ b/src/input/pointer/mod.rs
@@ -8,8 +8,8 @@ use std::{
 use crate::{
     backend::input::{Axis, AxisSource, ButtonState},
     input::{Seat, SeatHandler},
-    utils::Serial,
     utils::{IsAlive, Logical, Point},
+    utils::{RelativePoint, Serial},
 };
 
 mod cursor_image;
@@ -216,19 +216,13 @@ impl<D: SeatHandler + 'static> PointerHandle<D> {
     /// You provide the new location of the pointer, in the form of:
     ///
     /// - The coordinates of the pointer in the global compositor space
-    /// - The surface on top of which the cursor is, and the coordinates of its
-    ///   origin in the global compositor space (or `None` of the pointer is not
-    ///   on top of a client surface).
+    /// - The surface the cursor is on top of, and the relative location of the
+    ///   cursor within that surface.
     ///
     /// This will internally take care of notifying the appropriate client objects
     /// of enter/motion/leave events.
-    #[instrument(level = "trace", parent = &self.span, skip(self, data, focus), fields(focus = ?focus.as_ref().map(|(_, loc)| ("...", loc))))]
-    pub fn motion(
-        &self,
-        data: &mut D,
-        focus: Option<(<D as SeatHandler>::PointerFocus, Point<i32, Logical>)>,
-        event: &MotionEvent,
-    ) {
+    #[instrument(level = "trace", parent = &self.span, skip(self, data, focus), fields(focus = ?focus.as_ref().map(|focus| ("...", focus.loc))))]
+    pub fn motion(&self, data: &mut D, focus: Option<RelativePoint<D::PointerFocus>>, event: &MotionEvent) {
         let mut inner = self.inner.lock().unwrap();
         inner.pending_focus = focus.clone();
         let seat = self.get_seat(data);
@@ -242,11 +236,11 @@ impl<D: SeatHandler + 'static> PointerHandle<D> {
     /// This will internally send the appropriate button event to the client
     /// objects matching with the currently focused surface, if the client uses
     /// the relative pointer protocol.
-    #[instrument(level = "trace", parent = &self.span, skip(self, data, focus), fields(focus = ?focus.as_ref().map(|(_, loc)| ("...", loc))))]
+    #[instrument(level = "trace", parent = &self.span, skip(self, data, focus), fields(focus = ?focus.as_ref().map(|focus| ("...", focus.loc))))]
     pub fn relative_motion(
         &self,
         data: &mut D,
-        focus: Option<(<D as SeatHandler>::PointerFocus, Point<i32, Logical>)>,
+        focus: Option<RelativePoint<D::PointerFocus>>,
         event: &RelativeMotionEvent,
     ) {
         let mut inner = self.inner.lock().unwrap();
@@ -435,7 +429,12 @@ where
 {
     /// Retrieve the current pointer focus
     pub fn current_focus(&self) -> Option<<D as SeatHandler>::PointerFocus> {
-        self.inner.lock().unwrap().focus.clone().map(|(focus, _)| focus)
+        self.inner
+            .lock()
+            .unwrap()
+            .focus
+            .clone()
+            .map(|focus| focus.surface)
     }
 }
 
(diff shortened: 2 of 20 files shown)


colinmarc:
I can't get anvil to run on my machine for (it appears unrelated) reasons, so I haven't tested that part.

It occurs to me that this will probably result in confusing behavior for existing compositors who update without knowing. I don't really know how to mitigate that.

colinmarc:
Hm, this is kinda lackluster. Taking another stab at this with a newtype to flush out cases I missed.

colinmarc:
Ok, I think this is much better now. I introduced a type, `RelativePoint`, to add semantic meaning to `Option<(<D as SeatHandler>::PointerFocus, Point<i32, Logical>)>`, which was a bit of a mouthful.

I find the new API is a bit more intuitive (I got confused that `motion` wanted the surface origin), but I know it's opinionated, so please tell me if you'd prefer a different approach. One disadvantage is that `Window.surface_under` still returns `Option((surface, surface_origin))`, but I added `RelativePoint::on_focused_surface(pos, surface, surface_origin)` to make it easy to follow the happy path without doing any math yourself. This has the extra benefit that the calculation `pos - surface_origin.to_f64()` is only done there, instead of all over the library.

src/desktop/wayland/popup/grab.rs
```diff
@@ -303,7 +303,10 @@ where
                 // We set the focus to root as this will make
                 // sure the grab will stay alive until the
                 // toplevel is destroyed or the grab is unset
-                focus: Some((root.into(), (0, 0).into())),
+                focus: Some(RelativePoint {
+                    surface: root.into(),
+                    loc: (0f64, 0f64).into(),
```
colinmarc, line 308: I wanted to highlight this as something I really wasn't sure about.

Drakulix:
> That's relevant if, for example, you want to force a scale factor of 1 for Xwayland windows.

Before giving this a proper review, note that I would love to have proper API in smithay to support overriding scale factors of windows and/or globally for Xwayland.

src/utils/relative_point.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[cfg(feature = "wayland_frontend")]
+impl<S: WaylandFocus> RelativePoint<S> {
+    /// Calculates the relative position of a cursor on a target surface,
+    /// given a point and that  surface's origin coordinates, both in global
+    /// compositor space.
+    pub fn on_focused_surface(
+        point: Point<f64, Logical>,
+        focus: S,
+        focus_origin: Point<i32, Logical>,
```
Drakulix, line 24: Please use `impl Into<Point<...>>` as arguments whenever possibly to allow for example a tuble to be passed to these functions directly instead of just `Point`s.

anvil/src/winit.rs
```diff
@@ -94,7 +94,7 @@ pub fn run_winit() {
     let (mut backend, mut winit) = match winit::init::<GlesRenderer>() {
         Ok(ret) => ret,
         Err(err) => {
-            error!("Failed to initialize Winit backend: {}", err);
+            error!("Failed to initialize Winit backend: {:?}", err);
```
Drakulix, line 97: This change seems unrelated? (Though a good one, that isn't worth it's own PR, please put it in a separate commit.)

src/input/pointer/mod.rs
```diff
@@ -714,44 +714,43 @@ impl<D: SeatHandler + 'static> PointerInternal<D> {
...
-        if let Some((surface, surface_location)) = focus {
+        if let Some(focus) = focus {
             let entered = self.focus.is_none();
             // in all cases, update the focus, the coordinates of the surface
             // might have changed
-            self.focus = Some((surface, surface_location));
+            self.focus = Some(focus.clone());
             let event = MotionEvent {
-                location: event.location - surface_location.to_f64(),
+                location: focus.loc.to_f64(),
```
Drakulix, line 745: So we are not using the `MotionEvent` location anymore at all? I quite dislike that to be honest, because that means we pass in multiple locations, but ignore some of them. In that case please change the signature of `motion` to **not** take a `MotionEvent`, but directly the `serial` and `time`.

Drakulix (commented):
Couple of nitpicks. I like the general approach, this seems more flexible without introducing too much friction in the api.

src/utils/relative_point.rs
```diff
@@ -0,0 +1,31 @@
+#[cfg(feature = "wayland_frontend")]
+use crate::wayland::seat::WaylandFocus;
+
+use super::{Logical, Point};
+
+/// Represents a point on a surface, relative to that surface's origin in
+/// global compositor space.
+#[derive(Debug, Clone)]
+pub struct RelativePoint<S> {
```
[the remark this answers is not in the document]
colinmarc: That's what I originally had, but I realized I had to move it out of the `input::pointer` mod because it's used by `grab` code. Once I moved it into `utils`, it seemed appropriate to give it more of a general name (`Focus` being an input-specific concept).

Happy to place/name it whatever you like, just wanted to add that context. My suggestion given your constraints would be `SurfaceCoordinate` or `SurfacePoint` or similar.

anvil/src/winit.rs
colinmarc: Whoops, that snuck in.

src/input/pointer/mod.rs
colinmarc: `event.location` is still used to set `self.location`, which is provided by `PointerHandle.current_location`, for example.

src/utils/relative_point.rs
colinmarc: Done in 180b08e1126884846f1c50e9b610babad110aec8.

colinmarc:
Sorry for the delay! Updated to address comments.

src/utils/relative_point.rs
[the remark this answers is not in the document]
Drakulix: > SurfacePoint

That is also a good one, but given a focused element doesn't have to be a surface, `RelativePoint` does seem to capture this quite well. 😅

src/input/pointer/mod.rs
Drakulix: > event.location is still used to set self.location

Right... Maybe we should then at least rename `MotionEvent::location` to `global_location` or `absolute_location` to further differenciate the two values?

colinmarc: Done in e3a992725396e8aefb088d9836c962e6512375a3.

colinmarc:
Renaming `MotionEvent.location` made me realize that this line may no longer be correct:

https://github.com/Smithay/smithay/blob/master/src/desktop/wayland/layer.rs#L729

But I'm not sure.

Drakulix:
> Renaming `MotionEvent.location` made me realize that this line may no longer be correct:
> 
> https://github.com/Smithay/smithay/blob/859ed44a34ea4fbb66349b3ae55aa7c67bc3d443/src/desktop/wayland/layer.rs#L729
> 
> But I'm not sure.

It is correct, the issue is that `PointerTarget::enter/motion` both get a `MotionEvent` with relative coordinates, while the `PointerHandle` and grabs don't. But both take the same `MotionEvent`-type, which means the last commit made the situation both better and worse, depending on which part of the API you look at.

And I think this the core issue, that makes this api feel so weird. We probably don't want to use the same type here, as this is confusing *at best*. This fits quite nicely with a discussion we recently had of introducing `Local` or `Relative` coordinates as opposed to `Global` coordinates in addition to the `Logical`/`Physical` coordinate spaces. I'll get back to you on that one.

colinmarc:
I took a stab at making `MotionEvent` generic over relative/global coordinate spaces:

```rust
#[derive(Debug)]
pub struct GlobalOrigin;

#[derive(Debug)]
pub struct SurfaceOrigin;

/// Pointer motion event
#[derive(Debug)]
pub struct MotionEvent<Origin> {
    /// Location of the pointer. If `Origin` is `SurfaceOrigin`, the coordinates
    /// are relative to the origin of the surface. If `Origin` is `GlobalOrigin`, it is a
    /// location in global compositor space.
    pub location: Point<f64, Logical>,
    /// Serial of the event
    pub serial: Serial,
    /// Timestamp of the event, with millisecond granularity
    pub time: u32,

    _origin: std::marker::PhantomData<Origin>,
}
```

The problem is still that lots of functions take an optional `focus`, but the `MotionEvent` is not optional:

```rust
    /// Notify that the pointer moved
    ///
    /// You provide the new location of the pointer, in the form of:
    ///
    /// - The surface that the cursor is on top of, and
    /// - The coordinates of the cursor relative to the origin of that surface
    ///
    /// A value of None indicates that no surface is focused, and the
    /// coordinates of the MotionEvent will be ignored.
    ///
    /// This will internally take care of notifying the appropriate client
    /// objects of enter/motion/leave events.
    #[instrument(level = "trace", parent = &self.span, skip(self, data, focus))]
    pub fn motion(&self, data: &mut D, focus: Option<D::PointerFocus>, event: &MotionEvent<SurfaceOrigin>) {
        // ...
    }
```

Here, if `focus` is `None`, then the `location` field of the event is completely ignored. It's the same problem, but subtly different, and even more finicky, because you need to construct a `MotionEvent` in the backend code of your compositor ([anvil example](https://github.com/Smithay/smithay/blob/master/anvil/src/input_handler.rs#L617-L621)), and it's not clear what should be set for `location` if it's a relative one and there's no focus.

I see two possibilities:

1. Ditching `MotionEvent` altogether, adding `serial` etc as arguments in their own right, and keeping `RelativePoint` (from the [current version](https://github.com/Smithay/smithay/pull/1223/commits/9b5411ac287e595b18444ffcfa574700a5d5bed6#diff-4d2f1d428c6ceedf69db74c16fda79aa7ce3e8317c08456336a67f67b7ebcd13) of this PR)
2. Removing `location` from `MotionEvent` but otherwise keeping this PR as-is.

I think I like 2 the best. `MotionEvent` keeps its semantic meaning with a timestamp and serial, which is to say it's semi-opaque and gets passed through everywhere. But the `RelativePoint` gets more relative as we go down the tree.

Drakulix:
> I see two possibilities:
> 
>     1. Ditching `MotionEvent` altogether, adding `serial` etc as arguments in their own right, and keeping `RelativePoint` (from the [current version](https://github.com/Smithay/smithay/pull/1223/commits/9b5411ac287e595b18444ffcfa574700a5d5bed6#diff-4d2f1d428c6ceedf69db74c16fda79aa7ce3e8317c08456336a67f67b7ebcd13) of this PR)
> 
>     2. Removing `location` from `MotionEvent` but otherwise keeping this PR as-is.
> 
> 
> I think I like 2 the best. `MotionEvent` keeps its semantic meaning with a timestamp and serial, which is to say it's semi-opaque and gets passed through everywhere. But the `RelativePoint` gets more relative as we go down the tree.

I agree, that this sounds like a good solution.

ryanabx:
> Removing location from MotionEvent but otherwise keeping this PR as-is.

> I think I like 2 the best. MotionEvent keeps its semantic meaning with a timestamp and serial, which is to say it's semi-opaque and gets passed through everywhere. But the RelativePoint gets more relative as we go down the tree.

Naively trying to understand this -- If MotionEvent doesn't have location, how do we gather where the pointer moved in `fn motion`...

colinmarc:
> Naively trying to understand this -- If MotionEvent doesn't have location, how do we gather where the pointer moved in `fn motion`...

The `RelativePoint` has a location, but it's not in the global compositor space, it's relative to the surface. Turns out some of the APIs (grab intercepting for example) rely on getting the global cursor location out of the PointerHandle, so I got stuck.

chrisduerr:
I don't think `RelativePoint` contributes much here, especially since every point is relative.

With `MotionEvent` things are still just as confusing anyway, since what you want is probably position relative to the touch down surface, while the focus will potentially be on a separate surface. So I'd say location should **only** be in the event and not coupled to the focus.

PR #1224, opened 2023-11-10
data_device: get clipboard selection instead of primary selection

nicolasavru:
This appears to have been a copy/paste error, the primary selection accessor is in selection/primary selection,


Drakulix (approved):
Thanks!

PR #1226, opened 2023-11-12
wayland(dmabuf): ImportNotifier for asynchronous buffer import

### LINKED ISSUE #699: Dmabuf import should be async
At the time of writing, Smithay assumes that reporting whether importing a dmabuf was successful is immediate.

However it may be desirable to respond asynchronously whether dmabuf import was successful for one of the following reasons:
- Rendering is performed on a different thread, meaning the wayland frontend would need to block until the renderer has imported the dmabuf.
- The compositor may use a different process for rendering, meaning there must be a roundtrip before import is completed.

https://github.com/Smithay/smithay/blob/d133bef922b2a878ca6264241c9e29cafe2ec896/src/wayland/dmabuf/dispatch.rs#L244-L265

One solution would be to also provide a handle in the signature of `DmabufHandler::import_dmabuf` which is used to signal success or failure.

i509VCB:
This pull request adds the ImportNotifier type, which allows dmabuf import results to be delayed until processed at a later time. In my case this would be after being handled through a channel but can be extended to allow creating an importing thread.

Resolves #699

A quick test by smalling alacritty in smallvil shows this appears to work correctly.

### DIFF
--- a/src/wayland/dmabuf/dispatch.rs
+++ b/src/wayland/dmabuf/dispatch.rs
@@ -1,7 +1,5 @@
 use std::sync::{atomic::AtomicBool, Mutex};
 
-use tracing::error;
-
 use wayland_protocols::wp::linux_dmabuf::zv1::server::{
     zwp_linux_buffer_params_v1, zwp_linux_dmabuf_feedback_v1, zwp_linux_dmabuf_v1,
 };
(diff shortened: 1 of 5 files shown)


src/wayland/dmabuf/dispatch.rs
```diff
@@ -330,28 +310,18 @@ where
...
-                        }
+                        // The buffer isn't technically valid during data_init, but the client is not allowed to use the buffer until ready.
+                        let buffer = data_init.init(buffer_id, dmabuf.clone());
+
+                        let notifier = ImportNotifier {
+                            inner: params.clone(),
+                            display: dh.clone(),
+                            dmabuf: dmabuf.clone(),
+                            import: Import::Infallible(buffer),
+                            drop_ignore: false,
```
Drakulix, line 321: Could we have a constructor for this?

Drakulix (approved):
LGTM in general.

src/wayland/dmabuf/dispatch.rs
i509VCB: I assume you mean an internal constructor?

Drakulix: Yes. `drop_ignore` for example is always `false` and doesn't need to be part of it.

Drakulix (approved):

PR #1228, opened 2023-11-13
Feature/software renderer

cmeissl:
This PR adds a new pixman based software renderer

TODO:

- [x] Docs
- [ ] ~~Anvil~~ will be done in a separate PR
- [x] pixman-rs release
- [x] cleanup drm fb exporter

### DIFF
--- a/src/backend/allocator/dmabuf.rs
+++ b/src/backend/allocator/dmabuf.rs
@@ -12,6 +12,7 @@
 
 use calloop::generic::Generic;
 use calloop::{EventSource, Interest, Mode, PostAction};
+use rustix::ioctl::{Setter, WriteOpcode};
 
 use super::{Allocator, Buffer, Format, Fourcc, Modifier};
 use crate::utils::{Buffer as BufferCoords, Size};
@@ -244,6 +245,152 @@ impl Dmabuf {
         let source = DmabufSource::new(self.clone(), interest)?;
         let blocker = DmabufBlocker(source.signal.clone());
         Ok((blocker, source))
+    }
+
+    /// Map the plane at specified index with the specified mode
+    ///
+    /// Returns `Err` if the plane with the specified index does not exist or
+    /// mmap failed
+    pub fn map_plane(
+        &self,
+        idx: usize,
+        mode: DmabufMappingMode,
+    ) -> Result<DmabufMapping, DmabufMappingFailed> {
+        let plane = self
+            .0
+            .planes
+            .get(idx)
+            .ok_or(DmabufMappingFailed::PlaneIndexOutOfBound)?;
+
+        let size = rustix::fs::seek(&plane.fd, rustix::fs::SeekFrom::End(0)).map_err(std::io::Error::from)?;
+        rustix::fs::seek(&plane.fd, rustix::fs::SeekFrom::Start(0)).map_err(std::io::Error::from)?;
+
+        let len = (size - plane.offset as u64) as usize;
+        let ptr = unsafe {
+            rustix::mm::mmap(
+                std::ptr::null_mut(),
+                len,
+                mode.into(),
+                rustix::mm::MapFlags::SHARED,
+                &plane.fd,
+                plane.offset as u64,
+            )
+        }
+        .map_err(std::io::Error::from)?;
+        Ok(DmabufMapping { len, ptr })
+    }
+
+    /// Synchronize access for the plane at the specified index
+    ///
+    /// Returns `Err` if the plane with the specified index does not exist or
+    /// the dmabuf_sync ioctl failed
+    pub fn sync_plane(&self, idx: usize, flags: DmabufSyncFlags) -> Result<(), DmabufSyncFailed> {
+        let plane = self
+            .0
+            .planes
+            .get(idx)
+            .ok_or(DmabufSyncFailed::PlaneIndexOutOfBound)?;
+        unsafe { rustix::ioctl::ioctl(&plane.fd, Setter::<DmaBufSync, _>::new(dma_buf_sync { flags })) }
+            .map_err(std::io::Error::from)?;
+        Ok(())
+    }
+}
+
+bitflags::bitflags! {
+    /// Modes of mapping a dmabuf plane
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+    pub struct DmabufMappingMode: u32 {
+        /// Map the dmabuf readable
+        const READ = 0b00000001;
+        /// Map the dmabuf writable
+        const WRITE = 0b00000010;
+    }
+}
+
+impl From<DmabufMappingMode> for rustix::mm::ProtFlags {
+    fn from(mode: DmabufMappingMode) -> Self {
+        let mut flags = rustix::mm::ProtFlags::empty();
+
+        if mode.contains(DmabufMappingMode::READ) {
+            flags |= rustix::mm::ProtFlags::READ;
+        }
+
+        if mode.contains(DmabufMappingMode::WRITE) {
+            flags |= rustix::mm::ProtFlags::WRITE;
+        }
+
+        flags
+    }
+}
+
+/// Dmabuf mapping errors
+#[derive(Debug, thiserror::Error)]
+#[error("Mapping the dmabuf failed")]
+pub enum DmabufMappingFailed {
+    /// The supplied index for the plane is out of bounds
+    #[error("The supplied index for the plane is out of bounds")]
+    PlaneIndexOutOfBound,
+    /// Io error during map operation
+    Io(#[from] std::io::Error),
+}
+
+/// Dmabuf sync errors
+#[derive(Debug, thiserror::Error)]
+#[error("Sync of the dmabuf failed")]
+pub enum DmabufSyncFailed {
+    /// The supplied index for the plane is out of bounds
+    #[error("The supplied index for the plane is out of bounds")]
+    PlaneIndexOutOfBound,
+    /// Io error during sync operation
+    Io(#[from] std::io::Error),
+}
+
+bitflags::bitflags! {
+    /// Flags for the [`Dmabuf::sync_plane`](Dmabuf::sync_plane) operation
+    #[derive(Copy, Clone)]
+    pub struct DmabufSyncFlags: std::ffi::c_ulonglong {
+        /// Read from the dmabuf
+        const READ = 1 << 0;
+        /// Write to the dmabuf
+        #[allow(clippy::identity_op)]
+        const WRITE = 2 << 0;
+        /// Start of read/write
+        const START = 0 << 2;
+        /// End of read/write
+        const END = 1 << 2;
+    }
+}
+
+#[repr(C)]
+#[allow(non_camel_case_types)]
+struct dma_buf_sync {
+    flags: DmabufSyncFlags,
+}
+
+type DmaBufSync = WriteOpcode<b'b', 0, dma_buf_sync>;
+
+/// A mapping into a [`Dmabuf`]
+#[derive(Debug)]
+pub struct DmabufMapping {
+    ptr: *mut std::ffi::c_void,
+    len: usize,
+}
+
+impl DmabufMapping {
+    /// Access the raw pointer of the mapping
+    pub fn ptr(&self) -> *mut std::ffi::c_void {
+        self.ptr
+    }
+
+    /// Access the length of the mapping
+    pub fn length(&self) -> usize {
+        self.len
+    }
+}
+
+impl Drop for DmabufMapping {
+    fn drop(&mut self) {
+        let _ = unsafe { rustix::mm::munmap(self.ptr, self.len) };
     }
 }
 
--- a/src/backend/allocator/dumb.rs
+++ b/src/backend/allocator/dumb.rs
@@ -9,7 +9,6 @@ use tracing::instrument;
 
 use super::dmabuf::{AsDmabuf, Dmabuf, DmabufFlags};
 use super::{format::get_bpp, Allocator, Buffer, Format, Fourcc, Modifier};
-use crate::backend::drm::device::{DrmDevice, DrmDeviceInternal};
 use crate::backend::drm::DrmDeviceFd;
 use crate::utils::{Buffer as BufferCoords, Size};
 
@@ -29,7 +28,20 @@ impl fmt::Debug for DumbBuffer {
     }
 }
 
-impl Allocator for DrmDevice {
+/// Light wrapper around an [`DrmDeviceFd`] to implement the [`Allocator`]-trait
+#[derive(Debug)]
+pub struct DumbAllocator {
+    fd: DrmDeviceFd,
+}
+
+impl DumbAllocator {
+    /// Create a new [`DumbAllocator`] from a [`DrmDeviceFd`].
+    pub fn new(fd: DrmDeviceFd) -> Self {
+        DumbAllocator { fd }
+    }
+}
+
+impl Allocator for DumbAllocator {
     type Buffer = DumbBuffer;
     type Error = io::Error;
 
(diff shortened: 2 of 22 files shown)


src/backend/renderer/pixman/mod.rs
```diff
@@ -0,0 +1,1208 @@
...
+}
+
+fn dmabuf_sync_read_write_end(fd: impl AsFd) -> rustix::io::Result<()> {
+    let ioctl = DmabufIoctlSync {
+        sync: dma_buf_sync {
+            flags: DmaBufSyncFlags::READ | DmaBufSyncFlags::WRITE | DmaBufSyncFlags::END,
+        },
+    };
+    unsafe { rustix::ioctl::ioctl(fd, ioctl) }
+}
```
Drakulix, line 109: Can we put this somewhere more accessible and perhaps even make it public?

cmeissl: sure, as this is related to dmabuf we could put it there I guess?

Drakulix: sounds good.

src/backend/renderer/pixman/mod.rs
```diff
@@ -0,0 +1,1208 @@
...
+struct DmabufIoctlSync {
+    sync: dma_buf_sync,
+}
+
+unsafe impl rustix::ioctl::Ioctl for DmabufIoctlSync {
+    type Output = ();
+
+    const OPCODE: rustix::ioctl::Opcode = rustix::ioctl::Opcode::write::<dma_buf_sync>(b'b', 0);
+
+    const IS_MUTATING: bool = true;
```
ids1024, line 61: If this is a "write" ioctl, and the kernel doesn't modify the data, this can be `false`.

But it looks like this is something the `rustix::ioctl::Setter` type could be used for, instead of manually implementing this trait.

```rust
type DmaBufSync = WriteOpcode<b'b', 0, dma_buf_sync>;
ioctl(fd, Setter<DmaBufSync, _>::new(dma_buf_sync { .. }));
```

src/backend/allocator/dmabuf.rs
```diff
@@ -251,6 +253,141 @@ impl Dmabuf {
...
         Ok((blocker, source))
     }
+
+    /// Map this dmabuf readable
+    pub fn map_readable(&self) -> Result<DmabufMapping, DmabufMappingFailed> {
+        self.map(rustix::mm::ProtFlags::READ).map(DmabufMapping)
+    }
+
+    /// Map this dmabuf read/writeable
+    pub fn map_writable(&self) -> Result<DmabufMappingMut, DmabufMappingFailed> {
```
Drakulix, line 263: So... does this violate rust's safety constraints? Or would the mmap call error out mapping the same region twice?

ids1024: Creating `MAP_SHARED` mmaps of fds in Rust is only safe if either
* The fd is sealed against writes, or we trust any process with access to the fd not to write to it
* A slice is never created from the pointer

If the client modifies the contents of the dmabuf or shm fd while we have a slice into it, that is, at least technically, undefined behavior.

cmeissl: > So... does this violate rust's safety constraints? Or would the mmap call error out mapping the same region twice?

Why would it error out when mapping the same fd twice, it creates a mmap for each call? Regarding the safety constraints, maybe. But if we want to map it mutable there is no way around that imo as we already have `Clone` on `Dmabuf`. So I did not feel the need to make the API more complicated.

> Creating MAP_SHARED mmaps of fds in Rust is only safe if either
>
>    The fd is sealed against writes, or we trust any process with access to the fd not to write to it
>    A slice is never created from the pointer
>
> If the client modifies the contents of the dmabuf or shm fd while we have a slice into it, that is, at least technically, undefined behavior.

In the pixman renderer it is actually easier to use the ptr to initialize a image, so I am fine with not providing a slice at all. But we need `MAP_SHARED` for writing to the dmabuf or otherwise kms won't see our changes. For the read-only case we could use `MAP_PRIVATE`.

Drakulix: > But if we want to map it mutable there is no way around that imo as we already have Clone on Dmabuf.

Right, because we can already have multiple references to the same dmabuf, we can't enforce a single mutable reference here. I was just hoping we could catch that at runtime at least.

> In the pixman renderer it is actually easier to use the ptr to initialize a image, so I am fine with not providing a slice at all.

Getting a pointer seems to be a better solution.

src/backend/allocator/dumb.rs
```diff
@@ -96,7 +96,10 @@ impl AsDmabuf for DumbBuffer {
 
     #[profiling::function]
     fn export(&self) -> Result<Dmabuf, Self::Error> {
-        let fd = self.fd.buffer_to_prime_fd(self.handle.handle(), 0)?;
+        let fd = self.fd.buffer_to_prime_fd(
+            self.handle.handle(),
+            (rustix::fs::OFlags::CLOEXEC | rustix::fs::OFlags::RDWR).bits(),
```
bwidawsk, line 110: small nit, but I think these kind of flags should be moved over to the DRM_* variants:

https://github.com/torvalds/linux/blob/master/include/uapi/drm/drm.h#L851

src/backend/allocator/dmabuf.rs
```diff
@@ -251,6 +252,112 @@ impl Dmabuf {
...
+
+    /// Sync access to this dmabuf
+    pub fn sync(&self, flags: DmaBufSyncFlags) -> std::io::Result<()> {
+        for plane in self.0.planes.iter() {
+            unsafe { rustix::ioctl::ioctl(&plane.fd, Setter::<DmaBufSync, _>::new(dma_buf_sync { flags })) }?;
+        }
+        Ok(())
+    }
+
+    fn map(&self, prot: rustix::mm::ProtFlags) -> Result<DmabufMapping, DmabufMappingFailed> {
```
bwidawsk, line 274: Perhaps name this `map_linear_plane` or something less generic?

src/backend/renderer/pixman/mod.rs
```diff
@@ -0,0 +1,1183 @@
...
+        &mut self,
+        texture: &Self::TextureId,
+        src: Rectangle<f64, BufferCoords>,
+        dst: Rectangle<i32, Physical>,
+        damage: &[Rectangle<i32, Physical>],
+        src_transform: Transform,
+        alpha: f32,
+    ) -> Result<(), Self::Error> {
+        let Some(target_image) = self.renderer.target.as_ref().map(|target| target.image()) else {
+            return Ok(());
```
Drakulix, line 347: Why is this no error?

src/backend/allocator/dumb.rs
cmeissl: Yeah, unfortunately neither the `drm` nor the `drm-sys` crate expose those.

src/backend/allocator/dmabuf.rs
cmeissl: Maybe it would be a good idea to let the caller decide and provide some linear helper.

Something like:

```
pub fn map_plane(idx: usize, mode: Mode) -> Result<DmabufMapping, DmabufMappingFailed> { ... }

pub fn map_linear(mode: Mode) -> Result<DmabufMapping, DmabufMappingFailed> {
    ...checks for single plane linear
    self.map(0, mode)
}
```

sync analog to it

@ids1024 @Drakulix opinions?

src/backend/allocator/dumb.rs
cmeissl: @Drakulix are you fine if I open an issue in `drm` and leave a `FIXME` in the code for now?

src/backend/drm/dumb.rs
```diff
@@ -0,0 +1,185 @@
...
+            )
+        })
+    } else {
+        drm.add_planar_framebuffer(&buffer, flags).map(|fb| (fb, format))
+    };
+
+    let (fb, format) = match ret {
+        Ok(fb) => fb,
+        Err(source) => {
+            let fourcc = format.code;
```
bwidawsk, line 160: I think it'd be worth putting a debug!() here since most/all drivers we use should support addfb2, and this fallback *shouldn't* happen

src/backend/renderer/pixman/error.rs
```diff
@@ -0,0 +1,68 @@
+use drm_fourcc::{DrmFourcc, DrmModifier};
+use thiserror::Error;
+
+use crate::backend::{allocator::dmabuf::DmabufMappingFailed, SwapBuffersError};
+
+#[cfg(feature = "wayland_frontend")]
+use wayland_server::protocol::wl_shm;
+
+/// Error returned during rendering using GL ES
```
bwidawsk, line 9: s/GL ES/Pixman

src/backend/renderer/pixman/mod.rs
```diff
@@ -0,0 +1,1183 @@
...
+use super::{
+    sync::SyncPoint, Bind, DebugFlags, ExportMem, Frame, ImportDma, ImportMem, Offscreen, Renderer, Texture,
+    TextureFilter, TextureMapping, Unbind,
+};
+
+mod error;
+
+pub use error::*;
+
+lazy_static::lazy_static! {
```
bwidawsk, line 44: I think if we're willing to uprev the rust dependency to 1.70.0, OnceCell should contain all the functionality you need.

src/backend/renderer/pixman/mod.rs
```diff
@@ -0,0 +1,1183 @@
...
+        match self {
+            PixmanTarget::Image { image, .. } => &image.0.image,
+            PixmanTarget::RenderBuffer(render_buffer) => &render_buffer.0,
+        }
+    }
+}
+
+/// Offscreen render buffer
+#[derive(Debug)]
+pub struct PixmanRenderBuffer(pixman::Image<'static, 'static>);
```
Drakulix, line 86: Can't we make this take a proper lifetime?
I find it kinda annoying, that for example you couldn't use this renderer in combination with [`softbuffer`](https://docs.rs/softbuffer), as there is no way to `bind` the slice it provides you to render into.

src/backend/allocator/dumb.rs
Drakulix: sure

src/backend/allocator/dmabuf.rs
Drakulix: Yeah I like that idea. In case you have a tiled DumbBuffer (for whatever reason) and want to copy to another device, that can read it (or a vulkan buffer, whatever honestly), it might be useful to have that option and not be forced to only map linear buffers.

src/backend/renderer/pixman/mod.rs
```diff
@@ -0,0 +1,1183 @@
...
+    pub fn new(dmabuf: Dmabuf) -> std::io::Result<Self> {
+        dmabuf.sync(DmaBufSyncFlags::START | DmaBufSyncFlags::READ)?;
+        Ok(Self { dmabuf })
+    }
+}
+
+impl Drop for DmabufReadGuard {
+    #[profiling::function]
+    fn drop(&mut self) {
+        if let Err(err) = self.dmabuf.sync(DmaBufSyncFlags::END | DmaBufSyncFlags::READ) {
```
bwidawsk, line 169: Should this have `WRITE` also in case something happened after a SYNC_WRITE causes a drop?

src/backend/renderer/pixman/mod.rs
```diff
@@ -0,0 +1,1183 @@
...
+    size: Size<i32, Physical>,
+
+    finished: AtomicBool,
+}
+
+impl<'frame> PixmanFrame<'frame> {
+    fn draw_solid_color(
+        &mut self,
+        dst: Rectangle<i32, Physical>,
+        damage: &[Rectangle<i32, Physical>],
```
bwidawsk, line 236: For Smithay's case, does forcing the user to specify damage make sense?

src/backend/renderer/pixman/mod.rs
cmeissl: I think I had that as an error, but removed it to align the behavior with the gles renderer which does not error out in this case either.

Drakulix: I think for the GlesRenderer this is the case, as we can have an implicit framebuffer (EGLSurface). For this one there is no default target, right?

src/backend/renderer/pixman/mod.rs
cmeissl: That would require changing the `Bind` trait I guess, not sure if this should be part of this PR?
Probably also requires a bigger refactor of the rendering traits in general. The renderer currently stores
the bind target in the renderer. So if we introduce some lifetime here we would also need to introduce that on the renderer.

src/backend/renderer/pixman/mod.rs
cmeissl: Ah, yeah, good point. I will reintroduce the error then

src/backend/allocator/dmabuf.rs
cmeissl: I added more generic `map_plane` and `sync_plane` functions. The pixman renderer still only supports single plane buffers

src/backend/renderer/pixman/error.rs
cmeissl: thanks, done

src/backend/renderer/pixman/mod.rs
cmeissl: Added the error back

src/backend/renderer/pixman/mod.rs
cmeissl: created a separate issue for this

src/backend/renderer/pixman/mod.rs
cmeissl: created a separate issue for this

src/backend/allocator/dumb.rs
cmeissl: done https://github.com/Smithay/drm-rs/issues/182

src/backend/renderer/pixman/mod.rs
cmeissl: The `DmabufReadGuard` is only used for imported dmabufs, so only ever reads and never issues a WRITE

src/backend/renderer/pixman/mod.rs
cmeissl: Yeah, this is the contract for damage tracking. The renderer is not allowed to draw outside of the supplied damage rect(s) or otherwise you will get artifacts. This can also occur for a partial occluded solid element when the element above is transparent and triggers a redraw.

src/backend/drm/dumb.rs
cmeissl: You mean a log message? So more like a `tracing::warn`?

src/backend/allocator/dumb.rs
Drakulix: Given this is now fixed in drm-rs, we might able to solve this in this PR after all? https://github.com/Smithay/smithay/pull/1228

src/backend/drm/dumb.rs
Drakulix: If a warn!, we should make sure to not send this on every occurrence, but once.

src/backend/allocator/dumb.rs
cmeissl: Looks like this would require a release of `drm-rs`, the released version does not expose the flags

src/backend/drm/dumb.rs
cmeissl: Added a `warn` wrapped in an `Once::call_once` to warn once about the legacy fallback path

cmeissl:
> Before I forget, can we please reexport the pixman crate either in the reexports module or in `backend::renderer::pixman` given some types are part of the public api? (E.g. in `From<pixman::Image>` for `PixmanRenderBuffer`)

Added under `re-exports`

src/backend/allocator/dumb.rs
cmeissl: fixed

Drakulix (approved):
LGTM!

PR #1231, opened 2023-11-15
Add separate keymap for virtual keyboard

### LINKED ISSUE #1222: Properly track keymap when virtual keyboard is in use
- Have dedicated input types for both normal input and
  Virtual keyboard. Each type should track its keymap
  and modifiers state separatelly. They should also
  register into some global type that they were
  the last to send keys for the given client.

- When sending input events for the client,
  smithay should compare old != new source
  and resend keymap with the modifiers being
  active on that keymap, and then forward the
  state.

The tracking should be per-seat, not per-client
since seat can focus only one client at a time
and virtual keyboard just improves on normal
keyboard.

chrisduerr:
Smithay's virtual keyboard implementation previously stored a single keymap on the seat, updating it whenever either the real or virtual keyboard's keymap was changed. As a result the two keyboards could break each other by setting an incorrect keymap.

To allow both the seat's real keyboard and the virtual keyboard to function simultaneously, a separate keymap is stored for each and whenever an input event related to a keymap is handled, the keymap is updated.

Closes #1222.

---

I'm not actually sure if this is a valid approach or not and it's definitely possible that I've missed updating the keymap in some places, but this minimal fix has been confirmed to fix the issue for me.

### DIFF
--- a/src/input/keyboard/keymap_file.rs
+++ b/src/input/keyboard/keymap_file.rs
@@ -1,15 +1,20 @@
-use crate::utils::sealed_file::SealedFile;
-use tracing::error;
-use xkbcommon::xkb::{Keymap, KEYMAP_FORMAT_TEXT_V1};
-
 use std::ffi::CString;
 use std::os::unix::io::{AsFd, BorrowedFd};
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use tracing::error;
+use xkbcommon::xkb::{self, Keymap, KEYMAP_FORMAT_TEXT_V1};
+
+use crate::utils::sealed_file::SealedFile;
+
+static NEXT_ID: AtomicUsize = AtomicUsize::new(0);
 
 /// Wraps an XKB keymap into a sealed file or stores as just a string for sending to WlKeyboard over an fd
 #[derive(Debug)]
 pub struct KeymapFile {
     sealed: Option<SealedFile>,
     keymap: String,
+    id: usize,
 }
 
 impl KeymapFile {
@@ -23,14 +28,19 @@ impl KeymapFile {
             error!("Error when creating sealed keymap file: {}", err);
         }
 
+        let id = NEXT_ID.fetch_add(1, Ordering::Relaxed);
+
         Self {
             sealed: sealed.ok(),
             keymap,
+            id,
         }
     }
 
     #[cfg(feature = "wayland_frontend")]
-    pub(crate) fn change_keymap(&mut self, keymap: String) {
+    pub(crate) fn change_keymap(&mut self, keymap: &Keymap) {
+        let keymap = keymap.get_as_string(xkb::KEYMAP_FORMAT_TEXT_V1);
+
         let name = CString::new("smithay-keymap-file").unwrap();
         let sealed = SealedFile::with_content(name, CString::new(keymap.clone()).unwrap());
 
@@ -38,6 +48,7 @@ impl KeymapFile {
             error!("Error when creating sealed keymap file: {}", err);
         }
 
+        self.id = NEXT_ID.fetch_add(1, Ordering::Relaxed);
         self.sealed = sealed.ok();
         self.keymap = keymap;
     }
@@ -76,5 +87,10 @@ impl KeymapFile {
         self.with_fd(keyboard.version() >= 7, |fd, size| {
             keyboard.keymap(KeymapFormat::XkbV1, fd, size as u32);
         })
+    }
+
+    /// Get this keymap's unique ID.
+    pub(crate) fn id(&self) -> usize {
+        self.id
     }
 }
--- a/src/input/keyboard/keymap_file.rs
+++ b/src/input/keyboard/keymap_file.rs
@@ -7,6 +7,7 @@ use xkbcommon::xkb::{self, Keymap, KEYMAP_FORMAT_TEXT_V1};
 
 use crate::utils::sealed_file::SealedFile;
 
+/// Keymap ID, uniquely identifying the keymap without requiring a full content hash.
 static NEXT_ID: AtomicUsize = AtomicUsize::new(0);
 
 /// Wraps an XKB keymap into a sealed file or stores as just a string for sending to WlKeyboard over an fd
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -560,7 +560,7 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
     #[instrument(parent = &self.arc.span, skip(self, data, keymap))]
     pub(crate) fn change_keymap(&self, data: &mut D, keymap: &xkb::Keymap) {
         let mut keymap_file = self.arc.keymap.lock().unwrap();
-        keymap_file.change_keymap(&keymap);
+        keymap_file.change_keymap(keymap);
 
         let mods = self.arc.internal.lock().unwrap().mods_state;
         self.send_keymap(data, &keymap_file, mods);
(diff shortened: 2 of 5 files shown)


src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -78,15 +87,23 @@ where
...
-                if !data.has_keymap.load(Ordering::Relaxed) {
-                    virtual_keyboard.post_error(NoKeymap, "`key` sent before keymap.")
-                }
+                // Ensure keymap was initialized.
+                let virtual_data = data.handle.inner.lock().unwrap();
+                let keymap = match &virtual_data.keymap {
+                    Some(keymap) => keymap,
+                    None => {
+                        virtual_keyboard.post_error(NoKeymap, "`key` sent before keymap.");
+                        return;
```
chrisduerr, line 99: This is technically a change in behavior, but I'm not entirely sure why the existing code didn't abort on this error to begin with. It feels like this should either be legal, or the request should be ignored, so this change seems like an improvement to me.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -108,10 +125,20 @@ where
...
+                    Some(keymap) => keymap,
+                    None => {
+                        virtual_keyboard.post_error(NoKeymap, "`modifiers` sent before keymap.");
+                        return;
+                    }
+                };
+
+                // Ensure virtual keyboard's keymap is active.
                 let keyboard_handle = data.seat.get_keyboard().unwrap();
+                keyboard_handle.change_keymap(keymap);
```
chrisduerr, line 140: Just to be 100% clear, I wasn't actually able to verify `Modifiers` and `Key` requests in virtual keyboard still work, because the virtual keyboard I use don't use this (instead it inputs text through IME I believe).

If this would trigger `KeyboardHandle::input` this wouldn't work since that function would reset the keymap to the normal keyboard's keymap again. But it seems like this goes straight to the client so it should be fine?

src/input/keyboard/mod.rs
```diff
@@ -710,7 +714,11 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
         F: FnOnce(&mut D, &ModifiersState, KeysymHandle<'_>) -> FilterResult<T>,
     {
         trace!("Handling keystroke");
+
+        // Ensure keymap is up to date.
         let mut guard = self.arc.internal.lock().unwrap();
+        self.change_keymap(&guard.keymap);
```
kchibisov, line 720: This is incorrect place to do so. You should do that sort of thing when input goes straight into `WlKeyboard` client, thus it should be after `Filter` and `Grab` logic (if the target of the `Grab` is `WlKeyboard`).

Right now after each key with IME that does grabbing, you're sending keymap event, so the input looks more like
```
[2260896.167] wl_keyboard@25.keymap(1, fd 25, 67458)
[2260896.185] wl_keyboard@13.keymap(1, fd 19, 67458)
[2260897.689] wl_keyboard@25.keymap(1, fd 26, 67458)
[2260897.712] wl_keyboard@25.key(121, 49745, 36, 0)
[2260900.771] wl_keyboard@13.keymap(1, fd 25, 67458)
[2260900.799] wl_keyboard@13.key(120, 49745, 36, 0)
[2261204.985] wl_keyboard@25.keymap(1, fd 25, 67458)
[2261205.478] wl_keyboard@13.keymap(1, fd 19, 67458)
[2261207.746] wl_keyboard@25.keymap(1, fd 26, 67458)
[2261207.766] wl_keyboard@25.key(124, 50053, 125, 1)
[2261207.775] wl_keyboard@25.modifiers(126, 64, 0, 0, 0)
[2261209.798] wl_keyboard@13.keymap(1, fd 25, 67458)
[2261209.813] wl_keyboard@13.key(123, 50053, 125, 1)
[2261209.820] wl_keyboard@13.modifiers(125, 64, 0, 0, 0)
```

And it obviously doesn't work at all, because modifiers are all messed up. The keymap change should also lead to modifiers change event for the focused client?

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
kchibisov: That's fine.

src/wayland/virtual_keyboard/virtual_keyboard_handle.rs
```diff
@@ -78,15 +87,23 @@ where
...
+                    Some(keymap) => keymap,
+                    None => {
+                        virtual_keyboard.post_error(NoKeymap, "`key` sent before keymap.");
+                        return;
+                    }
+                };
+
+                // Ensure virtual keyboard's keymap is active.
                 let keyboard_handle = data.seat.get_keyboard().unwrap();
+                keyboard_handle.change_keymap(keymap);
```
kchibisov, line 105: You must resend modifiers on keymap change.

kchibisov (changes requested):
From testing with IME that actually grabs and sends back keys it doesn't work at all.

To explain that, the input goes like that:
```
libinput -> Keyboard::Handle::input -> YOU_CHANGE_KEYMAP -> Grab -> Ime -> VirtualKeyboard -> compositor -> YOU_CHANGE_KEYMAP -> WlKeyboard
```

You don't have to change keymap in this particular case, you should only do so when you touch `WlKeyboard` interface, which you either do via `input` and after you resolve all the filters, grabbing logic or from `VirtualKeyboard`, so all of that should be tracked on a lower (?) level than it is now.

You should also refresh modifiers when doing a keymap changes for the focused client like that, but it's what smithay code lacks in general and what was missed with the keymap update function.

chrisduerr:
@kchibisov Is the latest revision closer to what you would have expected?

src/input/keyboard/mod.rs
```diff
@@ -530,31 +531,51 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
...
-        let keymap = keymap.get_as_string(xkb::KEYMAP_FORMAT_TEXT_V1);
+    pub(crate) fn change_keymap(&self, keymap: &xkb::Keymap) {
         let mut keymap_file = self.arc.keymap.lock().unwrap();
-        keymap_file.change_keymap(keymap);
+        keymap_file.change_keymap(&keymap);
+
+        self.send_keymap(&keymap_file);
+    }
 
+    /// Send a new wl_keyboard keymap, without updating the internal keymap.
```
kchibisov, line 564: Should probably document that it's no-op when the keymap is the same.

src/input/keyboard/keymap_file.rs
```diff
@@ -1,15 +1,20 @@
...
 use std::ffi::CString;
 use std::os::unix::io::{AsFd, BorrowedFd};
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use tracing::error;
+use xkbcommon::xkb::{self, Keymap, KEYMAP_FORMAT_TEXT_V1};
+
+use crate::utils::sealed_file::SealedFile;
+
+static NEXT_ID: AtomicUsize = AtomicUsize::new(0);
```
kchibisov, line 10: should probably specify why are we doing all of that.

kchibisov (changes requested):
That works more or less like I'd expect it and nothing changed for my input (like it should).

The only issue I see is the modifiers state when you alter the keymap, since they are related to the keymap you're using, so you should resend them for the active client. Like if you have Alt+Shift pressed on normal keyboard and type on virtual one without any modifiers, they shouldn't affect each other, imho. The same for normal keyboard input. The mods are also tied to the keymap.

PR #1235, opened 2023-11-23
Fix a logic bug in the egl backend

mrhh69:
I don't know if creating a whole new `wrap_egl_call_*` function would be worth it just to fix this, let me know what you think!


Drakulix (approved):
Looks like a correct fix to me!

(And no, I don't think this single special function requires it's own `wrap_egl_call`-variant.)

Thanks!

PR #1239, opened 2023-11-28
Use pixman renderer for DrmCompositor's pointer

bwidawsk:
Primary motivation of this PR is to remove the dependency on `Offscreen` in the DrmCompositor.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -202,7 +202,7 @@ impl<B: Buffer> From<UnderlyingStorage> for ScanoutBuffer<B> {
     fn from(storage: UnderlyingStorage) -> Self {
         match storage {
             UnderlyingStorage::Wayland(buffer) => Self::Wayland(buffer),
-            UnderlyingStorage::Memory(_) => todo!(),
+            UnderlyingStorage::Memory { .. } => todo!(),
         }
     }
 }
@@ -287,7 +287,7 @@ impl From<&UnderlyingStorage> for ElementFramebufferCacheBuffer {
     fn from(storage: &UnderlyingStorage) -> Self {
         match storage {
             UnderlyingStorage::Wayland(buffer) => Self::Wayland(buffer.downgrade()),
-            UnderlyingStorage::Memory(_) => todo!(),
+            UnderlyingStorage::Memory { .. } => todo!(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/backend/drm/compositor/mod.rs
```diff
@@ -2971,89 +2980,117 @@ where
...
+            UnderlyingStorage::Memory((data, format, width, height)) => {
+                let stride =
+                    width * (format::get_bpp(format).expect("Format with unknown bits per pixel") / 8);
+                let size = stride * height;
+
+                let Ok(pixman_format) = pixman::FormatCode::try_from(format) else {
+                    debug!("No pixman format for {format}");
+                    return None;
+                };
+                match pixman::Image::from_slice_mut(
```
cmeissl, line 3008: Instead of `from_slice_mut` with `slice::from_raw_parts_mut` it should be possible to use
`Image::from_raw_mut`

src/backend/drm/compositor/mod.rs
```diff
@@ -2971,89 +2980,117 @@ where
...
+        let ret = cursor_buffer
+            .map_mut::<_, _, Result<_, <PixmanRenderer as Renderer>::Error>>(
+                &cursor_state.framebuffer_exporter,
+                0,
+                0,
+                cursor_buffer_size.w as u32,
+                cursor_buffer_size.h as u32,
+                |mbo| {
+                    let plane_pixman_format = pixman::FormatCode::try_from(DrmFourcc::Argb8888).unwrap();
+                    let cursor_dst = pixman::Image::from_slice_mut(
```
cmeissl, line 3037: same here

src/backend/drm/compositor/mod.rs
```diff
@@ -2971,89 +2980,117 @@ where
...
+                    let mut frame = self.pixman_renderer.render(self.cursor_size, output_transform)?;
+                    frame.clear(
+                        [0f32, 0f32, 0f32, 0f32],
+                        &[Rectangle::from_loc_and_size((0, 0), self.cursor_size)],
+                    )?;
+                    frame.render_texture_from_to(
+                        &cursor_texture,
+                        src,
+                        dst,
+                        element.damage_since(scale, None).as_slice(),
```
cmeissl, line 3062: Using `damage_since` is probably unnecessary here, you can just use `&[Rectangle::from_loc_and_size((0,0), dst.size)]`

cmeissl (changes requested):

PR #1246, opened 2023-12-04
High res scrolling support

ids1024:
I've been thinking about how to handle high res scroll. Might as well hack together an initial implementation. Very WIP, but shows what is still needed here.

Testing things, it seems `PointerScrollWheelEvent` is emitted whenever a scroll is sent from the device, while `PointerAxisEvent` is only sent after accumulating change. So this impacts even clients not using the new protocol version, so they can get more precise scrolling from `wl_pointer::axis`.

In particular, this works in Firefox.

Issues:

* For clients not using the new protocol version, we need to accumulate `axis_value120` events into a `axis_discrete` event
  - Where can this be handed? Probably somewhere in Smithay and not the compositor?
    * Not obviously something libinput backend can deal with. Handle in `PointerTarget<D> for WlSurface` somehow?
* Instead of one `PointerAxisEvent` with an `AxisSource` enum, there are multiple `PointerScroll*` event types
  - Currently this only handles `PointerScrollWheelEvent`
  - `smithay::backend::input` currently models its API here on libinput. So do we want to remove our `AxisSource` and split `type PointerAxisEvent` into 3 types?
* Requires libinput 1.19
  - May be okay? Otherwise, have feature flag, and fallback to generating events from `PointerAxisEvent`
* Needs implementation for X11/winit backend. Ideally using high-res scroll when available.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ drm-ffi = { version = "0.7.0", optional = true }
 errno = "0.3.5"
 gbm = { version = "0.14.0", optional = true, default-features = false, features = ["drm-support"] }
 glow = { version = "0.12", optional = true }
-input = { version = "0.8.2", default-features = false, features=["libinput_1_14"], optional = true }
+input = { version = "0.8.2", default-features = false, features=["libinput_1_19"], optional = true }
 indexmap = "2.0"
 lazy_static = "1"
 libc = "0.2.103"
@@ -94,7 +94,6 @@ backend_udev = ["udev", "input/udev"]
 backend_vulkan = ["ash", "scopeguard"]
 backend_session_libseat = ["backend_session", "libseat"]
 desktop = []
-libinput_1_19 = ["input/libinput_1_19"]
 renderer_gl = ["gl_generator", "backend_egl"]
 renderer_glow = ["renderer_gl", "glow"]
 renderer_multi = ["backend_drm"]
@@ -104,7 +103,7 @@ use_bindgen = ["drm-ffi/use_bindgen", "gbm/use_bindgen", "input/gen"]
 wayland_frontend = ["wayland-server", "wayland-protocols", "wayland-protocols-wlr", "wayland-protocols-misc", "tempfile"]
 x11rb_event_source = ["x11rb"]
 xwayland = ["encoding_rs", "wayland_frontend", "x11rb/composite", "x11rb/xfixes", "x11rb_event_source", "scopeguard"]
-test_all_features = ["default", "use_system_lib", "renderer_glow", "libinput_1_19", "renderer_test"]
+test_all_features = ["default", "use_system_lib", "renderer_glow", "renderer_test"]
 
 [[example]]
 name = "minimal"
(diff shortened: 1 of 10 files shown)


ids1024:
I guess accumulation for `axis_discrete` could be stored in `SurfaceData::data_map`, and reset to 0 on `axis_stop` or `leave`.

Drakulix:
>     * For clients not using the new protocol version, we need to accumulate `axis_value120` events into a `axis_discrete` event
>       
>       * Where can this be handed? Probably somewhere in Smithay and not the compositor?
>         
>         * Not obviously something libinput backend can deal with. Handle in `PointerTarget<D> for WlSurface` somehow?

> I guess accumulation for axis_discrete could be stored in SurfaceData::data_map, and reset to 0 on axis_stop or leave.

Yeah, that seems like a good approach to me. I would love to handle of the backwards compatibility inside the specific pointer targets (e.g. WlSurface's implementation) and just use the more precise values all out through smithay. That way we also don't have to replicate the `_120` naming.

This might be a bit annoying for toolkit integrations like smithay-egui or cosmic-comp's iced code, but they should be able to copy the WlSurface implementation, if they need to. Maybe we can make this conversion/caching into a helper function?

> 
>     * Instead of one `PointerAxisEvent` with an `AxisSource` enum, there are multiple `PointerScroll*` event types
>       
>       * Currently this only handles `PointerScrollWheelEvent`
>       * `smithay::backend::input` currently models its API here on libinput. So do we want to remove our `AxisSource` and split `type PointerAxisEvent` into 3 types?

Yes probably.

>     * Requires libinput 1.19

Totally fine imo, even debian stable already has libinput 1.22.

>     * Needs implementation for X11/winit backend. Ideally using high-res scroll when available.

Given not all compositors support this (and I am not sure if Xwayland does and will ever), we need a way to generate the new events from the old values. I guess that shouldn't be too difficult though, no?

ids1024:
> Given not all compositors support this (and I am not sure if Xwayland does and will ever), we need a way to generate the new events from the old values. I guess that shouldn't be too difficult though, no?

Conversion in that direction is much easier, I think. Nothing needs to be accumulated, and the discrete values can just be multiplied by 120 to get the "v120" value.

ids1024:
This should be working well now.

Instead of adding 2 more associated types, I used an `enum` to wrap the different types of scroll events in the libinput backend. Otherwise it would complicate the input handling in the compositor unnecessarily.

Drakulix (commented):
I am not super sure, we want the renaming from `discrete`->`v120` now. Is this intended to introduce breakage on purpose?

I don't find the name particular descriptive and the 120 values are still discrete, so if we want to rename, we might as well do `amount_discrete_v120` at least, no?

ids1024:
>  Is this intended to introduce breakage on purpose?

Leaving the name the same could/will break behavior. At a minimum the code in Anvil/Smallvil/Cosmic-Comp tries to use `amount_discrete * 3` if `amount` isn't defined. Which definitely isn't right after this without multiplying that by 120, though  I'm not sure that fallback is generally needed. (Are there circumstances where `amount_v120` is defined, by `amount` isn't?)

Changing the name seems appropriate if users are going to have to change the code using it anyway, to have correct behavior after this change.

As for the specific naming, Wayland named the event `axis_value120`. https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/72 originally called it `axis_v120`, but changed it after some discussion about the name. Libinput uses the name `scroll_value_v120`. So `amount_v120` seems most consistent with libinput. `amount_value120` would be more consistent with Wayland, but I'm not sure that's particularly better.

Drakulix (approved):
Sounds good. Given I got nothing on the actual code, I am going to merge this then.

PR #1248, opened 2023-12-07
gles2: Fix bgra on 2.0 context

Drakulix:
gles 2.0 doesn't support 10-bit, but `fourcc_to_gl_formats` might still return `BGRA_EXT` for `Abgr8888`.


cmeissl (approved):
Oh, I fixed that for shm but missed import_mem. LGTM

Drakulix:
Ignoring the WLCS tests while the ICE is fixed upstream for now.

PR #1249, opened 2023-12-08
xwm: Handle data in `WM_CHANGE_STATE`; support unminimize

ids1024:
Before this, `unminimize_request` was defined but not called anywhere.

This should match how ICCCM expects this to be implemented, and how wlroots handles it.

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -2135,9 +2135,14 @@ fn handle_event<D: XwmHandler + 'static>(
                     }
                 }
                 x if x == xwm.atoms.WM_CHANGE_STATE => {
+                    let data = msg.data.as_data32();
                     if let Some(surface) = xwm.windows.iter().find(|x| x.window_id() == msg.window).cloned() {
                         drop(_guard);
-                        state.minimize_request(xwm_id, surface);
+                        match data[0] {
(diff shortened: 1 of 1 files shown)


src/xwayland/xwm/mod.rs
```diff
@@ -2135,9 +2135,14 @@ fn handle_event<D: XwmHandler + 'static>(
                     }
                 }
                 x if x == xwm.atoms.WM_CHANGE_STATE => {
+                    let data = msg.data.as_data32();
                     if let Some(surface) = xwm.windows.iter().find(|x| x.window_id() == msg.window).cloned() {
                         drop(_guard);
-                        state.minimize_request(xwm_id, surface);
+                        match data[0] {
+                            1 => state.unminimize_request(xwm_id, surface),
```
Drakulix, line 2142: Out of curiosity: What is value 2?

Drakulix (approved):
LGTM

src/xwayland/xwm/mod.rs
ids1024: I was wondering that too, but strangely it isn't mentioned in https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html:

```
WithdrawnState		0
NormalState		1
IconicState		3 
```

Maybe it was used for something historically.

PR #1251, opened 2023-12-12
WIP: Text input v2

rano-oss:
QT apps use text input v2 for input method. The initial thought is to support the requests we support in v3 and ignore the rest, if possible.


rano-oss:
Because of https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/39#note_2367477 I can close this.

PR #1253, opened 2023-12-15
Implement `axis_relative_direction` (`wl_pointer` version 9)

ids1024:
The relative direction is "inverted" for natural scrolling, and "identical" otherwise.

With this, `wl_seat` is now on the latest version.


src/backend/input/mod.rs
```diff
@@ -357,6 +379,9 @@ pub trait PointerAxisEvent<B: InputBackend>: Event<B> {
 
     /// Source of the scroll event.
     fn source(&self) -> AxisSource;
+
+    /// Relative direction of phsical motion.
```
YaLTeR, line 383:
```suggestion
    /// Relative direction of physical motion.
```

src/backend/libinput/mod.rs
```diff
@@ -191,11 +193,21 @@ impl backend::PointerAxisEvent<LibinputInputBackend> for PointerScrollAxis {
...
+            Self::Wheel(_) => AxisSource::Wheel,
+            Self::Finger(_) => AxisSource::Finger,
+            Self::Continuous(_) => AxisSource::Continuous,
+        }
+    }
+
+    fn relative_direction(&self, axis: Axis) -> backend::AxisRelativeDirection {
+        let device = backend::Event::<LibinputInputBackend>::device(self);
+        match axis {
+            Axis::Vertical if device.config_scroll_natural_scroll_enabled() => {
```
YaLTeR, line 207: Does it only apply to the vertical axis? The explanation of natural scrolling, "matching the motion of the scroll device with the direction of the content", does not read like it should be restricted to the vertical axis.

src/wayland/seat/pointer.rs
```diff
@@ -305,9 +305,21 @@ where
...
+                        WlAxis::HorizontalScroll,
+                        details.relative_direction.0.into(),
+                    );
+                }
                 ptr.axis(details.time, WlAxis::HorizontalScroll, details.axis.0);
             }
             if details.axis.1 != 0.0 {
+                if ptr.version() >= 9 {
+                    ptr.axis_relative_direction(
+                        WlAxis::HorizontalScroll,
```
YaLTeR, line 319:
```suggestion
                        WlAxis::VerticalScroll,
```
well well well if it isn't the copy-paste mistake again

src/backend/libinput/mod.rs
ids1024: Seems you're right, when I test the natural scrolling option in libinput for either a touchpad or mouse.

ids1024: The Kwin implementation in https://invent.kde.org/plasma/kwin/-/merge_requests/3456 has a comment `Maps to wl_pointer.axis_relative_direction. Currently used for y axis only`, but I'm not sure that's true in any was as implemented there...

anvil/src/input_handler.rs
```diff
@@ -394,6 +394,7 @@ impl<BackendData: Backend> AnvilState<BackendData> {
         {
             let mut frame = AxisFrame::new(evt.time_msec()).source(evt.source());
             if horizontal_amount != 0.0 {
+                frame = frame.relative_direction(Axis::Horizontal, evt.relative_direction(Axis::Horizontal));
```
Drakulix, line 397: So should this now also be done for the vertical code path?

ids1024: Oh, I thought I had that here already. Added.

Drakulix (approved):

PR #1254, opened 2023-12-18
Fix: toggle decorations in anvil

tobiaslieber:
Fixes that toggling decorations may reach `unreachable!()` and thus panic `anvil`.


Drakulix (approved):

Drakulix:
Thanks!

PR #1256, opened 2023-12-19
Popup unconstraining

YaLTeR:
(includes #1255 for easier testing)

Popup unconstraining, now in Smithay proper. Good enough to be used in Smallvil even!

See Smallvil and Anvil impl, also [here's](https://github.com/YaLTeR/niri/compare/main...popups) a niri impl. No reactive updates in Smallvil and Anvil, but they are straightforward to add (niri has one on commit).

Loosely based on cosmic-comp's impl, used with @Drakulix's permission.

All you need to do is pass the target rect in the positioner's coordinate system. It handles the rest. If you don't want resizes, simply unset the resize bits on the positioner.

### DIFF
--- a/anvil/src/shell/xdg.rs
+++ b/anvil/src/shell/xdg.rs
@@ -2,8 +2,8 @@ use std::cell::RefCell;
 
 use smithay::{
     desktop::{
-        find_popup_root_surface, layer_map_for_output, space::SpaceElement, PopupKeyboardGrab, PopupKind,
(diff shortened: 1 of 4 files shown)


Drakulix (approved):
Looks good to me. Certainly very clear and understandable.

Drakulix:
> (includes https://github.com/Smithay/smithay/pull/1255 for easier testing)

Now that #1255 is merged, could you drop that commit from this branch?

YaLTeR:
Done.

PR #1258, opened 2023-12-20
Make Relocate/Rescale/CropRenderElement fields public

chrisduerr:
This patch changes the visibility of render element helpers to allow modifying existing transforms instead of having to rely on excessive nesting.

---

Haven't added docs yet because I'm curious if this is desired and if the implementation is appropriate. I think for scale/transform element there aren't any major drawbacks, but cropping's constructor does a little more than just store the rect.

My usecase for this is to move a texture in the later stages of my rendering pipeline:

```rust
                    texture.0.element.location.x += self.quickswitch_offset as i32;
                    texture.0.crop_rect.loc.x += self.quickswitch_offset as i32;
```

### DIFF
--- a/src/backend/renderer/element/utils/elements.rs
+++ b/src/backend/renderer/element/utils/elements.rs
@@ -11,9 +11,9 @@ use crate::{
 /// A element that allows to re-scale another element
 #[derive(Debug)]
 pub struct RescaleRenderElement<E> {
-    element: E,
-    origin: Point<i32, Physical>,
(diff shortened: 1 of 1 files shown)


Drakulix (commented):
I don't think there is a strong reason to disallow this, assuming especially more complex values like `crop_rect` get appropriate documentation.

chrisduerr:
Closing since I don't personally need it anymore.

PR #1259, opened 2023-12-20
Do not offset popup coordinates by the parent's geometry

YaLTeR:
The popup positioner coordinate system is relative to the parent's geometry, so we should not subtract it.

### DIFF
--- a/src/desktop/wayland/popup/manager.rs
+++ b/src/desktop/wayland/popup/manager.rs
@@ -4,9 +4,7 @@ use crate::{
     wayland::{
         compositor::{get_role, with_states},
         seat::WaylandFocus,
-        shell::xdg::{
-            SurfaceCachedState, XdgPopupSurfaceData, XdgPopupSurfaceRoleAttributes, XDG_POPUP_ROLE,
-        },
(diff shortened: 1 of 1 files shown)


Drakulix (commented):
> The popup positioner coordinate system is relative to the parent's geometry, so we should not subtract it.

Well depends on how you implement it, positioning the popup relative to the (0,0) coordinate of the bounding box is supposedly correct this way. Anyway, this is something downstream can do itself, if it needs to and I think the updated docs also made this clear.

Have you checked, if this breaks anvil/smallvil in any way or if their popups were mispositioned before?

YaLTeR:
> Have you checked, if this breaks anvil/smallvil in any way or if their popups were mispositioned before?

They were mispositioned before, and are now correct. This is because `Space` has window coordinates at geometry origin, which is the same as the new function's behavior.

YaLTeR:
You can check with adwaita-1-demo (which has big CSD shadows). The popup is touching the edge of the target area (smallvil window) as it should:

image

Drakulix:
> You can check with adwaita-1-demo (which has big CSD shadows). The popup is touching the edge of the target area (smallvil window) as it should:
> 
> image

Great! Good to merge then.

Drakulix (approved):

PR #1260, opened 2023-12-21
Fix 1257

bwidawsk:
popups: Emit a wl_pointer.leave when destroyed
    
    Wayland protocol indicates that a leave event must be generated before
    any subsequent enter event. WLCS (but true in general) have several
    tests where a popup is destroyed and it is expected to leave and
    re-enter the below surface.
    
    Half of the issue is addressed here, when a popup is explicitly
    destroyed the leave event will be generated, though the subsequent enter
    will not be generated until there is pointer motion. The enter event can
    be simulated within the compositor via a 0 length move event, but the
    leave event has no such workaround.
    
    This issue is potentially present for surfaces and subsurfaces as well,
    but WLCS doesn't seem to care (that I've yet seen) about those - so
    that's deferred until needed.

### DIFF
--- a/src/wayland/shell/xdg/handlers/surface/popup.rs
+++ b/src/wayland/shell/xdg/handlers/surface/popup.rs
@@ -2,9 +2,9 @@ use std::sync::atomic::Ordering;
 
 use crate::{
     input::SeatHandler,
-    utils::Serial,
+    utils::{Serial, SERIAL_COUNTER},
     wayland::{
-        compositor,
+        compositor, seat,
         shell::xdg::{SurfaceCachedState, XdgPopupSurfaceData, XdgPositionerUserData},
     },
 };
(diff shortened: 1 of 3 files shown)


src/wayland/shell/xdg/handlers/surface/popup.rs
```diff
@@ -1,38 +1,61 @@
...
+                    // leave event must be generated before the next enter.
+                    let mut ids = vec![];
+                    let _ = dh.backend_handle().with_all_objects_for(client.id(), |id| {
+                        ids.push(id);
+                    });
+                    for id in ids {
+                        let Ok(seat) = WlSeat::from_id(dh, id) else {
+                            continue;
+                        };
+                        let seat: Seat<D> = Seat::from_resource(&seat).unwrap();
```
Drakulix, line 51: *Untested*, but I think we can simplify this by doing something like this:

```suggestion

                    // If a pointer has focus (entered the popup's surface), protocol states that a
                    // leave event must be generated before the next enter.
                    let seat_state = state.seat_state();
                    for seat in &seat_state.seats {
```

Other than that, looks fine to me.

Drakulix (approved):
LGTM

PR #1269, opened 2024-01-05
drm: disable element direct scan-out on submit failure

cmeissl:
This marks the primary plane test as failed in case submit or page flip fails. The next call(s) to `render_frame`
will then skip the element on the primary plane until some properties change (src, dst, format, ...).

Just to note, this should not be necessary as an atomic commit test should already catch that.
But unfortunately there are broken drivers where the test succeeds, but the actual commit/flip fails :/


cmeissl:
@YaLTeR I will wait for you confirmation before merging this.

Drakulix (approved):
Code-wise this looks good to me.

PR #1271, opened 2024-01-06
drm: copy element to cursor plane when possible

cmeissl:
this adds a fast-path by copying shm buffers directly to the cursor plane bo whenever possible

some quick profile with tracy:

rendering with gl:
Screenshot from 2024-01-12 18-08-48

copy directly to bo:
Screenshot from 2024-01-12 18-09-25

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -165,6 +165,7 @@ use crate::{
     },
     output::{OutputModeSource, OutputNoMode},
     utils::{Buffer as BufferCoords, DevPath, Physical, Point, Rectangle, Scale, Size, Transform},
+    wayland::shm,
 };
 
 use super::{error::AccessError, DrmDeviceFd, DrmSurface, Framebuffer, PlaneClaim, PlaneInfo, Planes};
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
LGTM. Though I wonder if we need this, when #1239 exists?
I suppose that gives the same performance benefits, but allows for transformations/scaling/cropping?

(Though this technically can render any element, though cursor planes are likely too slow for animated cursors, so the question becomes, if it *should* support non-memory elements..)

cmeissl:
> LGTM. Though I wonder if we need this, when #1239 exists? I suppose that gives the same performance benefits, but allows for transformations/scaling/cropping?
> 
> (Though this technically can render any element, though cursor planes are likely too slow for animated cursors, so the question becomes, if it _should_ support non-memory elements..)

#1239 currently has a hard dependency on pixman, which is probably fine for most cases, but I really would like to make
it optional (enabled by default). So the idea of this PR is to allow a simple subset (no scaling, transform, whatever) in all cases.

We already introduced `ElementKind` to prevent stuff from being assigned/rendered on the cursor plane. So imo it should be fine to also allow hardware accelerated clients on the cursor plane.

cmeissl:
rebased on `master` and added two more trace points so that we can easier compare rendering vs copy

Drakulix (approved):

PR #1273, opened 2024-01-08
WIP linux-dmabuf vesion 6 (target device)

ids1024:
Basic implementation of https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/268.

Can be tested along with https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/26912.

Will require compositor to do something with the `target_device` for this to be useful. So far this and the Mesa PR don't do anything too interesting, but should be a good start for trying to improve how multiple GPUs are handled.


ids1024:
Rebased and updated to use `set_node`.

PR #1274, opened 2024-01-10
get_dmabuf_formats: Always allow implict modifiers

ids1024:
https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt

My interpretation is that `DRM_FORMAT_MOD_INVALID` is always supported to indicate "effective format modifier is implementation-defined". And is not listed by `eglQueryDmaBufModifiersEXT`.

So import with implicit modifiers was broken, if the EGL driver supported modifiers.

This seems to fix AMDVLK on cosmic-comp, which still uses `wl_drm`.

### DIFF
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -862,16 +862,7 @@ fn get_dmabuf_formats(
             }
         };
 
-        if num == 0 {
-            texture_formats.insert(DrmFormat {
-                code: fourcc,
-                modifier: Modifier::Invalid,
-            });
-            render_formats.insert(DrmFormat {
-                code: fourcc,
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
I wonder if we could and or should limit this to wl_drm exposed formats. Anyway this is probably still sensible.

ids1024:
The dmabuf protocol also allows implicit modifiers, and Gnome's dmabuf feedback also lists "Invalid" modifiers like this. I'm not sure there's any reason to limit it to `wl_drm`.

Clients that are capable of using modifiers probably should, but it's reasonable/correct to allow this anyway.

PR #1276, opened 2024-01-13
drm: fix cursor buffer write error check

### LINKED ISSUE #1275: No cursor in anvil
In anvil, the cursor is not visible. (The mouse works, since I can mark text using the mouse.) 

I get the following warning on the first startup of anvil (at subsequent starts of anvil I do not get this error):
`anvil::cursor: Unable to load xcursor: Theme has no default cursor, using fallback cursor`

cmeissl:
This should fix #1275 by recognizing `gbm_bo_write` failed, but only fixes number 3 as mentioned in  https://github.com/Smithay/smithay/issues/1275#issuecomment-1890473451

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -3076,7 +3076,12 @@ where
             }
         };
 
-        if let Err(err) = cursor_buffer.write(data) {
+        let Ok(res) = cursor_buffer.write(data) else {
+            info!("failed to write cursor buffer, device destroyed");
+            return None;
+        };
+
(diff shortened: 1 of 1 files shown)


cmeissl:
@AlexD97 can you please verify if this makes the cursor visible for you? 
This will prevent any direct scan-out if a cursor is visible until number 2 is fixed from the linked comment.

AlexD97:
Yeah, the cursor is visible.

cmeissl:
> Yeah, the cursor is visible.

Thanks for testing!

Drakulix (approved):

PR #1277, opened 2024-01-13
drm: do not silently drop gbm bo usage flags

### LINKED ISSUE #1262: Cursor corruption on AMD
I've tried Smithay-based compositors with two systems, first with an RTX 2060 and then with an RX 7800 XT. The AMD-based GPU has a cursor corruption issue that the NVIDIA-based one doesn't.

The NVIDIA GPU is of note only because the cursor was working fine (though it was not working flawlessly); all attached images and descriptions of the problem are with the RX 7800 XT.

Most often, it is a square of noise, mostly white/gray but with a few colorful pixels.

image

Sometimes on startup, it is more of a blocky mess with mostly red/green bars and lots of transparency.

image

These are pictures with my phone because screenshots do not capture the rendering issue, here is how it looks when i take a screenshot using `niri`'s built-in `screenshot;` command.

image

It is *possible* for this corruption to get "better" after a while of running, because it did that once. By this, i mean that the default state of the cursor was restored with some slight artifacts and the other states were still equally corrupted.

---

I experienced this first in `niri` but have narrowed it down to a Smithay issue, given that it does not affect any other compositors i've tried (GNOME, Hyprland, Sway) except for `anvil`.

The lovely people in the smithay matrix channel helped me try some workarounds: `ANVIL_DISABLE_DRM_COMPOSITOR=1` fixes the issue on `anvil` and for `niri`, instructions equivalent to the following patch were provided:

```diff
diff --git a/src/backend/tty.rs b/src/backend/tty.rs
index 01a5305..d1d3d74 100644
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -594,7 +594,7 @@ impl Tty {
             SUPPORTED_COLOR_FORMATS,
             render_formats.clone(),
             device.drm.cursor_size(),
-            Some(device.gbm.clone()),
+            None,
         )?;
 
         let dmabuf_feedback = DmabufFeedbackBuilder::new(device.id, texture_formats.clone())
```

Compiling `niri` with the above patch prevents the issue from occurring.

cmeissl:
in case we miss support fo create_buffer_object_with_modifiers2 we can only fall back to create_buffer_object_with_modifiers if we do not have gbm bo usage flags.
if we got flags we can fall back to create_buffer_object for invalid and linear modifiers, but not for others.

fixes cursor usage flags which is required for the bo on the cursor plane.

fixes number 2 from here https://github.com/Smithay/smithay/issues/1275#issuecomment-1890473451

(I guess this, together with [!1276](https://github.com/Smithay/smithay/pull/1276) should also fix #1262)

### DIFF
--- a/src/backend/allocator/gbm.rs
+++ b/src/backend/allocator/gbm.rs
@@ -71,10 +71,19 @@ impl<A: AsFd + 'static> GbmAllocator<A> {
             modifiers.iter().copied(),
             flags,
         );
+
         #[cfg(not(feature = "backend_gbm_has_create_with_modifiers2"))]
-        let result =
+        let result = if (flags & !(GbmBufferFlags::SCANOUT | GbmBufferFlags::RENDERING)).is_empty() {
             self.device
-                .create_buffer_object_with_modifiers(width, height, fourcc, modifiers.iter().copied());
+                .create_buffer_object_with_modifiers(width, height, fourcc, modifiers.iter().copied())
+        } else if modifiers.contains(&Modifier::Invalid) || modifiers.contains(&Modifier::Linear) {
+            return self.device.create_buffer_object(width, height, fourcc, flags);
+        } else {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "unsupported combination of flags and modifiers",
+            ));
+        };
 
         match result {
             Ok(bo) => Ok(bo),


src/backend/allocator/gbm.rs
```diff
@@ -71,10 +71,23 @@ impl<A: AsFd + 'static> GbmAllocator<A> {
             modifiers.iter().copied(),
             flags,
         );
+
         #[cfg(not(feature = "backend_gbm_has_create_with_modifiers2"))]
-        let result =
+        let result = if flags.is_empty()
+            || flags == GbmBufferFlags::SCANOUT
+            || flags == GbmBufferFlags::RENDERING
+            || flags == GbmBufferFlags::SCANOUT | GbmBufferFlags::RENDERING
```
Drakulix, line 79:
```suggestion
        let result = if (flags & !(GbmBufferFlags::SCANOUT | GbmBufferFlags::RENDERING)).is_empty()
```

Drakulix (approved):
LGTM

src/backend/allocator/gbm.rs
cmeissl: much better, thanks! :)

PR #1279, opened 2024-01-13
Take two on delegate macros

PolyMeilex:
So... I was not happy with #1090, it was a bit hacky and required the macro crate to depend on smithay just for module definitions.

So this is take two:
The proc macro allows us to delegate so-called bundles like so:
```rs
struct State<T> {
    my: usize,
    state: usize,
    generic: T,
}

smithay_macros::delegate_bundle!(
    impl<T> State<T> {},
    Bundle {
        dispatch_to: smithay::wayland::shm::ShmState,
        globals: [
            Global {
                interface: smithay::reexports::wayland_server::protocol::wl_shm::WlShm,
                data: (),
            }
        ],
        resources: [
            Resource {
                interface: smithay::reexports::wayland_server::protocol::wl_shm::WlShm,
                data: (),
            },
        ],
    },
);
```

The macro is intentionaly built from valid rust syntax, and therefore this piece of code is fully rustfmt compatible.

Tho, one can opt out of rustfmt compatibility by dropping some optional pieces of syntax for example all of those are valid:
- `impl<T> State<T> {}`
- `impl<T> State<T>`
- `State` (only if your state is not generic)

So, now that we have a nice bundle macro, Smithay can just wrap it like so:
```rs
macro_rules! delegate_shm {
    ($($params:tt)*) => {
        $crate::reexports::smithay_macros::delegate_bundle!(
            $($params)*,
            Bundle {
                dispatch_to: $crate::wayland::shm::ShmState,
                globals: [
                    Global {
                        interface: $crate::reexports::wayland_server::protocol::wl_shm::WlShm,
                        data: (),
                    }
                ],
                resources: [
                    Resource {
                        interface: $crate::reexports::wayland_server::protocol::wl_shm::WlShm,
                        data: (),
                    },
                ],
            },
        );
    };
}
```

And end user can use it more or less like the old macro:
```rs
// Non generic case
delegate_shm!(State);

// Generic case
delegate_shm!(impl<T> State<T>);
```

# Future

Derive macro could call those macros with proper generics automatically, it could look like this:
```rs
#[derive(AutoDelegate)]
#[delegate(
    smithay::delegate_compositor,
    smithay::delegate_output,
    smithay::delegate_shm,
)]
struct Anvil<T: Backend> {
    backend: T,
}

// Any delegates that can't be done automatically:
delegate_dmabuf!(AnvilState<UdevData>);
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [workspace]
 members = [
     "smithay-drm-extras",
+    "smithay-macros",
     "smallvil",
     "anvil",
     "wlcs_anvil"
@@ -69,6 +70,7 @@ profiling = "1.0"
 smallvec = "1.11"
 pixman = { version = "0.1.0", features = ["drm-fourcc"], optional = true }
 
+smithay-macros = { path = "./smithay-macros" }
 
 [dev-dependencies]
 clap = { version = "4", features = ["derive"] }
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -168,7 +168,7 @@ pub struct AnvilState<BackendData: Backend + 'static> {
     pub show_window_preview: bool,
 }
 
-delegate_compositor!(@<BackendData: Backend + 'static> AnvilState<BackendData>);
+delegate_compositor!(impl<BackendData: Backend + 'static> AnvilState<BackendData>);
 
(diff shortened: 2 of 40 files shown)


PolyMeilex:
I don't want to jinx it, but I found a way to get rid of those macros altogether, with minimal changes to wayland-rs so this might not be needed at all :crossed_fingers:

(no I'm not working on wayland-rs 0.40 kind of situation :laughing:, this should be durable without braking changes)

PR #1280, opened 2024-01-14
Filter closures for decorations

YaLTeR:
I want my CSD on my SDL2 windows, but it's borked on all current releases: https://github.com/libsdl-org/SDL/issues/8173

Thus, I shall hide the globals.

### DIFF
--- a/src/wayland/shell/kde/decoration.rs
+++ b/src/wayland/shell/kde/decoration.rs
@@ -39,7 +39,7 @@ use wayland_protocols_misc::server_decoration::server::org_kde_kwin_server_decor
 };
 use wayland_server::backend::GlobalId;
 use wayland_server::protocol::wl_surface::WlSurface;
-use wayland_server::{Dispatch, DisplayHandle, GlobalDispatch, WEnum};
+use wayland_server::{Client, Dispatch, DisplayHandle, GlobalDispatch, WEnum};
 
 /// KDE server decoration handler.
 pub trait KdeDecorationHandler {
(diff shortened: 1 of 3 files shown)


src/wayland/shell/xdg/decoration.rs
```diff
@@ -103,18 +104,30 @@ pub struct XdgDecorationState {
...
+pub struct XdgDecorationManagerGlobalData {
+    filter: Box<dyn for<'c> Fn(&'c Client) -> bool + Send + Sync>,
+}
+
 impl XdgDecorationState {
     /// Creates a new delegate type for handling xdg decoration events.
     ///
     /// A global id is also returned to allow destroying the global in the future.
-    pub fn new<D>(display: &DisplayHandle) -> XdgDecorationState
+    pub fn new<D, F>(display: &DisplayHandle, filter: F) -> XdgDecorationState
```
cmeissl, line 117: imo it would be nice to keep the current path and extend `XdgDecorationState` with a `with_filter` constructor like other protocol abstractions already provide. In most cases there is not need to filter out this global and if I understood the linked issue correct you also "just" need this as a workaround for a borked client implementation.

src/wayland/shell/kde/decoration.rs
```diff
@@ -83,17 +83,28 @@ pub struct KdeDecorationState {
...
+/// Data associated with a KdeDecorationManager global.
+#[allow(missing_debug_implementations)]
+pub struct KdeDecorationManagerGlobalData {
+    pub(crate) filter: Box<dyn for<'c> Fn(&'c Client) -> bool + Send + Sync>,
+}
+
 impl KdeDecorationState {
     /// Create a new KDE server decoration global.
-    pub fn new<D>(display: &DisplayHandle, default_mode: DefaultMode) -> Self
+    pub fn new<D, F>(display: &DisplayHandle, default_mode: DefaultMode, filter: F) -> Self
```
cmeissl, line 94: same here

anvil/src/state.rs
```diff
@@ -560,7 +560,7 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
         let shm_state = ShmState::new::<Self>(&dh, vec![]);
         let viewporter_state = ViewporterState::new::<Self>(&dh);
         let xdg_activation_state = XdgActivationState::new::<Self>(&dh);
-        let xdg_decoration_state = XdgDecorationState::new::<Self>(&dh);
+        let xdg_decoration_state = XdgDecorationState::new::<Self, _>(&dh, |_| true);
```
cmeissl, line 563: when keeping the new function we do not need this change here

PolyMeilex (changes requested):
I totally agree with cmeissl on this one, let's keep it optional whenever possible. Especially if we known that this is more or less for a workaround.

I would go for:
```rs
fn new<D>(dh: &DisplayHandle) {
    new_with_filter::<D, _>(dh, || true)
}
fn new_with_filter<D, F>(DisplayHandle, F) {}
```

YaLTeR:
Good idea, done.

cmeissl (approved):
LGTM

PolyMeilex (approved):

PolyMeilex:
Thanks!

PR #1281, opened 2024-01-14
Asynchronous input filtering

roscale:
The internals of `KeyboardHandle::input` have been exposed as 2 separate functions `KeyboardHandle::input_intercept` and `KeyboardHandle::input_forward` to allow for asynchronously deciding if a key event should be forwarded to the client or not.

It was not possible to do this using `KeyboardHandle::input` because this decision must be done in the `filter` closure before returning the result of the decision.

I'm using an external toolkit that handles keyboard shortcuts and I cannot synchronously decide what key events should be forwarded or not. I have to send a message to another thread and wait for the response to come on the event loop. Blocking in the `filter` closure and waiting for an answer is not an option because I'm also blocking the event loop and I will end up in a deadlock.

By having 2 separate functions that split the functionality of `KeyboardHandle::input`, I can:

- call `KeyboardHandle::input_intercept` when I receive a key event
- ask the toolkit if the event is consumed or not
- when I get an answer, call `KeyboardHandle::input_forward`

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -726,65 +726,6 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
         }
     }
 
-    /// TODO
-    pub fn intercept<T, F>(
-        &self,
-        data: &mut D,
-        keycode: u32,
-        state: KeyState,
-        filter: F,
-    ) -> (T, bool)
-        where F: FnOnce(&mut D, &ModifiersState, KeysymHandle<'_>) -> T,
-    {
-        trace!("Handling keystroke");
-
-        let mut guard = self.arc.internal.lock().unwrap();
-        let mods_changed = guard.key_input(keycode, state);
-        let key_handle = KeysymHandle {
-            // Offset the keycode by 8, as the evdev XKB rules reflect X's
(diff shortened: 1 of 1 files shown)


src/input/keyboard/mod.rs
```diff
@@ -751,6 +751,28 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
     ) -> Option<T>
     where
         F: FnOnce(&mut D, &ModifiersState, KeysymHandle<'_>) -> FilterResult<T>,
+    {
+        let (filter_result, mods_changed) = self.intercept(data, keycode, state, filter);
+        if let FilterResult::Intercept(val) = filter_result {
+            // the filter returned false, we do not forward to client
```
PolyMeilex, line 757: If we are moving this comment around we might as well fix it while at it, filter no longer returns bool so `the filter returned false` is outdated.

PolyMeilex (approved):
In general I like the idea.
Personally I would probably go for `input_intercept()` and `input_forward()` to be more explicit. 
But yeah that's probably a matter of preference so feel free to ignore that.

PolyMeilex:
Thanks!

PR #1284, opened 2024-01-14
logging intro

### LINKED ISSUE #1282: Write introduction on logging
When we switched to `tracing` I left a [TODO](https://github.com/Smithay/smithay/blob/164855cf4bb7c95a677f829a3392bd39a312d80e/src/lib.rs#L49).
Before creating a release we should fill that TODO.

cmeissl:
~depends on #1283~ merged

fixes #1282

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,20 @@
 //!
 //! ### Logging
 //!
-//! TODO: docs about tracing
+//! Smithay makes extensive use of [`tracing`] for its internal logging.
+//!
+//! For release builds it is recommended to limit the log level during compile time.
+//! This can be done by adding a dependency to [`tracing`] and enabling the corresponding features.
+//! For example to enable `trace` messages for debug builds, but limit release builds to `debug` add
+//! the following in your binary crate `Cargo.toml`:
+//!
+//! ```toml
+//! [dependencies]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -48,7 +48,17 @@
 //!
 //! ### Logging
 //!
-//! TODO: docs about tracing
+//! Smithay makes extensive use of [`tracing`] for it's internal logging.
```
YaLTeR, line 51:
```suggestion
//! Smithay makes extensive use of [`tracing`] for its internal logging.
```

cmeissl: done, thanks!

YaLTeR (approved):

Drakulix:
Could we maybe add something like:

If you do not want to use tracing for your compositor, refer to [`log` compatibility](https://docs.rs/tracing/latest/tracing/#log-compatibility) for how to forward smithays debug output to other `log` compatible frameworks.

cmeissl:
> Could we maybe add something like:
> 
> If you do not want to use tracing for your compositor, refer to [`log` compatibility](https://docs.rs/tracing/latest/tracing/#log-compatibility) for how to forward smithays debug output to other `log` compatible frameworks.

Sure, this is definitely a good idea, added.

Drakulix (approved):

PR #1285, opened 2024-01-15
Track entered surfaces in Output

### LINKED ISSUE #1138: Surfaces don't get WlOutput.enter() if they bind the global late
If I call `window.output_enter(output)` and the surface binds the global only afterwards, then it will never get the enter event on that global. This can happen for example when a compositor places existing windows on a newly connected monitor, calling `output_enter()` on them, before the windows had a chance to bind the global.

YaLTeR:
Fixes #1138

### DIFF
--- a/src/desktop/space/wayland/mod.rs
+++ b/src/desktop/space/wayland/mod.rs
@@ -1,10 +1,7 @@
-use std::{
-    cell::{RefCell, RefMut},
-    collections::{HashMap, HashSet},
-};
+use std::{cell::RefCell, collections::HashMap};
 
-use tracing::{instrument, trace};
-use wayland_server::{protocol::wl_surface::WlSurface, Resource, Weak as WlWeak};
+use tracing::instrument;
+use wayland_server::protocol::wl_surface::WlSurface;
 
 use crate::{
     backend::renderer::utils::RendererSurfaceStateUserData,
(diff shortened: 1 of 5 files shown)


cmeissl:
tbh I am not happy with creating a dependency from the generic wayland abstractions to the desktop module.
this can be quite surprising, hard to find and won't help if you do not use the desktop module.

YaLTeR:
How should this be done instead? Maybe `Output` itself should keep a list of surfaces that are currently "entered" on it, and send the events to those? That should make it work for everyone I think?

Drakulix:
> There already are output.enter()/leave() but they have the same problem that if you use them, you don't know when a new output is bound. Ideally Output itself should keep track of it.

Right, which is why those methods shouldn't just send the event, but internally track surfaces.

YaLTeR:
So I just replace OutputSurfacesUserdata with the same tracking on the output itself?

YaLTeR:
Did that, how does that look? There's no `Output::cleanup()` but desktop did it on enter/leave, so I just preserved that.

Drakulix (commented):
Looks like a proper fix to me. @cmeissl any additional thoughts?

cmeissl:
Nope, really happy how that turned out, thanks!

Drakulix (approved):

Issue #1263, opened 2024-01-03
Own `EGLDisplay` around `GpuManager::add_node()` drops the internal one


YaLTeR:
When using `GpuManager` you might have code like this:

```rust
let display = EGLDisplay::new(gbm.clone())?;
let egl_device = EGLDevice::device_for_display(&display)?;
let render_node = egl_device.try_get_render_node()??;
self.gpu_manager.as_mut().add_node(render_node, gbm.clone())?;

```

Unfortunately, when `display` is dropped at the end of the scope, it will also close `GpuManager`'s internal `EGLDisplay` since they seem to refer to the same object. This will break everything.

Workaround:

```rust
let display = EGLDisplay::new(gbm.clone())?;
let egl_device = EGLDevice::device_for_display(&display)?;

// HACK: There's an issue in Smithay where the display created by GpuManager will be the
// same as the one we just created here, so when ours is dropped at the end of the scope,
// it will also close the long-lived display in GpuManager. Thus, we need to drop ours
// beforehand.
drop(display);

let render_node = egl_device.try_get_render_node()??;
self.gpu_manager.as_mut().add_node(render_node, gbm.clone())?;

```

cc @Drakulix

cmeissl:
fixed by #1288

PR #1288, opened 2024-01-15
egl: Fix multiple references to same EGLDisplay

Drakulix:
Attempt at "fixing" (at least in the context of smithay) #1263 as reported by @YaLTeR.

Essentially this works around this part of the [`eglGetPlatformDisplay`-docs](https://registry.khronos.org/EGL/sdk/docs/man/html/eglGetPlatformDisplay.xhtml) as well as we can:

> Multiple calls made to eglGetPlatformDisplay with the same parameters will return the same EGLDisplay handle. 

There is no way to disallow other code to call `eglGetPlatformDisplay` behind our back, so I felt like this method warrents an `unsafe`. However this is debatable.

1. We do have places where `EGLDisplay::new` is called inside safe code, which is difficult to propagate (e.g. `EglGlesBackend::enumerate`).
2. Calling `eglGetPlatformDisplay` requires unsafe code in other code or libraries anyway.
3. `EGLDisplay`s broken this way will normally not crash the program, but calls using it will all error out.

Another possible solution would be to make calling `eglTerminate` explicit?

### DIFF
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -1,10 +1,10 @@
 //! Type safe native types for safe egl initialisation
 
 use std::ffi::{c_int, CStr};
+use std::hash::{Hash, Hasher};
 use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::sync::Arc;
-#[cfg(all(feature = "wayland_frontend", feature = "use_system_lib"))]
 use std::sync::{Mutex, Weak};
 use std::{
     collections::HashSet,
@@ -39,6 +39,9 @@ use tracing::{debug, error, info, info_span, instrument, trace, warn};
 lazy_static::lazy_static! {
     pub(crate) static ref BUFFER_READER: Mutex<Option<WeakBufferReader>> = Mutex::new(None);
 }
+lazy_static::lazy_static! {
+    static ref DISPLAYS: Mutex<HashSet<WeakEGLDisplayHandle>> = Mutex::new(HashSet::new());
+}
 
 /// Wrapper around [`ffi::EGLDisplay`](ffi::egl::types::EGLDisplay) to ensure display is only destroyed
 /// once all resources bound to it have been dropped.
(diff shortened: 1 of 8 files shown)


kchibisov:
As I side note because there's an interest in using glutin for that.

This issue is kind of solved since we encourage to pass `Display` which is `Arc` under the hood, so the entire thing is ref-counted. We don't disallow building the display again though, but we likely should.

>Another possible solution would be to make calling eglTerminate explicit?

I'd suggest to ref-count, there's even a EGL extension to do so, but it's not widely available. And to never call the `eglTerminate` if you initial display from other display (that's how most libraries do that, like gstreamer).


Also, if you're into safety, you may accept owned `T: HasDisplayHandle` and force the ref-count by that.

Drakulix:
> As I side note because there's an interest in using glutin for that.
> 
> This issue is kind of solved since we encourage to pass `Display` which is `Arc` under the hood, so the entire thing is ref-counted. We don't disallow building the display again though, but we likely should.

I mean it's ref-counted in smithay now as well, the issue is, that somebody could *also* use glutin in their compositor. So yes, long term it would definitely better to settle on a common EGL implementation.

> I'd suggest to ref-count, there's even a EGL extension to do so, but it's not widely available. And to never call the `eglTerminate` if you initial display from other display (that's how most libraries do that, like gstreamer).

That is a good call for our `FromRaw` method.

> Also, if you're into safety, you may accept owned `T: HasDisplayHandle` and force the ref-count by that.

Also a good idea, but I feel like introducing `raw-window-handle` in smithay in general is a bigger PR.

src/backend/egl/display.rs
```diff
@@ -84,9 +93,9 @@ pub struct EGLDisplay {
 }
 
 unsafe fn select_platform_display<N: EGLNativeDisplay + 'static>(
-    native: &N,
+    native: N,
     dp_extensions: &[String],
-) -> Result<(*const c_void, &'static str), Error> {
+) -> Result<(Arc<EGLDisplayHandle>, &'static str), Error> {
```
cmeissl, line 98: Would this benefit from internal ref-counting on `EGLDisplayHandle` instead of returning and using `Arc`?

src/backend/egl/display.rs
```diff
@@ -103,44 +112,82 @@ unsafe fn select_platform_display<N: EGLNativeDisplay + 'static>(
...
-            ffi::egl::GetPlatformDisplayEXT(
-                platform.platform,
-                platform.native_display,
-                platform.attrib_list.as_ptr(),
-            )
-        })
-        .map_err(Error::DisplayCreationError);
+        let mut hasher = DefaultHasher::new();
+        platform.hash(&mut hasher);
+        let hash = hasher.finish();
```
cmeissl, line 117: While unlikely couldn't that have collisions resulting in wrongly tracking displays, especially with something like the `DefaultHasher` that could change with a rust release?

From the docs it sounds like `eglGetPlatformDisplay` already tracks this correctly and would return the same handle, so couldn't we just keep calling `eglGetPlatformDisplay` and
test for equality on the returned handle?

Drakulix: > While unlikely couldn't that have collisions resulting in wrongly tracking displays, especially with something like the DefaultHasher that could change with a rust release?

Yeah good point. This is the result of me initially wanting to hash the parameters, but not being able to store the `EGLPlatform` inside the `HashMap` due to the native pointer contained.

If we just do a `HashSet<Arc<EGLDisplayHandle>>` instead and keep calling `getPlatformDisplay` I also don't have to change the signature of `select_platform_display`.

src/backend/egl/display.rs
```diff
@@ -39,19 +39,53 @@ use tracing::{debug, error, info, info_span, instrument, trace, warn};
...
 
+#[derive(Clone)]
+struct WeakEGLDisplayHandle {
+    handle: Weak<EGLDisplayHandle>,
+    ptr: u64,
+}
+
+impl From<Arc<EGLDisplayHandle>> for WeakEGLDisplayHandle {
+    fn from(other: Arc<EGLDisplayHandle>) -> Self {
+        let ptr = other.handle as *const c_void as u64;
```
cmeissl, line 67: Can we keep the original type instead of casting this to u64? imo it should be fine to use `ffi::egl::types::EGLDisplay`  for the ptr in `WeakEGLDisplayHandle` and implement `Send` for it.

Drakulix: Good call. Done. Also implemented `Sync` as we do that for the `EGLDisplayHandle` as well and there is no reason why it should be different for the weak variant.

src/backend/egl/native.rs
```diff
@@ -80,6 +82,14 @@ pub struct EGLPlatform<'a> {
...
 }
 
+impl<'a> Hash for EGLPlatform<'a> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.platform.hash(state);
+        self.native_display.hash(state);
+        self.attrib_list.hash(state);
+    }
+}
+
```
cmeissl, line 92: nit, it seems this change is no longer required

Drakulix: removed

PR #1290, opened 2024-01-16
multigpu: Allow api to trigger re-enumeration

Drakulix:
Right now the `GpuManager` only re-enumerates devices, if the user requests a node it doesn't know already.

This means it will lock resources, even if e.g. the node is removed from the `GbmGlesBackend`.

While this is fine for the `EglGlesBackend`, as there really is no notification of a removed rendering device (except for egl errors trying to use it) and enumeration being expensive, this is bad for the Gbm-backend, as for example the nvidia driver will keep the gpu active as long as EGL resources do exist.

As such introduce a way for the api to trigger re-enumeration on the next render step.

### DIFF
--- a/src/backend/renderer/multigpu/gbm.rs
+++ b/src/backend/renderer/multigpu/gbm.rs
@@ -30,7 +30,11 @@ use crate::{
 use gbm::Device as GbmDevice;
 #[cfg(all(feature = "wayland_frontend", feature = "use_system_lib"))]
 use std::borrow::BorrowMut;
-use std::{collections::HashMap, os::unix::prelude::AsFd};
+use std::{
+    collections::HashMap,
+    os::unix::prelude::AsFd,
+    sync::atomic::{AtomicBool, Ordering},
+};
 
 /// Errors raised by the [`GbmGlesBackend`]
 #[derive(Debug, thiserror::Error)]
@@ -62,6 +66,7 @@ pub struct GbmGlesBackend<R> {
     devices: HashMap<DrmNode, EGLDisplay>,
     factory: Option<Factory>,
     context_priority: Option<ContextPriority>,
+    needs_enumeration: AtomicBool,
     _renderer: std::marker::PhantomData<R>,
 }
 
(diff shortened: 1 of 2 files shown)


src/backend/renderer/multigpu/gbm.rs
```diff
@@ -129,6 +143,8 @@ impl<R: From<GlesRenderer> + Renderer<Error = GlesError>> GraphicsApi for GbmGle
     type Error = Error;
 
     fn enumerate(&self, list: &mut Vec<Self::Device>) -> Result<(), Self::Error> {
+        self.wants_enumeration.store(true, Ordering::SeqCst);
```
cmeissl, line 146: This probably was meant to clear `wants_enumeration` by storing `false`?

src/backend/renderer/multigpu/mod.rs
```diff
@@ -637,6 +640,10 @@ pub trait GraphicsApi {
     ///
     /// Existing devices are guranteed to be not recreated
     fn enumerate(&self, list: &mut Vec<Self::Device>) -> Result<(), Self::Error>;
+    /// Method to force a re-enumeration, e.g. to free resources
+    fn wants_enumeration(&self) -> bool {
```
cmeissl, line 644: If this has to force a re-enumeration maybe we should call the function just `force_enumeration`? Or alternative `needs_enumeration` which is a bit less strong.

cmeissl (changes requested):

src/backend/renderer/multigpu/gbm.rs
Drakulix: Yup. Copy&Paste mistake.

src/backend/renderer/multigpu/mod.rs
Drakulix: `force_enumeration` sounds like it does what `enumerate` does, just with skipping some checks. `needs_enumeration` sounds better to me.

cmeissl (approved):
LGTM

PR #1291, opened 2024-01-16
Fix deadlock in send_keymap when called form input

kchibisov:
Pass the actually required types we need locks for instead of ensuring that callers release the locks everywhere.

Fixes: 629975f714 (Fix deadlocks when updating xkb state)
Fixes: 5a0b08c804 (Add separate keymap for virtual keyboard)
Links: https://github.com/YaLTeR/niri/issues/103

--

cc @chrisduerr @YaLTeR


Drakulix (approved):
LGTM

ids1024:
This does seem to fix a deadlock with clients that use the virtual-keyboard protocol.

PR #1294, opened 2024-01-19
pixman rendered cursor plane

cmeissl:
Rebased and slightly modified version of #1239 

Main difference from #1239 is that pixman is still optional (enabled by default).

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -20,7 +20,7 @@ use smithay::backend::drm::compositor::PrimaryPlaneElement;
 #[cfg(feature = "egl")]
 use smithay::backend::renderer::ImportEgl;
 #[cfg(feature = "debug")]
-use smithay::backend::renderer::ImportMem;
+use smithay::backend::renderer::{multigpu::MultiTexture, ImportMem};
 use smithay::{
     backend::{
         allocator::{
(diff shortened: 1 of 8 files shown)


src/backend/renderer/element/mod.rs
```diff
@@ -100,6 +101,21 @@ pub enum UnderlyingStorage {
     /// A wayland buffer
     #[cfg(feature = "wayland_frontend")]
     Wayland(Buffer),
+    /// A memory backed buffer
+    Memory {
+        /// Data of the underlying buffer
+        data: *const u8,
```
Drakulix, line 107: So I don't think I have seen any null-ptr checks for this anywhere.

I assume there is a problem with using a proper lifetime here? If so, that's fine, but can we make this at least a little bit more safe?

Drakulix (commented):
Looks pretty good for a draft!

src/backend/renderer/element/mod.rs
cmeissl: It is only used to get a reference to the underlying storage of an element. If the element can't provide
some underlying storage it should return `None` and never a null-ptr. I can add some checks for that, but
I don't they are necessary and may also just be some debug_asserts.

Yeah, we can not (easily) introduce a proper lifetime handling here. What we could do imo, is to introduce
some small wrapper around some memory representing pixel data that can be cloned around. It could
be made COW and used in the memory element and returned in the underlying storage. That would be
way safer as it should guarantee that the memory is accessible for as long as you hold a reference to it.

Drakulix: >  It could
be made COW and used in the memory element and returned in the underlying storage. That would be
way safer as it should guarantee that the memory is accessible for as long as you hold a reference to it.


That sounds like a better solution. I am fine, if you prefer to do that in a follow up PR though.

cmeissl: okay, I drafted that and pushed a commit showing off the `MemoryBuffer`. 
If this is an acceptable solution I would fixup the other commits with these changes.

Drakulix: Looks like a good approach to me! :+1:

anvil/src/udev.rs
```diff
@@ -1441,19 +1441,16 @@ impl AnvilState<UdevData> {
                 }
             })
             .unwrap_or_else(|| {
-                let texture = TextureBuffer::from_memory(
-                    &mut renderer,
+                let buffer = MemoryRenderBuffer::from_slice(
                     &frame.pixels_rgba,
-                    Fourcc::Abgr8888,
+                    Fourcc::Argb8888,
```
Drakulix, line 1446: Are you certain this had to be swapped? `pixels_rgba` reads like we already had the correct order.

Drakulix (approved):

PR #1295, opened 2024-01-20
fix various drm issues

### LINKED ISSUE #1289: Monitors not lighting up DRM problems
cc @cmeissl 

I'm on a laptop with iGPU + dGPU, however for this test everything is connected to and rendered on the dGPU. Sometimes when switching TTY to my compositor, some monitors (usually external, but I saw the internal one too) fail to light up, or fail to start working. Switching TTY to a vconsole or to GDM always works even in those cases.

### Case 1

Compositor disconnects and then reconnects connectors upon TTY switch. Code: https://github.com/YaLTeR/niri/blob/e4643c6dbea64304c9e8b074699227e8654c989d/src/backend/tty.rs#L323-L345

Switching TTY fails to light up the monitor. Compositor log:
1. Disconnect succeeds.
2. Connect fails, probably at the `drm.create_surface()` call.

Replugging the monitor lights it up as normal.

DRM debug: [drm_debug-fail_to_turn_on-reconnect.log](https://github.com/Smithay/smithay/files/13946443/drm_debug-fail_to_turn_on-reconnect.log)

Points of interest:
```
[37255.783697] [drm:dm_crtc_helper_atomic_check [amdgpu]] Can't enable a CRTC without enabling the primary plane
[37255.784001] amdgpu 0000:03:00.0: [drm:drm_atomic_helper_check_planes] [CRTC:95:crtc-1] atomic driver check failed
[37255.784005] [drm:amdgpu_dm_atomic_check [amdgpu]] drm_atomic_helper_check_planes() failed
[37255.784289] [drm:amdgpu_dm_atomic_check [amdgpu]] Atomic check failed with err: -22
```
```
[37255.784703] amdgpu 0000:03:00.0: [drm:drm_atomic_helper_check_modeset] [CRTC:91:crtc-0] enabled/connectors mismatch
[37255.784706] [drm:amdgpu_dm_atomic_check [amdgpu]] drm_atomic_helper_check_modeset() failed
[37255.784987] [drm:amdgpu_dm_atomic_check [amdgpu]] Atomic check failed with err: -22
[37255.785268] amdgpu 0000:03:00.0: [drm:drm_atomic_check_only] atomic driver check for 00000000c31d0d76 failed: -22
```
```
[37255.829572] amdgpu 0000:03:00.0: [drm:drm_atomic_helper_check_modeset] [CRTC:91:crtc-0] mode changed
[37255.829577] amdgpu 0000:03:00.0: [drm:drm_atomic_helper_check_modeset] [CRTC:95:crtc-1] enabled/connectors mismatch
[37255.829468] [drm:gfx_v10_0_eop_irq [amdgpu]] IH: CP EOP
[37255.829581] [drm:amdgpu_dm_atomic_check [amdgpu]] drm_atomic_helper_check_modeset() failed
[37255.829935] [drm:amdgpu_ih_process [amdgpu]] amdgpu_ih_process: rptr 254016, wptr 254048
[37255.830079] [drm:amdgpu_dm_atomic_check [amdgpu]] Atomic check failed with err: -22
[37255.830621] amdgpu 0000:03:00.0: [drm:drm_atomic_check_only] atomic driver check for 00000000dd04f02a failed: -22
```
```
[37255.834543] amdgpu 0000:03:00.0: [drm:drm_atomic_helper_check_modeset] [CRTC:91:crtc-0] mode changed
[37255.834547] amdgpu 0000:03:00.0: [drm:drm_atomic_helper_check_modeset] [CRTC:95:crtc-1] enabled/connectors mismatch
[37255.834550] [drm:amdgpu_dm_atomic_check [amdgpu]] drm_atomic_helper_check_modeset() failed
[37255.834783] [drm:amdgpu_ih_process [amdgpu]] amdgpu_ih_process: rptr 254176, wptr 254208
[37255.835026] [drm:amdgpu_dm_atomic_check [amdgpu]] Atomic check failed with err: -22
[37255.835240] [drm:gfx_v10_0_eop_irq [amdgpu]] IH: CP EOP
[37255.835569] amdgpu 0000:03:00.0: [drm:drm_atomic_check_only] atomic driver check for 00000000dd04f02a failed: -22
```

### Case 2

Compositor calls device changed code upon TTY switch. Code: https://github.com/YaLTeR/niri/blob/e4643c6dbea64304c9e8b074699227e8654c989d/src/backend/tty.rs#L347-L360

Currently what I'm seeing is that it fails with two external monitors connected, but succeeds with just one. Replugging the monitor sometimes lights up both monitors, sometimes lights up just one (interestingly enough, not the one that is replugged), then upon second replug lights up the second one too.

Compositor log:
```
янв 16 10:50:53 niri[155321]: 2024-01-16T06:50:53.341558Z DEBUG niri::backend::tty: resuming session
янв 16 10:50:53 niri[155321]: 2024-01-16T06:50:53.603043Z DEBUG niri::backend::tty: device changed: 57856
янв 16 10:50:53 niri[155321]: 2024-01-16T06:50:53.603137Z DEBUG niri::backend::tty: device changed: 57857
янв 16 10:50:53 niri[155321]: 2024-01-16T06:50:53.619300Z ERROR niri::backend::tty: error queueing frame: The underlying drm surface encountered an error: Atomic Test failed for new properties on crtc (crtc::Handle(95))
янв 16 10:50:53 niri[155321]: 2024-01-16T06:50:53.619948Z ERROR niri::backend::tty: error queueing frame: The underlying drm surface encountered an error: Atomic Test failed for new properties on crtc (crtc::Handle(99))
янв 16 10:50:53 niri[155321]: 2024-01-16T06:50:53.634316Z ERROR niri::backend::tty: error queueing frame: The underlying drm surface encountered an error: Atomic Test failed for new properties on crtc (crtc::Handle(95))
янв 16 10:50:55 niri[155321]: 2024-01-16T06:50:55.918110Z ERROR niri::backend::tty: error queueing frame: The underlying drm surface encountered an error: Atomic Test failed for new properties on crtc (crtc::Handle(95))
```
That is, all that code succeeds, then every time we try to draw the monitor, it fails.

DRM debug:  [drm_debug-fail_to_turn_on-change-two_external_monitors.log](https://github.com/Smithay/smithay/files/13946460/drm_debug-fail_to_turn_on-change-two_external_monitors.log)

## DRM Info

- [drm_info-two_external_monitors-working.log](https://github.com/Smithay/smithay/files/13946470/drm_info-two_external_monitors-working.log)
- [drm_info-two_external_monitors-broken.log](https://github.com/Smithay/smithay/files/13946472/drm_info-two_external_monitors-broken.log): DRM info during the broken state, before replugging
- [drm_info-one_external_monitor-working.log](https://github.com/Smithay/smithay/files/13946473/drm_info-one_external_monitor-working.log): DRM info when it works, after replugging

cmeissl:
should fix #1289

### DIFF
--- a/src/backend/drm/device/mod.rs
+++ b/src/backend/drm/device/mod.rs
@@ -116,6 +116,7 @@ pub struct DrmDevice {
     cursor_size: Size<u32, Buffer>,
     resources: ResourceHandles,
     plane_claim_storage: PlaneClaimStorage,
+    surfaces: Vec<Weak<DrmSurfaceInternal>>,
 }
 
 impl AsFd for DrmDevice {
@@ -226,6 +227,7 @@ impl DrmDevice {
                 cursor_size,
                 resources,
                 plane_claim_storage: Default::default(),
+                surfaces: Default::default(),
             },
             DrmDeviceNotifier {
                 internal,
@@ -310,11 +312,13 @@ impl DrmDevice {
     ///     attached to a crtc in smithay.
     #[instrument(skip(self), parent = self.internal.span(), err)]
     pub fn create_surface(
-        &self,
+        &mut self,
         crtc: crtc::Handle,
         mode: Mode,
         connectors: &[connector::Handle],
     ) -> Result<DrmSurface, Error> {
+        self.surfaces.retain(|surface| surface.upgrade().is_some());
+
         if connectors.is_empty() {
             return Err(Error::SurfaceWithoutConnectors(crtc));
         }
@@ -365,12 +369,14 @@ impl DrmDevice {
                 connectors,
             )?)
         };
+        let internal = Arc::new(internal);
+        self.surfaces.push(Arc::downgrade(&internal));
 
         Ok(DrmSurface {
             dev_id: self.dev_id,
             crtc,
             planes,
-            internal: Arc::new(internal),
+            internal,
             plane_claim_storage: self.plane_claim_storage.clone(),
         })
     }
@@ -390,8 +396,9 @@ impl DrmDevice {
     /// This will cause the `DrmDevice` to avoid making calls to the file descriptor e.g. on drop.
     /// Note that calls directly utilizing the underlying file descriptor, like the traits of the `drm-rs` crate,
     /// will ignore this state. Use [`DrmDevice::is_active`] to guard these calls.
-    pub fn pause(&self) {
+    pub fn pause(&mut self) {
         self.set_active(false);
+        self.surfaces.retain(|surface| surface.upgrade().is_some());
         if self.device_fd().is_privileged() {
             if let Err(err) = self.release_master_lock() {
                 error!("Failed to drop drm master state Error: {}", err);
@@ -399,14 +406,24 @@ impl DrmDevice {
         }
     }
 
-    /// Actives a previously paused device.
-    pub fn activate(&self) {
+    /// Activates a previously paused device.
+    ///
+    /// Specifying `true` for `disable_connectors` will call [`DrmDevice::reset_state`] if
+    /// the device was not active before. Otherwise you need to make sure there are no
+    /// conflicting requirements when enabling or creating surfaces or you are prepared
+    /// to handle errors caused by those.
+    pub fn activate(&mut self, disable_connectors: bool) -> Result<(), Error> {
         if self.device_fd().is_privileged() {
             if let Err(err) = self.acquire_master_lock() {
                 error!("Failed to acquire drm master again. Error: {}", err);
             }
         }
-        self.set_active(true);
+        if !self.set_active(true) && disable_connectors {
+            self.reset_state()
+        } else {
+            self.surfaces.retain(|surface| surface.upgrade().is_some());
+            Ok(())
+        }
     }
 
     /// Returns if the device is currently paused or not.
@@ -417,10 +434,39 @@ impl DrmDevice {
         }
     }
 
-    fn set_active(&self, active: bool) {
+    /// Reset the state of this device
+    ///
+    /// This will disable all connectors and reset all planes.
+    /// Additional this will also reset the state on all known surfaces.
+    pub fn reset_state(&mut self) -> Result<(), Error> {
+        if !self.is_active() {
+            return Err(Error::DeviceInactive);
+        }
+
         match &*self.internal {
-            DrmDeviceInternal::Atomic(internal) => internal.active.store(active, Ordering::SeqCst),
-            DrmDeviceInternal::Legacy(internal) => internal.active.store(active, Ordering::SeqCst),
+            DrmDeviceInternal::Atomic(internal) => internal.reset_state(),
+            DrmDeviceInternal::Legacy(internal) => internal.reset_state(),
+        }?;
+
+        let mut i = 0;
+        while i != self.surfaces.len() {
+            if let Some(surface) = self.surfaces[i].upgrade() {
+                match &*surface {
+                    DrmSurfaceInternal::Atomic(surf) => surf.reset_state::<Self>(None),
+                    DrmSurfaceInternal::Legacy(surf) => surf.reset_state::<Self>(None),
+                }?;
+                i += 1;
+            } else {
+                self.surfaces.remove(i);
+            }
+        }
+        Ok(())
+    }
+
+    fn set_active(&self, active: bool) -> bool {
+        match &*self.internal {
+            DrmDeviceInternal::Atomic(internal) => internal.active.swap(active, Ordering::SeqCst),
+            DrmDeviceInternal::Legacy(internal) => internal.active.swap(active, Ordering::SeqCst),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/backend/drm/device/mod.rs
```diff
@@ -406,6 +406,19 @@ impl DrmDevice {
...
+        // longer like we left it before the session pause. The worst state is where the two CRTC got swapped. In this
+        // case we fail to update the state on the next commit because we currently always only update a single CRTC.
+        // This won't work because our CRTC is still bound to the other connector.
+        //
+        // I am not sure how we can solve this, but for now just clearing all CRTCs on a session resume should reliable solve this issue.
+        for crtc in self.crtcs() {
+            if let Err(err) = self.set_crtc(*crtc, None, (0, 0), &[], None) {
+                tracing::warn!(?err, ?crtc, "failed to reset crtc on active");
+            }
+        }
```
Drakulix, line 421: This looks a lot like the `reset_state` method we already have for our `AtomicDrmDevice` and `LegacyDrmDevice`, which are used for exactly that purpose when initializing the `DrmDevice` with `disable_connectors` set to `true`.

Can we re-use these here please?

cmeissl: wdym? reset_state gets called on session resume in anvil, but currently only reads the current state from drm. The actual change is done in commit, which faces exactly the described issue. Also reset_state is per surface, so this would rely on downstream call reset_state on all surfaces "atomically". Also it might not solve the issue in all cases I believe.

cmeissl: Ah, you are talking about the device, not the surface. Yeah, sure, sorry.

cmeissl: Hm, thinking about the whole thing again it might be better to make that explicit. Maybe by adding a disable_connectors parameter to activate which resets the state like in new. We could also make reset_state public, but I feel like that alone wouldn't be enough to make downstream aware of this issue. So maybe we should do both, add the parameter and break existing code making it aware of the issue and let downstream decide how to handle it.

YaLTeR:
I'm assuming the current version of the patch forces a modeset on all outputs right?

YaLTeR:
I'm currently running a build with this PR + changed code path, and on another TTY a regular build + changed code path, and so far the regular build already failed to bring the monitors up in a multitude of different ways, whereas with this PR I haven't seen a single failure yet.

cmeissl:
> I'm assuming the current version of the patch forces a modeset on all outputs right?

Yes, this always forces a modeset. Actually it will result in multiple modesets, the first round in activate disables all crtc, queue_frame will then enable them one by one again.

Resolving the connector mismatch requires a modeset on all crtcs that no longer have the expected state. There are multiple ways to do that, which also depends on the backend used (legacy vs atomic).
With atomic it should be possible to do a single commit to restore the expected state on all crtc, but that requires some manual intervention or refactoring of the current API (potentially putting all individual surfaces under one abstraction).
You could also just reset one crtc to resolve the tie and let the surface change the other. But there is no way around
a modeset if we have a crtc/connector mismatch.

But it might be possible to detect this situation and only reset/clear the crtcs in this case, which should at least result in
a smoother tty switch experience. If this is worth to do probably depends on your requirements.

cmeissl:
Actually it might also be possible to keep the happy path and just reset the drm device if we hit an error.

cmeissl:
Okay, there is still on issue. If we use the optimistic path, but never used a certain plane, we could end up with that plane showing a left over from a different session. But this will probably require more changes and this is probably good enough for now.

src/backend/drm/device/mod.rs
```diff
@@ -400,13 +407,23 @@ impl DrmDevice {
...
+    /// the device was not active before. Otherwise you need to make sure there are no
+    /// conflicting requirements when enabling or creating surfaces.
+    pub fn activate(&mut self, disable_connectors: bool) -> Result<(), Error> {
         if self.device_fd().is_privileged() {
             if let Err(err) = self.acquire_master_lock() {
                 error!("Failed to acquire drm master again. Error: {}", err);
             }
         }
+        if disable_connectors && !self.is_active() {
+            self.reset_state()?;
```
YaLTeR, line 421: I'm guessing this shouldn't early-return before `set_active(true)`? Otherwise it's kinda weird because if you don't request resetting the state then you'll get the device marked as active regardless.

Drakulix: I feel like, if the user specifically requests a reset and it fails, we shouldn't set it active, since the device likely disappeared or we lack permissions because it was moved between seats or something like that.

YaLTeR: But then if you activate(false) followed right away by reset_state(), then you will both get an error, and drm in active state

cmeissl: True, this is what I mean with better documenting the function. By providing false you guarantee that you already brought the device into a known state that will work (or will do so before calling any other drm related function).

cmeissl: But I get you point, it could be a surprising behavior and I should probably also gate `reset_state` begin `active`.
Doing so will make it impossible to call it before setting the device to active again.

cmeissl: Okay, so this now always sets the device to active and won't allow you to call `reset_state` on a suspended device.
If you receive an error from reset_state (or acticate with true for disable_connectors) you are doomed in most cases anyway.

Drakulix (approved):

PR #1297, opened 2024-01-22
Fix plane artifacts

cmeissl:
This should hopefully provide a fix for all kind of plane artifact issues after suspend/resume and dpms.

The comments on the single commits should be enough to describe the motivation behind the changes.

Just some warning, this breaks some existing behavior by fixing an issue I found in the `reset_state` function of the atomic surface. Before that change every call to `reset_state` would make the surface require a commit.
With this fixed `DrmCompositor` will no longer always submit a complete state after calling `reset_state`
which might result in artifacts.
The correct way to resolve this is to call `DrmCompositor::reset_state` instead directly on the surface.
This will make sure the next call to `render_frame`/`queue_frame` will produce a full update, updating
all planes.

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -357,7 +357,7 @@ pub fn run_udev() {
                         lease_global.resume::<AnvilState<UdevData>>();
                     }
                     for surface in backend.surfaces.values_mut() {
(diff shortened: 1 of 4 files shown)


Drakulix (commented):
LGTM

Drakulix (approved):

PR #1298, opened 2024-01-23
input: add support for keyboard led state

cmeissl:
An attempt to make the leds on the keyboard to show up.

~~As I am normally not working on the input side this might be complete bogus, but at least it seems to work for me.
What I am especially uncertain at is the `update_led` function on the keyboard input event, but it also did not
feel correct to add it to the generic device trait.~~

Ok, so this should now represent a much saner implementation. The `SeatHandler` now has a notify function which
is automatically called on changes to the led state. The only remaining thing downstream has to do is to track all
keyboards, listen to the led state updates and update all keyboard led states.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -58,6 +58,68 @@ enum GrabStatus<D> {
     Borrowed,
 }
 
+/// Mapping of the led of a keymap
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct LedMapping {
+    /// Index of the NUMLOCK led
+    pub num: Option<xkb::LedIndex>,
+    /// Index of the CAPSLOCK led
+    pub caps: Option<xkb::LedIndex>,
+    /// Index of the SCROLLLOCK led
+    pub scroll: Option<xkb::LedIndex>,
+}
+
+impl LedMapping {
+    /// Get the mapping from a keymap
+    pub fn from_keymap(keymap: &xkb::Keymap) -> Self {
+        Self {
+            num: match keymap.led_get_index(xkb::LED_NAME_NUM) {
+                xkb::LED_INVALID => None,
+                index => Some(index),
+            },
+            caps: match keymap.led_get_index(xkb::LED_NAME_CAPS) {
+                xkb::LED_INVALID => None,
+                index => Some(index),
+            },
+            scroll: match keymap.led_get_index(xkb::LED_NAME_SCROLL) {
+                xkb::LED_INVALID => None,
+                index => Some(index),
+            },
+        }
+    }
+}
+
+/// Current state of the led when available
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Default)]
+pub struct LedState {
+    /// State of NUMLOCK led
+    pub num: Option<bool>,
+    /// State of CAPSLOCK led
+    pub caps: Option<bool>,
+    /// State of SCROLLLOCK led
+    pub scroll: Option<bool>,
+}
+
+impl LedState {
+    /// Update the led state from an xkb state and mapping
+    ///
+    /// Returns whether the led state changed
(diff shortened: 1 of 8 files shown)


kchibisov:
I generally don't have a keyboard with leds, however I think it won't work in all cases, since you generally want to reload the leds after modifiers changes, and we can manually change the modifiers, etc.

cmeissl:
> I generally don't have a keyboard with leds, however I think it won't work in all cases, since you generally want to reload the leds after modifiers changes, and we can manually change the modifiers, etc.

Ok, yeah, so there needs to be same way to always update the led state in the backend.

kchibisov:
You may want to look into how wlroots is handling it, but basically each time you try to update keymap/modifiers you need to re-issues that, it's a bit messy with the current state of things in smithay. Maybe make a separate call that does kind of sets modifiers state separately? So users may call it each time they touch modifiers themselves?

cmeissl:
Yeah, I already did check wlroots before starting this PR, but wlroots seems way more tightly coupled between the input backend and the handling.  Smithay on the other hand decouples the backend completely from the generic input handling.

So we have `input::keyboard::KeyboardHandle` which seems to handle everything xkb related, which would include getting
the led state. And then we have `backend::input::InputBackend` which would be responsible to update the physical led state. 

It seems until now we only had to pump the events from the backend to the generic input handling, but with the led update we would have to communicate back to the backend. This either requires some coupling between the backend and the generic input handling or some more boiler plate code downstream. Maybe it is possible to provide some helper function to do the right thing.

src/input/keyboard/mod.rs
```diff
@@ -58,6 +58,68 @@ enum GrabStatus<D> {
...
 }
 
+/// Mapping of the led of a keymap
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct LedMapping {
+    /// Index of the NUMLOCK led
+    pub num: Option<xkb::LedIndex>,
+    /// Index of the CAPSLOCK led
+    pub caps: Option<xkb::LedIndex>,
+    /// Index of the CAPSLOCK led
```
YaLTeR, line 68:
```suggestion
    /// Index of the SCROLLLOCK led
```

src/input/keyboard/mod.rs
```diff
@@ -697,6 +786,12 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
             };
         }
 
+        if leds_changed {
+            let led_state = internal.led_state;
```
YaLTeR, line 790:
```suggestion
            let led_state = internal.led_state;
            std::mem::drop(internal);
```
Input deadlocks my beloved

src/input/mod.rs
```diff
@@ -139,6 +139,9 @@ pub trait SeatHandler: Sized {
 
     /// Callback that will be notified whenever a client requests to set a custom cursor image.
     fn cursor_image(&mut self, _seat: &Seat<Self>, _image: CursorImageStatus) {}
+
+    /// Callback that will be notified whenever the led state changed
```
YaLTeR, line 143:
```suggestion
    /// Callback that will be notified whenever the keyboard led state changes.
```

src/input/keyboard/mod.rs
cmeissl: thanks, fixed

src/input/mod.rs
cmeissl: thanks, done

src/input/keyboard/mod.rs
cmeissl: `internal` is a `&mut` in this function, I instead made sure the scope of `internal` ends before `data.led_state_changed` is invoked.

YaLTeR (approved):
Thanks, looks good at a glance

Drakulix (approved):

PR #1300, opened 2024-01-23
Expose focus events through the X11 backend

roscale:
Expose FocusIn and FocusOut events through the X11 backend. Focus was already exposed on the Winit backend, so why not have the same on the X11 one. Also, a helper function has been added that returns the key codes of currently pressed keys.

My use case:
When I run the compositor in a window under X11 and I Alt+Tab, the Alt key is permanently stuck in the pressed state because the host compositor never sends me a release event. When I lose focus, I want to iterate over all the keys that are being pressed and artificially release them, so that next time I focus the window no keys are stuck down.


Drakulix (approved):
Looks like a good approach to me. Thanks!

PR #1305, opened 2024-01-28
Send modifiers when creating IME grab

kchibisov:
This can happen when IME releases and takes back the grab on the fly, thus the modifier state may go out of sync with the keyboard modifiers.


kchibisov:
CI failure is unrelated to my changes.

Drakulix (approved):

PR #1306, opened 2024-01-29
chore: Fix deprecated indexmap remove method

Drakulix:
Looks like the IndexMap is used in all these places for a stable iterator order, possibly even z-order in the compositor, so shift_remove is likely the expected behavior here.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -1370,8 +1370,8 @@ impl OverlayPlaneElementIds {
     }
 
     fn remove_plane(&mut self, plane: &plane::Handle) {
-        if let Some(plane_id) = self.plane_plane_ids.remove(plane) {
-            self.plane_id_element_ids.remove(&plane_id);
+        if let Some(plane_id) = self.plane_plane_ids.swap_remove(plane) {
+            self.plane_id_element_ids.swap_remove(&plane_id);
         }
     }
 }
@@ -2119,7 +2119,7 @@ where
                 {
                     cursor_plane_element.take()
                 } else {
-                    overlay_plane_elements.remove(plane)
+                    overlay_plane_elements.shift_remove(plane)
                 };
 
                 // If we have no element on this plane skip the rest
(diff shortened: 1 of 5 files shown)


Drakulix:
This has turned into cleaning up various CI issues and nightly warnings.

src/wayland/shell/xdg/handlers/surface.rs
```diff
@@ -66,7 +66,7 @@ where
                 data.known_surfaces
                     .lock()
                     .unwrap()
-                    .remove(&xdg_surface.downgrade());
+                    .shift_remove(&xdg_surface.downgrade());
```
cmeissl, line 69: This does not seem to rely on the order, so `swap_remove` should be fine and has `O(1)` instead of `O(n)`.

src/backend/drm/compositor/mod.rs
```diff
@@ -1370,8 +1371,8 @@ impl OverlayPlaneElementIds {
     }
 
     fn remove_plane(&mut self, plane: &plane::Handle) {
-        if let Some(plane_id) = self.plane_plane_ids.remove(plane) {
-            self.plane_id_element_ids.remove(&plane_id);
+        if let Some(plane_id) = self.plane_plane_ids.shift_remove(plane) {
+            self.plane_id_element_ids.shift_remove(&plane_id);
```
cmeissl, line 1375: same here, `swap_remove` should be fine

src/wayland/shell/xdg/handlers/surface.rs
Drakulix: I was a bit confused why this uses an index map the first place and the only reasoning I could come up was, that it wants to be iterated in insertion order? Then we would break this here.

src/backend/drm/compositor/mod.rs
cmeissl: :+1: this is just used as a cache and lookup

src/wayland/shell/xdg/handlers/surface.rs
cmeissl: tbh, no idea, but all it does is to insert the surface on construction, remove it on destruction and enforce that are not surfaces alive on destruction of `xdg_wm_base` which would be a protocol violation.

PR #1307, opened 2024-01-29
Fix output transform issues

cmeissl:


YaLTeR:
Confirming this fixes both issues I encountered.

Drakulix (approved):

PR #1310, opened 2024-01-30
Add OutputHandler trait

YaLTeR:
Lets compositors know when a new `wl_output` is bound. Useful for implementing protocols with similar semantics to `wl_surface.enter` where newly bound `wl_output`s must participate in some events immediately.

### DIFF
--- a/src/wayland/output/handlers.rs
+++ b/src/wayland/output/handlers.rs
@@ -8,7 +8,7 @@ use wayland_server::{
     Client, DataInit, Dispatch, DisplayHandle, GlobalDispatch, New, Resource,
 };
 
-use super::{xdg::XdgOutput, Output, OutputManagerState, OutputUserData, WlOutputData};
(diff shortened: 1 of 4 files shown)


YaLTeR:
Example use: https://github.com/YaLTeR/niri/pull/142/commits/1ab6a11d8c4792830363202493771d20b450b9a6

src/wayland/output/handlers.rs
```diff
@@ -19,9 +19,10 @@ where
     D: GlobalDispatch<WlOutput, WlOutputData>,
     D: Dispatch<WlOutput, OutputUserData>,
     D: 'static,
+    D: OutputHandler,
```
cmeissl, line 22: some small style nit, I would prefer if we can keep the `'static` lifetime last

cmeissl (commented):
beside the small style nit, this look good to me

src/wayland/output/handlers.rs
YaLTeR: Done

cmeissl (approved):
thanks!

PR #1312, opened 2024-02-01
fix solid render element alpha

cmeissl:


YaLTeR (approved):
Works as intended now

Drakulix (approved):

PR #1317, opened 2024-02-06
fix: Don't unwrap RendererSurfaceState

Drakulix:
Makes `with_renderer_surface_state` not panic, should there have never been an import. (Which can be particularly easy to trigger with unmapped Xwayland windows.)

### DIFF
--- a/src/backend/renderer/utils/wayland.rs
+++ b/src/backend/renderer/utils/wayland.rs
@@ -409,18 +409,16 @@ impl SurfaceView {
 
 /// Access the buffer related states associated to this surface
 ///
-/// Calls [`compositor::with_states`] internally
-pub fn with_renderer_surface_state<F, T>(surface: &WlSurface, cb: F) -> T
+/// Calls [`compositor::with_states`] internally.
+///
+/// Returns `None`, if there never was an commit processed through `on_commit_buffer_handler`.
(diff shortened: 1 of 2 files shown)


cmeissl (approved):
Sounds like a good idea to not just blindly unwrap here 👍

src/backend/renderer/utils/wayland.rs
```diff
@@ -409,18 +409,20 @@ impl SurfaceView {
...
-            .get::<RendererSurfaceStateUserData>()
-            .unwrap()
-            .borrow_mut();
-        cb(&mut data)
+        if let Some(data) = states.data_map.get::<RendererSurfaceStateUserData>() {
+            let mut data = data.borrow_mut();
+            Some(cb(&mut data))
+        } else {
+            None
+        }
```
PolyMeilex, line 425: Looks like a good candidate for Option::map
```suggestion
        states
            .data_map
            .get::<RendererSurfaceStateUserData>()
            .map(|data| cb(&mut data.borrow_mut()))
```

Or just `?`

```suggestion
        let data = states.data_map.get::<RendererSurfaceStateUserData>()?;
        Some(cb(&mut data.borrow_mut()))
```

But yeah, in general I don't care, looks good to me anyway.

PolyMeilex (approved):

PolyMeilex (approved):

PR #1320, opened 2024-02-08
Implement ext-idle-notify

YaLTeR:
Taking over from #1021, cc @PolyMeilex @chrisduerr.

Changes:
1. Added `notify_activity(&Seat)`, kept the `WlSeat` method under a different name.
2. Fixed a bug where uninhibiting didn't reset idle state resulting in two subsequent `resumed` events (forbidden by the protocol).
3. Added dead `WlSeat` data cleanup.
4. Some other minor things.

Impl in niri: https://github.com/YaLTeR/niri/compare/main...idle-notify

### DIFF
--- a/src/wayland/idle_notify/mod.rs
+++ b/src/wayland/idle_notify/mod.rs
@@ -1,0 +1,359 @@
+//! This interface allows clients to monitor user idle status.
+//!
+//! ```
+//! # extern crate wayland_server;
+//! # #[macro_use] extern crate smithay;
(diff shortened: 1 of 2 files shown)


PolyMeilex (approved):
My only concern with this code was lack of compositor implementation, but if we have niri code, I'm fine with it now.

LGTM, not going to merge it tho, as that would kinda be me reviewing my own code :smiley:

chrisduerr (approved):

PolyMeilex:
Thanks!

PR #1324, opened 2024-02-09
Replace Option with Weak in content type protocol

PolyMeilex:
I have no idea why I added it there in the first place, it's never used, the value is always Some()

### DIFF
--- a/src/wayland/content_type/mod.rs
+++ b/src/wayland/content_type/mod.rs
@@ -58,7 +58,8 @@ use wayland_protocols::wp::content_type::v1::server::{
     wp_content_type_v1::{self, WpContentTypeV1},
 };
 use wayland_server::{
-    backend::GlobalId, protocol::wl_surface::WlSurface, Dispatch, DisplayHandle, GlobalDispatch,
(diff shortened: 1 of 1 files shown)


src/wayland/content_type/mod.rs
```diff
@@ -132,14 +132,14 @@ impl ContentTypeSurfaceData {
 
 /// User data of `WpContentTypeV1` object
 #[derive(Debug)]
-pub struct ContentTypeUserData(Mutex<Option<WlSurface>>);
+pub struct ContentTypeUserData(Mutex<WlSurface>);
```
Drakulix, line 135: From the protocol:
>  When the associated surface gets destroyed, this object becomes inert and the client should destroy it.

Given the *should* in there, I believe this should be a `Weak<WlSurface>` to not prevent the surface user data from being cleaned up.

Drakulix (changes requested):

src/wayland/content_type/mod.rs
PolyMeilex: Yep, sounds right, done

Drakulix (approved):

PR #1325, opened 2024-02-10
Implement wayland side of tearing control

PolyMeilex:
Probably needs some work on DRM side, but this does plumbing for the wayland side.

### DIFF
--- a/src/wayland/mod.rs
+++ b/src/wayland/mod.rs
@@ -80,6 +80,7 @@ pub mod shm;
 pub mod single_pixel_buffer;
 pub mod socket;
 pub mod tablet_manager;
+pub mod tearing_control;
 pub mod text_input;
 pub mod viewporter;
 pub mod virtual_keyboard;
--- a/src/wayland/tearing_control/dispatch.rs
+++ b/src/wayland/tearing_control/dispatch.rs
@@ -1,0 +1,143 @@
+use wayland_protocols::wp::tearing_control::v1::server::{
+    wp_tearing_control_manager_v1::{self, WpTearingControlManagerV1},
+    wp_tearing_control_v1::{self, WpTearingControlV1},
+};
+use wayland_server::{
+    backend::ClientId, Client, DataInit, Dispatch, DisplayHandle, GlobalDispatch, New, Resource,
+};
+
+use super::{
+    TearingControlState, TearingControlSurfaceCachedState, TearingControlSurfaceData, TearingControlUserData,
+};
+use crate::wayland::compositor;
+
+impl<D> GlobalDispatch<WpTearingControlManagerV1, (), D> for TearingControlState
+where
+    D: GlobalDispatch<WpTearingControlManagerV1, ()>,
+    D: Dispatch<WpTearingControlManagerV1, ()>,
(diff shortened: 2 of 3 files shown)


Drakulix:
Wayland side LGTM.

I *think* all this need for drm is:
- Check for the `ASyncPageFlip` capability (https://docs.rs/drm/latest/drm/enum.DriverCapability.html#variant.ASyncPageFlip) - This has to be done per-device, so this can't really affect the global being advertised.
- Set `PAGE_FLIP_ASYNC` on `page_flip`: https://docs.rs/drm/latest/drm/control/struct.AtomicCommitFlags.html#associatedconstant.PAGE_FLIP_ASYNC

The latter should only happen, when a surface with that hint set is fullscreen though. Possibly the `DrmCompositor` could extract that state from the surface render-elements (might need a new method for `RenderElement` and apply that automatically, when directly scanning out such a surface on the primary plane. (cc @cmeissl)

cmeissl:
> The latter should only happen, when a surface with that hint set is fullscreen though. Possibly the `DrmCompositor` could extract that state from the surface render-elements (might need a new method for `RenderElement` and apply that automatically, when directly scanning out such a surface on the primary plane. (cc @cmeissl)

I see a few different ways to implement that:
- Semi-automatically in the `DrmCompositor`, which like you said requires extending the trait by a function informing us about the preference. I believe we might be able to make the detection a bit more intelligent then basing it off primary plane scan-out. For example it might be ok to tear if only a single surface (that has tearing enabled) on an overlay plane is updated and everything else stays the same. Or a bit simpler and probably good enough, if all elements are directly scanned out and allow tearing.
- As a parameter/option of `DrmCompositor` and allow downstream decide based on whatever
- Combination of both, maybe some option on `DrmCompositor` which allows to select Auto, Disabled, Force or something like this.

YaLTeR:
I agree with @cmeissl; I'd like more manual tearing control in my compositor. So that I can implement the fullscreen with hint => tear policy, but also other policies, like maybe I want to enable it for windowed too under some conditions.

PolyMeilex:
> I agree with @cmeissl; I'd like more manual tearing control in my compositor. So that I can implement the fullscreen with hint => tear policy, but also other policies, like maybe I want to enable it for windowed too under some conditions.

For now I'm working on a simplest possible setup, you either queue a frame as VSync or ASync, [you can see it or even test it here](https://github.com/Smithay/smithay/pull/1335/files#diff-15b6cf0913d77e30afd2645e2b49a70f5d03ab9eecc20faef64df60dcea9d7ddR1729)
My kernel does not support Atomic async page flips yet, so I have not yet tested that code path, but legacy one with `SMITHAY_USE_LEGACY` env variable works already on my system.

PolyMeilex:
@cmeissl FYI this is now rebased and builds again, in case you need to test #1335 with real clients

relativemodder:
heyy, any progress?

vars1ty:
What's even left to do for this to be ready and merged?

candied-apple:
When will we get this?

khushmmehta:
Bump, can we get this merged please?

PR #1326, opened 2024-02-11
touch re-factor

cmeissl:
The current implementation has a few limitations and does not provide an abstraction on the same level as pointer or keyboard. For proper support including DnD we also need to support grabs. 

This also removes the implementation of `PointerTarget` and `KeyboardTarget` on the desktop abstractions. I described the motivation behind this [here](https://github.com/Smithay/smithay/pull/1334#issuecomment-1949422631)
The last commit also implements the base for the idea mentioned in the above comment.

### DIFF
--- a/anvil/src/focus.rs
+++ b/anvil/src/focus.rs
@@ -1,7 +1,5 @@
-use smithay::input::pointer::{
-    GestureHoldBeginEvent, GestureHoldEndEvent, GesturePinchBeginEvent, GesturePinchEndEvent,
-    GesturePinchUpdateEvent, GestureSwipeBeginEvent, GestureSwipeEndEvent, GestureSwipeUpdateEvent,
-};
+#[cfg(feature = "xwayland")]
+use smithay::xwayland::X11Surface;
 pub use smithay::{
     backend::input::KeyState,
     desktop::{LayerSurface, PopupKind},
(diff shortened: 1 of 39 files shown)


Drakulix:
As noted in matrix previously, I am not yet convinced, that dropping the `KeyboardHandler`/`PointerHandler` traits is the right move for more complex cases of in-compositor UI elements.

But I guess the only way to figure this out, is to implement it. And apart from that I do agree with the general structure.

anvil/src/focus.rs
```diff
@@ -14,46 +12,80 @@ pub use smithay::{
...
         data: &mut AnvilState<BackendData>,
         event: &MotionEvent,
     ) {
         match self {
-            FocusTarget::Window(w) => PointerTarget::enter(w, seat, data, event),
-            FocusTarget::LayerSurface(l) => PointerTarget::enter(l, seat, data, event),
-            FocusTarget::Popup(p) => PointerTarget::enter(p.wl_surface(), seat, data, event),
+            PointerFocusTarget::WlSurface(w) => PointerTarget::enter(w, seat, data, event),
+            #[cfg(feature = "xwayland")]
+            PointerFocusTarget::X11Surface(w) => PointerTarget::enter(w, seat, data, event),
```
ids1024, line 87: If `Window`/`LayerSurface`/`Popup` can be combined as just `WlSurface`, shouldn't that also cover `X11Surface`?

It seems `PointerTarget<D> for X11Surface` just calls `PointerTarget` methods on the underlying `WlSurface`, if it exists.

cmeissl: Yes, that would be possible, but it felt strange to leave the `KeyboardTarget` impl there (which does a few x related stuff) and only remove the `PointerTarget` impl.

ids1024: Ah, fair point.

cmeissl:
Linking to https://github.com/Smithay/smithay/pull/1334#issuecomment-1949422631 as it contains some background info about my motivation behind removing the `PointerTarget`\`KeyboardTarget` impls on `Window` and so on.
(Also includes some proposal on how to solve the focus issue, feedback welcome)

Drakulix (commented):
Liking this a lot, I am quite confident, that this state of the api will work with cosmic-comp.

ids1024:
So the idea is that a compositor (like cosmic-comp) would have some concept of a focus stack, without such a thing existing explicitly in smithay itself? Makes sense for that to be represented as a stack, though I don't know exactly how that will end up looking.

It's good to have touch brought up to parity with keyboard/pointer.

cmeissl:
> So the idea is that a compositor (like cosmic-comp) would have some concept of a focus stack, without such a thing existing explicitly in smithay itself? 

For the moment, yes. Though cnce we figure out a generic way to represent it we probably should add an abstraction to smithay.

Drakulix (approved):
Lacking a touch-device to properly test this, but code wise LGTM!

PR #1327, opened 2024-02-12
POC Begone delegate macros

PolyMeilex:
Based on https://github.com/Smithay/wayland-rs/pull/699

Removes all delegate macros, as well as redundant type constraints on `D`


ids1024:
This should genuinely make adding a simple protocol to Smithay take about half as much effort as it currently does. (The less annoying half.) It's great to see we can do this without radically changing the API!

PR #1329, opened 2024-02-13
Add `Frame::{set_,}{up,down}scale_filter()`

YaLTeR:
Allows changing the filter on the fly, especially useful within `RenderElements::draw()` where the renderer is inaccessible.

Example use: https://github.com/YaLTeR/niri/pull/189


SergioRibera:
Any update?

PR #1330, opened 2024-02-13
multigpu: use shadow copies over direct imports

Drakulix:
Sorry, I am making the MultiGPU code a bit more complex again..

## Overall idea/concept:

This in preparation for https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/268, but also solves a bunch of observed issues with the current code today.

As the discussion around the linked issue has brought up, directly importing buffers into different gpus can cause very expensive memory migrations.

The biggest issue regarding those migrations is, that this isn't a one-time cost payed at import, but the buffer will often be migrated to system-memory, where rendering for the application is more costly and sampling for the compositor as well, while mostly prohibiting scanout. Often rendering/sampling under those circumstances involve a bunch of copies handled transparently by the involved drivers.

Even in cases, where a more direct path (like pci-dma transfers between multiple amdgpu devices) is available, the rendering performance can be greatly reduced.

The idea of this patch set is to avoid importing client-provided dmabufs on any other GPU than the one originally allocating the buffer. To still make use of potentially faster direct-imports instead of relying on memory copies all the time (and potentially unnecessarily blocking valuable memory-bandwidth), we introduce shadow buffers to facilitate those copies ourselves.

Meaning we allocate a (likely migrated) dma buffer ourselves and blit/render the source buffers into those instead of importing them directly.

More predictable behaviour around imports also allows compositors to work-around bugs such as the iris driver simply hanging indefinitely when importing some nvidia buffers: https://gitlab.freedesktop.org/mesa/mesa/-/issues/10572 or better synchronization around drivers with broken implicit sync (*cough* also nvidia *cough*).

## Commits:

- 06afcfe9 - This commit adds a (rewritable) `DrmNode` to `Dmabuf`. This makes tracking the node, that should be used for a dmabuf much easier. We set the node in cases, where we know the origin (basically when we allocate the buffer ourselves). In other cases downstream can still set/override the node, e.g. anvil will always set this to it's render-device, since it knows that is the only node it advertises via the dmabuf protocol. Implementing dmabuf-v6 becomes trivial with this change.
- 659c00e9 - As a result of allocating the shadow buffers, we need more reliable access to an allocator for our renderer in the multigpu code. As such I am dropping the `EGLDevice`-backend (as it has no way to allocate with modifiers) and adjust the `Gbm`-backend to also expose the `GbmDevice` as an allocator. This removes the need for downstream to supply an allocator and all compositors used the `GbmGlesBackend` anyway. This also removes usage of the `VulkanAllocator` as that caused some very weird performance problems on some system. We can revisit this, once we have a proper vulkan-backend.
- 9fa71cd90 - Finally introduces the shadow-copies for imported dmabufs and cleans up a lot of logic surrounding imports and the target framebuffer.
- f2d1d2e42 - Optimization for anvil to avoid the MultiRenderer to try and import into multiple nodes.


Drakulix:
Draft while this still depends on unreleased gbm-rs code and was not yet tested with cosmic-comps more complex multi-gpu code.

src/backend/allocator/dmabuf.rs
```diff
@@ -233,6 +253,23 @@ impl Dmabuf {
...
+        *self.0.node.lock().unwrap()
+    }
+
+    /// Sets or unsets any node set for this dmabuf (see [`Dmabuf::node`]).
+    ///
+    /// May alter behavior of other parts of smithay using this as a hint.
+    #[cfg(feature = "backend_drm")]
+    pub fn set_node(&self, node: impl Into<Option<DrmNode>>) {
+        *self.0.node.lock().unwrap() = node.into();
+    }
```
cmeissl, line 271: Do we allow to change the node on `Dmabuf` to avoid more complex initialization in `create_dmabuf`/`dmabuf_imported` or because we have/expect a use-case for it?

src/backend/renderer/multigpu/gbm.rs
```diff
@@ -114,17 +121,23 @@ impl<R> GbmGlesBackend<R> {
         }
     }
 
+    /// Sets the default flags to use for allocating buffers via the [`GbmAllocator`]
+    /// provided by these backends devices.
+    ///
+    /// Only affects nodes added via [`add_node`] *after* calling this method.
+    pub fn set_allocator_flags(&mut self, flags: GbmBufferFlags) {
+        self.allocator_flags = flags;
+    }
```
cmeissl, line 130: This looks more like something to set during initialization. Maybe this shoule be another constructor like `GbmGlesBackend::with_allocator_flags`?
(Might be annoying to write as it would require to also support combinations with `with_factory` and `with_context_priority` or switch to a builder)

cmeissl (approved):
Okay, the structure and code looks good to me, nice work. :+1:
Just two API related questions, but nothing that has to change.

src/backend/renderer/multigpu/gbm.rs
Drakulix: Yes this was mostly to avoid all the permutations of existing constructors. If you have a better idea feel free, but a builder pattern also isn't exactly easy, given that with_factory and with_context_priority exclude each other.

src/backend/allocator/dmabuf.rs
Drakulix: Multiple things:

Firstly as you pointed out, we can't really set this during initialization in the dmabuf protocol, because while we do know the device for v3 and lower (and potentially for v6 and higher), we don't really for v4/v5. We can communicate a different device via the target_device of a tranche but we don't know, what the client does. As such this would need more complex handling during `dmabuf_imported` to ask downstream for the node without providing a `Dmabuf`, given that we need the node to construct in the first place. Allowing downstream to just set this afterwards felt easier.


Secondly this is not really a property of the buffer (again no api to query this), but state added by smithay and potentially consumed by smithay (e.g. the MultiRenderer). Since it alters behaviour it felt appropriate to allow downstream to override this (maybe they know better in some cases?). Having a buffer stuck on `None` for other potential api, where we can't know (but downstream might) feels wrong.

I feel like I also should point out, that the MultiRenderer actively sets this, if it gets a buffer with a node set to `None` to avoid testing all nodes next time. So again, more state than a property.

cmeissl: > I feel like I also should point out, that the MultiRenderer actively sets this, if it gets a buffer with a node set to None to avoid testing all nodes next time. So again, more state than a property.

Ah, missed that part. The rest would be solvable (in case there even is something to solve) by delaying the creation. But I agree that would greatly complicate the current API, so probably not worth it.

I am not totally happy with this, but I also don't see an easier way, so :+1:

src/backend/renderer/multigpu/gbm.rs
cmeissl: Okay, also not a big deal and it is actually documented.

src/backend/renderer/gles/mod.rs
```diff
@@ -1407,6 +1395,20 @@ impl ExportMem for GlesRenderer {
...
+    fn can_read_texture(&mut self, texture: &Self::TextureId) -> Result<bool, GlesError> {
+        let old_target = self.target.take();
+
+        // if we can't bind the texture, we can't read it
+        let res = self.bind(texture.clone()).is_ok();
+
+        // restore
+        self.unbind()?;
+        self.target = old_target;
+        self.make_current()?;
```
cmeissl, line 1407: This is probably not an issue, but when unbind fails it won't restore the target, in case make_current fails it already restored it. I guess in the end it won't matter, but just noticed that difference.

src/backend/renderer/mod.rs
```diff
@@ -623,6 +623,17 @@ pub trait ExportMem: Renderer {
         format: Fourcc,
     ) -> Result<Self::TextureMapping, Self::Error>;
 
+    /// Returns whether the renderer should be able to read-back from the given texture.
+    ///
+    /// No actual copying shall be performed by this function nor is a format specified,
+    /// so it is still legal for [`ExportMem::copy_texture`] to return and error, if this
```
cmeissl, line 629: `return and error` -> `return an error`?

src/backend/renderer/gles/mod.rs
Drakulix: Yes, but unbind also sets the `target`. Which is why it is fine imo to just error out. This *shouldn't* happen and if it does it will almost certainly have messed up the state.

cmeissl: Yeah, that was also my impression.

cmeissl (approved):
LGTM

YaLTeR:
I tried a bunch of weird things after applying this to niri and nothing seemed to break thus far

PR #1334, opened 2024-02-16
Make `PointerTarget`/`KeyboardTarget` object-safe; use in Anvil

ids1024:
I've been thinking about how to simplify this code for a while. Implementations of these traits become particularly verbose with a more complicated shell like cosmic-comp: https://github.com/pop-os/cosmic-comp/blob/master/src/shell/focus/target.rs.

It turns out it can be make object-safe, so we can have a `&dyn PointerTarget`, if `PartialEq` and `Clone` requirements are moved out of the trait itself.

I still wonder if it could be better to use enums for input events instead of a million methods. Anyway, something to think about.

Leaving as a draft pending the other changes in input targets in https://github.com/Smithay/smithay/pull/1326.


cmeissl:
So, the reason for the removal of the `PointerTarget`/`KeyboardTarget` implementation of `Window` /`Layer`/... in #1326 is that these bypass the `Grab`s. 

For example the default `ClickGrab` on the `PointerHandle` should make sure the pointer won't leave the surface that installed the grab. But with something like `Window` being the `PointerTarget` and dispatching the focus internally to different surfaces the grab only makes so the same `Window` is focused, but the actual rule is completely bypassed.

I did look into what cosmic is doing with the targets (turns out quite a lot in case of `PointerTarget`) for the past few days and procrastinated on how to resolve that without either making it a pain to use in cosmic or sacrifice the grab guarantees.

All I did so far in cosmic is shuffle around a few implementations and try out a few ideas.
My latest idea is based on the fact that the focus on the pointer is actually a stack and we might just should represent it as this. Basically what I have in mind is that we have some sort of focus stack where the top element is the final focus target. So if the pointer is above a surface inside a window inside a cosmic stack it would hold a reference from `WlSurface`->`Window`->`CosmicStack`. The focus stack would implement `PointerTarget` and its `PartialEq` implementation would compare equality of the last element of the stack. In addition to this `PointerTarget` would get a new function that is called when focus moves from one target to another instead of directly calling `PointerTarget::leave` and `PointerTarget::enter`. The stack class could also have some function to test if the stack contains some target, which would walk the stack and check if any items equals. The new focus moved function could then walk the old and new stack to find out what exactly left and entered.

This requires to resolve the whole focus stack when the pointer moves in cosmic, should make it possible to keep most of the other focus logic in cosmic as is while still fixing the grab stuff. 

Feedback very welcome :)

PR #1335, opened 2024-02-17
Initial support for DRM async page flip

PolyMeilex:
blocked by: https://github.com/Smithay/drm-rs/pull/190


nferhat:
~~Well I'm no Smithay maintainer but still curious whether this will help with integrating `tearing-control-v1` or not? Because it seems quite promising!~~

It actually does.

PolyMeilex:
> ~~Well I'm no Smithay maintainer but still curious whether this will help with integrating `tearing-control-v1` or not? Because it seems quite promising!~~
> 
> It actually does.

Yep, it's more or less the only use-case for this.

src/backend/drm/mod.rs
```diff
@@ -113,6 +113,16 @@ fn warn_legacy_fb_export() {
...
+/// Hint for DRM backend on how the surface should be presented
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PresentationMode {
+    /// Vertical synchronization
+    VSync,
+    /// Thearing presentationres
+    ASync,
+    // NOTE: Auto for detecting when thearing is desired would probably go here
+}
+
```
cmeissl, line 125: We might be able to get away without introducing some `Auto` variant. After all every frame needs to have a preference anyway. By adding `PresentationMode` to `queue_frame` this can be nicely communicated (nice work btw :+1: ) frame by frame. I think it would be nice to keep it this way and always let downstream decide without some internal mechanism to detect it. What we should do instead is to provide some helper representing some sane behavior that can be used when downstream has no preference on its own.

I was thinking about doing it this way:
- Moving the `PresentationMode` enum to our render element module and introducing `RenderElement::presentation_mode(&self) -> PresentationMode` function
- Adding a helper to `RenderFrameResult` of `DrmCompositor` to get a sane default `PresentationMode`

The `RenderFrameResult` already has a reference to all `Element`s assigned to planes, so it should be as easy
as to verify that all planes have something assigned that prefers tearing.

This way downstream should have to complete freedom to decide, but can opt in to auto easily by just forwarding the `PresentationMode` from `RenderFrameResult` to `DrmCompositor::queue_frame`.

Drakulix: Sounds good to me.

> The RenderFrameResult already has a reference to all Elements assigned to planes, so it should be as easy
as to verify that all planes have something assigned that prefers tearing.

This is a the part however that might warrent allowing customization in some way (perhaps with a closure, like we have the `default_primary_output_compare` and the likes?). E.g. some compositors might want to tear if a single element is added to the primary-plane that preferrs tearing (perhaps even dependant on the content-type?). To name a more concrete example, some FPS overlay on an overlay plane shouldn't be blocking tearing of an directly scanned-out fullscreen game on the primary plane. Of course interactions here can be extremly subtle, which is why I would also prefer to just delegate those issues to downstream.

src/backend/drm/mod.rs
```diff
@@ -113,6 +113,16 @@ fn warn_legacy_fb_export() {
     });
 }
 
+/// Hint for DRM backend on how the surface should be presented
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PresentationMode {
+    /// Vertical synchronization
+    VSync,
+    /// Thearing presentationres
+    ASync,
```
YaLTeR, line 122:
```suggestion
    Async,
```
It's one word after all, also line above Tearing :see_no_evil:

src/backend/drm/mod.rs
cmeissl: I mean, everything the helper function uses internally should be public accessible. So I don't think we need to complicate the default function, especially as it should be trivial to implement. The FPS overlay is a good example, so maybe just a single parameter if it should only check the primary plane or all assigned planes. Everything more complicated than that might be better handled completely downstream.

Drakulix: Sounds good to me

src/backend/drm/mod.rs
PolyMeilex: Done

src/backend/drm/mod.rs
PolyMeilex: Done, but on `Element` rather than `RenderElement`, as `R: Renderer` generic was kinda annoying to deal with.

Elements can choose to request VSync, Async, or set no preference. Based on that we search for common denominator.

The whole change is self-contained in this new commit: https://github.com/Smithay/smithay/pull/1335/commits/a7e1c32fdb9f3dd1e94c9f106cacac5e8c678d7a

PolyMeilex:
Ok, I belive this is ready.

Although I still can't test the atomic path, even tho my driver started to report `AtomicASyncPageFlip` as supported recently, any attempt to page flip results in `Invalid argument (os error 22)` (perhaps it's a code issue so any testing with presentation mode flipped to Async would be welcome, just flip VSync to Async in anvil's udev.rs)

Running with `SMITHAY_USE_LEGACY=1` on the other hand works flawlesly for me already (if one can use flawless to describe tearing :smile:).

src/backend/renderer/element/surface.rs
```diff
@@ -489,6 +489,12 @@ impl<R: Renderer + ImportAll> Element for WaylandSurfaceRenderElement<R> {
     fn kind(&self) -> Kind {
         self.kind
     }
+
+    fn presentation_mode(&self) -> Option<PresentationMode> {
+        // On Wayland assume VSync as default preference
+        Some(PresentationMode::VSync)
+        // TODO: Return Async in case wp_tearing is attatched to the surface
```
YaLTeR, line 496: This is waiting on the tearing control PR?

PolyMeilex: Yep, or the other way around, I don't care. Although the protocol PR is probably more merge ready. We could also just merge the PRs into one.

cmeissl:
Okay, so this is not going to work this way unfortunately. Async page flips have some special limitations we need to account for. For an async page flip to succeed the atomic commit is only allowed to change the framebuffer on the primary plane. Every other change will result in the page flip request to be declined. 

This unfortunately also includes `IN_FENCE_FD` required for explicit sync. But this is something that is being worked on, so imo we should not special case this and just wait for it to land. Currently we do not support explicit sync for direct scan-out anyway.

Not sure what to do about `FB_DAMAGE_CLIPS`, need to check what the plan for this is. (worst case is we disable it for async page flips).

There are also driver specific limitations like for example intel not supporting async page flips if the plane uses some compressed tiling format.

So imo the most sane way to handle this is to do best guessing in `render_frame` if an async page flip might work.
Then in `queue_frame` do trail and error with a retry without async flip and remember that it failed for this frame state.
`render_frame` would then use the state as a feedback loop to prevent endless retries. The failed state should reset in case the format or other stuff changes allowing to re-test.

I prototyped some parts of this and will post a patch when its ready.

cmeissl:
Okay, I put something together that *should* work (at least it seems to work on my system). 
https://github.com/cmeissl/smithay/tree/drm_async_page_flip (@PolyMeilex PR branch rebased on top of current master)

Curious if this also works on AMD and Nvidia now @PolyMeilex @YaLTeR @Drakulix 
Filtering for linear formats is probably not going to work on Nvidia and probably also not needed.
Intel also does support async page flips with some modifiers (probably only compressed one are not supported, which kind of makes sense), but I was too lazy to put together a list of working formats.

PolyMeilex:
@cmeissl Atomic backend now works on my AMD systems, looks promising.

Rebased on master, and cherrypicked the cmeissl's commit (with some minnor reverts of debug code edits)

PolyMeilex:
Confits resolved, everything seems to still work as expected on my AMD machines both with atomic and legacy

andry-dev:
Hi, what is the status for this PR? I would like it for `niri` (the compositor I currently use) to implement tearing, but I'm OOTL regarding what is missing to improve/merge this PR (and eventually submit a PR to niri), since it's been an year.

shibahaczix:
Are we tear yet?

cmeissl:
I will try to find some time in the next two weeks to go over the PR

@PolyMeilex are you fine with me updating the PR?

PolyMeilex:
> @PolyMeilex are you fine with me updating the PR?

For sure, go ahead

relativemodder:
just curious, any progress on this? it's really annoying that we don't have the tearing support in niri or cosmic because of this

shibahaczix:
I'm also curious. (also pretty annoying for me)

alexmozaidze:
Not to be that guy, but it's 2026 already. Gamers demand answers from the Niri gods. Will you bestow upon us the godly feature of tearing so we may dominate COD lobbies once more?

On a more serious note, what's missing as of now? What needs to be changed here or in the codebase so I could contribute? I know Rust and I want this feature badly enough to drop everything I do right now and try working on it.

nferhat:
> Not to be that guy, but it's 2026 already. Gamers demand answers from the Niri gods. Will you bestow upon us the godly feature of tearing so we may dominate COD lobbies once more?
> 
> On a more serious note, what's missing as of now? What needs to be changed here or in the codebase so I could contribute? I know Rust and I want this feature badly enough to drop everything I do right now and try working on it.

- #1325 To implement the protocol that lets clients choose for themselves if they want to have tearing presentation or not
- Rebasing this branch onto main again (smithay moves really quickly, and as a side effect of this leaving some fork/branch for a little could make it a hellish nightmare to rebase)
- Free time for smithay maintainers and compositors devs to write implementation in their respective compositors, test it, make sure there's nothing wrong, test it even more, point out flaws

And after that, if it doesn't require multiple rounds of such review, then it can potentially get merged. We all want lower latency in our games, me included, and I am avid to implement this and make use of it, but having a solid implementation backing it more important.

shibahaczix:
> Not to be that guy, but it's 2026 already. Gamers demand answers from the Niri gods. Will you bestow upon us the godly feature of tearing so we may dominate COD lobbies once more?
> 
> On a more serious note, what's missing as of now? What needs to be changed here or in the codebase so I could contribute? I know Rust and I want this feature badly enough to drop everything I do right now and try working on it.

Don't forget explicit sync.

gabriele2000:
> Don't forget explicit sync.

Explicit sync is in Smithay since 2024 and it has been perfected since 2025

PolyMeilex:
Ok, it took a bit of patience, but I managed to rebase this.
The anvil implementation did not survive the rebase tho, I did a minimal reimpl instead.

It's still possible to test this by editing this line: https://github.com/Smithay/smithay/blob/7ea3f73f8d38faa50a78f4035fe25c0b42a25337/src/backend/renderer/element/surface.rs#L432-L432
It's also worth noting that I still don't have a full grasp of what's going on in the Cmeissl's commit that made the atomic backend work, that's why I'm keeping it as a separate WIP commit.

Open question: Would dropping the seemingly more problematic atomic backend implementation and merging just the legacy impl make sense? Just to get things rolling, or should I rather just focus on trying to grasp what's needed for less hacky atomic impl?

cmeissl:
> Ok, it took a bit of patience, but I managed to rebase this. The anvil implementation did not survive the rebase tho.
> 
> It's still possible to test this by editing this line in anvil: https://github.com/PolyMeilex/smithay/blob/1072da79b699cbfb18ceac51541f847a4d907114/anvil/src/udev.rs#L1674 I will try to bring back RenderElement based presentation mode selection to Anvill soon'ish.
> 
> It's also worth noting that I still don't have a full grasp of what's going on in the Cmeissl's commit that made the atomic backend work, that's why I'm keeping it as a separate WIP commit.
> 
> Open question: Would dropping the seemingly more problematic atomic backend implementation and merging just the legacy impl make sense? Just to get things rolling, or should I rather just focus on trying to grasp what's needed for less hacky atomic impl?

Thanks again for taking the time to get this forward. imo it does not make sense without the atomic backend. Also some of the things  of my commit should not longer be required as some of the limitations have been lifted in the kernel. But I need some time to look into this again :(

PolyMeilex:
> Also some of the things of my commit should not longer be required as some of the limitations have been lifted in the kernel. But I need some time to look into this again :(

The one thing I noticed is that your Linear format filters did not survive the rebase, but the impl still seems to work fine, so perhaps that's one of the limitations lifted?

cmeissl:
> > Also some of the things of my commit should not longer be required as some of the limitations have been lifted in the kernel. But I need some time to look into this again :(
> 
> The one thing I noticed is that your Linear format filters did not survive the rebase, but the impl still seems to work fine, so perhaps that's one of the limitations lifted?

Possible, I don't remember all details. But one thing that should be lifted is the IN_FENCE_FD iirc.

PR #1341, opened 2024-02-24
fix buffer offset handling

cmeissl:
fixes the regression introduced in https://github.com/Smithay/smithay/pull/1123#issuecomment-1953661036
(not that is was working correctly before, but pointer was broken at bit less)

what I do not like is all the manual plumbing in the Dnd handling, handling this in smithay will require re-structuring a few things. From the spec of DnD start:

> The icon surface is an optional (can be NULL) surface that provides an icon to be moved around with the cursor. Initially, the top-left corner of the icon surface is placed at the cursor hotspot, but subsequent wl_surface.attach request can move the relative position.

According to this we need the cursor hotspot at the start of the grab to initialize the DnD icon offset. But as the hotspot is stored in the current pointer surface we can no access it in smithay. On commit of the DnD icon we need to be able to change the current DnD offset by the surface offset if any. But again we can not do this internally as we have no access to the icon.

We can probably do better by either:
- moving some parts that are now in anvil/downstream into smithay
- providing some helpers
- requiring access to some state held by downstream (similar to `SeatState`)

based on https://github.com/Smithay/smithay/pull/1340

### DIFF
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -152,6 +152,22 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
             }
             if let Some(window) = self.window_for_surface(&root) {
                 window.0.on_commit();
+
+                if &root == surface {
+                    let buffer_offset = with_states(surface, |states| {
+                        states
+                            .cached_state
+                            .get::<SurfaceAttributes>()
+                            .current()
+                            .buffer_delta
+                            .take()
+                    });
+
+                    if let Some(buffer_offset) = buffer_offset {
+                        let current_loc = self.space.element_location(&window).unwrap();
+                        self.space.map_element(window, current_loc + buffer_offset, false);
+                    }
+                }
             }
         }
         self.popups.commit(surface);
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -178,6 +178,21 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
             });
         }
 
+        if matches!(&self.dnd_icon, Some(icon) if &icon.surface == surface) {
+            let dnd_icon = self.dnd_icon.as_mut().unwrap();
+            with_states(&dnd_icon.surface, |states| {
+                let buffer_delta = states
+                    .cached_state
+                    .get::<SurfaceAttributes>()
+                    .current()
+                    .buffer_delta
+                    .take()
+                    .unwrap_or_default();
+                tracing::trace!(offset = ?dnd_icon.offset, ?buffer_delta, "moving dnd offset");
+                dnd_icon.offset += buffer_delta;
+            });
+        }
+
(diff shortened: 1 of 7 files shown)


cmeissl:
@YaLTeR would be nice to get some feedback from you

cmeissl:
@ids1024 @Drakulix @PolyMeilex would love to get some feedback on this topic, maybe we can find some better solution together.

YaLTeR:
Test cases:
1. GTK4: [Identity](https://flathub.org/apps/org.gnome.gitlab.YaLTeR.Identity). Open a video, play it, DnD while playing, the DnD icon should be a centered video.
2. [wleird-attach-delta-loop](https://gitlab.freedesktop.org/emersion/wleird/-/blob/master/attach-delta-loop.c?ref_type=heads), see what it does in Weston for instance.

cmeissl:
> Test cases:
> 
>     1. GTK4: [Identity](https://flathub.org/apps/org.gnome.gitlab.YaLTeR.Identity). Open a video, play it, DnD while playing, the DnD icon should be a centered video.
> 
>     2. [wleird-attach-delta-loop](https://gitlab.freedesktop.org/emersion/wleird/-/blob/master/attach-delta-loop.c?ref_type=heads), see what it does in Weston for instance.

Number 2 is actually a good example on why and what it is currently broken in smithay. The current implementation offsets all surfaces by the buffer offset, but weston actually modifies the position of the toplevel. Without SSD (which weston does not have) it just looks the same, but it actually does something completely different.

The offset should (and this is how it is done in weston) be handled in a role specify way. For pointer it defines that it should decrement the hotspot, dnd defines it as adding it to the current offset, toplevel will be moved, ...
The hotspot for pointer is reset when the client sets the cursor image, this also explains why pointer is currently so broken in the current implementation.
(Also a good explanation on why the surface will not move in sway and probably also most other tiling compositors. So not sure what you want to do in niri, but for toplevels it should be compositor policy and nothing smithay should internally handle)

I hope this helps to clarify the motivations behind these changes and why it looked like it was working, while it was actually broken.

cmeissl:
Okay, I moved the whole buffer offset handling into anvil. I do not like to have more `Mutex` stuff for this in `smithay` itself and only handling the cursor hotspot in smithay is inconsistent.

anvil/src/shell/mod.rs
```diff
@@ -152,6 +152,22 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
             }
             if let Some(window) = self.window_for_surface(&root) {
                 window.0.on_commit();
+
+                if &root == surface {
```
YaLTeR, line 156: For subsurfaces it is supposed to do nothing?

cmeissl: That's not so easy to answer, it might have been intended to also apply to subsurfaces. But as mutter and wlroots always ignored the offset for subsurface [weston has been changed](https://gitlab.freedesktop.org/wayland/wayland/-/issues/389) to also ignore it. So imo we should follow that and just accept this as the common behavior.

PR #1342, opened 2024-02-24
[wayland] Implement XdgForeign

PolyMeilex:
This also includes a SCTK test client, not sure if we want that in the codebase, but I committed it for now, we can just drop the commit as needed.


Drakulix (approved):
LGTM (waiting for a decision regarding the example before merging)

PolyMeilex:
Test client dropped, SCTK sounds like a better place, as there it can serve as both example and a test client at the same time.

PR #1343, opened 2024-02-25
[wayland.compositor] Removable hooks

PolyMeilex:
[[wayland.compositor] Removable hooks](https://github.com/Smithay/smithay/commit/2f1833259db1737c3613bf4b3344221570e60fee) - Adds `HookId`, and a way to use it to unregister the hooks. Also brings destruction hook inline with other hooks by unlocking the mutex during a hook call.
[[wayland.compositor] Use lock_user_data for all methods](https://github.com/Smithay/smithay/commit/a7e52006726aceb3070b8e0e5140d990dfc64a7e) - While at it, use user data getter added in previous commit in other unrelated functions to clean up the code a bit.

### DIFF
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -134,7 +134,7 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
                     }
                 }
             }
-        })
+        });
     }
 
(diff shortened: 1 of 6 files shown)


Drakulix (approved):
LGTM

PR #1346, opened 2024-02-25
Perf/reduce mem allocations

cmeissl:
attempt on reducing mem allocations in hot-code paths and some performance improvements

not sure how to best handle the `Space::refresh`/`Output::enter`/`Output::leave` stuff. this is called way to often
imo and we need to get the alive checks out of enter/leave.

### DIFF
--- a/src/backend/renderer/damage/mod.rs
+++ b/src/backend/renderer/damage/mod.rs
@@ -789,7 +789,8 @@ impl OutputDamageTracker {
         <R as Renderer>::TextureId: Texture,
         F: FnOnce(&mut R) -> Result<(), <R as Renderer>::Error>,
     {
-        let (output_size, output_scale, output_transform) = self.mode.clone().try_into()?;
+        let (output_size, output_scale, output_transform) =
+            std::convert::TryInto::<(Size<i32, Physical>, Scale<f64>, Transform)>::try_into(&self.mode)?;
 
         // Output transform is specified in surface-rotation, so inversion gives us the
         // render transform for the output itself.
--- a/src/backend/renderer/damage/mod.rs
+++ b/src/backend/renderer/damage/mod.rs
@@ -282,6 +282,9 @@ pub struct OutputDamageTracker {
     mode: OutputModeSource,
     last_state: RendererState,
     damage_shaper: DamageShaper,
+    damage: Vec<Rectangle<i32, Physical>>,
+    element_damage: Vec<Rectangle<i32, Physical>>,
+    opaque_regions: Vec<(usize, Vec<Rectangle<i32, Physical>>)>,
     span: tracing::Span,
 }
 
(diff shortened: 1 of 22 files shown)


src/backend/renderer/gles/mod.rs
```diff
@@ -2324,6 +2329,7 @@ impl Renderer for GlesRenderer {
 
         let current_projection = flip180 * transform.matrix() * renderer;
         let span = span!(parent: &self.span, Level::DEBUG, "renderer_gles2_frame", current_projection = ?current_projection, size = ?output_size, transform = ?transform).entered();
+        let vertices = std::mem::take(&mut self.vertices);
```
Drakulix, line 2332: I am not sure, why we have to move this, given the frame borrows the renderer anyway?

src/backend/renderer/damage/mod.rs
```diff
@@ -454,17 +461,22 @@ impl OutputDamageTracker {
...
+        self.damage.clear();
+        self.element_damage.clear();
+        self.opaque_regions.clear();
+
         let mut element_render_states = RenderElementStates {
             states: HashMap::with_capacity(elements.len()),
         };
 
+        let mut damage = std::mem::take(&mut self.damage);
+        let mut element_damage = std::mem::take(&mut self.element_damage);
```
Drakulix, line 478: This will cause new Vec's to be created, no? Why are we not directly using `self.damage` here, we just cleared it anyway?

Drakulix (commented):
A couple of small nitpicks and questions, but otherwise this looks like a really nice improvement so far.

src/backend/renderer/gles/mod.rs
cmeissl: Yeah, no reason to do this. I had something in mind, but can't exactly remember :shrug:

src/backend/renderer/damage/mod.rs
cmeissl: It is returned here: https://github.com/Smithay/smithay/blob/13bc4241c52be424fdacf2a60000a45b9e9dae87/src/backend/renderer/damage/mod.rs#L681 and an empty Vec does not allocate. Anyway, this is also just some leftover from searching for a error and we can just use `self.damage`, yeah.

Drakulix: Right and empty Vec does nothing, but it is still confusing to read. If there is no reason to do it this way, I would prefer to just use the value directly in all these cases.

src/backend/renderer/gles/mod.rs
cmeissl: done

src/backend/renderer/damage/mod.rs
cmeissl: mostly done, we need to take `element_damage` as we move it around a bit

Drakulix (approved):

PR #1349, opened 2024-03-01
Update `calloop` to 0.13.0

ids1024:


Drakulix:
Looks like we need a version specifier again for calloop on the docs job.

PR #1350, opened 2024-03-02
Run Xwayland using `-displayfd` and a pipe

colinmarc:
This method, while undocumented, is a lot more robust than an intermediate shell. It works by passing a pipe to the Xwayland cmdline invocation. When Xwayland is ready, it writes the display number and a newline to the pipe.

While we're at it, we can remove the hard dependency on Calloop, so that the Xwayland instance can be used with other event loops (my compositor uses `mio`). This isn't enough to achieve that completely, since `X11Wm::new` takes a `LoopHandle` as an argument. I will address that in a separate PR.

c.f. weston's implementation: https://gitlab.freedesktop.org/wayland/weston/-/blob/main/frontend/xwayland.c

### DIFF
--- a/src/xwayland/xserver.rs
+++ b/src/xwayland/xserver.rs
@@ -1,231 +1,373 @@
-/*
- * Steps of XWayland server creation
- *
- * Sockets to create:
- * - a pair for XWayland to connect to smithay as a wayland client, we use our
- *   end to insert the XWayland client in the display
- * - a pair for smithay to connect to XWayland as a WM, we give our end to the
- *   WM and it deals with it
- * - 2 listening sockets on which the XWayland server will listen. We need to
- *   bind them ourselves so we know what value put in the $DISPLAY env variable.
- *   This involves some dance with a lockfile to ensure there is no collision with
- *   an other starting xserver
- *   if we listen on display $D, their paths are respectively:
- *   - /tmp/.X11-unix/X$D
- *   - @/tmp/.X11-unix/X$D (abstract socket)
- *
- * The XWayland server is spawned via an intermediate shell
- * -> wlroots does a double-fork while weston a single one, why ??
- *    -> https://stackoverflow.com/questions/881388/
- * -> once it is started, it will check if SIGUSR1 is set to ignored. If so,
- *    if will consider its parent as "smart", and send a SIGUSR1 signal when
- *    startup completes. We want to catch this so we can launch the VM.
- * -> we need to track if the XWayland crashes, to restart it
- *
- * cf https://github.com/swaywm/wlroots/blob/master/xwayland/xwayland.c
- *
- * Setting SIGUSR1 handler is complicated in multithreaded program, because
- * Xwayland will send SIGUSR1 to the process, and if a thread cannot handle
- * SIGUSR1, that thread will be killed.
- *
- * Double-fork can tackle this issue, but this is also very complex in a
- * a multithread program, after forking only signal-safe functions can be used.
- * The only workaround is to fork early before any other thread starts, but
- * doing so will expose an unsafe interface.
- *
- * We use an intermediate shell to translate the signal to simple fd IO.
- * We ask sh to setup SIGUSR1 handler, and in a subshell mute SIGUSR1 and exec
- * Xwayland. When the SIGUSR1 is received, it can communicate to us via redirected
- * STDOUT.
- */
 use std::{
     env,
     ffi::OsStr,
-    io::{self, Read},
+    os::fd::{BorrowedFd, OwnedFd},
     os::unix::{
-        io::{AsRawFd, BorrowedFd, RawFd},
+        io::{AsRawFd, RawFd},
         net::UnixStream,
         process::CommandExt,
     },
-    process::{ChildStdout, Command, Stdio},
+    process::Command,
     sync::{Arc, Mutex},
 };
 
-use calloop::{
-    channel::{self, sync_channel, Channel, SyncSender},
-    generic::Generic,
-    Interest, LoopHandle, Mode,
-};
-use wayland_server::{
-    backend::{ClientData, ClientId, DisconnectReason},
-    Client, DisplayHandle,
-};
+use tracing::{info, trace};
+use wayland_server::backend::ClientData;
+use wayland_server::backend::DisconnectReason;
+use wayland_server::{Client, DisplayHandle};
 
-use tracing::{error, info, instrument};
+use crate::{utils::user_data::UserDataMap, wayland::compositor::CompositorClientState};
 
 use super::x11_sockets::{prepare_x11_sockets, X11Lock};
-use crate::utils::user_data::UserDataMap;
-#[cfg(feature = "wayland_frontend")]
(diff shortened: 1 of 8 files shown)


src/xwayland/xserver.rs
```diff
@@ -1,509 +1,378 @@
...
+        let (wl_x11, wl_me) = UnixStream::pair()?;
+
+        let (lock, listen_sockets) = prepare_x11_sockets(display.into(), open_abstract_socket)?;
+        let display_number = lock.display_number();
+
+        // XWayland writes the the display number and a newline to this pipe when it's ready.
+        let (displayfd_recv, displayfd_send) =
+            rustix::pipe::pipe_with(rustix::pipe::PipeFlags::NONBLOCK | rustix::pipe::PipeFlags::CLOEXEC)?;
+
+        let mut command = Command::new("/home/colinmarc/src/xserver/out/bin/Xwayland");
```
i509VCB, line 131: This needs to be fixed before merge

src/xwayland/xserver.rs
```diff
@@ -1,509 +1,378 @@
...
+///
+/// To shut down the instance, dropping this handle is generally sufficient,
+/// along with the X11 connection that was passed to the window manager
+/// implementation, if any. The process will die once the connections to it
+/// are closed.
 #[derive(Debug)]
 pub struct XWayland {
-    inner: Arc<Mutex<Inner>>,
+    inner: Arc<Mutex<Instance>>,
+    source: calloop::generic::Generic<calloop::generic::FdWrapper<RawFd>>,
```
i509VCB, line 59: Is there a reason OwnedFd can't be used here?

src/xwayland/xserver.rs
```diff
@@ -1,509 +1,378 @@
...
-pub struct XWaylandSource {
-    channel: Channel<XWaylandEvent>,
+struct Instance {
+    display_lock: X11Lock,
+    x11_socket: Option<UnixStream>,
+    display_fd: OwnedFd,
 }
 
-impl calloop::EventSource for XWaylandSource {
+impl calloop::EventSource for XWayland {
```
i509VCB, line 262: The display number is technically inaccessible unless you use a Dispatcher since LoopHandle::insert_source needs to own the event source. I think a way around that could be providing XWayland via the metadata associated type but that may be an issue with double borrows.

src/xwayland/xserver.rs
colinmarc: Agh, thank you for catching that!

src/xwayland/xserver.rs
colinmarc: The event has the display number on it, or you can fetch it before you `insert_source`.

src/xwayland/xserver.rs
colinmarc: Both the handle and the source need the fd, and it made more sense to me for the handle to own it.

Drakulix (approved):
LGTM, just waiting on @i509VCB to resolve the open comments once those are fixed.

src/xwayland/xserver.rs
```diff
@@ -1,509 +1,382 @@
...
+///               │              │
+///               │              │
+///  ┌────────────▼─────────────┼────────────┐
+///  │  Compositor               │            │
+///  │                         ┌─┴───────┐    │
+///  │                         │ XWM     │    │
+///  │                         │         │    │
+///  │                         └─────────┘    │
+///  │                                        │
+///  └────────────────────────────────────────┘
```
i509VCB, line 46: How does this render in docs? iirc a `text` code block is needed for this type of graphics?

src/xwayland/xserver.rs
```diff
@@ -1,509 +1,382 @@
...
-///
-/// This is a very low-level interface, only notifying you when the connection
-/// with XWayland is up, or when it terminates.
-///
-/// Your WM code must be able to handle the XWayland server connecting then
-/// disconnecting several time in a row, but only a single connection will
-/// be active at any given time.
+/// Events generated by an [XWayland] instance.
 #[derive(Debug)]
 #[must_use = "Connection events must be handled to prevent fd leaking"]
```
i509VCB, line 68: Now that a RawFd is not placed on this any more, the must use warning is not applicable.

src/xwayland/xserver.rs
```diff
@@ -1,509 +1,382 @@
...
 
-/// Events generated by the XWayland manager
-///
-/// This is a very low-level interface, only notifying you when the connection
-/// with XWayland is up, or when it terminates.
-///
-/// Your WM code must be able to handle the XWayland server connecting then
-/// disconnecting several time in a row, but only a single connection will
-/// be active at any given time.
+/// Events generated by an [XWayland] instance.
```
i509VCB, line 66:
```suggestion
/// Events generated by an [`XWayland`] instance.
```

src/xwayland/xserver.rs
```diff
@@ -1,509 +1,382 @@
...
+    /// `DISPLAY` environment variable.
+    pub fn display_number(&self) -> u32 {
+        self.inner.lock().unwrap().display_lock.display_number()
     }
 
-    info!("Starting XWayland");
+    /// Returns a file descriptor which can be polled for readiness. When the fd
+    /// is readable, the XWayland server's readiness can be checked with
+    /// [take_socket](Self::take_socket).
+    pub fn poll_fd(&self) -> BorrowedFd<'_> {
```
i509VCB, line 235: This could just be an `AsFd` implementation?

src/xwayland/xserver.rs
```diff
@@ -1,509 +1,382 @@
...
+    /// Calling `take_socket` successfully transfers ownership of the connection
+    /// to the caller. After returning `Some` the first time, it will always
+    /// return `None`.
+    ///
+    /// An `Err` result is only returned if the XWayland instance has exited
+    /// unexpectedly.
+    ///
+    /// This is a low-level method. Using the instance as an event source is the
+    /// recommended way to interact with it.
+    pub fn take_socket(&mut self) -> std::io::Result<Option<UnixStream>> {
```
i509VCB, line 254: AsFd would make this redundant.

colinmarc: I don't think so - just because of the `take` part.

src/xwayland/xserver.rs
colinmarc: This matches `DisplayHandle`, fwiw.

src/xwayland/xserver.rs
colinmarc: Fixed. It's pretty now.

src/xwayland/xserver.rs
colinmarc: Fixed.

src/xwayland/xserver.rs
colinmarc: The backticks aren't necessary, and it looks worse to me, but I can change it if you really want.

colinmarc:
@i509VCB I think I addressed all your comments, if you don't mind taking another look.

i509VCB (approved):
Sorry it took some time to review this. Seems good to me.

PR #1351, opened 2024-03-02
Implement the xwayland shell protocol

colinmarc:
Depends on #1350 (but only trivially).

The xwayland shell protocol moves the responsibility of setting the "xwayland_surface" role on the wl_surface from the x11 side of the connection to the wayland side. Among other things, this avoids the xwm racing the lifecycle of the surface.

Since we can check the serial with a commit hook, this also removes the need for a ugly idle callback in the `Xwm`, and may allow for better separation of concerns in the future, for example moving the association state tracking out of Xwm.

Without the idle callback, we may be able to remove the hard dependency on calloop in the xwayland stack (which is my ulterior motive). It's still required for `send_selection` - I'll attempt to figure those issues out in another PR.

### DIFF
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -2,6 +2,7 @@ use std::cell::RefCell;
 
 #[cfg(feature = "xwayland")]
 use smithay::xwayland::{X11Wm, XWaylandClientData};
+
 use smithay::{
     backend::renderer::utils::on_commit_buffer_handler,
     desktop::{
@@ -141,7 +142,7 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
 
     fn commit(&mut self, surface: &WlSurface) {
         #[cfg(feature = "xwayland")]
-        X11Wm::commit_hook::<AnvilState<BackendData>>(surface);
+        X11Wm::commit_hook(self, surface);
 
         on_commit_buffer_handler::<Self>(surface);
         self.backend_data.early_import(surface);
--- a/anvil/src/shell/x11.rs
+++ b/anvil/src/shell/x11.rs
@@ -17,6 +17,7 @@ use smithay::{
             },
             SelectionTarget,
         },
+        xwayland_shell::{XWaylandShellHandler, XWaylandShellState},
     },
     xwayland::{
         xwm::{Reorder, ResizeEdge as X11ResizeEdge, XwmId},
@@ -41,6 +42,12 @@ impl OldGeometry {
 
     pub fn restore(&self) -> Option<Rectangle<i32, Logical>> {
         self.0.borrow_mut().take()
+    }
+}
+
+impl<BackendData: Backend> XWaylandShellHandler for AnvilState<BackendData> {
+    fn xwayland_shell_state(&mut self) -> &mut XWaylandShellState {
+        &mut self.xwayland_shell_state
     }
 }
 
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -99,10 +99,11 @@ use crate::{
 };
 #[cfg(feature = "xwayland")]
 use smithay::{
-    delegate_xwayland_keyboard_grab,
+    delegate_xwayland_keyboard_grab, delegate_xwayland_shell,
     utils::{Point, Size},
     wayland::selection::{SelectionSource, SelectionTarget},
     wayland::xwayland_keyboard_grab::{XWaylandKeyboardGrabHandler, XWaylandKeyboardGrabState},
+    wayland::xwayland_shell,
     xwayland::{X11Wm, XWayland, XWaylandEvent},
 };
 
@@ -147,6 +148,8 @@ pub struct AnvilState<BackendData: Backend + 'static> {
     pub presentation_state: PresentationState,
     pub fractional_scale_manager_state: FractionalScaleManagerState,
     pub xdg_foreign_state: XdgForeignState,
+    #[cfg(feature = "xwayland")]
+    pub xwayland_shell_state: xwayland_shell::XWaylandShellState,
 
     pub dnd_icon: Option<WlSurface>,
 
@@ -519,6 +522,9 @@ impl<BackendData: Backend + 'static> XWaylandKeyboardGrabHandler for AnvilState<
 #[cfg(feature = "xwayland")]
 delegate_xwayland_keyboard_grab!(@<BackendData: Backend + 'static> AnvilState<BackendData>);
 
+#[cfg(feature = "xwayland")]
+delegate_xwayland_shell!(@<BackendData: Backend + 'static> AnvilState<BackendData>);
+
 impl<BackendData: Backend> XdgForeignHandler for AnvilState<BackendData> {
     fn xdg_foreign_state(&mut self) -> &mut XdgForeignState {
         &mut self.xdg_foreign_state
@@ -615,6 +621,12 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
 
         let keyboard_shortcuts_inhibit_state = KeyboardShortcutsInhibitState::new::<Self>(&dh);
 
+        #[cfg(feature = "xwayland")]
+        let xwayland_shell_state = xwayland_shell::XWaylandShellState::new::<Self>(&dh.clone());
+
+        #[cfg(feature = "xwayland")]
+        XWaylandKeyboardGrabState::new::<Self>(&dh.clone());
+
         AnvilState {
             backend_data,
             display_handle: dh,
@@ -648,6 +660,8 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
             clock,
 
             #[cfg(feature = "xwayland")]
+            xwayland_shell_state,
+            #[cfg(feature = "xwayland")]
             xwm: None,
             #[cfg(feature = "xwayland")]
             xdisplay: None,
@@ -661,8 +675,6 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
     pub fn start_xwayland(&mut self) {
         use std::process::Stdio;
 
-        XWaylandKeyboardGrabState::new::<Self>(&self.display_handle.clone());
-
         let (xwayland, client) = XWayland::spawn(
             &self.display_handle,
             None,
@@ -674,7 +686,6 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
         )
         .expect("failed to start XWayland");
 
-        let dh = self.display_handle.clone();
         let ret = self
             .handle
             .insert_source(xwayland, move |event, _, data| match event {
@@ -682,7 +693,7 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
                     x11_socket,
                     display_number,
                 } => {
-                    let mut wm = X11Wm::start_wm(data.handle.clone(), dh.clone(), x11_socket, client.clone())
+                    let mut wm = X11Wm::start_wm(data.handle.clone(), x11_socket, client.clone())
                         .expect("Failed to attach X11 Window Manager");
 
                     let cursor = Cursor::load();
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -1,12 +1,33 @@
 //!
 //! Xwayland Window Manager module
 //!
-//! Provides an [`X11Wm`] type, which will register itself as a window manager for a previously spawned Xwayland instances,
-//! allowing backwards-compatibility by seemlessly integrating X11 windows into a wayland compositor.
+//! Provides an [`X11Wm`] type, which will register itself as a window manager
+//! for a previously spawned Xwayland instances, allowing
+//! backwards-compatibility by seamlessly integrating X11 windows into a wayland
+//! compositor.
 //!
-//! To use this functionality you must first spawn an [`XWayland`](super::XWayland) instance to attach a [`X11Wm`] to.
+//! To use this functionality you must first spawn an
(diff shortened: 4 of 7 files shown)


src/xwayland/xwm/mod.rs
```diff
@@ -2118,22 +2098,50 @@ fn handle_event<D: XwmHandler + 'static>(
...
+
+                        let serial_lo = msg.data.as_data32()[0];
+                        let serial_hi = msg.data.as_data32()[1];
+                        let serial = u64::from(serial_lo) | (u64::from(serial_hi) << 32);
+
+                        let surface_state = xsurface.state.clone();
+                        let mut guard = surface_state.lock().unwrap();
+                        guard.wl_surface_serial = Some(serial);
+
+                        xwm.unpaired_surfaces.insert(serial, msg.window);
```
colinmarc, line 2129: This is really unfortunate. We should only be setting this in the `else` below, but that would mean double-borrowing `state`. I stared at this for tens of minutes without seeing a way around it.

src/xwayland/xwm/mod.rs
```diff
@@ -2125,22 +2117,50 @@ fn handle_event<D: XwmHandler + 'static>(
...
+
+                        if let Some(wl_surface) =
+                            xwayland_shell::XWaylandShellHandler::xwayland_shell_state(state)
+                                .surface_for_serial(serial)
+                                .clone()
```
Drakulix, line 2153: I think if you changed this to `cloned`, you can avoid the double borrow.

(Referencing https://github.com/Smithay/smithay/pull/1351/commits/64ebcfa9a3229c6d916f194481a2389ac0c9fcec#r1510053974)

anvil/src/shell/x11.rs
```diff
@@ -44,9 +45,126 @@ impl OldGeometry {
     }
 }
 
+impl<BackendData: Backend> XWaylandShellHandler for CalloopData<BackendData> {
+    fn xwayland_shell_state(&mut self) -> &mut XWaylandShellState {
+        &mut self.state.xwayland_shell_state
+    }
+}
+
+impl<BackendData: Backend> XWaylandShellHandler for AnvilState<BackendData> {
```
Drakulix, line 54: I don't think we want a second trait and force downstream to awkwardly forward calls like this. I think it is fine to just drop the old matching logic and require a recent enough Xwayland version supporting this protocol.

src/xwayland/xwm/mod.rs
colinmarc: I can check later (mobile), but I'm pretty sure the issue is using `xwm` again after using `&mut state` at all, even in a temporary variable - there's an implicit `drop(xwm)` which makes the code work right now.

anvil/src/shell/x11.rs
colinmarc: This hack predates my PR, I just made it explicit. I'll try to explain. It has nothing to do with the old matching logic.

This is necessary because Anvil has an `EventLoop<CalloopData>` but a different state object (`AnvilState`), on which most handlers are implemented. Because `Xwm` requires `LoopHandle<D> where D: XwmHandler`, the existing code implemented `XwmHandler` directly on `CalloopData` instead of `AnvilState`. But then it had to call into the Xwm from one of those other handlers on `AnvilState`:

https://github.com/Smithay/smithay/blob/418190e4992ce642e6bac873307d4fc4fa9a1e89/anvil/src/shell/mod.rs#L142

I changed the commit hook to take a `D` as a parameter, so that broke. I actually think this hack is cleaner, in a way, although it's terrible on its face.

Note that this all goes away if remove the tight integration with calloop that xwm has. I would propose doing that by adding a `poll_fd()` and a `dispatch` method, mirroring the wayland side.

By the second trait, do you mean `XWaylandShellHandler`? I think having that separated makes sense, since it's just for the protocol itself. And that means people can supply their own xwm implementation.

Drakulix: > Note that this all goes away if remove the tight integration with calloop that xwm has. I would propose doing that by adding a poll_fd() and a dispatch method, mirroring the wayland side.

Good idea.

> By the second trait, do you mean XWaylandShellHandler? I think having that separated makes sense, since it's just for the protocol itself. And that means people can supply their own xwm implementation.

Then maybe we need to split it. But the duplicated x-wm callbacks are not nice to use.

colinmarc: Sorry, I'm a bit lost. What do you mean by duplicated callbacks?

Drakulix: Having the `XWaylandShellHandler` call into `XwmHandler`.

colinmarc: > Having the XWaylandShellHandler call into XwmHandler.

I don't see where it does that. The two traits are independent, except that `XwmHandler` requires `D` to also implement `XWaylandShellHandler`, since `X11wm` uses both to do the matching. (I should actually move the constraint to `X11Wm` to make that clearer, on reflection, but that's a trivial change).

What's happening here (where this comment is) is that we're implementing `XWaylandShellHandler` for both `CalloopData<BackendData>` and `AnvilState<BackendData`. The forwarding is just between two implementations of the same trait, because of an interaction between `X11Wm` weirdness and `Anvil` weirdness. There's no cross-pollination between `XWaylandShellHandler` and `XwmHandler`.

I'm really sorry if I'm just being thick! I have a head cold at the moment and having trouble focusing :)

anvil/src/shell/mod.rs
```diff
@@ -139,7 +138,7 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
 
     fn commit(&mut self, surface: &WlSurface) {
         #[cfg(feature = "xwayland")]
-        X11Wm::commit_hook::<CalloopData<BackendData>>(surface);
+        X11Wm::commit_hook(self, surface);
```
colinmarc, line 141: Referencing the discussion about trait forwarding below: https://github.com/Smithay/smithay/pull/1351/files/79648e06a287286f00520afc3d12efbdd4c38277..64ebcfa9a3229c6d916f194481a2389ac0c9fcec#r1511009746

Removing this hack (because of the addition of the `D` parameter to `X11Wm::commit_hook`) is what made the trait forwarding necessary, just for anvil.

anvil/src/shell/x11.rs
Drakulix: > I don't see where it does that.

Ah my god, the diff view had me confused... sorry!

Yeah it was the two `XwmHandler` implementations (once for `CalloopData<AnvilState>` and the other for `AnvilState`), that were throwing me off.

I don't think we can really keep those in anvil, even if a normal compositor wouldn't have to do this, because this encourages people to just copy this hacky approach.

We should get rid of CalloopData in anvil, is that something you feel like possibly tackling?

colinmarc: > I should actually move the constraint to X11Wm to make that clearer, on reflection, but that's a trivial change

Addressed in 6c675b5.

> We should get rid of CalloopData in anvil, is that something you feel like possibly tackling?

Between that and removing `X11Wm`'s dependency on `LoopHandle<D>`, and given that either would free us of this hack - I would rather work on the latter. :sweat_smile:

src/xwayland/xwm/mod.rs
```diff
@@ -1265,12 +1271,17 @@ impl X11Wm {
...
+fn handle_event<D>(
     loop_handle: &LoopHandle<'_, D>,
     state: &mut D,
     xwm_id: XwmId,
     event: Event,
-) -> Result<(), ReplyOrIdError> {
+) -> Result<(), ReplyOrIdError>
+where
+    D: XwmHandler,
+    D: xwayland_shell::XWaylandShellHandler,
```
Drakulix, line 1282: If we require this here, we don't really support compositors using the old association method, do we?

src/xwayland/xwm/mod.rs
```diff
@@ -2118,22 +2129,50 @@ fn handle_event<D: XwmHandler + 'static>(
...
-                                drop(_guard);
-                                X11Wm::new_surface(state, xwm_id, surface, wl_surface);
-                            }
+                        // This is the old, deprecated method for associating a
+                        // wl_surface with an X11 window. It can race with the
+                        // lifecycle of the wl_surface. Still, we can make it
+                        // available to compositors who want to support old
+                        // versions of xwayland by manually matching on ID in an
+                        // idle callback.
+                        surface.state.lock().unwrap().wl_surface_id = Some(wid);
```
Drakulix, line 2138: Referencing the previous comment.

src/xwayland/xwm/mod.rs
colinmarc: We require compositors to support the new association method. Whether they also handle the legacy method for old xwayland versions by checking for id-based associations in an idle callback (as I mention in the comment) is optional.

anvil/src/shell/x11.rs
Drakulix: So what needs to be done to get rid of this double implementation in anvil?

colinmarc: I took a look at changing `Xwm`'s hard dependency on calloop, and it was a bit too tangled up for me. I would love some help with that, or with ditching `CalloopData` in anvil. Either would remove the need for this hack.

colinmarc:
Rebased and removed the anvil hack. This is only blocked on #1350 now (although I could separate it out if you really want).

anvil/src/state.rs
```diff
@@ -616,40 +620,7 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
-                XWaylandEvent::Exited => {
-                    let _ = data.xwm.take();
-                }
-            });
-            if let Err(e) = ret {
-                tracing::error!("Failed to insert the XWaylandSource into the event loop: {}", e);
-            }
-            xwayland
-        };
+        let xwayland_shell_state = xwayland_shell::XWaylandShellState::new::<Self>(&dh.clone());
```
Drakulix, line 623: I guess we could move this into start_xwayland, just like the XwaylandKeyboardGrab protocol?

src/xwayland/xwm/mod.rs
```diff
@@ -1,18 +1,45 @@
...
+//! When an X11 window is created, XWayland will associate it with a
+//! `wl_surface` that it creates via the wayland connection. This happens in two
+//! steps:
+//!
+//!  - Via the [xwayland shell](crate::wayland::xwayland::shell) protocol, a
+//!    serial number is set on the surface, and it is given the
+//!    "xwayland_surface" role.
+//!  - Via the X11 connection, a matching `WL_SURFACE_SERIAL` atom is set on the
+//!    X11 window, which can be queried with
+//!    [`X11Surface::wl_surface_serial()`].
```
Drakulix, line 23: I don't think this description makes it clear to the user, what they are supposed to do. You describe an internal mechanism from the users perspective. Open questions are:
- Are they supposed to match the surfaces themselves? Or does smithay match them?
- Does this module work without the XwaylandShell protocol?

Drakulix (commented):
Some nitpicks, overall this is quite clean and I like the implementation a lot.

anvil/src/state.rs
colinmarc: It means making that field an `Option`. What if I just move `XwaylandKeyboardGrab` protocol out here?

src/xwayland/xwm/mod.rs
colinmarc: I rewrote this part, let me know what you think.

anvil/src/state.rs
Drakulix: sure

src/xwayland/xwm/mod.rs
Drakulix: I guess you could re-borrow the `xwm` after getting the surface, but meh.. this is fine.

src/xwayland/xwm/surface.rs
```diff
@@ -241,7 +248,7 @@ impl X11Surface {
     /// Returns if the window is currently mapped or not
     pub fn is_mapped(&self) -> bool {
         let state = self.state.lock().unwrap();
-        state.wl_surface.is_some() || state.mapped_onto.is_some()
+        state.mapped_onto.is_some()
```
Drakulix, line 251: I am not sure this is right for override-redirect windows. We don't set the mapped_onto for those, but those can still be mapped/unmapped in which case Xwayland will remove/re-create the surface.

In general it is not a correct assumption, that an X-window will only have a single wl_surface associated during it's whole lifetime.

colinmarc: I'm not sure about what the context for this function is, but it seems like it was wrong before as well?

Drakulix: yeah I guess just checking for the surface to exist would be enough?

src/xwayland/xwm/mod.rs
Drakulix: sounds good now!

src/xwayland/xwm/surface.rs
colinmarc: Changed, but I'm not sure I understand. :smile:

anvil/src/state.rs
colinmarc: Done.

colinmarc:
CI failures appear unrelated.

src/xwayland/xwm/surface.rs
Drakulix: You did in fact understand me correctly. :D

Drakulix:
CI failure should be fixed by https://github.com/Smithay/smithay/pull/1424/files, can you rebase?

Drakulix (approved):
Good to go, once CI passes. Thanks for putting in all the work!!

Issue #1352, opened 2024-03-05
With multiple subsurface, only last created (topmost) subsurface and parent get pointer input


ids1024:
This issue occurs in Anvil or cosmic-comp, and probably other compositors.

`Window::surface_under` calls `under_from_surface_tree`, which uses `from_surface_downward`. Then:

* The `filter` is called on the parent surface. It sets `found`, since the pointer is within the input region of the parent surface.
* The `post_filter` isn't called yet; it's called after iterating over subsurfaces.
* It then iterates through subsurfaces in stacking order. Starting with the most recently created subsurface, if `place_above`/`place_below` haven't been used to change the stacking order. This may or may not set `found`, depending on whether or not the pointer is over the subsurface.
* At the end of the recursive call, the `post_filter` is invoked, and returns `false`. Since `found` has been set (either to the parent surface, or the first subsurface)
* It never gets to the remaining subsurfaces.

So there seem there are two issues here:
* `from_surface_downward` iterates over subsurfaces before calling the `post_filter`, which doesn't seem to be how `under_surface_from_tree` expects the function to behave. Is this is correct for `from_surface_downward`, `under_surface_from_tree` could be changed to use `TraversalAction::SkipChildren` when `found` is set.
* But more fundamentally: it doesn't seem to have any way to handle stacking order between subsurfaces and the parent surface
  - [wl_subsurface::place_above](https://wayland.app/protocols/wayland#wl_subsurface:request:place_above) and [wl_subsurface::place_above](https://wayland.app/protocols/wayland#wl_subsurface:request:place_above) mention that "The reference surface must be one of the sibling surfaces, or the parent surface", so I think it needs to keep track of which subsurfaces are above and below the parent, and iterate over those above the parent before giving pointer focus the parent surface?

Not sure how correct the other functions dealing with the surface tree are. More testing could be helpful; not sure what tests wlcs has for that.

Drakulix:
Working on a fix right now, `under_from_surface_tree` simply shouldn't handle everything in the filter, but use the processor properly.

`children` contains the surface itself, so no this is otherwise right, the handling in the filter just breaks the order.

PR #1354, opened 2024-03-07
Fix various slight tree-iteration issues

Drakulix:
Stumbled upon these looking into https://github.com/Smithay/smithay/issues/1352.

Some clear mistakes (like only using the `filter` closure in `under_for_surface_tree`), but mostly just some inconsistencies handling `WindowSurfaceType`.

Draft, because this needs more testing.


ids1024:
Testing with Anvil, this seems to work well. The pointer enters and leaves subsurfaces as expected, or the parent surface when there's not a subsurface in front of it.

PR #1356, opened 2024-03-08
Add linux-drm-syncobj-v1 protocol

ids1024:
So far just exposes the protocol but doesn't do anything. https://github.com/Smithay/smithay/pull/1327 does indeed make that part mostly painless instead of having impl bounds to deal with and a delegate macro.

### DIFF
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -2,6 +2,10 @@ use std::cell::RefCell;
 
 #[cfg(feature = "xwayland")]
 use smithay::xwayland::XWaylandClientData;
+
+#[cfg(feature = "udev")]
+use smithay::wayland::drm_syncobj::DrmSyncobjCachedState;
+
 use smithay::{
     backend::renderer::utils::on_commit_buffer_handler,
     desktop::{
@@ -112,7 +116,17 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
 
     fn new_surface(&mut self, surface: &WlSurface) {
         add_pre_commit_hook::<Self, _>(surface, move |state, _dh, surface| {
+            #[cfg(feature = "udev")]
+            let mut acquire_point = None;
             let maybe_dmabuf = with_states(surface, |surface_data| {
+                #[cfg(feature = "udev")]
+                acquire_point.clone_from(
+                    &surface_data
+                        .cached_state
+                        .get::<DrmSyncobjCachedState>()
+                        .pending()
+                        .acquire_point,
+                );
                 surface_data
                     .cached_state
                     .get::<SurfaceAttributes>()
@@ -125,6 +139,21 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
                     })
             });
             if let Some(dmabuf) = maybe_dmabuf {
+                #[cfg(feature = "udev")]
+                if let Some(acquire_point) = acquire_point {
+                    if let Ok((blocker, source)) = acquire_point.generate_blocker() {
+                        let client = surface.client().unwrap();
+                        let res = state.handle.insert_source(source, move |_, _, data| {
+                            let dh = data.display_handle.clone();
+                            data.client_compositor_state(&client).blocker_cleared(data, &dh);
+                            Ok(())
+                        });
+                        if res.is_ok() {
+                            add_blocker(surface, blocker);
+                            return;
+                        }
+                    }
+                }
                 if let Ok((blocker, source)) = dmabuf.generate_blocker(Interest::READ) {
                     if let Some(client) = surface.client() {
                         let res = state.handle.insert_source(source, move |_, _, data| {
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -131,7 +131,7 @@ use std::{
 };
 
 use drm::{
-    control::{connector, crtc, framebuffer, plane, Mode, PlaneType},
+    control::{connector, crtc, framebuffer, plane, Device as _, Mode, PlaneType},
     Device, DriverCapability,
 };
 use drm_fourcc::{DrmFormat, DrmFourcc, DrmModifier};
@@ -202,6 +202,22 @@ enum ScanoutBuffer<B: Buffer> {
     Wayland(crate::backend::renderer::utils::Buffer),
     Swapchain(Slot<B>),
     Cursor(GbmBuffer),
+}
+
+impl<B: Buffer> ScanoutBuffer<B> {
+    fn acquire_point(
+        &self,
+        signaled_fence: Option<&Arc<OwnedFd>>,
+    ) -> Option<(SyncPoint, Option<Arc<OwnedFd>>)> {
+        if let Self::Wayland(buffer) = self {
+            // Assume `DrmSyncobjBlocker` is used, so acquire point has already
+            // been signaled. Instead of converting with `SyncPoint::from`.
+            if buffer.acquire_point().is_some() {
+                return Some((SyncPoint::signaled(), signaled_fence.cloned()));
+            }
+        }
+        None
+    }
 }
 
 impl<B: Buffer> ScanoutBuffer<B> {
@@ -1567,6 +1583,7 @@ where
     supports_fencing: bool,
     direct_scanout: bool,
     reset_pending: bool,
+    signaled_fence: Option<Arc<OwnedFd>>,
 
     framebuffer_exporter: F,
 
@@ -1633,6 +1650,24 @@ where
         cursor_size: Size<u32, BufferCoords>,
         gbm: Option<GbmDevice<G>>,
     ) -> FrameResult<Self, A, F> {
+        let signaled_fence = match surface.create_syncobj(true) {
+            Ok(signaled_syncobj) => match surface.syncobj_to_fd(signaled_syncobj, true) {
+                Ok(signaled_fence) => {
+                    let _ = surface.destroy_syncobj(signaled_syncobj);
+                    Some(Arc::new(signaled_fence))
+                }
+                Err(err) => {
+                    tracing::warn!(?err, "failed to export signaled syncobj");
+                    let _ = surface.destroy_syncobj(signaled_syncobj);
+                    None
+                }
+            },
+            Err(err) => {
+                tracing::warn!(?err, "failed to create signaled syncobj");
+                None
+            }
+        };
+
         let span = info_span!(
             parent: None,
             "drm_compositor",
@@ -1735,6 +1770,7 @@ where
                         primary_is_opaque: is_opaque,
                         direct_scanout: true,
                         reset_pending: true,
+                        signaled_fence,
                         current_frame,
                         pending_frame: None,
                         queued_frame: None,
@@ -4095,7 +4131,10 @@ where
             buffer: element_config.buffer.clone(),
             damage_clips,
             plane_claim,
-            sync: None,
+            sync: element_config
+                .buffer
+                .buffer
+                .acquire_point(self.signaled_fence.as_ref()),
         };
 
         let is_compatible = previous_state
--- a/src/backend/renderer/utils/wayland.rs
+++ b/src/backend/renderer/utils/wayland.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "backend_drm")]
+use crate::wayland::drm_syncobj::{DrmSyncPoint, DrmSyncobjCachedState};
 use crate::{
     backend::renderer::{buffer_dimensions, buffer_has_alpha, element::RenderElement, ImportAll, Renderer},
     utils::{Buffer as BufferCoord, Coordinate, Logical, Physical, Point, Rectangle, Scale, Size, Transform},
@@ -53,12 +55,24 @@ unsafe impl Send for RendererSurfaceState {}
 unsafe impl Sync for RendererSurfaceState {}
 
 #[derive(Debug)]
-struct InnerBuffer(WlBuffer);
+struct InnerBuffer {
+    buffer: WlBuffer,
+    #[cfg(feature = "backend_drm")]
+    acquire_point: Option<DrmSyncPoint>,
+    #[cfg(feature = "backend_drm")]
+    release_point: Option<DrmSyncPoint>,
+}
 
 impl Drop for InnerBuffer {
     #[inline]
     fn drop(&mut self) {
-        self.0.release();
+        self.buffer.release();
+        #[cfg(feature = "backend_drm")]
+        if let Some(release_point) = &self.release_point {
+            if let Err(err) = release_point.signal() {
+                tracing::error!("Failed to signal syncobj release point: {}", err);
+            }
+        }
     }
 }
 
@@ -68,12 +82,11 @@ pub struct Buffer {
     inner: Arc<InnerBuffer>,
 }
 
-impl From<WlBuffer> for Buffer {
-    #[inline]
-    fn from(buffer: WlBuffer) -> Self {
-        Buffer {
-            inner: Arc::new(InnerBuffer(buffer)),
-        }
+impl Buffer {
+    #[cfg(feature = "backend_drm")]
+    #[allow(dead_code)]
+    pub(crate) fn acquire_point(&self) -> Option<&DrmSyncPoint> {
+        self.inner.acquire_point.as_ref()
     }
 }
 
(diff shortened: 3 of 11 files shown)


src/wayland/drm_syncobj/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+        Self {
+            acquire_point: None,
+            release_point: None,
+        }
+    }
+
+    fn merge_into(self, into: &mut Self, _dh: &DisplayHandle) {
+        if self.acquire_point.is_some() && self.release_point.is_some() {
+            into.acquire_point = self.acquire_point;
+            into.release_point = self.release_point;
```
Drakulix, line 53: In case we unblock a bunch of commits at once, we still need to make sure to signal the release_point of unused buffers.

Drakulix:
> Reading a bit more about `IN_FENCE_FD` and `OUT_FENCE_PTR`, I think we need to use both if we want DRM scanout to not use implicit sync?

I am not sure, if you can use `OUT_FENCE_PTR` without `IN_FENCE_FD`. It wouldn't really make sense for the driver to support one but not the other. We probably should just set them both all the time.

> https://github.com/ValveSoftware/gamescope/blob/master/src/drm.cpp uses an `g_nAlwaysSignalledSyncFile` as the `IN_FENCE_FD`. Which is probably what we want as well (for direct scanout) if we only commit the buffer when we have already waited for it to be ready.

Yes, if we do CPU latching (that is polling the fence), then this seems to be the right approach for direct-scanout.

Though I would still just send the fence for composited frames, although we obviously could apply the principle there as well, as I tried here: https://github.com/pop-os/cosmic-comp/pull/291

However I don't see a benefit outside of cases, where the driver supports polling fences, but not `IN_FENCE_FD`. In those we could skip a CPU blocking `glFinish`, which also becomes a non-issue once we have a thread per surface.

Perhaps there are some advantages for VRR use cases, but I don't think that is worth the trouble for a first iteration.

> Then we can poll the fence produced by `OUT_FENCE_PTR`, at not signal release points for buffers used in direct scanout until that.

:+1:

src/wayland/drm_syncobj/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+use crate::backend::drm::DrmDeviceFd;
+
+mod sync_point;
+pub use sync_point::{DrmSyncPoint, DrmTimeline};
+
+/// Handler trait for DRM syncobj protocol.
+pub trait DrmSyncobjHandler {
+    // TODO better way to deal with this?
+    /// DRM device for importing syncobj file descriptors
+    fn import_device(&self) -> &DrmDeviceFd;
```
Drakulix, line 27: Can't we at least pass the `Client` here? I am not sure, if any part of the stack expects a particular device, but I would like to make this the same device we advertise via wl_drm or as a main device in cosmic's implementation.

src/backend/drm/compositor/mod.rs
```diff
@@ -203,6 +203,15 @@ enum ScanoutBuffer<B: Buffer> {
     Cursor(GbmBuffer),
 }
 
+impl<B: Buffer> ScanoutBuffer<B> {
+    fn acquire_point(&self) -> Option<SyncPoint> {
+        if let Self::Wayland(buffer) = self {
+            return buffer.acquire_point().cloned().map(SyncPoint::from);
```
Drakulix, line 209: So if a blocker is used this point will always be signalled, right? Then we don't need to create our own signalled fence.

ids1024: That was my question yesterday.

If we assume every smithay-based compositor will use `DrmSyncpointBlocker`, we don't need to get a fence from the sync point. But should create a fence that is already signaled to use as an `IN_FENCE_FD` to make sure no implicit sync occurs.

If `DmabufBlocker` is used for any surface that isn't using explicit sync, we can also submit a signaled `IN_FENCE_FD` for implict sync surfaces. (So we can just do that for every direct scanout plane.)

(If some Smithay compositors might not want to use blockers like that, they'll need a way to submit an `IN_FENCE_FD` from the acquire point.)

Drakulix: > If we assume every smithay-based compositor will use DrmSyncpointBlocker, we don't need to get a fence from the sync point. But should create a fence that is already signaled to use as an IN_FENCE_FD to make sure no implicit sync occurs.

If DmabufBlocker is used for any surface that isn't using explicit sync, we can also submit a signaled IN_FENCE_FD for implict sync surfaces. (So we can just do that for every direct scanout plane.)

So maybe we can make these blockers attach a hint on the `WlSurface`, that signals the `DrmCompositor`, that the buffer should be ready? And just fall back to implicit sync (no `IN_FENCE_FD`) otherwise?

I suppose no smithay compositor would want to use a different fence for a buffer. At least not when using the DrmCompositor, which just does direct scanout or submits an `IN_FENCE_FD` for compositing.

Anything else that might need custom fences (e.g. multi-gpu transfers directly scanned out or composing onto a plane) is not supported right now by the `DrmCompositor` anyway.

Drakulix: > Though it might be simpler for DrmCompositor to take a bool of whether or not to assume buffers are already ready.

Yeah, though then we need good docs to explain when it is safe to pass that bool. (And use the fences from the acquire points otherwise?)

src/wayland/drm_syncobj/mod.rs
ids1024: Good point. I *think* it shouldn't matter? In which case it would be simpler to take a `DrmDeviceFd` when creating the global.

But if it does matter, that a reason to stick with a trait method like this. Although, we probably would want to derive the node from the `Dmabuf`. Which should handle that, but also a device specified with https://github.com/Smithay/smithay/pull/1430. Right?

We could pass the `WlBuffer`. But if the implementation here assumes the buffer is created with `zwp_linux_dmabuf_v1`, we could pass the `Option<DrmNode>`.

Drakulix: > [...] we could pass the `Option<DrmNode>`

That would only be set for dmabuf <= 3 (and potentially >= 6), right? We have more information given the main device in cosmic. So imo the `WlBuffer` is the most flexible, as it also allows to get the `Client`.

Whether we make use of this in the end is a separate question, but I think we should offer this in the api.

ids1024:
Perhaps not using a blocker could make sense for a `cage`-like compositor, or as an optimization when there is only a single fullscreen application. Though I'm not sure what issues there might be if the `IN_FENCE_FD` isn't signaled for a long time (say, a broken client that never signals it?). At some point we'll want to do testing and see what works and if it can improve things.

Without blockers, a renderer may also need to handle the acquire points. A Vulkan renderer should import the timeline as a `VkSemaphore`. I think we also would need to use `EGL_ANDROID_native_fence_sync` in the EGL renderer, and wait on the acquire fence? I don't think we can rely on implicit sync.

(Perhaps it would be best to document that the blocker (or something equivalent) is required for Smithay's syncobj-v1 implementation, and generalize things later?)

Testing this: I see `vkgears` on cosmic-comp is only using direct scanout to an overlay plane when full-screened. In which case, `export_sync_file` runs successfully. Though I can't really tell if implicit sync is being avoided, and that this is preventing sync errors, since at least with `vkgears` I don't see artifacts if I removed the blocker and use a signaled in fence.

Wayland EGL on the Nvidia 555 driver, and Mesa Vulkan on git, both freeze if `BufferInner::drop` doesn't signal the release fence. So as long as we make sure to keep a reference to `Buffer` as long as we need the buffer, that seems to be working. Which seems correct for direct-scanout if we hold it until we get a page flip event.

Drakulix:
> Perhaps not using a blocker could make sense for a `cage`-like compositor, or as an optimization when there is only a single fullscreen application. Though I'm not sure what issues there might be if the `IN_FENCE_FD` isn't signaled for a long time (say, a broken client that never signals it?). At some point we'll want to do testing and see what works and if it can improve things.

Given commits could be blocking, this might permanently lock up a display without a timeout. It might be reasonable to never send any user-provided fences and instead require adding a Blocker, if one chooses the implement linux-drm-syncobj.

> Without blockers, a renderer may also need to handle the acquire points. A Vulkan renderer should import the timeline as a `VkSemaphore`. I think we also would need to use `EGL_ANDROID_native_fence_sync` in the EGL renderer, and wait on the acquire fence? I don't think we can rely on implicit sync.

We kinda do that already? SyncPoints can be awaited via the GPU driver in smithay currently and we use that for synchronizing multi-gpu copies and block for compositing before submitting.

> (Perhaps it would be best to document that the blocker (or something equivalent) is required for Smithay's syncobj-v1 implementation, and generalize things later?)

Yes, agreed.

> Testing this: I see `vkgears` on cosmic-comp is only using direct scanout to an overlay plane when full-screened. In which case, `export_sync_file` runs successfully. Though I can't really tell if implicit sync is being avoided, and that this is preventing sync errors, since at least with `vkgears` I don't see artifacts if I removed the blocker and use a signaled in fence.

That is the issue with races, I guess.

> Wayland EGL on the Nvidia 555 driver, and Mesa Vulkan on git, both freeze if `BufferInner::drop` doesn't signal the release fence. So as long as we make sure to keep a reference to `Buffer` as long as we need the buffer, that seems to be working. Which seems correct for direct-scanout if we hold it until we get a page flip event.

Sounds good!

ids1024:
> Given commits could be blocking, this might permanently lock up a display without a timeout. It might be reasonable to never send any user-provided fences and instead require adding a Blocker, if one chooses the implement linux-drm-syncobj.

Apparently fences have timeouts? Since the kernel wants to prevent an indefinite lock like this: https://docs.kernel.org/driver-api/dma-buf.html#indefinite-dma-fences

Presumably the timeout behavior would be similar to implict sync without a `DmabufBlocker`? In which case that's just consistent with the expected problems when not using a blocker.

So that does sound like something that could be a viable optimization for things like fullscreen applications, even if a general purpose compositor probably want to use blockers under most circumstances (for implicit or explicit sync). But it's an optimization that can be handled later, with more testing.

> We kinda do that already? SyncPoints can be awaited via the GPU driver in smithay currently and we use that for synchronizing multi-gpu copies and block for compositing before submitting.

Ah right, we already have support for that in the renderer. So if we want to use the `acquire_fence ` without a CPU wait, we'll want to pass the fence to EGL there. But that can be handled later.

> * Never pass user-provided fences to IN_FENCE_FD, only for compositing like we do now (and denylist the nvidia driver here for now).
> * Pass a signalled fence for direct-scanout, if downstream signals that it is using blockers (either automagically or with a flag on the DrmCompositor).
> * Otherwise just fallback to implicit sync for direct-scanout.

I'm not sure if we can automagically fallback to implicit sync. We could just not pass an `IN_FENCE_FD`, and ignore the acquire fence, but I don't think ignoring the explicit fence is guaranteed to work (with all drivers)?

I think we need to either:
1. Document that, at least for now, use of `DrmSyncPointBlocker` is required with the Smithay syncobj-v1 protocol.
   - If `drmSyncobjEventfd` isn't supported by the kernel, the protocol shouldn't be exposed.
   - I think this matches what Mutter is doing.
3. If a blocker isn't used, use `IN_FENCE_FD`, or import the fence into EGL
  - Since `IN_FENCE_FD` doesn't work on Nvidia, for now, `DrmSyncPointBlocker` is the only way to get proper synchronization with the Nvidia driver?

Drakulix:
> So that does sound like something that could be a viable optimization for things like fullscreen applications

Right. For fullscreen surfaces, we probably want to just pass through the fence, not for normal desktop usage though. I agree this is a later optimization.

> Ah right, we already have support for that in the renderer. So if we want to use the acquire_fence without a CPU wait, we'll want to pass the fence to EGL there. But that can be handled later.

Yes, but we kinda don't want that (except for the fullscreen case again), because then we would block compositing for an undefined amount of time. The goal is to render as late as possible and use the latest ready buffer.

> We could just not pass an IN_FENCE_FD, and ignore the acquire fence, but I don't think ignoring the explicit fence is guaranteed to work (with all drivers)?

Yes we can absolutely do that, that is what we are doing right now. We would only block drivers not supporting implicit sync (nvidia). Anyone serious about nvidia support, just needs to implement the protocol and I don't want to mandate that.

DemiMarie:
My understanding is that DRM syncobjs are, at least in the future, not guaranteed to ever signal.  This is because future drivers may support Userspace Memory Fences (UMFs) and userspace can deadlock themselves with them.  Therefore, I think Smithay needs to do an (asynchronous) CPU-side wait (possibly with a timeout) so that the display cannot be locked up forever.

ids1024:
I've rebased this (and the cosmic-comp PR) and made a few small improvements to the doc comments.

I believe this reflects how we agreed the implementation/API should work, and as far as I can tell works correctly. Though it's hard to test thoroughly. The latest version of nvidia/egl-wayland should fix the issues that NVIDIA drivers were having on certain clients with compositors that supported explicit sync.

If no one else sees issues with this, it should be good to merge.

cmeissl:
Thanks for working on this! 

I will try to review and also test it on a few different devices during the next few days.

src/backend/renderer/utils/wayland.rs
```diff
@@ -68,12 +82,11 @@ pub struct Buffer {
     inner: Arc<InnerBuffer>,
 }
 
-impl From<WlBuffer> for Buffer {
-    #[inline]
-    fn from(buffer: WlBuffer) -> Self {
-        Buffer {
-            inner: Arc::new(InnerBuffer(buffer)),
-        }
```
cmeissl, line 76: Removing the `From<WlBuffer>` implementation will make it impossible to construct `Buffer` outside of this crate.
As `Buffer` represents a public API used in different modules we have to provide a way constructing it.

cmeissl (changes requested):

anvil/src/shell/mod.rs
```diff
@@ -125,6 +139,21 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
...
+                            data.client_compositor_state(&client).blocker_cleared(data, &dh);
+                            Ok(())
+                        });
+                        if res.is_ok() {
+                            add_blocker(surface, blocker);
+                            return;
+                        }
+                    }
+                }
                 if let Ok((blocker, source)) = dmabuf.generate_blocker(Interest::READ) {
```
Drakulix, line 157: Should we wait on both of them, or is this unnecessary, if we got an acquire point? I was under the assumption the later is the case.

src/backend/renderer/utils/wayland.rs
Drakulix: Yeah, I think it would be nice to have a `Buffer::from_implicit_buffer(WlBuffer)`-method as a replacement for the old `From`-implementation and perhaps a new `from_explicit_buffer`-method that could also take fences of some form.

anvil/src/shell/mod.rs
ids1024: If it adds an acquire point blocker, it has an early `return;`.

Drakulix: skipped over that. seems fine.

src/wayland/compositor/tree.rs
```diff
@@ -150,9 +152,25 @@ impl PrivateSurfaceData {
         if let Some(BufferAssignment::NewBuffer(buffer)) = guard.pending().buffer.take() {
             buffer.release();
         };
+        drop(guard);
+        #[cfg(feature = "backend_drm")]
+        let mut guard = my_data.public_data.cached_state.get::<DrmSyncobjCachedState>();
```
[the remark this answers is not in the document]
ids1024: Good point. The syncobj protocol does need to be handled in `RendererSurfaceState::update_buffer`, which can't obviously be done with any kind of hook, but this part could probably use `add_destruction_hook`. (After all, this code is right before the hooks are run.)

src/backend/renderer/utils/wayland.rs
ids1024: Ah, I didn't think about possible uses of this outside of the crate. It would be easy enough to have some kind of public constructor(s) if that's needed.

cmeissl: Yeah, iirc at least `catacomb` relies on this as they are doing their own buffer management, but still use the rendering abstractions and `DrmCompositor`.

PolyMeilex:
Tested a few clients that make use of syncobj on my AMD systems, everything works as expected. :ok_hand:

cmeissl (approved):
Okay, so while I would like to see a more generic solution allowing control over passing fences directly to `IN_FENCE_FD` I agree that we can do that later. For overlay planes, especially when having multiple, no one will probably want to just pass the fences. So the only real use-case would be fullscreen direct scan-out. As most compositors outside of kiosk use-cases will have to handle blockers anyway there is no real point for special casing fullscreen anyway.

src/wayland/drm_syncobj/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+        _state: &mut D,
+        _client: &Client,
+        _resource: &WpLinuxDrmSyncobjSurfaceV1,
+        request: wp_linux_drm_syncobj_surface_v1::Request,
+        data: &DrmSyncobjSurfaceData,
+        _dh: &DisplayHandle,
+        _data_init: &mut DataInit<'_, D>,
+    ) {
+        let Ok(surface) = data.surface.upgrade() else {
+            return;
```
Drakulix, line 326: Isn't this potentially a `NoSurface` error?

Drakulix (approved):

src/wayland/drm_syncobj/mod.rs
ids1024: Ah true, it does look like the protocol spec states this should return a `no_surface` for `set_acquire_point` or `set_release_point`.

Drakulix (approved):

PR #1358, opened 2024-03-10
Fix the X11 window instance getter returning the class instead of the instance

roscale:
I guess this is a typical copy-paste bug as someone forgot to change the body of this function to return the right thing.


Drakulix (approved):
Nice catch! Thanks. :)

PR #1359, opened 2024-03-11
desktop: clamp the window geometry to the bbox

### LINKED ISSUE #1357: Possibly wrong window geometry size with wp-viewporter
I tried adding wp-viewporter to niri and found ~~two~~ one issue so far.

If you run `weston-scaler`, then its `window.geometry().size` will not match its destination rectangle. You can tell by the border + background size around the window.

Screenshot from 2024-03-10 18-04-35

I'm not sure how to tell what Weston and sway do exactly, but `weston-scaler -h` says "window size 220x308 px" which to me implies `window.geometry().size`.

image

The Anvil "window preview" seems to work fine and that uses `window.bbox()` rather than geometry from what I can tell?

The second issue was originally that Chromium fails to accept pointer input beyond its initial configure window size if resized bigger than that, but it seems that giving it a (0, 0) initial configure makes that go away. Non-zero initial configure claims another victim it seems... Although, I'll have to test this more; I think I saw the bug trigger even with a (0, 0) initial configure once.

It appears that without viewporter, or with viewporter and initial configure = 0, Chromium/Electron sets a NULL input region, but with viewporter and nonzero initial configure, it sets a specific input region, and then never updates it?

cmeissl:
The spec defines the effective geometry as the set geometry clamped to the bounding rectangle of the surface and its subsurfaces.

fixes #1357


YaLTeR:
Seems to fix both the weston-scaler issue and the Chromium input issue. Xwayland and SDL fullscreen still work fine. Also somehow this fixed Chromium nested popup positioning?

Drakulix (approved):

PR #1364, opened 2024-03-13
trigger `BufferHandler::buffer_destroyed` on client death

colinmarc:
Currently, the callback is called when the client makes a destroy request, but not when the client disconnects and everything is destroyed implicitly.


Drakulix (approved):
Thanks!

PR #1365, opened 2024-03-17
input: Add "ISO level 3 shift" modifier

Trundle:
I guess its absence was an oversight / nobody needed it so far and not a deliberate choice.


Drakulix (approved):
Thanks!

PR #1366, opened 2024-03-17
Fix typo

T0mstone:
Sorry for the comically small contribution, but this one confused me for a bit, so I think the correction is warranted.


Drakulix (approved):

Drakulix:
Thanks!

PR #1370, opened 2024-03-21
anvil: Drop `CalloopData`

Drakulix:
@colinmarc There you go.

Ref https://github.com/Smithay/smithay/pull/1351#discussion_r1534530909


PolyMeilex (approved):
Cool, that's a lot nicer to use

PR #1371, opened 2024-03-23
{an,small}vil: Fix winit scroll values

YaLTeR:
The usual value for one wheel tick is 15, not 3. This can be verified with Firefox by scrolling on a TTY backend vs. on the winit backend (since this particular fallback is used on the winit backend).


Drakulix (approved):

Drakulix:
Could you rebase now that https://github.com/Smithay/smithay/pull/1373 is merged for ci?

PR #1373, opened 2024-03-24
Fixup clippy lints

PolyMeilex:


src/wayland/shm/pool.rs
```diff
@@ -1,4 +1,4 @@
-#![forbid(unsafe_op_in_unsafe_fn)]
```
PolyMeilex, line 1: forbid was causing a conflict somewhere inside of thread_local macro, so I changed it to deny

src/wayland/shm/pool.rs
```diff
@@ -44,7 +44,7 @@ static DROP_THIS: Lazy<SyncSender<InnerPool>> = Lazy::new(|| {
     tx
 });
 
-thread_local!(static SIGBUS_GUARD: Cell<(*const MemMap, bool)> = Cell::new((ptr::null_mut(), false)));
+thread_local!(static SIGBUS_GUARD: Cell<(*const MemMap, bool)> = const { Cell::new((ptr::null_mut(), false)) });
```
i509VCB, line 47: This changes the MSRV?

PolyMeilex: Not 100% sure, but I don't think so: this is 3 years old https://github.com/dtolnay/rust/commit/c6eea222a9cafeeeedfa1767232328710b350447

Drakulix (approved):

anvil/src/input_handler.rs
```diff
@@ -1306,6 +1306,7 @@ impl AnvilState<UdevData> {
 }
 
 /// Possible results of a keyboard action
+#[allow(dead_code)] // some of these are only read if udev is enabled
```
chrisduerr, line 1309: Are they always written to? Why not feature gate the variants behind the backend?

Drakulix: Yes the corresponding key binding will still generate the KeyAction. Of course we could feature-gate all of that as well, but I am not sure that is cleaner than this short line.

PolyMeilex:
Your commits look fine to me, so I'm just going to merge this.

PR #1375, opened 2024-03-28
pointer: Reset image on focus change

Drakulix:
Smithay doesn't reset the cursor image, when we replace focus. This can be the default thing that happens with e.g. swaybg or cosmic-bg running, causing the last cursor image to be stuck, if any newly entered surface doesn't care to set a new image.

Lets reset on swapping focus as well, not just when nothing is focused.

### DIFF
--- a/src/input/pointer/mod.rs
+++ b/src/input/pointer/mod.rs
@@ -150,6 +150,7 @@ where
         event: &MotionEvent,
     ) {
         PointerTarget::<D>::leave(&replaced, seat, data, event.serial, event.time);
+        data.cursor_image(seat, CursorImageStatus::default_named());
         PointerTarget::<D>::enter(self, seat, data, event);
     }
 }


src/input/pointer/mod.rs
```diff
@@ -742,6 +742,7 @@ impl<D: SeatHandler + 'static> PointerInternal<D> {
                     focus.motion(seat, data, &event);
                 }
                 (focus, Some((old_focus, _))) => {
+                    data.cursor_image(seat, CursorImageStatus::default_named());
```
cmeissl, line 745: Maybe it would be better to place that inside the default impl of `PointerTarget::replace`? Between the leave and enter of the new target.

cmeissl (commented):
Looks like I may have misinterpreted the spec here which defines

> When a seat's focus enters a surface, the pointer image is undefined and a client should respond to this event by setting an appropriate pointer image with the set_cursor request.

Drakulix:
> Looks like I may have misinterpreted the spec here which defines
> 
> > When a seat's focus enters a surface, the pointer image is undefined and a client should respond to this event by setting an appropriate pointer image with the set_cursor request.

Right, so the application can't expect any specific state. But it seems like it isn't required for the application to set a new pointer image (just highly suggested), so smithay probably wants to define something for downstream in that case, that isn't "keep the last icon".

I guess technically this could also be a callback, but... what  for?

src/input/pointer/mod.rs
Drakulix: I guess so. I wanted to keep it local to the other part resetting the image (in the `(focus, None)` case), since I don't really see this as the responsibility of `focus`.

cmeissl: imo it makes sense to keep it local to `leave`. by moving it to the default impl downstream could also override this behavior (though I am not sure this will ever be needed).

Drakulix: Makes sense. Updated

cmeissl (approved):

PR #1377, opened 2024-03-31
Add a handler for when an X11 window property changes

roscale:
On Wayland, we can detect when toplevel properties like the app id and title are updated. We have functions called `app_id_changed` and `title_changed` defined in `XdgShellHandler`. `XwmHandler`, however, does not have any function like that. The code only updates the surface properties with new values without notifying the user.

This PR adds a function `property_notify` with an empty default implementation in the trait `XwmHandler`. One of its arguments specify which property has been updated.

Use case:

We have an external renderer (Flutter) running in another thread from Smithay, and I would like to react to protocol state changes and send relevant information to the renderer.

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -291,7 +291,15 @@ pub trait XwmHandler {
         geometry: Rectangle<i32, Logical>,
         above: Option<X11Window>,
     );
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
LGTM! Thanks :)

PR #1381, opened 2024-04-05
gles: keep blending enabled for shadow buffer blit

### LINKED ISSUE #1378: Cursor and overlay elements overwrite primary plane in `RenderFrameResult::blit_frame_result`
Background: I'm implementing wlr-screencopy by blitting the frame result to a `GlesRenderbuffer` and giving the bytes to the shm buffer.

When doing the blit, any element on the cursor or overlay planes completely overwrites pixels underneath it, regardless of its transparency. Below is a sample screencopy (screenshotted from feh to show transparency):
image

I'd expect the background behind the Alacritty on the overlay plane to still be there, but it's been overwritten to just the terminal's pixels. Similarly there's a transparent hole around the cursor for the same reason.

Disabling the overlay and cursor planes fixes the problem. Digging into the method, I think it's because elements on those planes are being `RenderElement::draw`n into the target buffer after blitting the primary plane in. Is that supposed to overwrite pixels already in the buffer even if what's getting drawn in has transparency?

cmeissl:
when rendering in multiple passes we potentially need to blend the shadow buffer with the current content, so lets keep blending active until we are finished.

fixes #1378

### DIFF
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -2533,11 +2533,6 @@ impl<'frame> GlesFrame<'frame> {
             return Ok(SyncPoint::signaled());
         }
 
-        unsafe {
-            self.renderer.gl.Disable(ffi::SCISSOR_TEST);
-            self.renderer.gl.Disable(ffi::BLEND);
-        }
-
         if let Some(target) = self.renderer.target.as_ref() {
             if let Some(shadow) = target.get_shadow() {
                 target.make_current_no_shadow(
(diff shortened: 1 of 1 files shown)


Drakulix:
Doesn't this mess with damage tracking for transparent surfaces? The idea of blitting the shadow buffer into the real framebuffer was to override all parts we touched (which is what we use the stencil buffer for).

Obviously this is only a problem, when we rely on previous contents of the buffer and both can be transparent, so overlay/cursor planes. I feel like this might blend with the previous framebuffer contents, instead of replacing them, when necessary.

PR #1383, opened 2024-04-06
Fix outgoing xwm incremental selection

### LINKED ISSUE #1380: Pasting images from wayland -> x11 truncates/corrupts the image
Reproducible in Anvil, cosmic-comp, and Pinnacle across my desktop and laptop (though they all use the same code for Wayland -> x11 selections so I guess that's expected).

Pasting to other Wayland clients works fine, but pasting the exact same image contents into an xwayland client results in a truncated or corrupted image, on my machines at least. The image gets pasted at the right dimensions, but they seem to be corrupted, filled with fully transparent/NUL bytes after the first few rows of pixels, truncated, or some combination of those.

The following is a PNG screenshot of a screencopy that was pasted into Discord running under X11 as it appears in element-desktop. It's filled with what seems to be transparency. Pasting into GIMP will fill the same parts with hex #979899, but I think that's a GIMP thing.

image

Another user has the same problem as reported in pinnacle-comp/pinnacle#192, also PNGs.

They also gave me a screencopy that appears to have a chunk of corrupted bytes after the normal looking chunk, but I haven't been able to reproduce that:

<details>

<summary><b>Corrupted screencopy</b></summary>

image

</details>

Any copied image has the same problem, not just screencopies, like ones copied off of Google images.

## JPEGs

Forcing the image to paste as a JPEG using `wl-paste --type image/jpeg | wl-copy` results in weird tiling artifacts:

<details>

<summary><b>Original cat JPEG</b></summary>

image

</details>

<details>

<summary><b>Messed up cat JPEG pasted into X11 GIMP (this is a screenshot)</b></summary>

image

</details>

Ottatop:
Fixes two internal issues:
1. There's a `self.source_data = Vec::new()` that clears the internal selection buffer that probably isn't supposed to be there, and
2. Fixing the above revealed that sometimes the final 0-byte data chunk to complete the transfer wasn't always getting sent. I'm still not sure of the code path that causes this to happen.

This was the simplest way I could fix the problem. If there's a better way let me know!

Fixes #1380

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -472,6 +472,8 @@ struct OutgoingTransfer {
 
     property_set: bool,
     flush_property_on_delete: bool,
+    /// The final 0-byte data chunk has been sent, denoting the completion of this transfer
+    sent_finished: bool,
 }
 
 impl fmt::Debug for OutgoingTransfer {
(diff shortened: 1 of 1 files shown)


Ottatop:
Actually this is preventing non-incremental transfers from being dropped properly, need to fix that

Drakulix (approved):
Thanks for figuring this out! Looks like a perfectly fine workaround to me.

PR #1386, opened 2024-04-12
Add `unset` method to input grab traits

ids1024:
`PointerGrab::unset` is called when a grab is removed or replaced with another grab.

This fixes https://github.com/pop-os/cosmic-comp/issues/403. I've seen issues with one or two of cosmic-comp's grabs if something isn't cleaned up correctly when `unset` is called, so this should help with them, as well as some other Smithay grabs.

I still need to add this for keyboard and touch grabs, but it seems workable.


Drakulix:
Seems like a good idea to me!

Drakulix (approved):

PR #1387, opened 2024-04-12
Make a few rendering things public

YaLTeR:
Useful/required for implementing custom shaders. The safety comments are a bit non-specific but I'm not sure what the specific conditions are.


Drakulix (approved):
LGTM

PR #1388, opened 2024-04-14
Ei protocol support using `reis`

ids1024:
The Ei protocol is needed for the xdg-desktop-portal `RemoteDesktop` portal to emulate input devices, as well as by the `InputCapture` portal for Synergy-like uses (input-leap supports Wayland with this portal).

This is quite incomplete, but the `type-text` example in `reis` now works in Anvil. Reis could still use somewhat better higher-level servere-side APIs.

It might make sense if ei exposed seats and devices matching those in Smithay (or would compositors not always want to do that?). Not sure how best to handle that.

Receiver contexts for the `InputCapture` portal are also a bit more complicated to implement. Those involve capturing input once the cursor crosses outside the display. That isn't implemented at all here yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,7 @@ aliasable = { version = "0.1.3", optional = true }
 atomic_float = "1.1.0"
 sha2 = "0.10.9"
 tracy-client = { version = "0.18.4", default-features = false, optional = true }
+reis = { version = "0.6.0", features = ["calloop"] }
 
 [dev-dependencies]
 clap = { version = "4", features = ["derive"] }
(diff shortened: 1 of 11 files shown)


axelkar:
Oops, I've been doing the same thing in Niri. I ruled out upstreaming to Smithay because I thought the implementation would have more Niri and `org.gnome.Mutter.RemoteDesktop`-specific features.

Some comments on your code: (`smithay::backend::input` shortened to `sbi`.. please name that module more uniquely next time)

I wrote an adapter struct between `EisRequest` and `sbi::Event` because I thought I had to avoid Rust's orphan rule (turns out just having the input backend defined by the same crate makes it work).

Doesn't Smithay specify that IDs must be unique? I currently pass a `session_id` (`org.gnome.Mutter.RemoteDesktop` session ID) through the adapter to `sbi::Event::device` and `sbi::Device::id`.

I can also provide a true `sbi::KeyboardKeyEvent::count`, and I combine axis requests into a single frame, similar to the wlr-virtual-pointer implementation in Niri.

ids1024:
> please name that module more uniquely next time

Yeah, the same module or type name being used for different things in different parts of `smithay` can be a bit confusing sometimes.

> Doesn't Smithay specify that IDs must be unique? I currently pass a session_id (org.gnome.Mutter.RemoteDesktop session ID) through the adapter to sbi::Event::device and sbi::Device::id.

https://docs.rs/smithay/0.7.0/smithay/backend/input/trait.Device.html says "Unique id of a single device at a point in time." and "Note: This means ids may be re-used by the backend for later devices.".

I suppose with ei, the protocol object id would be unique in that manner.

ids1024:
Probably the biggest question here is exactly how much of it Smithay should handle/abstract. But I think we probably should just clean up any issues with the event conversion, and merge a minimal version of this. Then see what further abstractions are useful to add in either `smithay` or `reis` later.

ids1024:
I think I've come up with a reasonable API for input emulation using libei. (Input capture for something like the Input Capture portal may have different requirements). Mirroring the Smithay seat API a bit, to provide an API that should be flexible enough without requiring the compositor to do too much.

I've also moved some things into `reis` so they don't have to be dealt with here.

ids1024:
I guess this will also need to send `DeviceAdded` and `DeviceRemoved` events.

For a unique device `.id()`, it could combine the fd and protocol id. I guess it just needs to not race a device removal with a device add on a different socket that connects with the same fd...

Edit: I guess fds racing isn't an issue, because as long as we're processing the event source, the fd isn't closed.

ids1024:
Merged in https://github.com/Smithay/smithay/pull/2051. Any future improvements can now be done in separate PRs.

PR #1390, opened 2024-04-15
add #[must_use] to AxisFrame

m4rch3n1ng:
it's very easy to accidentally forget the `frame =` reassignment when working with an `AxisFrame`, especially because it doesn't actually "consume" it, since it is `Copy`.

```rs
let mut frame = AxisFrame::new(evt.time_msec()).source(evt.source());
if horizontal_amount != 0.0 {
    // this line does nothing but it doesn't warn me
    frame.relative_direction(Axis::Horizontal, evt.relative_direction(Axis::Horizontal));
    frame = frame.value(Axis::Horizontal, horizontal_amount);
}
```

### DIFF
--- a/src/input/pointer/mod.rs
+++ b/src/input/pointer/mod.rs
@@ -970,7 +970,7 @@ pub struct ButtonEvent {
 ///     .value(Axis::Vertical, 30, time)
 ///     .stop(Axis::Vertical);
 /// ```
-#[must_use]
+#[must_use = "AxisFrame uses a builder-like pattern, so its result must be used"]
 #[derive(Copy, Clone, Debug)]
 pub struct AxisFrame {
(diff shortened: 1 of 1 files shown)


ids1024:
I think it would also work to add a `#[must_use]` annotation to the `AxisFrame` type itself. Then any function returning `AxisFrame` will raise this warning.

PolyMeilex (approved):
Message would be welcome, but I don't have any good ideas, perhaps  `AxisFrame uses builder-like pattern, therefore not using it's result is a noop`?

Either way LGTM

m4rch3n1ng:
added a message and rebased to fix the clippy warnings

Drakulix (approved):
Thanks!

PR #1391, opened 2024-04-15
More fixes around `set_grab`/`unset_grab`

ids1024:
This is a followup to https://github.com/Smithay/smithay/pull/1386, which I noticed an issue with: cancelled drags in clients like Firefox and Nautilus caused the window to act like it didn't have pointer focus until it lost and regained it. It seems the the previous change in `KeyboardGrab` also wasn't quite right.

See commit messages for more specific details.

This is a breaking API change requiring passing additional arguments, but although a bit annoying to update for, it's pretty straightforward.


Drakulix (approved):
Seems sensible to me.

PR #1395, opened 2024-04-18
elements/surface: Snapshot render state

Drakulix:
Attempt at solving parts of the discussion in https://github.com/Smithay/smithay/pull/1393 and observed profiling hot-paths, by copying from the `RendererSurfaceState` once on render element creation instead of accessing the surface's user-data all the time for the `WaylandSurfaceRenderElement`.

### DIFF
--- a/src/backend/renderer/element/surface.rs
+++ b/src/backend/renderer/element/surface.rs
@@ -218,17 +218,17 @@
 //! }
 //! ```
 
-use std::{fmt, marker::PhantomData};
+use std::fmt;
 
 use tracing::{instrument, warn};
 use wayland_server::protocol::wl_surface;
 
 use crate::{
     backend::renderer::{
-        utils::{DamageSet, RendererSurfaceStateUserData},
+        utils::{Buffer, DamageSet, DamageSnapshot, RendererSurfaceStateUserData, SurfaceView},
         Frame, ImportAll, Renderer, Texture,
     },
-    utils::{Buffer, Physical, Point, Rectangle, Scale, Size, Transform},
+    utils::{Buffer as BufferCoords, Logical, Physical, Point, Rectangle, Scale, Size, Transform},
     wayland::compositor::{self, SurfaceData, TraversalAction},
 };
 
@@ -247,7 +247,7 @@ pub fn render_elements_from_surface_tree<R, E>(
 ) -> Vec<E>
 where
     R: Renderer + ImportAll,
-    <R as Renderer>::TextureId: 'static,
+    <R as Renderer>::TextureId: Clone + 'static,
     E: From<WaylandSurfaceRenderElement<R>>,
 {
     let location = location.into().to_f64();
@@ -290,7 +290,8 @@ where
                     match WaylandSurfaceRenderElement::from_surface(
                         renderer, surface, states, location, alpha, kind,
                     ) {
-                        Ok(surface) => surfaces.push(surface.into()),
+                        Ok(Some(surface)) => surfaces.push(surface.into()),
+                        Ok(None) => {} // surface is not mapped
                         Err(err) => {
                             warn!("Failed to import surface: {}", err);
                         }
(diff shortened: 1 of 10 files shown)


src/backend/renderer/element/surface.rs
```diff
@@ -305,21 +307,29 @@ where
 }
 
 /// A single surface render element
-pub struct WaylandSurfaceRenderElement<R> {
+pub struct WaylandSurfaceRenderElement<R: Renderer> {
     id: Id,
     location: Point<f64, Physical>,
     alpha: f32,
-    surface: wl_surface::WlSurface,
     renderer_type: PhantomData<R>,
```
cmeissl, line 314: Is this still needed now that we use `R` for `texture`?

src/backend/renderer/element/surface.rs
```diff
@@ -305,21 +307,29 @@ where
...
     kind: Kind,
+
+    view: Option<SurfaceView>,
+    current_commit: CommitCounter,
+    buffer: Option<Buffer>,
+    buffer_scale: i32,
+    buffer_transform: Transform,
+    buffer_dimensions: Option<Size<i32, BufferCoords>>,
+    damage: DamageBag<i32, BufferCoords>,
```
cmeissl, line 323: Instead of `DamageBag` this could use `DamageSnapshot` which provides a read-only view of the damage in the `DamageBag` and was introduced to support exactly such use-cases. The snapshot can be obtained from `DamageBag::snapshot()`.

src/backend/renderer/element/surface.rs
```diff
@@ -305,21 +307,29 @@ where
...
+pub struct WaylandSurfaceRenderElement<R: Renderer> {
     id: Id,
     location: Point<f64, Physical>,
     alpha: f32,
-    surface: wl_surface::WlSurface,
     renderer_type: PhantomData<R>,
     kind: Kind,
+
+    view: Option<SurfaceView>,
+    current_commit: CommitCounter,
```
cmeissl, line 318: We do not have to store the `CommitCounter` separately from the `damage`, `DamageSnapshot::current_commit` can directly be used in `Element::current_commit`.

src/backend/renderer/utils/mod.rs
```diff
@@ -75,6 +75,15 @@ pub struct DamageSnapshot<N, Kind> {
...
 }
 
+impl<N, Kind> Clone for DamageBag<N, Kind> {
+    fn clone(&self) -> Self {
+        Self {
+            limit: self.limit,
+            state: self.state.clone(),
+        }
+    }
+}
```
cmeissl, line 85: I intentionally did not implement `Clone` on `DamageBag`. `DamageBag::snapshot` should instead be used to get a read-only view of the current damage.

src/backend/renderer/element/surface.rs
```diff
@@ -336,31 +346,54 @@ impl<R: Renderer + ImportAll> WaylandSurfaceRenderElement<R> {
...
             renderer_type: PhantomData,
             kind,
+            view: data.view(),
+            current_commit: data.current_commit(),
+            buffer: data.buffer().cloned(),
+            buffer_scale: data.buffer_scale,
+            buffer_transform: data.buffer_transform,
+            buffer_dimensions: data.buffer_dimensions,
+            damage: data.damage.clone(),
+            opaque_regions: data.opaque_regions.clone(),
```
cmeissl, line 378: This is more a note to myself, we should probably also provide a COW structure and maybe apply the same optimizations like we did with `DamageSet` for opaque regions.

cmeissl:
Thanks for working on this! :smile:

src/backend/renderer/element/surface.rs
cmeissl: moved to https://github.com/Smithay/smithay/issues/1396

src/backend/renderer/element/surface.rs
```diff
@@ -336,31 +345,52 @@ impl<R: Renderer + ImportAll> WaylandSurfaceRenderElement<R> {
...
+        let data_ref = states
+            .data_map
+            .get::<RendererSurfaceStateUserData>()
+            .map(|data| data.borrow());
+        let mut fallback_data = None;
+
+        let data = data_ref.as_deref().unwrap_or_else(|| {
+            fallback_data = Some(RendererSurfaceState::default());
+            fallback_data.as_ref().unwrap()
+        });
```
cmeissl, line 362: I just realized that we could get rid of this here and also most `Option` handling if we just return `Result<Option<Self>, <R as Renderer>::Error>` in case the surface is not mapped. `render_elements_from_surface_tree` already skips all surfaces that have no `view` anyway. This could result in cleaner code in the `Element` implemetation.

cmeissl (approved):
Sorry for not realizing this earlier, feel free to ignore that comment as it is just a small optimization. I can refactor that in a later step if needed/desired.

Otherwise LGTM, thanks!

src/backend/renderer/element/surface.rs
Drakulix: Not a bad idea. I'll try that asap

cmeissl: thanks!

PR #1397, opened 2024-04-19
Notify when InputMethod::PopupSurface changes location

kchibisov:
Each call to `set_text_input_rectangle` changes the position of the popup and also will result in popup repositioning, thus notify users when such request occurs.

The location and text cursor area where separated, since one is used for the rendering, and the other one is original location.

--

Use case: https://github.com/YaLTeR/niri/pull/295

### DIFF
--- a/src/wayland/input_method/input_method_popup_surface.rs
+++ b/src/wayland/input_method/input_method_popup_surface.rs
@@ -25,10 +25,10 @@ pub struct PopupSurface {
     /// The surface role for the input method popup
     pub surface_role: ZwpInputPopupSurfaceV2,
     surface: WlSurface,
-    // NOTE the popup position could change at any time, so the popup manager should get updates,
-    // thus use shared storage for the `cloned` data to automatically apply .
-    /// Rectangle with position and size of  text cursor, used for placement of popup surface.
+    /// Protected cursor area.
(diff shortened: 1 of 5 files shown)


src/wayland/input_method/input_method_popup_surface.rs
```diff
@@ -25,10 +25,10 @@ pub struct PopupSurface {
     /// The surface role for the input method popup
     pub surface_role: ZwpInputPopupSurfaceV2,
     surface: WlSurface,
-    // NOTE the popup position could change at any time, so the popup manager should get updates,
-    // thus use shared storage for the `cloned` data to automatically apply .
-    /// Rectangle with position and size of  text cursor, used for placement of popup surface.
+    /// Protected cursor area.
     pub(crate) rectangle: Arc<Mutex<Rectangle<i32, Logical>>>,
+    /// Location of the popup surface.
+    location: Arc<Mutex<Point<i32, Logical>>>,
```
rano-oss, line 31: Any reason this has to be an Arc<Mutex< ? Isn't it only updated in one function?
Otherwise it looks good too me :smile_cat:

kchibisov: because of how it works with popup manager, I don't like it as well, but it's just the way it's.

Basically one can force re-insert of popup into popup manager once the position changes...

Drakulix (approved):
LGTM

PR #1398, opened 2024-04-19
Update `drm`, `gbm`, `image`; `xkbcommon`/`x11rb` deps of Anvil

ids1024:
There's also a new `ash` release, but that seems to require some changes to use. So I left that for now.


ids1024:
Okay, excluded `use_bindgen` from `cargo hack check` and this now builds fine.

Drakulix (approved):

PR #1400, opened 2024-04-21
backend/x11: Include `window_id` with `X11Event::Input` when possible.

nferhat:
Useful for people like me who want to test multi-output setups but don't have access to one. With this you can create multiple surfaces and dispatch input events accordingly based on the window ID.

Example implementation using this: [`fht-compositor/src/backend/x11.rs`](https://github.com/nferhat/fht-shell/blob/355f0df9bda980070a52291d3c99e31fbb7b9df8/fht-compositor/src/backend/x11.rs)

### DIFF
--- a/src/backend/x11/mod.rs
+++ b/src/backend/x11/mod.rs
@@ -138,7 +138,15 @@ pub enum X11Event {
     Focus(bool),
 
     /// An input event occurred.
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Thanks!

PR #1403, opened 2024-04-25
impl Neg for Point

m4rch3n1ng:
i'm not entirely sure if this is the correct way to implement that, but i have ran into this a few times and have always had to resolve to a more clunky `Point::from((0,0)) - offset`.


m4rch3n1ng:
i think "negating a point" makes the same amount of sense as "subtracting a point from (0, 0)", which i think makes enough sense to warrant this.

on another note i actually managed to remove both of the `Point::from((0, 0)) - offset` by just inverting the previous subtraction (i don't know how i didn't think of that earlier lol), so i am actually more skeptical if there is a lot of practical application for this

PR #1404, opened 2024-04-28
Fix `Point::constrain` clamping at wrong values

Ottatop:
This fixes the method clamping down using the size, not with the positions of the right/bottom edges.


Drakulix (approved):
Thanks for spotting this!

PR #1413, opened 2024-05-01
wayland.shm: Zero-Initialize private fields of `libc::sigaction`

### LINKED ISSUE #1412: Build error on s390x: src/wayland/shm/pool.rs:311:15  ..mem::zeroed() field `__glibc_reserved0` is private
Context: I'm working on releasing cosmic-comp in Fedora Rawhide (https://src.fedoraproject.org/rpms/cosmic-comp)

The build currently fails on 2 arches: i686 and s390x

i686 doesn't matter, I can choose not to build for that arch, s390x is required though, unfortunately.
https://koji.fedoraproject.org/koji/taskinfo?taskID=117083063

Looks like there's a member in smithay that's private on s390x:

screenshot-2024-05-01-18-31-10

PolyMeilex:
`sigaction` can contain private fields on some targets, so we can't use the `..mem::zeroed()` syntax.

fixes #1412

### DIFF
--- a/src/wayland/shm/pool.rs
+++ b/src/wayland/shm/pool.rs
@@ -305,11 +305,12 @@ unsafe fn nullify_map(ptr: *mut u8, size: usize) -> Result<(), ()> {
 unsafe fn place_sigbus_handler() {
     // create our sigbus handler
     unsafe {
-        let action = libc::sigaction {
-            sa_sigaction: sigbus_handler as _,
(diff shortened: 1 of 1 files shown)


Drakulix (changes requested):
This is a quite subtle change, mind adding a little explainer comment, so nobody accidentally changes this back again?

PolyMeilex:
Done

PR #1416, opened 2024-05-02
Avoid unwraping clients from surfaces

### LINKED ISSUE #1407: `wlr_layer`'s `pre_commit_hook` can post an error, disconnecting the client, leading to a panic later on in the `commit` handling
While running wlcs tests (namely `LayerSurfaceTest.specifying_zero_size_without_anchors_is_an_error`) against [Pinnacle](https://github.com/pinnacle-comp/pinnacle), I ran across a panic in smithay.

The root cause is that the test requests a zero-width window without anchors, causing the `pre_commit_hook` in `wlr_layer` to (correctly) signal a protocol error for the surface, disconnecting the client:
https://github.com/Smithay/smithay/blob/7544c4d4b9ef89e2945bb626dd47518fedc5f660/src/wayland/shell/wlr_layer/handlers.rs#L132-L138

The panic then occurs in `commit` function for `PrivateSurfaceData` when `unwrap`ing the surface's client here:
https://github.com/Smithay/smithay/blob/7544c4d4b9ef89e2945bb626dd47518fedc5f660/src/wayland/compositor/tree.rs#L292

My first intuition for a solution would be to cleanly handling the case, that `client()` returns `None`, cleanly in the `commit` function, but I am not sure what exactly that entails.

For reference, the panic in the wlcs test output looks like this, but that's only marginally helpful: https://gist.github.com/LogicalOverflow/804f83c870da2a7667fb9f6d3d915043#file-pinnacle-wlcs-log-L449

(https://github.com/pinnacle-comp/pinnacle/pull/204 to make this issue show up in there)

LogicalOverflow:
Even when handling request for a surface, `surface.client()` may return `None`. This occurs, for example, if a protocol error is signaled; the surface remains alive for some time, but `client()` returns `None` immediately after `post_error` was called.

I just added early returns in the three places I found in the codebase, where a client for a surface was previously being unwrapped.

I didn't add the check, if a surface `is_alive()` before calling the `commit`, discussed in #1407, as I didn't see value in doing the extra check anymore. Let me know if I should still add the check anyways.

Fixes #1407

### DIFF
--- a/anvil/src/shell/mod.rs
+++ b/anvil/src/shell/mod.rs
@@ -124,14 +124,15 @@ impl<BackendData: Backend> CompositorHandler for AnvilState<BackendData> {
             });
             if let Some(dmabuf) = maybe_dmabuf {
                 if let Ok((blocker, source)) = dmabuf.generate_blocker(Interest::READ) {
-                    let client = surface.client().unwrap();
-                    let res = state.handle.insert_source(source, move |_, _, data| {
-                        let dh = data.display_handle.clone();
-                        data.client_compositor_state(&client).blocker_cleared(data, &dh);
-                        Ok(())
-                    });
-                    if res.is_ok() {
-                        add_blocker(surface, blocker);
+                    if let Some(client) = surface.client() {
+                        let res = state.handle.insert_source(source, move |_, _, data| {
+                            let dh = data.display_handle.clone();
+                            data.client_compositor_state(&client).blocker_cleared(data, &dh);
+                            Ok(())
+                        });
+                        if res.is_ok() {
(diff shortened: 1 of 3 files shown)


LogicalOverflow:
Clippy failure is unrelated, seems to be cause by the rust version having switched from 1.77.2 to 1.78.0

PolyMeilex:
> Clippy failure is unrelated, seems to be cause by the rust version having switched from 1.77.2 to 1.78.0

Fixed, rebase should make it pass.

LogicalOverflow:
> Fixed, rebase should make it pass.

Just rebased the branch

Drakulix (approved):

PR #1423, opened 2024-05-07
Add `shape`/`orientation` to `TouchHandle` and `TouchGrab`

ids1024:
The event structs were already defined and used in `TouchTarget`, but there was no way to actually send them.

I would also add backend support for getting these events from libinput, but apparently that still doesn't exist:
https://gitlab.freedesktop.org/libinput/libinput/-/issues/746.


cmeissl (approved):
thanks :+1:

PR #1425, opened 2024-05-08
wayland: allow to send touch frame for touch up

cmeissl:
sending wl_touch.frame for a wl_touch.up is valid and we should not prevent it. wl_touch.cancel explicitly states that no frame should be sent.

cc @chrisduerr


chrisduerr (approved):
Can confirm this fixes my issues.

PR #1429, opened 2024-05-11
Another round of performance optimizations

cmeissl:
Okay, this is now ready for review. I am still working on other optimizations, but I won't touch this PR to get some stuff merged eventually.

based on https://github.com/Smithay/smithay/pull/1427

### DIFF
--- a/src/backend/allocator/gbm.rs
+++ b/src/backend/allocator/gbm.rs
@@ -117,6 +117,7 @@ impl<A: AsFd + 'static> Allocator for GbmAllocator<A> {
 }
 
 impl<T> Buffer for GbmBuffer<T> {
+    #[inline]
     fn size(&self) -> Size<i32, BufferCoords> {
         (
             self.width().unwrap_or(0) as i32,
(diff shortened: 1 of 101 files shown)


src/backend/allocator/gbm.rs
```diff
@@ -11,10 +11,108 @@ use super::{
...
 use std::os::unix::io::{AsFd, BorrowedFd};
 use tracing::instrument;
 
+/// A GBM buffer object
+#[derive(Debug)]
+pub struct GbmBuffer {
+    bo: BufferObject<()>,
+    size: Size<i32, BufferCoords>,
+    format: Format,
+}
```
Drakulix, line 27: This is a great commit, because I just ran into an issue with virgl, that wlroots works around by overriding the modifier: https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/bf86110fc5615bfd1af46e95cf81ac720ecac307

This is a great place to do the same, if we end up calling `gbm_bo_create`.

Drakulix (approved):
Looks all good! Thanks :)

PapyElGringo:
This is great, good work !

Shouldn't we apply the #581c750 improvement to xwayland for X11Surface too?

PR #1430, opened 2024-05-15
dmabuf v6

Drakulix:
Implementation of https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/268

Superseeds https://github.com/Smithay/smithay/pull/1273

### DIFF
--- a/src/wayland/dmabuf/dispatch.rs
+++ b/src/wayland/dmabuf/dispatch.rs
@@ -1,5 +1,6 @@
 use std::sync::{Arc, Mutex, atomic::AtomicBool};
 
+use libc::dev_t;
 use wayland_protocols::wp::linux_dmabuf::zv1::server::{
     zwp_linux_buffer_params_v1, zwp_linux_dmabuf_feedback_v1, zwp_linux_dmabuf_v1,
 };
@@ -75,6 +76,7 @@ where
                         formats: self.formats.clone(),
                         modifier: Mutex::new(None),
                         planes: Mutex::new(Vec::with_capacity(MAX_PLANES)),
+                        node: Mutex::new(None),
                     },
                 );
             }
@@ -288,6 +290,17 @@ where
                 }
             }
 
+            zwp_linux_buffer_params_v1::Request::SetSamplingDevice { device } => {
+                if let Ok(array) = device.try_into() {
+                    *self.node.lock().unwrap() = Some(dev_t::from_ne_bytes(array));
+                } else {
+                    params.post_error(
+                        zwp_linux_buffer_params_v1::Error::InvalidDevTSize,
+                        "Could not parse the dev_t provided via SetSamplingDevice",
+                    );
+                }
+            }
+
             zwp_linux_buffer_params_v1::Request::Create {
                 width,
                 height,
@@ -295,7 +308,9 @@ where
                 flags,
             } => {
                 // create_dmabuf performs an implicit ensure_unused function call.
-                if let Some(dmabuf) = self.create_dmabuf(params, width, height, format, flags, None) {
+                if let Some(dmabuf) =
+                    self.create_dmabuf(params, width, height, format, flags, *self.node.lock().unwrap())
+                {
                     if state.dmabuf_state().globals.contains_key(&self.id) {
                         let notifier = ImportNotifier::new(
                             params.clone(),
@@ -320,7 +335,9 @@ where
             } => {
                 // Client is killed if the if statement is not taken.
                 // create_dmabuf performs an implicit ensure_unused function call.
-                if let Some(dmabuf) = self.create_dmabuf(params, width, height, format, flags, None) {
+                if let Some(dmabuf) =
+                    self.create_dmabuf(params, width, height, format, flags, *self.node.lock().unwrap())
+                {
                     if state.dmabuf_state().globals.contains_key(&self.id) {
                         // The buffer isn't technically valid during data_init, but the client is not allowed to use the buffer until ready.
                         let buffer = data_init.init(buffer_id, dmabuf.clone());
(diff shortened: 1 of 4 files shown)


gabriele2000:
Any update on this?

Drakulix:
Needs https://github.com/Smithay/wayland-rs/pull/928. Ping @ids1024

ids1024:
I've released the `wayland-protocols` crate, since that doesn't have breaking API changes (while wayland-rs `master` does in other crates). It will need a bump to `rust-version = "1.86"`.

Drakulix:
Thanks @ids1024. Updated wayland-protocols and remove the draft-status. @cmeissl any objection to the MSRV bump?

cmeissl:
> Thanks @ids1024. Updated wayland-protocols and remove the draft-status. @cmeissl any objection to the MSRV bump?

nope, everything up to 1.92 is fine for me :)

src/wayland/dmabuf/mod.rs
```diff
@@ -445,15 +444,21 @@ impl PartialEq for DmabufFeedback {
...
         );
 
         for tranche in self.0.tranches.iter() {
             feedback.tranche_target_device(tranche.target_device.to_ne_bytes().to_vec());
-            feedback.tranche_flags(tranche.flags);
+            if feedback.version() <= 5 && tranche.flags.contains(TrancheFlags::Scanout) {
+                feedback.tranche_flags(TrancheFlags::Scanout);
+            } else {
+                feedback.tranche_flags(tranche.flags);
+            }
```
cmeissl, line 473: Not sure I get this right, but to me it looks like this would send the new sampling flag to old clients?

Drakulix: Good catch, I must have flatten the if at some point incorrectly. No client seemed to complain though. :sweat_smile: 

Changed the approach to something that is a bit more future-proof.

cmeissl: XD I guess most clients use mesa and that probably just discards unknown flags (iirc it just tests for the scan-out flag during buffer allocation)

src/wayland/dmabuf/dispatch.rs
```diff
@@ -288,14 +289,20 @@ where
                 }
             }
 
+            zwp_linux_buffer_params_v1::Request::SetSamplingDevice { device } => {
+                *self.node.lock().unwrap() = device[0..8].try_into().ok().map(u64::from_ne_bytes);
```
cmeissl: This should emit `invalid_dev_t_size` in the error case instead of silently dropping the device.

ids1024: This should also handle architectures/OSes where `dev_t` isn't 64-bit.

Something like

```rust
if let Ok(array) = device.try_into() {
    *self.node.lock().unwrap() = Some(dev_t::from_ne_bytes(array));
} else {
    // Send protocol error
}
```

ids1024:
Is there a branch with the cosmic-comp changes using this?

I'm wondering how these changes, and the compositor changes using it, affect clients still using v5. This ends up advertising the same tranches to v5 and v6 clients, other than omitting the `Sampling` flag? Could that cause any issues? Wasn't there some reason we're not already advertising a tranche per device?

Drakulix:
> Is there a branch with the cosmic-comp changes using this?

There is now: https://github.com/pop-os/cosmic-comp/pull/2513

> I'm wondering how these changes, and the compositor changes using it, affect clients still using v5. This ends up advertising the same tranches to v5 and v6 clients, other than omitting the `Sampling` flag? Could that cause any issues? Wasn't there some reason we're not already advertising a tranche per device?

Yeah it might. As far as I remember e.g. mesa didn't even check the device on scan-out tranches, as it would only try to match new modifiers.

I don't think there is a big reason to skip these for the default-feedback, but it wouldn't hurt, while for the surface feedback we definitely want more control.

It might be enough to simply skip tranches containing unsupported flags, but then you couldn't e.g. advertise a branch without any flags on v5, but replace it with a different configuration on v6. I don't think there is a use-case for this now, but perhaps we should think about an api in the future to allow specifying feedback per global version? Or is that overkill? Any opinion are welcome.

ids1024:
> I don't think there is a use-case for this now, but perhaps we should think about an api in the future to allow specifying feedback per global version? Or is that overkill? Any opinion are welcome.

We could definitely have a way to add tranches with version ranges. But yeah, I guess all we really need is whatever seems to make the existing Mesa and Nvidia implementations behave themselves. (AMDVK is discontinued now. I guess if there are any proprietary drivers on ARM hardware, they'll generally be single-GPU systems.)

Drakulix:
> > I don't think there is a use-case for this now, but perhaps we should think about an api in the future to allow specifying feedback per global version? Or is that overkill? Any opinion are welcome.
> 
> We could definitely have a way to add tranches with version ranges. But yeah, I guess all we really need is whatever seems to make the existing Mesa and Nvidia implementations behave themselves. (AMDVK is discontinued now. I guess if there are any proprietary drivers on ARM hardware, they'll generally be single-GPU systems.)

Still needs some testing, but I added a commit, that should allow us to limit specific tranches to specific version ranges.

src/wayland/dmabuf/mod.rs
```diff
@@ -339,15 +342,15 @@ impl DmabufFeedbackBuilder {
     pub fn add_preference_tranche(
         mut self,
         target_device: libc::dev_t,
-        flags: Option<zwp_linux_dmabuf_feedback_v1::TrancheFlags>,
+        flags: zwp_linux_dmabuf_feedback_v1::TrancheFlags,
         formats: impl IntoIterator<Item = Format>,
+        version: impl Into<Range<u32>>,
```
ids1024: This should use `RangeInclusive<u32>`, like `wayland-client` and `sctk` use.

Drakulix: fixed

src/wayland/dmabuf/dispatch.rs
Drakulix: fixed

ids1024 (approved):

PR #1434, opened 2024-05-16
Release overriden buffers (again)

cmeissl:
Revert "Call `wl_buffer::release` only in drop impl of `renderer::utils::Buffer`"

`SurfaceAttributes::merge_into` is called during `WlSurface::commit`.
So if no one (like `on_buffer_commit_handler`) processed an existing buffer in the mean time we have to
send a release or no one will do that for a committed buffer.

This reverts commit f25bcd892e6785ec9ce6b57e55843545722921f6.

### DIFF
--- a/src/backend/renderer/utils/wayland.rs
+++ b/src/backend/renderer/utils/wayland.rs
@@ -60,9 +60,6 @@ impl Drop for InnerBuffer {
 }
 
 /// A wayland buffer
-///
-/// Reference counted, calling `wl_buffer::release` when last reference is
-/// dropped.
 #[derive(Debug, Clone)]
 pub struct Buffer {
     inner: Arc<InnerBuffer>,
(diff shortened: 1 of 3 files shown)


ids1024 (approved):
As I understand it now, it is correct to have `wl_buffer::release` calls here.

Though we should do more testing at some point to make sure the behavior is correct when the same buffer is committed twice in a row.

cmeissl:
> As I understand it now, it is correct to have `wl_buffer::release` calls here.
> 
> Though we should do more testing at some point to make sure the behavior is correct when the same buffer is committed twice in a row.

Totally agree, though this case *should* already be handled. The commit handler and on_commit_buffer_handler won't release a duplicate buffer already. But afaik it would also be allowed to commit the same buffer to different surfaces, which currently would result in receive multiple release events.

Drakulix (approved):

PR #1435, opened 2024-05-17
session_lock: Verify `BufferAssignment` is `NewBuffer` in attached buffer check

Ottatop:
gtklock is getting the `already_constructed` protocol error without an attached buffer because it attaches a null one before calling `get_lock_surface`. This causes the `.is_some()` check for the pending buffer to trigger as it becomes `Some(BufferAssignment::Removed)`. This fix explicitly checks for `Some(BufferAssignment::NewBuffer)` before raising the protocol error.

Ref the gtklock part of YaLTeR/niri#341

### DIFF
--- a/src/wayland/session_lock/lock.rs
+++ b/src/wayland/session_lock/lock.rs
@@ -67,8 +67,14 @@ where
                 // Ensure surface has no existing buffers attached.
                 let has_buffer = compositor::with_states(&surface, |states| {
                     let cached = &states.cached_state;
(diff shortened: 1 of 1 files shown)


YaLTeR (approved):
The protocol seems to say only that the surface must have no buffer attached before the get_lock_surface call, so I guess this makes sense.

Drakulix (approved):

Drakulix:
Thanks!

PR #1436, opened 2024-05-21
xwm: Fix missing map_window_notify callback

Drakulix:
https://github.com/Smithay/smithay/pull/1351 accidentally dropped the `XwmHandler::map_window_notify` (which admittedly was at a very unintuitive place, this is now much cleaner), breaking compositors making use of it. Lets reintroduce it.


ids1024 (approved):
Seems to fix cosmic-comp, and this looks like an appropriate place to call it now.

cmeissl (approved):

PolyMeilex (approved):

PR #1437, opened 2024-05-22
compositor: Disable `IN_FENCE_FD` on Nvidia

ids1024:
Atomic commits fail if this is set on the Nvidia driver, it seems: https://github.com/NVIDIA/open-gpu-kernel-modules/issues/622

Kwin also has a check for this https://invent.kde.org/plasma/kwin/-/merge_requests/4770.

This fixes Anvil and cosmic-comp on the Nvidia 555 beta driver. (I'm not sure why it only started to be a problem; `EGL_ANDROID_native_fence_sync` is present in the 550 driver, etc.)

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -1640,6 +1640,22 @@ where
             .overlay
             .sort_by_key(|p| std::cmp::Reverse(p.zpos.unwrap_or_default()));
 
+        let driver = surface.get_driver().map_err(|err| {
+            FrameError::DrmError(DrmError::Access(AccessError {
+                errmsg: "Failed to query drm driver",
+                dev: surface.dev_path(),
+                source: err,
+            }))
+        })?;
+        // `IN_FENCE_FD` makes commit fail on Nvidia driver
+        // https://github.com/NVIDIA/open-gpu-kernel-modules/issues/622
+        let is_nvidia = driver.name().to_string_lossy().to_lowercase().contains("nvidia")
+            || driver
+                .description()
+                .to_string_lossy()
+                .to_lowercase()
+                .contains("nvidia");
+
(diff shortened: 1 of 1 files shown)


src/backend/drm/compositor/mod.rs
```diff
@@ -1640,6 +1640,22 @@ where
...
+            }))
+        })?;
+        // `IN_FENCE_FD` makes commit fail on Nvidia driver
+        // https://github.com/NVIDIA/open-gpu-kernel-modules/issues/622
+        let is_nvidia = driver.name().to_string_lossy().to_lowercase().contains("nvidia")
+            || driver
+                .description()
+                .to_string_lossy()
+                .to_lowercase()
+                .contains("nvidia");
```
Drakulix, line 1657: Does the open-driver also set `nvidia` or is it possibly `nvidia-open` and we need to filter for that as well?

Drakulix (commented):
> I'm not sure why it only started to be a problem; EGL_ANDROID_native_fence_sync is present in the 550 driver, etc.

Possibly because of the DrmCapability SyncObj now being announced?

ids1024:
Ah yeah, it's probably the `DriverCapability::SyncObj`.

src/backend/drm/compositor/mod.rs
ids1024: https://github.com/NVIDIA/open-gpu-kernel-modules/blob/083cd9cf17ab95cd6f9fb50a5349c21eaa2f7d4b/kernel-open/nvidia-drm/nvidia-drm-drv.c#L1666-L1669

Looks like NVIDIA's open kernel modules just advertise the driver as `"nvidia-drm"`, `"NVIDIA DRM driver"`, `"20160202"`. Like the closed source kernel driver does.

Drakulix (approved):

PR #1438, opened 2024-05-24
style: Use `AnvilState<BackendData>` everywhere

kenoss:
Before this patch, `AnvilState<BackendData>`,
`AnvilState<Backend: crate::state::Backend>`, and `AnvilState<B>` are mixed up. This patch uses `AnvilState<BackendData>` for consistency.


Drakulix (approved):
Thanks!

PR #1439, opened 2024-05-28
Fix the xwayland shell - X11Window to WlSurface association

PapyElGringo:
Since the `map_window_notify` doesn't ensure that a valid association is made between a X11Surface and a WlSurface we need to have a new way to associate them. But currently the `surface_associated` doesn't provide arguments that allow this association in the implementation side.]

This PR aim to fix thoses issues.

Note: It's currently doesn't work as expected since the execution order seems to be odd


Drakulix (approved):
Thanks for working on this!

PR #1444, opened 2024-06-05
xwayland: Make hook implementation-detail

Drakulix:
Also fixes, that the commit hook is added on every SetSerial call (xwayland re-uses surfaces).

cc @PapyElGringo, @colinmarc


PapyElGringo:
:+1:

PolyMeilex (approved):
Looks reasonable, and a lot nicer.

PR #1445, opened 2024-06-05
Move pointer protocol extensions logic to their corresponding modules

PolyMeilex:
Maybe it's just a personal preference, but every time I open seat logic code, and see some new protocol extension added in there it starts to overwhelm me a bit.
So spliting it up per protocol seems resonable.

Also this adresses my other main problem with reading seat code, the fact that data of the protocol is defined in `input::` module rather than next to the protocol implementation, which leads to constant jumping around. With this change state type definition is stored next to the logic.


Drakulix (approved):
Don't see any problems with this and certainly structures stuff better.

PR #1448, opened 2024-06-06
Multiply alpha modifier with alpha in `WaylandSurfaceRenderElement`

ids1024:
Previously the alpha modifier was not used anywhere within Smithay. The compositor possibly could handle the alpha value itself, but it's not clear how to do that when using `render_elements_from_surface_tree`. Nor is there an obvious reason not to handle this in Smithay.

The protocol leaves the exact blending space undefined. But multiplying here seems like a correct implementation.


PolyMeilex (approved):
Nice and straight forward, thanks!

PR #1449, opened 2024-06-07
Perf/tiled renderers

cmeissl:
This extends the renderer api to take the opaque regions which can be used to optimize rendering.
On certain hardware it is important to disable blending when not necessary to achieve the full potential.
This is especially important on tile based renderer which might otherwise have to do an expensive read-back
from dram to the tile cache.

### DIFF
--- a/src/backend/renderer/multigpu/mod.rs
+++ b/src/backend/renderer/multigpu/mod.rs
@@ -1212,6 +1212,7 @@ where
                             Rectangle::from_loc_and_size((0, 0), buffer_size).to_f64(),
                             Rectangle::from_loc_and_size((0, 0), self.size),
                             &damage,
+                            &[Rectangle::from_loc_and_size((0, 0), self.size)],
                             Transform::Normal,
                             1.0,
                         )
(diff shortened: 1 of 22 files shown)


src/backend/renderer/multigpu/mod.rs
```diff
@@ -1212,6 +1212,7 @@ where
                             Rectangle::from_loc_and_size((0, 0), buffer_size).to_f64(),
                             Rectangle::from_loc_and_size((0, 0), self.size),
                             &damage,
+                            &[],
```
Drakulix, line 1215: Well I guess, everything is opaque in this case, but it doesn't matter because we cleared it before?

Drakulix (approved):
LGTM

src/backend/renderer/multigpu/mod.rs
cmeissl: Right, I added the size as opaque here. We might want to use a different shader in future for drawing opaque regions ignoring the alpha channel.

(I also changed it in the same function for the mem copy case)

PR #1452, opened 2024-06-11
Refactor fractional-scale to store state persistently

YaLTeR:
Did you know about `map_or()`? It's my new favorite function for reducing line count


Drakulix (approved):
Makes sense imo

PR #1456, opened 2024-06-14
gles: disable opaque short-cut for custom shaders

### LINKED ISSUE #1455: Custom texture shader regression from implicit opaque check
This check https://github.com/Smithay/smithay/commit/1859d4d28e326ea26b0d8d0b2f7c46e2fc405d7e#diff-4a2a291b9d780fc4d734426015f88a0f865d8ab8a3996e88a771fc52adac256eR2906 assumes the texture shader does not mess with the transparency, which is not a given. I have a rounded corners shader which correctly sets its opaque regions, but that check does not take them into account, which results in artifacts.

cc @cmeissl

cmeissl:
our assumption on implicit opaque rendering might
not hold for external shaders. in this case disable the short-cut and just use the default opaque region
logic to decide of blending should be enabled or not.

fixes #1455 

@YaLTeR please test

### DIFF
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -2874,13 +2874,13 @@ impl<'frame> GlesFrame<'frame> {
         non_opaque_damage.clear();
         opaque_damage.clear();
 
-        // If drawing is implicit opaque we can skip some logic and save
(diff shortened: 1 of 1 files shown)


YaLTeR:
Works fine, thanks

Drakulix (approved):

PR #1458, opened 2024-06-18
gles: clean-up on unbind

cmeissl:
this prevents resources from being kept alive unnecessarily

@ids1024 @Drakulix please test if this fixes the multi-gpu leak

### DIFF
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -2028,6 +2028,7 @@ impl Unbind for GlesRenderer {
         }
         unsafe { self.gl.BindFramebuffer(ffi::FRAMEBUFFER, 0) };
         self.target = None;
+        self.cleanup();
         self.egl.unbind()?;
         Ok(())
     }


ids1024:
Updating cosmic-comp for Smithay changes (https://github.com/pop-os/cosmic-comp/pull/561) then pointing it at `fix/copy_texture_resource_leak`, I'm still seeing memory usage climbing with an Nvidia client running on an Intel monitor.

cmeissl:
> Updating cosmic-comp for Smithay changes ([pop-os/cosmic-comp#561](https://github.com/pop-os/cosmic-comp/pull/561)) then pointing it at `fix/copy_texture_resource_leak`, I'm still seeing memory usage climbing with an Nvidia client running on an Intel monitor.

:/ Okay, so it probably does not hit that path and instead in `MultiGpu::import_dmabuf_internal`/`MultiGpu::texture_copy` it uses dmabuf import.

Drakulix:
this version fixes it.

PR #1459, opened 2024-06-19
Keep dmabuf formats in order

cmeissl:
This PR changes the dmabuf format handling so that the list populated over zwp_dmabuf stays in the same order as returned from the driver. Yes, this should in theory not make a difference, but it does on some drivers.
I noticed this because on a particular binary driver egl clients, using dmabuf v3, misbehaved. Further debugging
has shown that it semi-randomly selected formats, sometimes even YUV formats. This caused everything from wrong
color to performance issues.

Before this PR formats were only stable for a single run of the process as we used a HashSet which internally uses `RandomState`, resulting in different order each time we built the formats.

(Some of the changes are probably not needed, like the change to the plane format ordering, and just me being paranoid after debugging this...)

### DIFF
--- a/src/wayland/dmabuf/mod.rs
+++ b/src/wayland/dmabuf/mod.rs
@@ -188,7 +188,7 @@
 mod dispatch;
 
 use std::{
-    collections::{HashMap, HashSet},
+    collections::HashMap,
     ffi::CString,
     ops::Sub,
     os::unix::io::AsFd,
@@ -198,7 +198,7 @@ use std::{
     },
 };
 
-use indexmap::IndexSet;
+use indexmap::{IndexMap, IndexSet};
 use rustix::fs::{seek, SeekFrom};
 use wayland_protocols::wp::linux_dmabuf::zv1::server::{
     zwp_linux_buffer_params_v1::{self, ZwpLinuxBufferParamsV1},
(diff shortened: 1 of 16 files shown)


src/wayland/dmabuf/mod.rs
```diff
@@ -617,7 +621,7 @@ impl DmabufState {
     pub fn create_global_with_filter<D, F>(
         &mut self,
         display: &DisplayHandle,
-        formats: Vec<Format>,
+        formats: impl IntoIterator<Item = Format>,
```
Drakulix, line 624: So why do we switch this constructor to use an Iterator and not the other we are internally calling?

Drakulix (approved):
Looks mostly good to me.

I wonder, if we really need the `FormatSet` type, but then again it is probably a lot nicer to use than an `Arc<IndexMap<Format>>`. And at least the usage is consistent now.

cmeissl:
> Looks mostly good to me.
> 
> I wonder, if we really need the `FormatSet` type, but then again it is probably a lot nicer to use than an `Arc<IndexMap<Format>>`. And at least the usage is consistent now.

Yeah, the Arc was actually just a side effect. The main reason was to not have `IndexSet` in the public API.

src/wayland/dmabuf/mod.rs
cmeissl: The whole change is not really needed for this, I just thought it might make the API a bit nicer and more consistent with the other functions. The reason for not changing the internal function is that it uses an `Option` and `Option<impl IntoIterator<Item = Format>>` is a pain to call (there are ways around that, but I did not feel like there is a need for a nicer solution)

PR #1460, opened 2024-06-19
Update `calloop`, `ash`, `winit`

ids1024:


Drakulix (approved):
LGTM

ids1024:
I guess to fix build with `-Z minimial-versions` it needs to force a newer version of wayland-client that doesn't use `nix` without specifying certain required features...

PR #1461, opened 2024-06-21
perf: use WlWeak::is_alive instead of upgrading

cmeissl:
this has been introduced in the latest wayland-rs release and can safe a few cycles, especially for lists holding a lot of weak resources.


PolyMeilex (approved):
That's nice

PR #1463, opened 2024-06-22
drm: pick one primary plane if we have multiple

### LINKED ISSUE #1462: Primary plane selection broken on some devices
It seems we forgot handling devices where the same primary plane can be used on different CRTC. In this case we might end up using the same primary plane multiple times, which apparently can not work.

cmeissl:
some drivers might offer multiple candidates as the primary plane for a CRTC. pick one and claim it so that we do not accidentially select the same plane multiple times.

fixes #1462

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -512,9 +512,24 @@ impl DrmLeaseHandler for AnvilState<UdevData> {
                 builder.add_connector(conn);
                 builder.add_crtc(*crtc);
                 let planes = backend.drm.planes(crtc).map_err(LeaseRejected::with_cause)?;
-                builder.add_plane(planes.primary.handle);
-                if let Some(cursor) = planes.cursor {
-                    builder.add_plane(cursor.handle);
+                let (primary_plane, primary_plane_claim) = planes
+                    .primary
+                    .iter()
+                    .find_map(|plane| {
+                        backend
+                            .drm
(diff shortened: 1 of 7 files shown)


anvil/src/udev.rs
```diff
@@ -523,7 +523,9 @@ impl DrmLeaseHandler for AnvilState<UdevData> {
                 builder.add_connector(conn);
                 builder.add_crtc(*crtc);
                 let planes = backend.drm.planes(crtc).map_err(LeaseRejected::with_cause)?;
-                builder.add_plane(planes.primary.handle);
+                for plane in &planes.primary {
```
cmeissl, line 526: @Drakulix I am not sure if this is correct?

Drakulix: No it isn't. All resources passed here, are given to the client and I don't think any client has use for multiple primary planes. We should just pick one here (ideally a free one :sweat_smile: ).

Drakulix (commented):
Looks good on first glace

anvil/src/udev.rs
cmeissl: I somewhat expected this to be incorrect :smile: 

Just picking a currently free one might again result in duplicate plane usage. I added the claim logic to the
lease builder. Afaict this should be good?

Drakulix: Looks correct now :+1:

Drakulix (approved):
LGTM!

PR #1465, opened 2024-06-26
Expose _NET_WM_PID of XWayland Surface

PapyElGringo:


src/xwayland/xwm/surface.rs
```diff
@@ -829,6 +851,24 @@ impl X11Surface {
...
+        let reply = match conn
+            .get_property(false, self.window, atom, AtomEnum::CARDINAL, 0, 1)?
+            .reply_unchecked()
+        {
+            Ok(Some(reply)) => reply,
+            Ok(None) | Err(ConnectionError::ParseError(_)) => return Ok(None),
+            Err(err) => return Err(err),
+        };
+
+        let Some(u32) = reply.value32().unwrap().next() else {
```
Drakulix, line 865: `u32` is a valid identifier? Certainly interesting, if that is the case, but please use something like `value` instead.

Drakulix (commented):
Looks good! Just a small nitpick

Drakulix (approved):

PR #1466, opened 2024-06-26
renderer/gles: Drop shadow buffer

Drakulix:
As discussed previously, lets drop the shadow buffer code. We can scanout to 10-bit without it and any color-aware rendering is not going to end up in this renderer, so it just hurts performance. Which is especially bad considering this is supposed to be the most compatible renderer (apart from software rendering).

Also fixes https://github.com/Smithay/smithay/issues/1378


anvil/src/udev.rs
```diff
@@ -244,14 +244,7 @@ pub fn run_udev() {
...
         let egl_context = EGLContext::new_with_priority(egl_display, ContextPriority::High)?;
-        let disable_color_transform = std::env::var("ANVIL_DISABLE_COLOR_TRANSFORM").is_ok();
-        unsafe {
-            let capabilities = GlesRenderer::supported_capabilities(&egl_context)?
-                .into_iter()
-                .filter(|c| !disable_color_transform || *c != Capability::ColorTransformations);
-
-            Ok(GlesRenderer::with_capabilities(egl_context, capabilities)?)
-        }
+        unsafe { Ok(GlesRenderer::new(egl_context)?) }
```
cmeissl, line 247: This could now use `GbmGlesBackend::with_context_priority(ContextPriority::High)`

anvil/src/udev.rs
```diff
@@ -244,14 +244,7 @@ pub fn run_udev() {
 
     let gpus = GpuManager::new(GbmGlesBackend::with_factory(|egl_display| {
         let egl_context = EGLContext::new_with_priority(egl_display, ContextPriority::High)?;
-        let disable_color_transform = std::env::var("ANVIL_DISABLE_COLOR_TRANSFORM").is_ok();
```
cmeissl, line 247: The `README` of anvil also mentions  `ANVIL_DISABLE_COLOR_TRANSFORM`

cmeissl (changes requested):

anvil/src/udev.rs
Drakulix: Good call, fixed.

anvil/src/udev.rs
Drakulix: Dropped from the readme.

cmeissl (approved):

ids1024:
The shadow buffer code stood out on profiling, so this seems like a good change.

PR #1470, opened 2024-06-28
drm: always submit the complete state

cmeissl:
while it should be totally fine to only
update the planes that have actually changed
it seems some drivers might break by doing so.

after some trial and error it seems that it might
be enough to always include the primary plane
in the update to workaround this issues. But
to be safe for now and until this has resolved
we just include all used planes in the update.

might fix/workaround https://github.com/pop-os/cosmic-comp/issues/499

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -784,7 +784,22 @@ impl<B: Framebuffer> FrameState<B> {
                 // actually used. We can skip getting a claim here if we have a
                 // config as this means we already claimed the plane for us.
                 if allow_partial_update {
-                    !state.skip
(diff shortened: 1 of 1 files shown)


ids1024:
Testing this commit and the previous one a couple times with the steps in https://github.com/pop-os/cosmic-comp/issues/499#issuecomment-2161436025, I'm able to consistently reproduce that issue before this change, and this seems to fix it. Along with the persistent duplication on the screen there seems to be some less consistent flickering that can happen and may also be fixed here.

So at least with the behavior I've seen, this seems clearly helpful as a workaround.

ids1024 (approved):

Drakulix (approved):

PR #1472, opened 2024-06-30
Use CStr literals everywhere possible

linkmauve:
This addition to the 2021 edition makes it possible to generate a `&CStr` at compile-time, with no need for `unwrap()`/`expect()` or `unsafe`, making it much more ergonomic to pass to C APIs.

I’ve also modified the `SealedString` constructors to take a `&CStr` instead of a `CString` since it only needs to use it, not keep ownership of it.

This requires to bump the MSRV to 1.77 so I just did.

### DIFF
--- a/examples/buffer_test.rs
+++ b/examples/buffer_test.rs
@@ -1,4 +1,4 @@
-use std::{ffi::CStr, fmt, fs::File, os::unix::io::OwnedFd, path::PathBuf};
+use std::{fmt, fs::File, os::unix::io::OwnedFd, path::PathBuf};
 
 use clap::{Args, Parser, Subcommand, ValueEnum};
 use smithay::{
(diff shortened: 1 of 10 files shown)


examples/buffer_test.rs
```diff
@@ -189,11 +189,7 @@ fn buffer_test(args: TestArgs) {
                 Instance::new(Version::VERSION_1_2, None).expect("Unable to create vulkan instance");
             let physical_device = PhysicalDevice::enumerate(&instance)
                 .expect("Failed to enumerate physical devices")
-                .filter(|phd| {
-                    phd.has_device_extension(unsafe {
-                        CStr::from_bytes_with_nul_unchecked(b"VK_EXT_physical_device_drm\0")
-                    })
-                })
+                .filter(|phd| phd.has_device_extension(c"VK_EXT_physical_device_drm"))
```
i509VCB, line 192: We could consider using the `NAME` constants defined in ash. But this is fine.

linkmauve: I’ve used this constant, I didn’t know about that in ash.

cmeissl:
Thanks, looks good in general. Unfortunately bumping the MSRV is not an option at the moment. I will try to make sure we can do this in the near future, but this has to wait until we are able to bump the MSRV in some projects that depend on `smithay`.

Drakulix:
We bumped the MSRV as part of https://github.com/Smithay/smithay/pull/1576.

Could you drop https://github.com/Smithay/smithay/pull/1472/commits/2eac2b258dfadfb9ae1b32c461ea80252bfe1f16 and rebase this PR?

linkmauve:
dbe96e2037eede10005b64a168fb74dc03f2870c did the same changes, closing.

PR #1473, opened 2024-07-01
pixman: re-map resized shm pool images

cmeissl:
The actual mmap can change if the shm pool gets
resized. In this case we have to re-map our image
to prevent accessing a unmapped memory region.


Drakulix (approved):
LGTM

PR #1475, opened 2024-07-02
shm/pool: Map as writable in `nullify_map`

ids1024:
The actual mapping is mutable, so the placeholder should also be. Or writes will trigger a `SEGV` due to the page being unmapped when execution is resumed after the `BUS` handler.

This fixes uses of `with_buffer_contents_mut` if the file provided by the client isn't large enough. In particular the segfault reported in https://github.com/pop-os/cosmic-comp/issues/48.


Drakulix (approved):
Seems sensible, thanks!

PR #1478, opened 2024-07-03
wayland: only send dnd_drop_performed for accepted offers

### LINKED ISSUE #1474: Can't click in chromium tab after selecting and dragging text
I've confirmed with the folks over on the niri matrix that this issue happens in both niri, and cosmic-comp, so it's likely smithay related.

Whenever you select text in chromium, and drag it around (even just for 1 pixel), you now lose the ability to click in that tab, and have to close it.

Does not happen in non smithay based compositors. 

I hope the description is enough!

cmeissl:
fixes #1474

### DIFF
--- a/src/wayland/selection/data_device/dnd_grab.rs
+++ b/src/wayland/selection/data_device/dnd_grab.rs
@@ -237,11 +237,10 @@ where
             }
         }
         if let Some(ref source) = self.data_source {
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
LGTM

YaLTeR:
Giving this some testing in niri, seems so far so good.

PR #1479, opened 2024-07-04
seat/kbd: Allow overriding replace behaviour

Drakulix:
Similar to `PointerTarget::replace` this allows overriding the behaviour for custom targets.


cmeissl (approved):
LGTM but clippy disagrees

PR #1480, opened 2024-07-05
drm: pick one cursor plane if we have multiple

cmeissl:
some drivers might offer multiple cursor planes per CRTC. in this case try to use the first unclaimed plane and try to keep the same for as long as possible


Drakulix (commented):
LGTM on first glance

Drakulix (approved):

PR #1482, opened 2024-07-10
xwm: Rename HIDDEN state from minimized to suspended

Drakulix:
`NET_WM_STATE_HIDDEN` represents what xdg-shell calls `suspended`, not minimized. X11 windows typically expect to be unmapped and remapped, when asking for a specific minimized state, not for `_NET_WM_STATE_HIDDEN`. So lets rename this to be less confusing and more consistent.

### DIFF
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -492,11 +492,11 @@ impl X11Surface {
         Ok(())
     }
 
-    /// Sets the window as minimized or not.
+    /// Sets the window as suspended/hidden or not.
     ///
-    /// Allows the client to e.g. stop rendering while minimized.
-    pub fn set_minimized(&self, minimized: bool) -> Result<(), ConnectionError> {
-        if minimized {
+    /// Allows the client to e.g. stop rendering.
+    pub fn set_suspended(&self, suspended: bool) -> Result<(), ConnectionError> {
+        if suspended {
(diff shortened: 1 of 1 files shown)


ids1024:
https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html

> _NET_WM_STATE_HIDDEN should be set by the Window Manager to indicate that a window would not be visible on the screen if its desktop/viewport were active and its coordinates were within the screen bounds. The canonical example is that minimized windows should be in the _NET_WM_STATE_HIDDEN state. Pagers and similar applications should use _NET_WM_STATE_HIDDEN instead of WM_STATE to decide whether to display a window in miniature representations of the windows on a desktop. 
>
> Implementation note: if an Application asks to toggle _NET_WM_STATE_HIDDEN the Window Manager should probably just ignore the request, since _NET_WM_STATE_HIDDEN is a function of some other aspect of the window such as minimization, rather than an independent state. 

EWMH seems to think that minimized windows should be set to `_NET_WM_STATE_HIDDEN`, but also emphasizes that it isn't equivalent to minimization, but a separate state influenced by "some other aspect of the window" including minimization.

So not calling this "minimized" seems correct, at least, from the perspective of the spec. (I'm not sure exactly when it is in practice set by modern compositing WMs. Always when windows are minimized? For anything else?)

cmeissl (approved):

PR #1485, opened 2024-07-13
Return real alpha in SolidColorRenderElement

YaLTeR:
Without this, changing just the alpha in SolidColorRenderElement::from_buffer() doesn't damage the element.

I didn't test this in Smithay, but I bumped into this problem in my in-tree copy in niri.


Drakulix (approved):
Yeah this seems very sensible

PR #1491, opened 2024-07-26
fix for new clippy lints (1.80.0)

cmeissl:


PolyMeilex (approved):
LGTM

If `mutable_key_type` lint has so many [false positives](https://github.com/rust-lang/rust-clippy/issues/13109) in our case, I wonder if it would make sense to just disable it for the whole crate.

cmeissl:
> LGTM
> 
> If `mutable_key_type` lint has so many [false positives](https://github.com/rust-lang/rust-clippy/issues/13109) in our case, I wonder if it would make sense to just disable it for the whole crate.

Might be worth if we hit more false positives, but with only these 3 I would keep it enabled for now.

Drakulix (approved):

PR #1492, opened 2024-07-26
drm: remove implicit egl fence wait

cmeissl:
always testing if the fence has been signaled
might hurt performance. we either use the exported fence fd or wait on it on the cpu anyway


Drakulix (approved):
Sure, lets hope all KMS drivers are smart enough to figure this out.

PR #1494, opened 2024-07-28
Single pixel buffer (take 2)

cmeissl:
This PR adds support for the `wp_single_pixel_buffer` protocol, based on #719 and  #988

Main difference to #988 is that this PR leaves the `WaylandSurfaceRenderElement` in tact instead of replacing it
with an enum holding a `TextureRenderElement` and `SolidRenderElement`. Some code might already depend
on the `WaylandSurfaceRenderElement` and moving everything over to the `TextureRenderElement` in a generic
way would just complicate the code. This also allows us to keep the wayland surface specific rounding while we
could lift that for the other render elements.

### DIFF
--- a/src/backend/renderer/color.rs
+++ b/src/backend/renderer/color.rs
@@ -1,0 +1,77 @@
+use std::ops::Mul;
+
+/// A four-component color representing pre-multiplied RGBA color values
+#[derive(Debug, Copy, Clone, Default, PartialEq)]
+pub struct Color32F([f32; 4]);
+
+impl Color32F {
+    /// Initialize a new [`Color`]
+    #[inline]
+    pub const fn new(r: f32, g: f32, b: f32, a: f32) -> Self {
+        Self([r, g, b, a])
+    }
+}
+
+impl Color32F {
+    /// Transparent color
+    pub const TRANSPARENT: Color32F = Color32F::new(0.0, 0.0, 0.0, 0.0);
+
+    /// Solid black color
+    pub const BLACK: Color32F = Color32F::new(0f32, 0f32, 0f32, 1f32);
+}
+
+impl Color32F {
+    /// Red color component
+    #[inline]
+    pub fn r(&self) -> f32 {
+        self.0[0]
+    }
+
+    /// Green color component
+    #[inline]
+    pub fn g(&self) -> f32 {
+        self.0[1]
+    }
+
+    /// Blue color component
+    #[inline]
+    pub fn b(&self) -> f32 {
+        self.0[2]
+    }
+
+    /// Alpha color component
+    #[inline]
+    pub fn a(&self) -> f32 {
+        self.0[3]
+    }
+
+    /// Color components
+    #[inline]
+    pub fn components(self) -> [f32; 4] {
+        self.0
+    }
+}
+
+impl Color32F {
+    /// Test if the color represents a opaque color
+    #[inline]
+    pub fn is_opaque(&self) -> bool {
+        self.a() == 1f32
+    }
+}
+
+impl From<[f32; 4]> for Color32F {
+    #[inline]
+    fn from(value: [f32; 4]) -> Self {
+        Self(value)
+    }
+}
+
+impl Mul<f32> for Color32F {
+    type Output = Color32F;
+
+    #[inline]
+    fn mul(self, rhs: f32) -> Self::Output {
+        Self::new(self.r() * rhs, self.g() * rhs, self.b() * rhs, self.a() * rhs)
+    }
+}
--- a/src/backend/renderer/element/solid.rs
+++ b/src/backend/renderer/element/solid.rs
@@ -6,7 +6,7 @@
 //! # use smithay::{
 //! #     backend::{
 //! #         allocator::Fourcc,
-//! #         renderer::{DebugFlags, Frame, ImportMem, Renderer, Texture, TextureFilter, sync::SyncPoint},
+//! #         renderer::{Color32F, DebugFlags, Frame, ImportMem, Renderer, Texture, TextureFilter, sync::SyncPoint},
 //! #     },
 //! #     utils::{Buffer, Physical, Rectangle, Transform},
 //! # };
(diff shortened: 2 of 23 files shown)


src/backend/renderer/element/surface.rs
```diff
@@ -530,14 +547,17 @@ where
...
+            WaylandSurfaceTexture::Texture(ref texture) => frame.render_texture_from_to(
+                texture,
+                src,
+                dst,
+                damage,
+                opaque_regions,
+                self.buffer_transform,
+                self.alpha,
+            ),
+            WaylandSurfaceTexture::SolidColor(color) => frame.draw_solid(dst, damage, color),
```
Drakulix, line 560: The wp-alpha-modifier protocol can be used with single-pixel buffers. We need to accommodate for that here, by blending `self.alpha` with `color`.

src/backend/renderer/element/solid.rs
```diff
@@ -251,10 +253,10 @@ impl SolidColorRenderElement {
...
         let geo = Rectangle::from_loc_and_size(location, buffer.size.to_physical_precise_round(scale));
         let color = [
-            buffer.color[0] * alpha,
-            buffer.color[1] * alpha,
-            buffer.color[2] * alpha,
-            buffer.color[3] * alpha,
+            buffer.color.r() * alpha,
+            buffer.color.g() * alpha,
+            buffer.color.b() * alpha,
+            buffer.color.a() * alpha,
```
Drakulix, line 259: Nitpick: Couldn't multiplying by an alpha be a method of `Color32F`, so we can then use `components`?

src/backend/renderer/color.rs
```diff
@@ -0,0 +1,66 @@
+/// A four-component color
```
Drakulix, line 1: Nitpick: Can we document that this is supposed to be pre-multiplied?
I don't think we handle colors in any other way anywhere in smithay and once we do I would use the type-system anyway to make misuse harder.

Drakulix (approved):
Looks (mostly) good to me.

src/backend/renderer/color.rs
cmeissl: Yeah, I started with some generic `Color` type, but opted out pretty quickly... :sweat_smile:

src/backend/renderer/element/solid.rs
cmeissl: Sure, I thought about adding a `Mul` impl or do you prefer some explicit function for that?
(No need to use components, `draw_solid` also takes `Color32F`)

Drakulix: I think `Mul` is good enough.

YaLTeR:
As far as I understand, this does not yet add any special cases for direct scanout when a smaller-than-monitor surface is shown on top of a black single-pixel-buffer background? I'm mentioning this because this seems to be one of the intended use-cases for single-pixel-buffer, e.g. https://docs.gtk.org/gtk4/method.GraphicsOffload.set_black_background.html

YaLTeR:
To elaborate: for overlay planes it'll work as is naturally, but I've seen discussions about using it for primary plane fullscreen scanout. If I'm not mistaken, it was something along: with color management, black background may mean "display this specific shade of black" resulting in two layers while in fullscreen, while the solid pixel buffer black can be assumed to be the "unspecified plane background", thereby resulting in only a single layer while in fullscreen.

Drakulix:
> To elaborate: for overlay planes it'll work as is naturally, but I've seen discussions about using it for primary plane fullscreen scanout. If I'm not mistaken, it was something along: with color management, black background may mean "display this specific shade of black" resulting in two layers while in fullscreen, while the single pixel buffer black can be assumed to be the "unspecified plane background", thereby resulting in only a single layer while in fullscreen.

Yeah, I guess that would need filtering in the `DrmCompositor`. Essentially any `[0.0, 0.0, 0.0, 0.0]`-single-pixel-buffers from the bottom of the render-elements can be dropped and then we should end up with a single layer for primary plane scanout? (Once we support primary planes with different dst sizes.)

cmeissl:
> > To elaborate: for overlay planes it'll work as is naturally, but I've seen discussions about using it for primary plane fullscreen scanout. If I'm not mistaken, it was something along: with color management, black background may mean "display this specific shade of black" resulting in two layers while in fullscreen, while the single pixel buffer black can be assumed to be the "unspecified plane background", thereby resulting in only a single layer while in fullscreen.
> 
> Yeah, I guess that would need filtering in the `DrmCompositor`. Essentially any `[0.0, 0.0, 0.0, 0.0]`-single-pixel-buffers from the bottom of the render-elements can be dropped and then we should end up with a single layer for primary plane scanout? (Once we support primary planes with different dst sizes.)

I thought about that a bit the last few days and one way to do this would be extend UnderlyingStorage and when detecting this situation to override the clear color internally. We just have to be careful to not break underlay assumptions.

cmeissl:
So basically to extend the first loop in render_elements to test id we found the last visible element on that output and then do the rest of the necessary checks. If every requirement is fulfilled skip thus element and override the clear color.

cmeissl:
Okay, I implemented the special casing in https://github.com/Smithay/smithay/pull/1494/commits/4af22e3f15684af1a3e16575b48bf36a96d666bd

I tested this by using swaybg to set the background color to all black and running glmark2 on the output. Without swaybg (or non black color) glmark2 is placed on an overlay plane. With all black spb from swaybg it successfully gets scanned out on the primary plane.

(This also revealed a bug in the `OutputDamageTracker`  which did not clear everything in case the clear color changed)

src/backend/renderer/element/solid.rs
cmeissl: done

src/backend/renderer/color.rs
cmeissl: done

src/backend/renderer/color.rs
```diff
@@ -0,0 +1,77 @@
...
+        self.a() == 1f32
+    }
+}
+
+impl From<[f32; 4]> for Color32F {
+    #[inline]
+    fn from(value: [f32; 4]) -> Self {
+        Self(value)
+    }
+}
```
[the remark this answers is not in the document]
cmeissl: I omitted that in preference to the explicit components function. The From impl was mostly for compatibility reasons.

PR #1495, opened 2024-07-29
Add `from_resource` to `KeyboardHandle`/`PointerHandle`/`TouchHandle`

ids1024:
It seems most protocols take a `wl_seat` as an argument rather than a capability type, but it seems reasonable for Smithay to expose this in some way to compositors still.


Drakulix (approved):
Ah yeah, I have also stumbled upon this in the past. LGTM!

PR #1496, opened 2024-08-01
fix: check viewporter dest when validating the lock surface size

wash2:


src/wayland/session_lock/lock.rs
```diff
@@ -128,12 +129,27 @@ where
...
-                                        let scale = surface_attrs.buffer_scale;
-                                        let transform = surface_attrs.buffer_transform.into();
-                                        let surface_size = buf_size.to_logical(scale, transform);
-
-                                        let surface_size =
-                                            Size::from((surface_size.w as u32, surface_size.h as u32));
+                                        let viewport = states
+                                            .data_map
+                                            .get::<ViewporterSurfaceState>()
+                                            .unwrap()
```
Drakulix, line 135: I am not sure we even attach a `ViewportSurfaceState`, if the client never binds the viewporter protocol. So this `unwrap` could potentially explode and should be treated like the if-condition further down.

wash2: Ahh, ok, I copied that from the viewport commit hook. I'll change it :sweat_smile:.

Drakulix (approved):

PR #1497, opened 2024-08-05
WIP anvil/udev: Dynamic support for pixman rendererer

ids1024:
With the changes here, the anvil udev backend will use pixman for rendering if `ANVIL_USE_PIXMAN` is set.

To do:
- Figure out how to deal with `Send` bound on `MemoryRenderBuffer`, and no `Send` for `PixmanTexture`
- Fix build with `debug` feature (need to abstract texture)
- Make render node optional, and automatically use pixman if hardware acceleration isn't present
- Pixman support in x11/winit backends

It's not entirely clear how best to abstract renderers dynamically:
- Should an enum wrapper implement the `Renderer` trait?
  * How does that impact performance or binary size compare to dispatching to a render function twice like this currently does?
  * Will also need enums over texture, frame, error; and probably a panic if these types are mismatched in some way


cmeissl:
I tried to create a wrapper enum implementing the renderer traits. `Bind` is the most "problematic" as this can be different per renderer.

Macro based implementation: https://github.com/Smithay/smithay/compare/master...cmeissl:smithay:feature/software_renderer_rework#diff-a0c304b3769f4f73ae1f61dea898368741b6e582845e806a5fe84c06fbec967e
Usage of the enum: https://github.com/Smithay/smithay/compare/master...cmeissl:smithay:feature/software_renderer_rework#diff-a6fa9efbdd07cfc0b5edaadc6a11ccc76ca84c16664de459c08b0b7bd3b63ba5R28

ids1024:
I was thinking a macro might make sense. That looks handy. Although macro rules macros like that get a bit complicated.

ids1024:
Updated on top of https://github.com/Smithay/smithay/pull/1600.

Testing this I see a crash:

```
2024-12-04T18:08:44.814186Z  WARN anvil::udev: Error during rendering: ContextLost(BufferDestroyed)
thread 'main' panicked at anvil/src/udev.rs:1613:30:
Rendering loop lost: The underlying buffer has been destroyed
```

I guess that's an existing issue with the pixman renderer, and not something changed here? Hm.

ids1024:
Updated so the `debug` feature builds, and all the feature combinations tested on CI now work. Still needs a fix for that crash.

Exactly how best to abstract over multiple renderers remains an open question.

ids1024:
This seems to be working correctly, including with the `debug` feature.

We may want to change how renderers are abstracted dynamically in the future (and provide some helper for this in Smithay), but this seems good for now, to have a way to test Pixman rendering properly.

PR #1501, opened 2024-08-09
drm/gbm: Don't use the legacy code path for import client dmabufs

Drakulix:
The legacy code path will succeed and then the nvidia driver will accept the framebuffer even though being unable to scan it out and never deliver another vblank event...

We probably shouldn't even try that code path for client-side buffers, so lets skip it.


cmeissl (approved):
Might still be worth raising an issue in nvidia drivers because the driver should not return OK for an error case imo.

PR #1505, opened 2024-08-18
utils: `id_gen!` optimization

Paraworker:
The current id generation logic may return an ID that is already in the HashSet.

I've also simplified the usage of `id_gen!`.

### DIFF
--- a/src/utils/ids.rs
+++ b/src/utils/ids.rs
@@ -1,36 +1,32 @@
 macro_rules! id_gen {
-    ($func_name:ident, $id_name:ident, $ids_name:ident) => {
-        static $id_name: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
-        lazy_static::lazy_static! {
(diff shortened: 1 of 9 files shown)


src/utils/ids.rs
```diff
@@ -1,36 +1,32 @@
...
-                            id += 1;
-                        }
-                        id += 1;
-                        Some(id)
-                    },
-                );
-                if let Ok(id) = new_id {
-                    break id;
+                while !id_set.insert(*counter) {
+                    *counter += 1;
```
Drakulix, line 18: Looking at this again, this should probably be a wrapping_add, just like the next increment.

Drakulix (commented):
Thanks! The module generation makes this indeed much cleaner to use.

src/utils/ids.rs
Paraworker: Oh thanks! Forgot the overflow panic in Rust.

Paraworker: Done.

Drakulix (approved):
LGTM now. Thanks!

PR #1506, opened 2024-08-18
expose multitexture get method

nongio:
The Multitexture get method, is useful for accessing the internal texture reference for a given DrmNode and a specific Renderer.
Exposing this method will allow Custom Renders to leverage the smithay automatic import of buffer while implementing their own rendering logic on the output elements

### DIFF
--- a/src/backend/renderer/multigpu/mod.rs
+++ b/src/backend/renderer/multigpu/mod.rs
@@ -1449,7 +1449,7 @@ impl MultiTexture {
         })))
     }
 
-    fn get<A: GraphicsApi + 'static>(
+    pub fn get<A: GraphicsApi + 'static>(
         &self,
         render: &DrmNode,
     ) -> Option<<<A::Device as ApiDevice>::Renderer as Renderer>::TextureId>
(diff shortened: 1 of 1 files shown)


src/backend/renderer/multigpu/mod.rs
```diff
@@ -1449,7 +1449,7 @@ impl MultiTexture {
         })))
     }
 
-    fn get<A: GraphicsApi + 'static>(
+    pub fn get<A: GraphicsApi + 'static>(
```
Drakulix, line 1452: Could you add a doc-comment for this function please?

```suggestion
    /// Retrieves the internal texture for a given render node, if any.
    pub fn get<A: GraphicsApi + 'static>(
```

nongio: i just saw that you wrote a suggestion for the doc, i wrote something looking at other public get methods, please let me know if it is ok or you prefer the shorter version

Drakulix: Nah seems good!

Drakulix (approved):
Thanks!

PR #1507, opened 2024-08-20
Apply protocol state in post-commit hooks

YaLTeR:
State should be applied on the actual commit. Pre-commit is too early, because the actual commit can and will be delayed by blockers. This would lead to inconsistent state where the current surface state is still old, but the protocol surface state is already new. Post-commit hook is the right place for this, it runs right after the current surface state is updated, and before the compositor commit handler is called.

After this change, my resize transactions impl in niri seems to work. Also, haven't run into any issues yet I think. But would be good for someone else to give this a try before merging. No compositor side changes are needed.


Drakulix (approved):
LGTM.

I'll definitely have to look at niri's implementation, as even with this cosmic-comps tiling algorithm (that also tries to use blockers for this very thing) doesn't work nearly as well as niri seems to do in your videos. But nothing regressed from what I can tell, so this is good to merge imo.

PR #1511, opened 2024-08-23
layer: Take popup geometry loc into account

YaLTeR:
Fixes offset cursor position in layer-shell popup menus.


Drakulix (approved):
Thanks!

PR #1515, opened 2024-08-26
WIP winit git

ids1024:
I wanted to see how the changes in Winit git would work here. So of course this depends on that being released.

Depends on https://github.com/Smithay/smithay/pull/1514. It doesn't seem to pose a particular challenge beyond that. Updating compositors for that change will be needed to test this...


ids1024:
Closing in favor of https://github.com/Smithay/smithay/pull/1869. Which at least works for testing with the latest winit beta.

PR #1522, opened 2024-08-31
libdisplay-info integration

cmeissl:
Rough integration of libdisplay-info into anvil.

Do we want to provide an easier integration? Maybe replacing `edid-rs` in `smithay-drm-extras`?
If yes it might make sense to limit this to returning the `Info` from `libdisplay-info` instead of `EdidInfo`.
`Info` already has accessor for model/make.

@PolyMeilex @Drakulix @YaLTeR


YaLTeR:
> Do we want to provide an easier integration? Maybe replacing edid-rs in smithay-drm-extras?
If yes it might make sense to limit this to returning the Info from libdisplay-info instead of EdidInfo.
Info already has accessor for model/make.

I think this makes sense. Looking at the rustdoc, there's `Info::edid()` which seems to get you access to the rest of the information?

Drakulix:
> Rough integration of libdisplay-info into anvil.
> 
> Do we want to provide an easier integration? Maybe replacing `edid-rs` in `smithay-drm-extras`? If yes it might make sense to limit this to returning the `Info` from `libdisplay-info` instead of `EdidInfo`. `Info` already has accessor for model/make.
> 
> @PolyMeilex @Drakulix @YaLTeR

Agreeing with @YaLTeR here, I think smithay-drm-extras should just return `libdisplay_info::Info` instead of `edid-rs::EdidInfo`.

PolyMeilex:
> I think smithay-drm-extras should just return libdisplay_info::Info

Sounds good to me, if we are going to need C dependency for display info anyways, then there is no need to keep edid-rs around.

cmeissl:
> Looks great, anything else you want to do with this before it is ready to merge?

nope, should be fine as is :)

PR #1523, opened 2024-09-02
Add accessor for FormatSet's underlying storage 

chrisduerr:
This gives read-only access to the underlying storage of `FormatSet`, to make it simpler for consumers to combine formats without having to reallocate.

### DIFF
--- a/src/backend/allocator/format.rs
+++ b/src/backend/allocator/format.rs
@@ -402,6 +402,11 @@ impl FormatSet {
             inner: self.formats.intersection(&other.formats),
         }
     }
+
+    /// Get access to the underlying storage.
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
I mean sure, but given this is read-only access, how does this avoid a reallocation?

chrisduerr:
> I mean sure, but given this is read-only access, how does this avoid a reallocation?

The problem is that there's no access to the underlying `IndexSet`, as such any method attempting to work with an indexset directly would have to call `iter`. The `FormatSet` only has a tiny fraction of the API surface of `IndexSet` and this API allows working with the many IndexSet functions. This is especially useful because many of them don't require ownership.

PR #1525, opened 2024-09-03
specify `ExactSizeIterator` for `Space::elements()`

m4rch3n1ng:
i don't know what else i could add there that makes sense. maybe FusedIterator.

### DIFF
--- a/src/desktop/space/mod.rs
+++ b/src/desktop/space/mod.rs
@@ -158,7 +158,7 @@ impl<E: SpaceElement + PartialEq> Space<E> {
     }
 
     /// Iterate elements in z-order back to front
-    pub fn elements(&self) -> impl DoubleEndedIterator<Item = &E> {
+    pub fn elements(&self) -> impl DoubleEndedIterator<Item = &E> + ExactSizeIterator {
         self.elements.iter().map(|e| &e.element)
     }
 


Drakulix:
The more rigorously we specify these constraints the less options we have to change this internally later. What is your use-case for `ExactSizeIterator`? Just optimizations?

m4rch3n1ng:
mostly optimizations, yes. i do use the count() method quite a bit, but i have realized in the past week that i can't use ExactSize half the time anyway, so this is a little pointless for me

> the less options we have to change this internally later

that is fair, but since most std data structures (i.e. HashSet, LinkedList, BtreeMap, etc) have an ExactSize iterator i don't think you would ever run into a problem with this, unless you want to like filter or flatten the Iterator.

i can close this and it would be fine bc i can't use it half the time anyway

Drakulix:
> that is fair, but since most std data structures (i.e. HashSet, LinkedList, BtreeMap, etc) have an ExactSize iterator i don't think you would ever run into a problem with this, unless you want to like filter or flatten the Iterator.

I was not questioning `ExactSize` in particular, I think that totally makes sense. I was just curious of the reasoning and cautious about adding further bounds.

Drakulix (approved):

PR #1527, opened 2024-09-06
xwayland: Some improvements for handling of xserver crashes

ids1024:
This logs a tracing error if the X server is disconnected with a protocol  error, and logs the return status of the `Xwayland` subprocess (which also means properly reaping the pid). Having these logged (and logged alongside any other compositor errors) will hopefully help with debugging some issues.

This also adds a `disconnected` method to `XwmHandler` so the compositor knows when we lose connection to the X server. (See https://github.com/pop-os/cosmic-comp/pull/823 for a use case.)

### DIFF
--- a/src/backend/x11/mod.rs
+++ b/src/backend/x11/mod.rs
@@ -578,8 +578,11 @@ impl EventSource for X11Backend {
 
         let post_action = self
             .source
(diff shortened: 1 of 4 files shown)


Drakulix (approved):
LGTM

Issue #1374, opened 2024-03-28
`prop_mapping` cache in `AtomicDrmDevice` doesn't handle connector hotplug


ids1024:
Cosmic-comp panics with `Unknown handle` in `AtomicDrmDevice::reset_state` on the Surface Pro 9 when:
* The compositor is started with no USB DisplayPort device
* Monitors are connected over USB-C; which adds DRM connectors
* TTY is switched away from cosmic-comp, and back

Presumably `prop_mapping` needs to be updated in this case.

PR #1531, opened 2024-09-09
drm/atomic: Update connector prop mapping in `reset_state()`, and refactor prop mapping code a bit

ids1024:
Fixes https://github.com/Smithay/smithay/issues/1374.

### DIFF
--- a/src/backend/drm/surface/atomic.rs
+++ b/src/backend/drm/surface/atomic.rs
@@ -92,16 +92,7 @@ impl State {
         // make sure the mapping is up to date
         map_props(fd, res_handles.connectors(), &mut prop_mapping.connectors)?;
         for conn in res_handles.connectors() {
-            let crtc_prop = prop_mapping
-                .connectors
-                .get(conn)
-                .expect("Unknown handle")
-                .get("CRTC_ID")
-                .ok_or_else(|| Error::UnknownProperty {
-                    handle: (*conn).into(),
-                    name: "CRTC_ID",
-                })
-                .copied()?;
+            let crtc_prop = prop_mapping.conn_prop_handle(*conn, "CRTC_ID")?;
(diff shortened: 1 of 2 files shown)


YaLTeR:
If I store some prop handles in my compositor (i.e. GAMMA_LUT) does that mean I need to refresh those at some point too, and when?

src/backend/drm/surface/atomic.rs
```diff
@@ -90,18 +90,9 @@ impl State {
         // If they don't match, `commit_pending` will return true and they will be changed on the next `commit`.
         let mut current_connectors = HashSet::new();
         // make sure the mapping is up to date
-        map_props(fd, res_handles.connectors(), &mut prop_mapping.0)?;
+        map_props(fd, res_handles.connectors(), &mut prop_mapping.connectors)?;
```
Drakulix, line 93: If we want be consistent here, we should probably clear `prop_mapping.connectors` first to not keep old entries around of properties missing for a new connector, no?

Drakulix:
> If I store some prop handles in my compositor (i.e. GAMMA_LUT) does that mean I need to refresh those at some point too, and when?

You should remove and re-evaluate them on hotplug, so if you are using smithay-drm-extras on `connector_changed` (otherwise in `UdevHandler::device_changed`, for any new connectors).

Note this only applies to connector properties, crtcs and planes are static anyhow.

ids1024:
> If I store some prop handles in my compositor (i.e. GAMMA_LUT) does that mean I need to refresh those at some point too, and when?

This is specifically an issue with connectors, which on some cards can be added and removed. I think once a drm object is created the properties it has are static? And I don't know if planes or CRTCs are ever added dynamically...

YaLTeR:
Okay I'm not sure I actually store any connector properties, but either way I would store them for each connector as it is connected (and remove that info when it is disconnected). But I do call `drm_compositor.reset_state()` which is mentioned in the title of this PR, which is why I asked.

Drakulix (approved):
LGTM now

PR #1533, opened 2024-09-13
Don't hold lock in `input_intercept` callback and don't expose `xkb::` types in safe API

ids1024:
See commits messages for descriptions of the changes here. The last commit is potentially more controversial than the earlier ones; if it's an issue, I think the `Xkb` changes in the others are a good improvement at least.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -208,7 +208,7 @@ pub(crate) struct KbdInternal<D: SeatHandler> {
     pub(crate) pressed_keys: HashSet<Keycode>,
     pub(crate) forwarded_pressed_keys: HashSet<Keycode>,
     pub(crate) mods_state: ModifiersState,
-    xkb: Xkb,
+    xkb: Arc<Mutex<Xkb>>,
     pub(crate) repeat_rate: i32,
     pub(crate) repeat_delay: i32,
     led_mapping: LedMapping,
@@ -255,11 +255,11 @@ impl<D: SeatHandler + 'static> KbdInternal<D> {
             pressed_keys: HashSet::new(),
             forwarded_pressed_keys: HashSet::new(),
             mods_state: ModifiersState::default(),
-            xkb: Xkb {
+            xkb: Arc::new(Mutex::new(Xkb {
                 context,
                 keymap,
                 state,
-            },
+            })),
             repeat_rate,
             repeat_delay,
             led_mapping,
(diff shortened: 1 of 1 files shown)


src/input/keyboard/mod.rs
```diff
@@ -162,6 +162,30 @@ impl Xkb {
...
+            .unwrap_or_default()
+    }
+
+    /// Get the human readable name for the layout.
+    pub fn layout_name(&self, layout: Layout) -> &str {
+        self.keymap.layout_get_name(layout.0)
+    }
+
+    /// Iterate over layouts present in the keymap.
+    pub fn layouts(&self) -> Box<dyn Iterator<Item = Layout>> {
```
YaLTeR, line 180: I suppose this Box and dyn can be removed now?

src/input/keyboard/mod.rs
```diff
@@ -130,15 +130,61 @@ impl LedState {
...
+            .field("context", &self.context.get_raw_ptr())
+            .field("keymap", &self.keymap.get_raw_ptr())
+            .field("state", &self.state.get_raw_ptr())
+            .finish()
+    }
+}
+
+// This is OK because all parts of `xkb` will remain on the
+// same thread
+unsafe impl Send for Xkb {}
```
Drakulix, line 203: But `Xkb` is part of smithay's public api right? So this is only safe, because all the accessors aren't?

In that case wouldn't it make more sense to wrap this in a way, where we don't expose the safety issues to the outside?

Something like:

```rust
pub struct Xkb(XkbInternal);

struct XkbInternal {
    context: xkb::Context,
    keymap: xkb::Keymap,
    state: xkb::State,
}

unsafe impl Send for XkbInternal {}
```

src/input/keyboard/mod.rs
```diff
@@ -477,13 +480,15 @@ pub struct XkbContext<'a> {
 
 impl<'a> XkbContext<'a> {
     /// Get the reference to the xkb state.
-    pub fn xkb(&self) -> &Xkb {
+    pub fn xkb(&self) -> &Mutex<Xkb> {
```
Drakulix, line 483: Would it be cleaner to already return a `MutexGuard`?

Drakulix (commented):
Seems sensible, this deadlock has been bugging me a lot.

src/input/keyboard/mod.rs
ids1024: I'm not sure if there's any general convention for this sort of thing. If so maybe name it `.lock_xkb()` so it's visible that this is holding onto a lock of something.

src/input/keyboard/mod.rs
Drakulix: I think we can leave it with accessors in that case.

src/input/keyboard/mod.rs
Drakulix: I don't mind either way too much to be honest. You can't do anything nasty with a read-only reference of a Mutex compared to a MutexGuard anyway.

ids1024: Then I think it can be left returning a `&Mutex<Xkb>`. I think that's good for now.

ids1024:
Hopefully we can improve things here more in the future, since none of this seems perfect, but the changes here at least clean up confusing behavior in the API around mutex locking (and the technical soundness issue of providing safe xkb accessors) without any significant change to the API.

Drakulix (approved):
Needs a rebase, but other than that, I'd say this is ready to go.

It is definitely an improvement over the current state of things.

PR #1535, opened 2024-09-16
x11: Use xinput2, and ignore key repetition

ids1024:
The winit backend (and libinput backend) already don't produce key repeat events like this.


Drakulix (approved):
Needs rebase

PR #1536, opened 2024-09-17
Use `Keycode` type consistently for representing keycode

ids1024:
It seems good to use a dedicated type for keycodes, particularly given there's two conventions for keycodes, offset by 8. So there's ambiguity about what the numberic value of the keycode actually represents.

We might prefer to use the Wayland value rather than the X11 version like `xkeysym::KeyCode` uses. But it would complicate things to have two different keycode types, and the code that calls `kbd.key` already is converting from `Keycode`.


Drakulix (approved):
Seems like a sensible idea. :+1:

PR #1537, opened 2024-09-18
ext_foreign_toplevel_list: Allow to retrieve resources

Drakulix:
This is useful for other protocols building on top of `ext_foreign_toplevel_list` to figure out, for which `ExtForeignToplevelHandleV1`s events need to be send.

### DIFF
--- a/src/wayland/foreign_toplevel_list/mod.rs
+++ b/src/wayland/foreign_toplevel_list/mod.rs
@@ -162,6 +162,26 @@ impl ForeignToplevelHandle {
         resource.data::<Self>().cloned()
     }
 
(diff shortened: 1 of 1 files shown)


PolyMeilex (approved):
Looks good, but why not expose all the resources while at it, so users can just filter whatever they need?

(Obviously, we can still have a `_for_client` helper as well on top of that)

Drakulix:
> Looks good, but why not expose all the resources while at it, so users can just filter whatever they need?
> 
> (Obviously, we can still have a `_for_client` helper as wel)

Yeah, that is probably right. Fixed it.

PolyMeilex (approved):

PR #1541, opened 2024-09-19
xdg-shell Fix definition of initial_decoration_configure_sent

PolyMeilex:
It looks like `initial_decoration_configure_sent` field ended up in both `popup` and `toplevel` states, which is wrong. Decoration configure refers only to toplevels


Drakulix (approved):
Makes sense. Ty!

PR #1542, opened 2024-09-20
xdg shell: Introduce `reset_initial_configure_sent()` 

PolyMeilex:
As far as I can tell, the xdg initial configure dance has to be done every time a xdg surface remaps itself.

So this MR:
- introduced a simple `reset_initial_configure_sent` helper method for xdg_surface, that allows to reset `initial_configure_sent`/`initial_decoration_configure_sent` to `false`.
- fixes remapping in anvil
- introduces a simple test client collection crate, for now it has a remap cycle client, that does the remapping in an endless loop.

---
Spec quote relating to this behavior:
```
For an xdg_surface to be mapped by the compositor, the following conditions must be met:
(1) the client has assigned an xdg_surface-based role to the surface
(2) the client has set and committed the xdg_surface state and the role-dependent state to the surface
(3) the client has committed a buffer to the surface

A newly-unmapped surface is considered to have met condition (1) out of the 3 required conditions
for mapping a surface if its role surface has not been destroyed,
i.e. the client must perform the initial commit again before attaching a buffer.
```

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -1547,6 +1547,8 @@ impl ToplevelSurface {
     ///
     /// This method is used to mark a surface for reinitialization.
     ///
+    /// NOTE: If you are using smithay's rendering abstractions you don't have to call this manually
+    ///
     /// Calls [`compositor::with_states`] internally.
     pub fn reset_initial_configure_sent(&self) {
         compositor::with_states(&self.wl_surface, |states| {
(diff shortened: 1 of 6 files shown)


YaLTeR:
Another test client: wleird-unmap

Drakulix:
Hmm, why can't we do this automatically inside smithay, when we get a `null`-buffer attached?

PolyMeilex:
> Hmm, why can't we do this automatically inside smithay, when we get a `null`-buffer attached?

Done, as long as the user uses Smithay's rendering abstractions this will now happen automatically. For any other case, the manual reset methods remain.

src/wayland/shell/xdg/mod.rs
```diff
@@ -1542,6 +1542,27 @@ impl ToplevelSurface {
         })
     }
 
+    /// A newly-unmapped toplevel surface has to perform the initial commit-configure sequence as if it was
+    /// a new toplevel.
+    ///
+    /// This method is used to mark a sufrace for reinitialization.
```
Drakulix, line 1548:
```suggestion
    /// This method is used to mark a surface for reinitialization.
```

Drakulix (approved):
Thanks!

PR #1543, opened 2024-09-20
stop deadlocking in `raw_latin_sym_or_raw_current_sym`

m4rch3n1ng:
since https://github.com/Smithay/smithay/commit/3293482ca42258fe2b04145dc15e4b7490feaa58 `KeysymHandle::raw_latin_sym_or_raw_current_sym` causes a deadlock unconditionally due to a double mutex lock.


ids1024 (approved):
Oops; good catch. This seems like a reasonable fix.

cmeissl (approved):
Thanks!

PR #1544, opened 2024-09-21
fix deadlock in AtomicDrmDevice

m4rch3n1ng:
https://github.com/Smithay/smithay/commit/88cdf9bd7d9c3550d16aa7b291486b9b4471e1c0 introduced a deadlock when creating AtomicDrmDevice, as https://github.com/Smithay/smithay/blob/88cdf9bd7d9c3550d16aa7b291486b9b4471e1c0/src/backend/drm/device/atomic.rs#L122 causes a deadlock in https://github.com/Smithay/smithay/blob/88cdf9bd7d9c3550d16aa7b291486b9b4471e1c0/src/backend/drm/device/atomic.rs#L185, since the `mapping` is dropped too late. this changes it to be dropped earlier.

### DIFF
--- a/src/backend/drm/device/atomic.rs
+++ b/src/backend/drm/device/atomic.rs
@@ -138,6 +138,8 @@ impl AtomicDrmDevice {
         dev.old_state = old_state;
         trace!("Mapping: {:#?}", mapping);
 
+        drop(mapping);
+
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Thanks!

PR #1545, opened 2024-09-22
multigpu: reset texture on surface buffer format change

cmeissl:
A surface itself does not have a format, it is implicitly defined by the currently attached buffer. So whenever we cache something that depends on that format we have to clear the cache if the format changes.

~~This might explain all kind of strange issues when going between direct scan-out and composition where it is not
unlikely that the format changes.~~ Might not happen so often as mesa does not implicitly change format afaik

@YaLTeR Thanks for reporting the detailed reproduction in `#smithay`


YaLTeR:
Fixes the issue in niri!

Drakulix (approved):
Seems very sensible!

PR #1546, opened 2024-09-22
bump msrv to 1.78

cmeissl:

PR #1548, opened 2024-09-25
shm: Use `OnceLock` instead of `static mut`

ids1024:
Fixes warning on nightly Rust.


ids1024:
I see https://github.com/Smithay/smithay/pull/1546 also has a fix for this.

But using `OnceLock` to avoid a `static mut` entirely is probably the cleaner solution, unless anyone sees a problem with this.

cmeissl (approved):
Yeah, this is a cleaner solution

PR #1550, opened 2024-09-25
shm: simplify sigbus init

Paraworker:
Since we use `OnceLock` for `OLD_SIGBUS_HANDLER`, the `SIGBUS_INIT` is redundant I think.


ids1024 (approved):
Ah yeah, using `get_or_init` (and ignoring the return value) seems like a reasonable way to simplify this.

Drakulix (approved):

Issue #1553, opened 2024-09-27
DPMS Support


ids1024:
https://github.com/pop-os/cosmic-comp/pull/741 has code setting the `DPMS` connector property to disable monitors. But the `DPMS` property is deprecated, not supported on atomic mode-setting, and for some reason setting it seems to take .5 to 1 second.

> Legacy property for setting the power state of the connector. For atomic drivers this is only provided for backwards compatibility with existing drivers, it remaps to controlling the “ACTIVE” property on the CRTC the connector is linked to. Drivers should never set this property directly, it is handled by the DRM core by calling the [drm_connector_funcs.dpms](https://www.kernel.org/doc/html/latest/gpu/drm-kms.html#c.drm_connector_funcs) callback. For atomic drivers the remapping to the “ACTIVE” property is implemented in the DRM core.
>
> Note that this property cannot be set through the MODE_ATOMIC ioctl, userspace must use “ACTIVE” on the CRTC instead.

https://www.kernel.org/doc/html/latest/gpu/drm-kms.html

I think to handle DPMS properly, it should be handled as part of the atomic commit within Smithay.

Apparently toggling the `ACTIVE` property of a CRTC should always succeed, if no other properties are changed. We *can* simply set `ACTIVE` to false, and then set it on again when we want to show things on screen again. (But not try to do more page flips with `ACTIVE` off, which errors.)

I think though that we also want to release buffers used in direct scanout? So probably also a commit removing planes...

The compositor will also want to stop sending frame callbacks or throttle their rate, however it normally handles surfaces that aren't visible. But that doesn't require Smithay changes.

I guess if I right about how this should work, and there aren't other considerations I'm missing, the question is how to fit this in the API of `DrmCompositor`...

ids1024:
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/drm/atomic.c?ref_type=heads#L412-472

Wlroots does seem to disable planes when it sets `ACTIVE` to false.

https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/backends/native/meta-kms-impl-device-atomic.c?ref_type=heads#L390-470

Mutter seems to set `MODE_ID` to `0` when it sets `ACTIVE` to `0`? I don't see anything handling `active` in `process_plane_assignment`, so I don't know if it's doing anything different with planes when active is false.

Drakulix:
I wonder if the easiest solution would be to just drop the `DrmSurface`.

[...]

PR #1561, opened 2024-10-10
Add a `disable` method to `DrmSurface`/`DrmCompositor` for implementing DPMS

ids1024:
As discussed in https://github.com/Smithay/smithay/issues/1553.

This seems to be working okay, on atomic or legacy DRM. I'm still finishing the cosmic-comp support.

Not sure if `DrmCompositor::disable` should clear any other state. Or if `AtomicDrmSurface::clear_state` does anything that is unneeded here. Other compositors appear to vary slightly in which properties their atomic backends set/unset for DPMS.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -4216,6 +4216,24 @@ where
             Err(Some(RenderingReason::ScanoutFailed))
         }
     }
+
+    /// Clear the surface, setting DPMS state to off, disabling all planes,
+    /// and clearing the pending frame.
+    ///
+    /// Calling [`queue_frame`][Self::queue_frame] will re-enable.
+    pub fn clear(&mut self) -> Result<(), DrmError> {
+        self.surface.clear()?;
+
+        self.current_frame
+            .planes
+            .iter_mut()
+            .for_each(|(_, state)| *state = Default::default());
+        self.pending_frame = None;
+        self.queued_frame = None;
+        self.next_frame = None;
+
+        Ok(())
+    }
 }
 
 #[inline]
--- a/src/backend/drm/surface/mod.rs
+++ b/src/backend/drm/surface/mod.rs
@@ -441,6 +441,17 @@ impl DrmSurface {
             DrmSurfaceInternal::Legacy(surf) => &surf.span,
         }
     }
+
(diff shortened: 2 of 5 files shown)


src/backend/drm/compositor/mod.rs
```diff
@@ -4216,6 +4216,24 @@ where
...
+    pub fn disable(&mut self) -> Result<(), DrmError> {
+        self.surface.disable()?;
+
+        self.current_frame
+            .planes
+            .iter_mut()
+            .for_each(|(_, state)| *state = Default::default());
+        self.pending_frame = None;
+        self.queued_frame = None;
+        self.next_frame = None;
```
Drakulix, line 4233: Ping @cmeissl

Does this look fine to you?

cmeissl:
I will do a review Sunday, latest Monday when I am back home.

src/backend/drm/surface/mod.rs
```diff
@@ -441,6 +441,17 @@ impl DrmSurface {
             DrmSurfaceInternal::Legacy(surf) => &surf.span,
         }
     }
+
+    /// Disable the output, setting DPMS state to off.
```
cmeissl, line 445: imo the name disable might be misleading, or at lest the comment might be. it does not only set the dpms state, but also clears the whole state including all planes.

src/backend/drm/compositor/mod.rs
cmeissl: This depends on how it should behave, this should be fine for disabling alone, yes.
We could also reset the element states to drop all framebuffers we imported, this might allow to lower
the memory footprint. Same for the swapchain which could be reset.

cmeissl (commented):
The whole thing looks more like a reset/clear to me. Disable suggests that the surface will be disabled until some non-existing enable imo.
I would expect it to track and handle the disable state internally, potentially returning errors for some functions like `render_frame`.
In its current state tracking still has to be done externally, which is fine imo, but maybe not expected from the name disable.

ids1024:
I've renamed the function to `clear` and changed the doc comment slightly. Hopefully that's a bit better.

cmeissl (approved):
LGTM thanks!

PR #1563, opened 2024-10-18
WIP `fifo` and `commit-timing` protocols

ids1024:
This doesn't yet implement the correct behavior; it just has the boilerplate to expose the protocol.

I've been meaning to see what implementing `fifo` would take. And now it's provided in `wayland-protocols`. I guess it basically amounts to having a `Blocker` for a transaction that delays a transaction with a  `wait_barrier` until the previous transaction with `set_barrier` has passed? I think that should be easy...

For `commit_timing`, I guess we need a `Blocker` that just checks if a certain monotonic time has passed yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,3 +140,10 @@ harness = false
 [profile.release-with-debug]
 inherits = "release"
 debug = true
+
+[patch.crates-io]
+wayland-protocols = { git = "https://github.com/Smithay/wayland-rs" }
+wayland-client = { git = "https://github.com/Smithay/wayland-rs" }
+wayland-server = { git = "https://github.com/Smithay/wayland-rs" }
+wayland-backend = { git = "https://github.com/Smithay/wayland-rs" }
+wayland-sys = { git = "https://github.com/Smithay/wayland-rs" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ udev = { version = "0.9.0", optional = true }
 wayland-client = { version = "0.31.3", optional = true }
 wayland-cursor = { version = "0.31.3", optional = true }
 wayland-egl = { version = "0.32.0", optional = true }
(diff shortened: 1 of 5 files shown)


YaLTeR:
Not sure if it's for this PR, but there should be some better way to compose blockers. In niri I have a dmabuf blocker and a tiling transaction blocker, and I had to manually write code to combine them correctly in my pre commit hook. Things like, signal the surface's transaction readiness, but only after the dmabuf blocker completes if there is one, but make sure not to unblock the surface itself until the rest of the transaction is ready. Extra blockers like fifo or commit timing need to somehow be joined with the dmabuf blocker in a way that doesn't cause a combinatorial explosion in the code.

ids1024:
I guess `PresentationState::new` takes a clock id, and we'll need to use the same clock. (Or remove that argument, and say that `CLOCK_MONOTONIC` is always used here in Smithay, if there's no reason to do otherwise.)

> Extra blockers like fifo or commit timing need to somehow be joined with the dmabuf blocker in a way that doesn't cause a combinatorial explosion in the code.

I'm not sure I understand. `add_blocker` pushes the blocker to a vector of blockers associated with the transaction, which `Transaction::state` fold into a single `BlockerState`. So I think it's supposed to do the right thing if this just calls `add_blocker` an additional time?

YaLTeR:
I believe the thing I was missing is a place to run code after some blockers are done but still before the commit. For a niri tiling transaction (not the smithay internal commit transaction), I need to signal the surface's readiness once all other blockers are done on the surface, but the tiling transaction itself should keep blocking the surface until all other surfaces in the transaction are ready. Currently I would have to manually watch each of the blockers for completion to signal the surface readiness once they're all done.

So maybe what would work is a method that smithay calls after all blockers clear out, but from that method you can add some more blockers and keep waiting. In this case, I could first add dmabuf, fifo, commit timing blockers, then inside that new method I would signal the surface readiness in its tiling transaction and add the tiling transaction itself as a blocker to keep waiting for it to complete.

PR #1565, opened 2024-10-20
output: fix scale clamping

### LINKED ISSUE #1559: Regression in dynamic buffer scale changes
Introduction of 9a7afcf7b107992182d4998d198412c68b49bf3c has broken the ability to manually set the preferred buffer scale for my surfaces. If I change a surface from scale 2 to 1, then back to 2, it will stay at 1 forever.

cmeissl:
we want to make sure the output integer scale
is always >= 1 after applying the client scale.
using min will do the opposite, so change to max
which will make sure we always return at least 1
or whatever scale we calculated

fixes #1559

### DIFF
--- a/src/wayland/output/mod.rs
+++ b/src/wayland/output/mod.rs
@@ -254,7 +254,7 @@ impl Output {
                 inner.send_geometry_to(&output);
             }
             if (new_scale.is_some() || scale_changed) && output.version() >= 2 {
-                let scale = (inner.scale.integer_scale() / client_scale as i32).min(1);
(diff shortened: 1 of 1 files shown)


YaLTeR:
Seems to fix the issue that I observed.

cmeissl:
@chrisduerr does this also fix the regression for you?

PolyMeilex (approved):

chrisduerr:
> @chrisduerr does this also fix the regression for you?

I'm currently a bit busy with other stuff, but I can give it a test. Feel free to just close the issue and I'll let you know if it wasn't fixed.

Drakulix (approved):
Thanks!

PR #1566, opened 2024-10-24
xdg_system_bell: Implement v1 of xdg system bell

PolyMeilex:


ids1024:
I really hope at some point something like https://github.com/Smithay/wayland-rs/pull/699 can let us have less boilerplate for a protocol that provides a single global with a single request...

But this looks good.

Drakulix (approved):

PR #1567, opened 2024-10-24
Add Pointer/Touch/KeyboardHandle::with_grab() and downcasting

YaLTeR:
Allows the compositor to inspect which grab is currently active.

Use case 1: in niri I want to scroll the workspace left and right when hovering the mouse close to output edges with a DnD grab, but not with a regular click grab.

Use case 2: avoiding a weird case with libadwaita DnD/move conflict https://gitlab.gnome.org/GNOME/gtk/-/issues/7113 by checking if the active grab is a DnD grab specifically and denying the move request in that case.

```rust
let mut is_dnd_grab = false;
pointer.with_grab(|_, grab| {
    is_dnd_grab = grab.as_any().downcast_ref::<DnDGrab<Self>>().is_some();
});
```

Debug impls are auto-generated with rust-analyzer (they want `D: 'static`).

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -2,6 +2,7 @@
 
 use crate::backend::input::KeyState;
 use crate::utils::{IsAlive, Serial, SERIAL_COUNTER};
+use downcast_rs::{impl_downcast, Downcast};
 use std::collections::HashSet;
 #[cfg(feature = "wayland_frontend")]
 use std::sync::RwLock;
@@ -590,7 +591,7 @@ impl<D: SeatHandler + 'static> Clone for GrabStartData<D> {
 /// When your grab ends (either as you requested it or if it was forcefully cancelled by the server),
 /// the struct implementing this trait will be dropped. As such you should put clean-up logic in the destructor,
 /// rather than trying to guess when the grab will end.
-pub trait KeyboardGrab<D: SeatHandler> {
+pub trait KeyboardGrab<D: SeatHandler>: Downcast {
     /// An input was reported.
     ///
     /// `modifiers` are only passed when their state actually changes. The modifier must be
@@ -622,6 +623,8 @@ pub trait KeyboardGrab<D: SeatHandler> {
     /// The grab has been unset or replaced with another grab.
     fn unset(&mut self, data: &mut D);
 }
+
+impl_downcast!(KeyboardGrab<D> where D: SeatHandler);
 
 /// An handle to a keyboard handler
 ///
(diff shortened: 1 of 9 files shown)


src/input/keyboard/mod.rs
```diff
@@ -918,6 +921,14 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
         }
     }
 
+    /// Calls `f` with the active grab, if any.
+    pub fn with_grab(&self, f: impl FnOnce(Serial, &dyn KeyboardGrab<D>)) {
+        let guard = self.arc.internal.lock().unwrap();
+        if let GrabStatus::Active(s, g) = &guard.grab {
+            f(*s, &**g)
+        }
+    }
```
cmeissl, line 930: From the example it might make sense to have the function return something maybe?

Something like:

```rs
    pub fn with_grab<R>(&self, f: impl FnOnce(Serial, &dyn KeyboardGrab<D>) -> Option<R>) {
        let guard = self.arc.internal.lock().unwrap();
        if let GrabStatus::Active(s, g) = &guard.grab {
            Some(f(*s, &**g))
        } else {
           None
        }
    }
```

Could then make the example a bit more convenient: 

```rs
let mut is_dnd_grab = 
pointer.with_grab(|_, grab| {
    grab.as_any().downcast_ref::<DnDGrab<Self>>().is_some()
}).unwrap_or(false);
```

cmeissl (commented):
If the main goal is to just be able check if a certain grab is active it might just make sense to add something like `has_grab<T>(&self) -> bool`?

YaLTeR:
> If the main goal is to just be able check if a certain grab is active it might just make sense to add something like `has_grab<T>(&self) -> bool`?

I figured I might as well just add the full function if a use case turns up later on.

src/input/keyboard/mod.rs
YaLTeR: I was debating between this and the simpler one in the PR, kind of 50/50 between them.

cmeissl: Afaik most function in smithay that operate on a closure allow to return something, so I would prefer we add it here too.

cmeissl:
> > If the main goal is to just be able check if a certain grab is active it might just make sense to add something like `has_grab<T>(&self) -> bool`?
> 
> I figured I might as well just add the full function if a use case turns up later on.

Not sure there is a use-case where you would need access to the `dyn Trait` object. I guess in most cases you either
want to know the specific grab that is active and maybe some info from it. So maybe with_grab could already take the
expected type?

(Just some idea, I am also fine with accessing the dyn Trait)

YaLTeR:
Well, with dyn Trait you can check if a grab is active and if it's a particular grab type in one go more or less.

src/input/keyboard/mod.rs
YaLTeR: Did that

Drakulix (approved):
Seems good to me

Issue #1552, opened 2024-09-27
wayland.pointer_constrains: Potential reference cycles


PolyMeilex:
I don't have the time to investigate this closely atm, so I'll post this as a future reminder.

Those two look fishy:

```mermaid
flowchart TD
    ZwpLockedPointerV1 --> PointerConstraintUserData --> WlSurface
    WlSurface --> PointerConstraintData --> LockedPointer --> ZwpLockedPointerV1
```
```mermaid
flowchart TD
    ZwpConfinedPointerV1 --> PointerConstraintUserData --> WlSurface
    WlSurface --> PointerConstraintData --> ConfinedPointer --> ZwpConfinedPointerV1
```

PolyMeilex:
Checked closer, this one looks fine, we are calling `remove_constraint` on object destroy, so the reference cycle is broken as soon as the constraint is destroyed.

PR #1568, opened 2024-10-24
wayland.pointer_constraints: Remove constraints in the destructor

PolyMeilex:
I'm pretty sure this works fine as is, but just to be 100% sure that the reference cycle (see  #1552) gets broken properly, let's remove the constraint in the destructor rather than on event.

`remove_constraint` removes the constraint object from `wlsurface` user data, and as the constraint object contains the surface in its own user data this makes sure that the cycle is broken properly.

### DIFF
--- a/src/wayland/pointer_constraints.rs
+++ b/src/wayland/pointer_constraints.rs
@@ -482,11 +482,22 @@ where
                     }
                 });
             }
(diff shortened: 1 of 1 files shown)


Drakulix (approved):

Drakulix:
CI doesn't seem to be happy

PR #1571, opened 2024-10-28
Fix initial configure reset logic

### LINKED ISSUE #1570: Popup positioning regression
de5ad9eac90e250cc938025f2b1ea19ce738a059 regresses popups. Starting from that commit they appear at (0, 0) of the window geometry for one frame. In case of GTK 3 tooltips, they remain at (0, 0) indefinitely.

To test: run gtk4-demo in anvil, right click on text to open a popup menu, notice it renders at window geometry (0, 0) for one frame.

cc @PolyMeilex

PolyMeilex:
fixes #1570

The `initial_configure_sent` should be reset only for surfaces that are already mapped, to prevent interference with the initial commit dance.

@YaLTeR Can you check if this fixes the problem in niri as well?

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -205,6 +205,8 @@ macro_rules! xdg_role {
                 pub last_acked: Option<$state>,
                 /// Holds the current state after a successful commit.
                 pub current: $state,
+                /// Does the surface have a buffer (updated on every commit)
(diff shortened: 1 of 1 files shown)


YaLTeR:
Appears to work fine for GTK 3, 4 and layer-shell GTK 3.

cmeissl (approved):

PolyMeilex:
Thanks!

PR #1576, opened 2024-11-03
DrmOutputManager

cmeissl:
highly wip of the ideas described here: https://github.com/pop-os/cosmic-comp/pull/969#issuecomment-2450833400

it implements most of the stuff, except:
- the capabilities stuff
- the actual format selection
- some things are stubs and need to be implemented properly

not extensively tested, but anvil seems to still be able to launch.
returning an error from `DrmCompositor::new` also successfully triggers the format selection
logic

## TODO

A lot...

- [x] Verify that this does not mess up the render logic with multiple commits triggering vblanks
       If this turns out to be problematic we can disable the event for `commit_frame`, but we need to make sure
       to kick off rendering in this case somehow..

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -2429,6 +2429,41 @@ where
         Ok(())
     }
 
+    pub fn commit_frame(&mut self) -> FrameResult<(), A, F> {
+        if !self.surface.is_active() {
+            return Err(FrameErrorType::<A, F>::DrmError(DrmError::DeviceInactive));
+        }
+
+        let mut prepared_frame = self.next_frame.take().ok_or(FrameErrorType::<A, F>::EmptyFrame)?;
+        if prepared_frame.is_empty() {
+            return Err(FrameErrorType::<A, F>::EmptyFrame);
+        }
+
+        if let Some(plane_state) = prepared_frame.frame.plane_state(self.surface.plane()) {
+            if !plane_state.skip {
+                let slot = plane_state.buffer().and_then(|config| match &config.buffer {
+                    ScanoutBuffer::Swapchain(slot) => Some(slot),
+                    _ => None,
+                });
+
+                if let Some(slot) = slot {
+                    self.swapchain.submitted(slot);
+                }
+            }
+        }
+
+        let flip = prepared_frame
+            .frame
+            .commit(&self.surface, self.supports_fencing, false, false);
+
+        if flip.is_ok() {
+            self.queued_frame = None;
+            self.pending_frame = None;
+        }
+
+        self.handle_flip(prepared_frame, None, flip)
+    }
+
     /// Re-evaluates the current state of the crtc and forces calls to [`render_frame`](DrmCompositor::render_frame)
     /// to return `false` for [`RenderFrameResult::is_empty`] until a frame is queued with [`queue_frame`](DrmCompositor::queue_frame).
     ///
@@ -2462,13 +2497,22 @@ where
                 .page_flip(&self.surface, self.supports_fencing, allow_partial_update, true)
         };
 
+        self.handle_flip(prepared_frame, Some(user_data), flip)
+    }
+
+    fn handle_flip(
+        &mut self,
+        prepared_frame: PreparedFrame<A, F>,
+        user_data: Option<U>,
+        flip: Result<(), crate::backend::drm::error::Error>,
+    ) -> FrameResult<(), A, F> {
         match flip {
             Ok(_) => {
                 if prepared_frame.kind == PreparedFrameKind::Full {
                     self.reset_pending = false;
                 }
 
-                self.pending_frame = Some(PendingFrame {
+                self.pending_frame = user_data.map(|user_data| PendingFrame {
                     frame: prepared_frame.frame,
                     user_data,
                 });
@@ -4295,4 +4339,17 @@ fn nvidia_drm_version() -> Option<(u32, u32, u32)> {
     let minor = u32::from_str(components.next()?).ok()?;
     let patch = u32::from_str(components.next()?).ok()?;
     Some((major, minor, patch))
+}
+
+#[test]
+fn drm_compositor_is_send() {
+    use std::marker::PhantomData;
+
+    use crate::backend::drm::DrmDeviceFd;
+
+    fn is_send<T: Send>() {
+        let _ = PhantomData::<T>;
+    }
+
+    is_send::<DrmCompositor<GbmAllocator<DrmDeviceFd>, GbmDevice<DrmDeviceFd>, (), DrmDeviceFd>>();
 }
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -1288,6 +1288,7 @@ where
     /// - `modifiers` is the set of modifiers allowed, when allocating buffers with the specified color format
     /// - `cursor_size` as reported by the drm device, used for creating buffer for the cursor plane
     /// - `gbm` device used for creating buffers for the cursor plane, `None` will disable the cursor plane
+    #[allow(clippy::too_many_arguments)]
     pub fn with_format(
         output_mode_source: impl Into<OutputModeSource> + Debug,
         surface: DrmSurface,
@@ -1514,10 +1515,9 @@ where
             Ok(None) => return Err((swapchain.allocator, FrameError::NoFramebuffer)),
             Err(err) => return Err((swapchain.allocator, FrameError::FramebufferExport(err))),
         };
-        buffer.userdata().insert_if_missing(|| {
-            let init = CachedDrmFramebuffer::new(DrmFramebuffer::Exporter(fb_buffer));
-            init
-        });
+        buffer
+            .userdata()
+            .insert_if_missing(|| CachedDrmFramebuffer::new(DrmFramebuffer::Exporter(fb_buffer)));
 
         let mode = drm.pending_mode();
         let handle = buffer
@@ -1559,7 +1559,7 @@ where
             }),
         };
 
-        match current_frame_state.test_state(&drm, supports_fencing, drm.plane(), plane_state, true) {
+        match current_frame_state.test_state(drm, supports_fencing, drm.plane(), plane_state, true) {
             Ok(_) => Ok((swapchain, use_opaque)),
             Err(err) => {
                 warn!(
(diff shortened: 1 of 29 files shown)


src/backend/drm/output.rs
```diff
@@ -0,0 +1,459 @@
...
+use super::{
+    compositor::{
+        DrmCompositor, FrameError, FrameMode, FrameResult, RenderFrameError, RenderFrameErrorType,
+        RenderFrameResult,
+    },
+    exporter::ExportFramebuffer,
+    DrmDevice, DrmError, Planes,
+};
+
+type CompositorList<A, F, U, G> = Arc<RwLock<HashMap<crtc::Handle, RefCell<DrmCompositor<A, F, U, G>>>>>;
```
Paraworker, line 32: If i understand correctly, `RefCell` is `!Sync`, which would make `CompositorList` both `!Sync` and `!Send`, making `Arc<RwLock<>>` pointless.

cmeissl: Indeed, sigh. `!Sync` is intended/expected, it should be `Send` , I guess this is blocked on the `PixmanRenderer` not being `Send` :(

Paraworker: Sorry I'm a little confused. `Arc<T>`  is `Sync` or `Send` requires `T` to be both `Sync` and `Send`, but `RefCell<T>` is unconditional `!Sync`. I think `RefCell<T>` should be removed here.

Drakulix: `RwLock` or `Mutex` add the `Sync` bound, so our `T` is just required to be `Send`, which is indeed blocked on pixman-rs

cmeissl: Looking at it again we indeed need to use Mutex instead of the Refcell because RwLock requires Sync for read access.

This is the easy part, unfortunately DrmCompositor is not Send atm. Currently looking into this.

src/backend/drm/compositor/mod.rs
```diff
@@ -1561,6 +1041,22 @@ where
...
+    pub struct FrameMode: u32 {
+        /// Allow to realize the frame by compositing elements on the primary plane
+        const COMPOSITE = 1;
+        const PRIMARY_PLANE_SCANOUT = 2;
+        const OVERLAY_PLANE_SCANOUT = 4;
+        const CURSOR_PLANE_SCANOUT = 8;
+        /// Allow to realize the frame by assigning elements on planes
+        const SCANOUT = Self::PRIMARY_PLANE_SCANOUT.bits() | Self::OVERLAY_PLANE_SCANOUT.bits() | Self::CURSOR_PLANE_SCANOUT.bits();
+
+        const ALL = Self::COMPOSITE.bits() | Self::SCANOUT.bits();
```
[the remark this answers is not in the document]
cmeissl: tbh I don't think this will fit in the current DrmCompositor API. This will require allowing custom assign algorithms. What I have in mind is that we make something similar to the FrameState public and actually externalize the whole assigning algorithm.

src/backend/drm/output.rs
```diff
@@ -0,0 +1,459 @@
...
+                    let (elements, clear_color) = render_elements(handle);
+                    compositor.reset_buffer_ages();
+                    compositor
+                        .render_frame(renderer, elements, clear_color, FrameMode::COMPOSITE)
+                        .map_err(DrmOutputManagerError::RenderFrame)?;
+                    compositor.commit_frame().map_err(DrmOutputManagerError::Frame)?;
+
+                    let current_format = compositor.format();
+                    if let Err(err) =
+                        compositor.set_format(self.allocator.clone(), current_format, [DrmModifier::Invalid])
```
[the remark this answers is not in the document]
cmeissl: Sure, this is more to provide a PoC

src/backend/drm/output.rs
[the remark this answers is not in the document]
cmeissl: Not primarily, but this was a nice side effect and is already used in use_mode.

The main reason was faking atomic access when changing the active outputs. So basically whenever a output is added or removed. The second was ownership.

We need to be able to access all active crtc while making sure we have unique access. Using Arc and cloning it around makes it hard to track aliveness of the output. We could store a Weak in the output manager, but tbh I really want to avoid that. Working with Weak is a pain sometimes. Also it would require locking all outputs before access, making this unnecessary complicated. The idea was that RwLock gives us that for free, concurrent read access and unique write access. Also it makes it easier to reason about who owns the outputs imo.

src/backend/drm/compositor/mod.rs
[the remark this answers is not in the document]
Drakulix: Sounds good. (And is also not a very important problem atm.)

src/backend/drm/output.rs
cmeissl: Okay, `DrmCompositor` is now `Send` as long as the allocator buffer and the framebuffer are `Send + Sync`. This allows `DrmOutputManager` to be `Send` and `DrmOutput` to implement `Send` and `Sync`.

## Notes
- [ ] This currently needs `unsafe impl Sync` for `GbmBuffer` which should be removed when [this](https://github.com/Smithay/gbm.rs/pull/45) lands
- [ ] The unsafe `Sync` impl on `PixmanRenderer` should be fine atm, but we need to address this in [`pixman-rs`](https://github.com/cmeissl/pixman-rs/pull/18)

Paraworker:
Hey guys I have an idea about getting rid of the `Mutex` in `CompositorList`.

We access `DrmCompositor` through two ways, under `RwLock::write()` or `RwLock::read()`, let me discuss separately:

### RwLock::write() ###
In this case, we can be sure that access is unique, `Mutex` is redundant here.

### RwLock::read() ###
This is the case we really need to care about. From the code I can see `RwLock::read()` only happens in `DrmOutput::with_compositor()`. If we can mark the `DrmOutput` with `!Sync` or mark `DrmOutput::with_compositor()` with mut, then no race condition exists since `DrmOutput` only access it's own `DrmCompositor` under `RwLock::read()`.

While the `Mutex` is not need, we do still need interior mutability. Something like `DrmCompositorCell` can be considered, it wraps an `UnsafeCell` but provide `Sync`:

```
struct DrmCompositorCell<A, F, U, G> {
    inner: UnsafeCell<DrmCompositor<A, F, U, G>>,
}

impl<A, F, U, G> DrmCompositorCell<A, F, U, G> {
    fn new(value: DrmCompositor<A, F, U, G>) -> Self {
        Self { inner: UnsafeCell::new(value) }
    }

    /// SAFETY: make sure unique access
    unsafe fn get(&self) -> &mut DrmCompositor<A, F, U, G> {
        &mut *self.inner.get()
    }
}

unsafe impl<A, F, U, G> Sync for DrmCompositorCell<A, F, U, G> {}
```

Then the `CompositorList` would be like:

```
type CompositorList<A, F, U, G> = Arc<RwLock<HashMap<crtc::Handle, DrmCompositorCell<A, F, U, G>>>>;
```

All these are just theoretical and may be too tricky😮‍💨. I don't know how you guys think about it.

src/backend/drm/output.rs
Drakulix: gbm 0.16.1 published

Drakulix:
> ## TODO
> 
> A lot...
> 
>     * [ ]  Verify that this does not mess up the render logic with multiple commits triggering vblanks
>       If this turns out to be problematic we can disable the event for `commit_frame`, but we need to make sure
>       to kick off rendering in this case somehow..

Addressed in https://github.com/Smithay/smithay/pull/1576/commits/3fe19dd794900c04f70b27ec557e4a841af1fc5b. While it works okay-ish for anvil, it definitely would mess up state tracking in niri and cosmic-comp and while potentially work-arounds exists, at least in the latter with multi-threading this requires ugly synchronization, while the point of this api is to hide that.

Drakulix:
All the code paths seem to now work pretty nicely in https://github.com/pop-os/cosmic-comp/pull/1020 (though I don't have any devices that actually have to use these paths, I just forced them manually). This just needs a new pixman-rs release to pass tests.

src/backend/drm/compositor/mod.rs
```diff
@@ -3152,6 +2930,21 @@ where
             true,
         )?;
 
+        if let ScanoutBuffer::Swapchain(slot) = &frame_state
+            .plane_buffer(self.surface.plane())
+            .expect("We have a buffer for the primary plane")
+            .buffer
+        {
+            if slot.format() != element_config.properties.format {
```
cmeissl, line 2938: This needs to be configurable from the outside, I am just not sure how and in what detail. Having this restriction will kill performance on lower end hardware. In my case I would only enable this on general purpose x68 hardware and only when more than a single output is enabled.

src/backend/drm/output.rs
```diff
@@ -0,0 +1,908 @@
...
+}
+
+impl<R, E> Default for DrmOutputRenderElements<R, E>
+where
+    E: RenderElement<R>,
+    R: Renderer + Bind<Dmabuf>,
+    <R as Renderer>::TextureId: Texture + 'static,
+    <R as Renderer>::Error: Send + Sync + 'static,
+{
+    fn default() -> Self {
```
cmeissl, line 850: While not super important we should consider adding a `with_capacity` function allowing to pre-allocate the intern hash map.

cmeissl (commented):
So, in general I believe the changes, except the one limiting the primary format, are pretty harmless. The `DrmOutputManager` is a really nice optional addition which will definitely need some more work in the future. But I feel like this provides a solid base to do so. I am really happy how this turned out, thanks!

I really like the approach with `DrmOutputRenderElements`, nice work :)

cmeissl:
pixman-rs 0.2.1 is out

src/backend/drm/compositor/mod.rs
Drakulix: Hmm, maybe just have this be another bit for `FrameMode`?  We might also have a need for a new flag [here](https://github.com/Smithay/smithay/issues/1610#issuecomment-2546155469). So maybe `FrameMode` should become `FrameFlags` and have something like `ALLOW_PRIMARY_SCANOUT` instead of `PRIMARY_SCANOUT.

Drakulix: I have pushed a commit doing this: https://github.com/Smithay/smithay/pull/1576/commits/54379b2e86ed3fbb1b55127bca658706e3128fbe

cmeissl: looks fine, thanks!

cmeissl (commented):
LGTM

PR #1577, opened 2024-11-03
wayland.selection: Introduce ext data control protocol

PolyMeilex:
- `wayland.selection: ExtDataControl` Adds support for the new ext data control protocol
- `wayland.selection: Introduce DataDeviceKind` Removes `TypeId` in favor of `DataDeviceKind` enum

### DIFF
--- a/src/wayland/selection/device.rs
+++ b/src/wayland/selection/device.rs
@@ -8,7 +8,6 @@ use wayland_server::Resource;
 
 use super::data_device::DataDeviceUserData;
 use super::ext_data_control::ExtDataControlDeviceUserData;
-use super::offer::DataControlOffer;
 use super::offer::SelectionOffer;
 use super::primary_selection::PrimaryDeviceUserData;
 use super::private::selection_dispatch;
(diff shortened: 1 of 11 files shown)


ids1024:
`wayland-protocols` 0.32.6 is now released on crates.io.

PolyMeilex:
This is now ready.

src/wayland/selection/offer.rs
```diff
@@ -55,9 +55,27 @@ impl<U: Clone + Send + Sync + 'static> OfferReplySource<U> {
...
+    /// Advertise offered MIME type
+    ///
+    /// Sent immediately after creating the wlr_data_control_offer object. One event per offered MIME type.
+    fn offer(&self, mime_type: String) {
+        match self {
+            Self::Wlr(obj) => obj.offer(mime_type),
+            Self::Ext(obj) => obj.offer(mime_type),
+        }
+    }
+}
```
kchibisov, line 74: I feel like it shouldn't be in this file, but rather in the place where you wrap data controls.

src/wayland/selection/device.rs
```diff
@@ -9,16 +7,89 @@ use wayland_server::protocol::wl_seat::WlSeat;
...
+        }
+    }
+
+    fn version(&self) -> u32 {
+        match self {
+            Self::Wlr(obj) => obj.version(),
+            Self::Ext(obj) => obj.version(),
+        }
+    }
+}
```
kchibisov, line 86: I'd probably move it somewhere else to data_control only related file. Though, as alternative, you could write it with `WlrDataControl` and `ExtDataControl` as top-level types, would be much easier on a review, since the code will be pretty much duplicated.

And in fact, it's already pretty much duplicated, so really seems like there's no point in having a wrapper? Though, I could be wrong.

src/wayland/selection/seat_data.rs
```diff
@@ -4,14 +4,14 @@ use wayland_server::{Client, DisplayHandle, Resource};
...
+use super::device::{DataControlDevice, SelectionDevice};
 use super::offer::{OfferReplySource, SelectionOffer};
 use super::{SelectionHandler, SelectionTarget};
 
 /// Seat data used to handle regular selection operations.
 ///
-/// The data is shared accross primary, data device, and data control selections.
-pub struct SeatData<U: Clone + Sync + Send + 'static> {
+/// The data is shared across primary, data device, and data control selections.
+pub(crate) struct SeatData<U: Clone + Sync + Send + 'static> {
```
kchibisov, line 14: I think it's a part of the public API? At least all the `Data` was usually a public API.

kchibisov (commented):
Would assume that the implementation of request handling was copy pasted 1 to 1 from the Wlr one, with the exception of removed version checks.

src/wayland/selection/seat_data.rs
PolyMeilex: The `seat_data` module is private, and the type is not reexported. So I just marked it as pub(crate) to make that fact more clear.

src/wayland/selection/offer.rs
PolyMeilex: Wrappers removed.

src/wayland/selection/device.rs
PolyMeilex: Wrappers removed.

Drakulix (approved):
LGTM as well :)

PR #1578, opened 2024-11-08
drm: Add vrr property helpers

Drakulix:
Draft while this is missing docs and more testing.

Adds helper methods to make it possible to set VRR per frame, which makes more dynamic VRR policies like "only enable, when we have a fullscreen application" more easily possible. Also simplifies VRR checks.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -172,7 +172,10 @@ use crate::{
     wayland::{shm, single_pixel_buffer},
 };
 
-use super::{error::AccessError, DrmDeviceFd, DrmSurface, Framebuffer, PlaneClaim, PlaneInfo, Planes};
+use super::{
+    error::AccessError, surface::VrrSupport, DrmDeviceFd, DrmSurface, Framebuffer, PlaneClaim, PlaneInfo,
+    Planes,
+};
 
 pub mod dumb;
 mod elements;
@@ -2662,6 +2665,24 @@ where
             PrimaryPlaneElement::Element(primary_plane_scanout_element.unwrap())
         };
 
+        // if the update only contains a cursor position update, skip it for vrr
+        if self.vrr_enabled()
+            && allow_partial_update
+            && next_frame_state.planes.iter().all(|(plane, state)| {
+                state.skip
+                    || (self.planes.cursor.iter().any(|p| *plane == p.handle)
+                        && state.buffer().map(|b| &b.fb)
+                            == previous_state.plane_buffer(*plane).map(|b| &b.fb))
+            })
+        {
+            for plane in self.planes.cursor.iter() {
+                let Some(state) = next_frame_state.plane_state_mut(plane.handle) else {
+                    continue;
+                };
+                state.skip = true;
+            }
+        }
+
         let next_frame = PreparedFrame {
             kind: if allow_partial_update {
                 PreparedFrameKind::Partial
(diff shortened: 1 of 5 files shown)


Drakulix:
Now in use here and seems to work well: https://github.com/pop-os/cosmic-comp/pull/1004

cmeissl (commented):
wp_presentation version 1 requires 0 to be sent for vrr

> If the output does not have a constant refresh rate, explicit video mode switches excluded, then the refresh argument must be zero.

src/backend/drm/compositor/mod.rs
```diff
@@ -2357,6 +2362,17 @@ where
             }
         }
 
+        // If we only have a cursor update and vrr is enabled,
```
cmeissl, line 2365: I don't understand the comment here, what does update refer to here? This would mean that there is nothing shown except a cursor. Is this a realistic scenario?

src/backend/drm/compositor/mod.rs
```diff
@@ -2930,6 +2946,26 @@ where
         Ok(())
     }
 
+    /// Returns if Variable Refresh Rate is advertised as supported by the given connector.
+    ///
+    /// See [`DrmSurface::vrr_supported`] for more details.
+    pub fn vrr_supported(&self, conn: connector::Handle) -> FrameResult<VrrSupport, A, F> {
```
cmeissl, line 2952: Maybe we can add a `vrr_enabled` function, returning the current vrr state. Could be useful when sending presentation feedback.

src/backend/drm/compositor/mod.rs
Drakulix: The case here being "nothing is being updated, but the cursor". Shouldn't we already have filtered out elements without any damage?

The main use-case here is a single fullscreen surface + a hardware cursor. Moving the cursor shouldn't dictate the frame rate in that case.

src/backend/drm/compositor/mod.rs
Drakulix: I would expect reading back the property on every frame to be quite slow and instead expect compositors to track the state themselves. Is that a wrong assumption?

Drakulix: > But that actually made me question how we would handle wp-presentation v2, which requires us to send the minimum refresh.

Oh right, good idea.

> But that actually made me question how we would handle wp-presentation v2, which requires us to send the minimum refresh. So maybe it already has to be tracked somewhere else?

Maybe we should just merge #1588, so I can more easily experiment with this. But it might make sense to add this to the `RenderFrameResult`. Compositors would then need to pass this through the `queue_frame` user-data to have access to this after the vblank-event.

src/backend/drm/compositor/mod.rs
cmeissl: What happens is that we build a new state and while we do this we compare with the previous one, including the commit id for direct scan-out. If nothing changed we still build the state, but skip the plane.

Composition on the primary plane might complicate things a bit, depending on how we handle the cursor update skip. We only know after calling into the damage tracked rendered if something has been drawn.

I haven't checked, but it should work to put the logic to after rendering to the primary plane and evaluate if only the cursor plane is not skipped, in which case we can override it to skip. This should automatically put our frame to "empty".

Though we either have to update the current frame to not loose the rendered cursor buffer and re-draw it every call to render_frame. Or just not skip if the cursor plane has been rendered to, which might be way easier imo.
This should be doable by comparing the buffer on the cursor plane with the previous frame.

cmeissl: We only re-draw the cursor plane if absolutely necessary, so it should not happen often and already exclude movement alone. So maybe it would be okay to not skip in this case? Otherwise I fear it might be too easy to end in an unexpected cursor redraw-loop.

src/backend/drm/compositor/mod.rs
cmeissl: Yeah, sounds good. I will it merge it now :)

cmeissl: The min refresh rate is defined in the monitor EDID iirc. libdisplayinfo should already expose it.

Drakulix: > For version 2 and later, if the output does not have a constant refresh rate, explicit video mode switches excluded, then the refresh argument must be either an appropriate rate picked by the compositor (e.g. fastest rate), or 0 if no such rate exists.

Doesn't that mean the value should simply be that of the `Mode`?

cmeissl:
> And I added an example to anvil for sending the right PresentationTime refresh rate, though I would prefer to drop that commit given anvil doesn't implement VRR at all. It is just there to show off how easy this is to do with the `vrr_enabled`-method. I thought about adding it to the `FrameResult` (as we only really know the refresh rate mode we send off to the hardware on `submit`, but making the into a proper struct with meta data felt overkill, when the check is this simple.

Looks goods and I agree that we can remove that from anvil as it does not support vrr at all. I hope that the breaking change for the `refresh` parameter is enough to get people implement it correctly.

src/backend/drm/compositor/mod.rs
```diff
@@ -2930,6 +2953,31 @@ where
...
+        self.surface.vrr_supported(conn).map_err(FrameError::DrmError)
+    }
+
+    /// Returns if Variable Refresh Rate is currently enabled for frames composed by this [`DrmCompositor`].
+    pub fn vrr_enabled(&self) -> bool {
+        self.vrr
```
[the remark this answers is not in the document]
Drakulix: No, I don't think so. Will do

Drakulix:
@cmeissl Merged your changes, removed `self.vrr`, dropped the anvil commit and squashed. Ready for merge?

cmeissl (approved):

cmeissl:
> @cmeissl Merged your changes, removed `self.vrr`, dropped the anvil commit and squashed. Ready for merge?

Yes :)

Drakulix:
> > @cmeissl Merged your changes, removed `self.vrr`, dropped the anvil commit and squashed. Ready for merge?
> 
> Yes :)

Thanks! :)

PR #1586, opened 2024-11-14
zero-saturating size sub

m4rch3n1ng:
this pr is two commits

- in [b5b9410409f2de6eb42084754451fa7fcbf37e17] i `impl Sub for Size`, that just mirrors what the `SubAssign` already does

- [ecb3f8565594bbdbd325ce54c6cd345e8072e186] might be a little controvertial: `Sub` and `SubAssign` operations on `Size` are already saturating, but they saturate to `i32::MIN`, which i don't personally think is very useful, because a `Size` can (should) not be negative, which means that i always have to check before subtracting two sizes. i changed it so they saturate to `0`.

to saturate to 0, i added an assossiated `const ZERO: Self` to the `Coordinate` trait.

if you prefer, i could move them to a `Size::saturating_sub(other: Size)` instead, but i think that this change makes `Size` more convenient to use.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -943,6 +943,25 @@ impl<N: Coordinate, Kind> AddAssign for Size<N, Kind> {
     }
 }
 
+impl<N: Coordinate, Kind> Sub for Size<N, Kind> {
+    type Output = Size<N, Kind>;
+    #[inline]
+    fn sub(self, rhs: Self) -> Self::Output {
+        debug_assert!(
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
I can't find any reason against this, given we already had the `debug_assert`.

@cmeissl @ids1024 @PolyMeilex @i509VCB if there aren't any complains in the next couple of days, I am going to merge this.

cmeissl (approved):
LGTM

Drakulix:
Thanks @m4rch3n1ng!

PR #1587, opened 2024-11-14
Allow constructing an XKB keymap directly from a file

ndreas:
Hello there!

I was testing https://github.com/YaLTeR/niri/ when I discovered that I couldn't set the XKB keymap directly from a file. This is something that can be done in Sway and Hyprland by providing the path to a `.xkb` file.

The `Keymap` struct from `xkbcommon` [has a constructor](https://docs.rs/xkbcommon/0.8.0/xkbcommon/xkb/struct.Keymap.html) that takes a `std::fs::File` and compiles a keymap from that, so I figured I might as well add it here.

Let me know if the approach is OK, I opted to go with an optional field on the `XKBConfig` struct instead of a completely separate struct.

### DIFF
--- a/src/input/keyboard/xkb_config.rs
+++ b/src/input/keyboard/xkb_config.rs
@@ -1,3 +1,4 @@
+use std::fs::File;
 pub use xkbcommon::xkb;
 
 /// Configuration for xkbcommon.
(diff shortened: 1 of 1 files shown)


src/input/keyboard/xkb_config.rs
```diff
@@ -27,19 +28,34 @@ pub struct XkbConfig<'a> {
...
-            self.model,
-            self.layout,
-            self.variant,
-            self.options.clone(),
-            xkb::KEYMAP_COMPILE_NO_FLAGS,
-        )
-        .ok_or(())
+        match &self.file {
+            Some(f) => {
+                let mut file = File::open(f).map_err(|_| ())?;
```
ids1024, line 40: It's probably not good to have a `File::open` where errors are just mapped to `()`, without at least logging something. That could be confusing if the file simply doesn't exist.

ids1024:
We do have a way to set the keymap from a string, which could also be used for this: https://smithay.github.io/smithay/smithay/input/keyboard/struct.KeyboardHandle.html#method.set_keymap_from_string. That I guess `add_keyboard` still need to be used with an `XkbConfig` first.

> I opted to go with an optional field on the XKBConfig struct instead of a completely separate struct.

If it only uses the `file`, or the other fields, but not both, an `enum` would probably make more sense. If it is combined into one type.

ndreas:
Thank you for the feedback! I should have found the `set_keyboard_from_string()` if I had looked more closely. I could use that to make a PR for Niri instead.

If you still want a feature like this, I can improve the PR by using an enum, introduce a proper error from the function that wraps the file, and expose that in the error enum here: https://github.com/Smithay/smithay/blob/ace2e6a9197f4b4173dea13d53d280d82f3a99c6/src/input/keyboard/mod.rs#L347

Otherwise I'll close this PR. Your call!

ids1024:
I'm not sure what anyone else thinks of this, but maybe it's best to just avoid complicating the API more, if `set_keymap_from_string` works well here.

PR #1588, opened 2024-11-14
Support fifo_v1 protocol

cmeissl:
WIP support for the new fifo protocol


src/wayland/compositor/tree.rs
```diff
@@ -306,7 +306,8 @@ impl PrivateSurfaceData {
         }
         my_data
             .pending_transaction
-            .insert_state(surface.clone(), my_data.current_txid);
+            .insert_state(surface.clone(), current_txid);
+        my_data.current_txid.0 = my_data.current_txid.0.wrapping_add(1);
```
cmeissl, line 310: @elinorbgr @Drakulix @ids1024 Not sure this is correct, but without incrementing the transaction id the blocked state will not be serialized but directly merged into any already blocked transaction. This means that for example the buffer will be overwritten and might get a release making fifo unusable with blockers (same for explicit sync, but I guess there it might not be an issue). It would also break my fifo barrier logic because it would allow to override an already pending barrier.

src/wayland/fifo/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+        let Some(barrier) = self.barrier.take() else {
+            return;
+        };
+
+        if into.barrier.as_ref() == Some(&barrier) || barrier.is_signaled() {
+            return;
+        }
+
+        if let Some(barrier) = into.barrier.replace(barrier) {
+            barrier.signal();
```
cmeissl, line 347: I tried to come up with a reason why this might be problematic, but so far couldn't find one.

anvil/src/state.rs
```diff
@@ -760,85 +773,183 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
...
+                    .as_ref()
+                    .map(|o| o == output)
+                    .unwrap_or(true)
+                {
+                    let fifo_barrier = states
+                        .cached_state
+                        .get::<FifoBarrierCachedState>()
+                        .current()
+                        .barrier
+                        .take();
```
Drakulix, line 919: What about fifo-barriers or commit-timers for DnD and Cursor surfaces?

src/wayland/fifo/mod.rs
```diff
@@ -0,0 +1,406 @@
...
+
+                        if let Some(barrier) = fifo_barrier {
+                            // If multiple consecutive commits only call wait_barrier, but not set_barrier
+                            // we might end up with the same barrier in multiple commits. It could happen
+                            // that the barrier is already signaled in which case there is no need to
+                            // further delay this commit
+                            //
+                            // In addition the spec also defines that the constraint must be ignored for
+                            // sync subsurfaces
+                            if !barrier.is_signaled() || is_sync_subsurface(surface) {
```
Drakulix, line 254: What if the surface is set sync afterwards? Given the docs state that `wait_barrier` should be the even ignored, I feel like there are cases, where we need to clear the blocker later (or the blocker needs to check the surface's sync state upon evaluation somehow).

Drakulix (commented):
The fifo-protocol states that:
>  If the surface is not being updated by the compositor (off-screen, occluded) the compositor may ignore the constraint.

I guess that would just mean immediately clearing the blocker on `commit`, right?

src/wayland/fifo/mod.rs
cmeissl: Are there any guarantees for a surface being set sync afterwards? This is already racy imo as the blocker could already be cleared and the state applied. So not sure we really need to handle that. Checking in the blocker won't help as we only evaluate it on `blocker_cleared`. I would rather not add some kind of callback logic for surfaces being set sync. We theoretically could check it in either `pre_paint` or `post_paint`.

But I will ask in #wayland for clarification

cmeissl: okay, so I was actually redirected to this issue :) https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/362 which also references this one https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/361

while the latter has been closed and the former is not merged yet we can conclude that commit time is the correct interpretation (unless I misunderstood something in the discussion of 362, got a bit lost tbh in between)

anvil/src/state.rs
cmeissl: :upside_down_face: done (which should also finally resolve the missing frame callbacks...)

src/wayland/fifo/mod.rs
cmeissl: nice, I actually got the check backwards. should be fixed now and skip the blocker if the barrier is already signaled or the surface is sync

Drakulix (approved):

PR #1589, opened 2024-11-16
output: store a weak ref in the global to avoid...

### LINKED ISSUE #1584: `update_from_render_element_states` relies on `strong_count` to determine if `Output` is still valid
https://github.com/Smithay/smithay/blob/28417d22202a10e24de53b8994f5044aa042dcc5/src/backend/renderer/element/mod.rs#L194-L197

`WeakOutput::is_alive()` simply checks if the strong count is `0`. In `cosmic-comp` I seem to be seeing an issue with it. Perhaps a strong reference is being held somewhere it shouldn't, either in Smithay or cosmic-comp.

Careful use of `WeakOutput` should avoid that, but I wonder if it would also be a good idea to explicitly track if the output is valid, instead of relying on reference count.

A compositor can set a `compare` callback that has an additional check to see if the old `Output` is still valid. But `default_primary_scanout_output_compare` won't do that. So this may be causing issues for any Smithay compositors that don't have a custom callback with some check for old removed/disabled outputs.

cmeissl:
...a cycle keeping the output alive indefinitely

Should get rid of the last cycle in smithay as mentionen in #1585  and hopefully fix #1584

### DIFF
--- a/src/wayland/output/handlers.rs
+++ b/src/wayland/output/handlers.rs
@@ -38,16 +38,22 @@ where
         data_init: &mut DataInit<'_, D>,
     ) {
         let client_scale = state.client_compositor_state(client).clone_client_scale();
-        let output = data_init.init(
+
+        let Some(output) = global_data.output.upgrade() else {
+            tracing::warn!("trying to bind a destroyed output, forgot to disable a output global?");
+            return;
+        };
+
+        let wl_output = data_init.init(
             resource,
             OutputUserData {
-                output: global_data.output.downgrade(),
+                output: global_data.output.clone(),
                 last_client_scale: AtomicU32::new(client_scale.load(Ordering::Acquire)),
                 client_scale,
             },
         );
 
-        let mut inner = global_data.output.inner.0.lock().unwrap();
+        let mut inner = output.inner.0.lock().unwrap();
 
         let span = warn_span!("output_bind", name = inner.name);
         let _enter = span.enter();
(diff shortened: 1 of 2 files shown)


ids1024:
Ah. So when does global user data like this actually get freed? Only once every client that has had a registry with the global present has disconnected?

src/wayland/output/handlers.rs
```diff
@@ -38,16 +38,22 @@ where
         data_init: &mut DataInit<'_, D>,
     ) {
         let client_scale = state.client_compositor_state(client).clone_client_scale();
-        let output = data_init.init(
+
+        let Some(output) = global_data.output.upgrade() else {
+            tracing::warn!("trying to bind a destroyed output, forgot to disable a output global?");
+            return;
```
Drakulix, line 44: I don't think we should and can just return here. Any request to the `output` would otherwise trigger a panic, given we didn't initialize it's data.

The warning is fine (and a good idea), but I think we should just initialize the `OutputUserData` with `Weak::new`.

Drakulix (changes requested):

cmeissl:
> Ah. So when does global user data like this actually get freed? Only once every client that has had a registry with the global present has disconnected?

The global data itself is freed on `remove_global`. I actually started this before pulling in your changes, so looking at it again, the change here is unnecessary. Just tried it out and the `primary_scanout_output` in `PrimaryScanoutOutput::update_from_render_element_states` seems to work fine in anvil on master.

ids1024:
Yeah, as long as this is freed immediately when `remove_global` is called, it makes a lot of sense for the global to hold a strong reference.

PR #1594, opened 2024-11-21
Update `gbm` to 0.18

ids1024:
Depends on https://github.com/Smithay/gbm.rs/pull/25.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ drm-fourcc = "^2.2.0"
 drm = { version = "0.14.0", optional = true }
 drm-ffi = { version = "0.9.0", optional = true }
 errno = "0.3.5"
-gbm = { version = "0.16.0", optional = true, default-features = false, features = ["drm-support"] }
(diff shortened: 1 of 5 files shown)


Drakulix:
> I updated this, but I noticed there's still an unused generic: [Smithay/gbm.rs#48](https://github.com/Smithay/gbm.rs/pull/48).
> 
> It could be merged as is now, though it has to pass something explicitly as that generic. Annoying that clippy doesn't say anything about unused generics.

v0.18 is inbound here https://github.com/Smithay/gbm.rs/actions/runs/12126008784

ids1024:
> It seems package 'gbm-sys' modified since '0.3.1' so new version should be published

Drakulix:
> > It seems package 'gbm-sys' modified since '0.3.1' so new version should be published

https://github.com/Smithay/gbm.rs/actions/runs/12143920979/job/33862081369

Drakulix (approved):
Looks good!

PR #1599, opened 2024-12-02
Fix clippy/fmt with latest nightly

ids1024:


src/backend/renderer/multigpu/mod.rs
```diff
@@ -723,16 +723,16 @@ impl<'render, 'target, R: GraphicsApi, T: GraphicsApi> fmt::Debug for MultiRende
...
+impl<R: GraphicsApi, T: GraphicsApi> AsRef<<R::Device as ApiDevice>::Renderer>
+    for MultiRenderer<'_, '_, R, T>
 {
     fn as_ref(&self) -> &<R::Device as ApiDevice>::Renderer {
         self.render.renderer()
     }
 }
 
-impl<'render, 'target, R: GraphicsApi, T: GraphicsApi> AsMut<<R::Device as ApiDevice>::Renderer>
-    for MultiRenderer<'render, 'target, R, T>
```
Drakulix, line 735: I am not sure I like this clippy lint telling us to remove explicitly named lifetimes. The names are descriptive.. I guess it is fine, as long as it is only doing that for `impl`s though?

ids1024: Good point. But yeah, I'm not sure the named lifetimes add anything in things like `AsRef`/`AsMut` impls, where the lifetime doesn't appear anywhere in the body. So it's probably good here.

Drakulix (approved):
Seems good

PR #1600, opened 2024-12-03
Update to `pixman` 0.2, and make `PixmanTexture` thread-safe

ids1024:
Needs testing. Compiles, but I haven't checked that the `borrow_mut` calls can't panic.

Requires https://github.com/cmeissl/pixman-rs/pull/19.


ids1024:
I think if the pixman-rs PR is merged and released, everything here is good.

cmeissl:
`pixman-rs` 0.2.0 is out https://crates.io/crates/pixman/0.2.0

PR #1601, opened 2024-12-05
Add `GpuManager::devices` methods

ids1024:
It doesn't seem `GpuManager` has a way to get a list of all backend renders; or other information about devices. (Calling `GraphicsApi::enumerate` in multiple places would not be correct.)

This seems useful in some cases, and easy to support.

(My use case is to dynamically configure a shader, and compile the shader on all GPUs when the shader is changed. Which needs a way to iterate over the `GlesRenderer`s).


Drakulix (approved):
Seems fine to me. Given this is read-only, there shouldn't be a way to mess up internal state with this.

PR #1605, opened 2024-12-12
Add a `GpuManager::device_mut()` method

ids1024:
Renderers require mut references for most things. Allocators also need an `&mut` to allocate. The `node` isn't mutable, though `mem::swap` could replace the whole `GbmGlesDevice` (already possible elsewhere with the just renderer).

### DIFF
--- a/src/backend/renderer/multigpu/mod.rs
+++ b/src/backend/renderer/multigpu/mod.rs
@@ -239,6 +239,14 @@ impl<A: GraphicsApi> GpuManager<A> {
         Ok(&self.devices)
     }
 
+    /// Get all devices enumerated by the API.
+    pub fn devices_mut(&mut self) -> Result<&mut [A::Device], A::Error> {
+        if self.api.needs_enumeration() {
+            self.api.enumerate(&mut self.devices)?;
+        }
+        Ok(&mut self.devices)
+    }
+
     /// Create a [`MultiRenderer`] from a single device.
     ///
     /// This a convenience function to deal with the same types even, if you only need one device.
--- a/src/backend/renderer/multigpu/mod.rs
+++ b/src/backend/renderer/multigpu/mod.rs
@@ -232,19 +232,19 @@ impl<A: GraphicsApi> GpuManager<A> {
     }
 
     /// Get all devices enumerated by the API.
-    pub fn devices(&mut self) -> Result<&[A::Device], A::Error> {
+    pub fn devices(&mut self) -> Result<impl Iterator<Item = &A::Device>, A::Error> {
         if self.api.needs_enumeration() {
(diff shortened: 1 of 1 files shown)


Drakulix:
> Renderers require mut references for most things. Allocators also need an `&mut` to allocate. The `node` isn't mutable, though `mem::swap` could replace the whole `GbmGlesDevice` (already possible elsewhere with the just renderer).

Uhmm, I really don't like this. With a mutable slice you can do other ugly things like swap the order of elements around. Could we at least make `devices` and `devices_mut` return an iterator?

ids1024:
Using an iterator won't really change that, in principle. If it returns `Result<impl Iterator<Item=&mut A::Device>, A::Error>`, you can just call `.devices_mut().unwrap().collect::<Vec<_>>()` or otherwise store multiple entries in the iterator to access at once. I guess internal iteration with a callback that takes a device would avoid that? But you could still `mem::swap` with a separate device that isn't from the iterator.

For my immediate use of this, a `renderer_mut()` function returning an iterator of renderers would also work (same considerations for being able to swap renders, but not other parts of the device). But I assume the other device fields could be useful to access as well.

ids1024:
(Ideally the solution here would be for Rust to have a separate concept of "movable references" as https://without.boats/blog/pinned-places suggests, and then the places an `&mut Renderer` is available couldn't abuse it with swapping. Maybe something would be possible now by making every renderer method take `Pin<&mut Self>` but that sounds rather painful to deal with.)

Drakulix:
> Using an iterator won't really change that, in principle. If it returns `Result<impl Iterator<Item=&mut A::Device>, A::Error>`, you can just call `.devices_mut().unwrap().collect::<Vec<_>>()` or otherwise store multiple entries in the iterator to access at once. I guess internal iteration with a callback that takes a device would avoid that? But you could still `mem::swap` with a separate device that isn't from the iterator.
> 
> For my immediate use of this, a `renderer_mut()` function returning an iterator of renderers would also work (same considerations for being able to swap renders, but not other parts of the device). But I assume the other device fields could be useful to access as well.

Yes it allows `mem::swap` but not [`slice::swap`](https://doc.rust-lang.org/std/primitive.slice.html#method.swap). (I know that is pedantic, but I feel like hiding the collection type at least is good practice here.)

ids1024:
Yeah, I don't really see any issue with using `impl Iterator` here. It just doesn't technically impact what you can do with the function.

Using an iterator is probably a good practice from an API stability standpoint. (Not that this is a huge concern for Smithay at the moment.)

ids1024:
Updated both to use iterators.

Drakulix (approved):

PR #1609, opened 2024-12-14
Toplevel parent fixes, checks, and callback

YaLTeR:
Several things here:
1. Add a self-parent and a parent loop check when setting a parent, corresponding to the protocol:

    > The parent toplevel must not be one of the child toplevel's descendants, and the parent must be different from the child toplevel, otherwise the invalid_parent protocol error is raised.
1. Fix inverted logic in xdg-foreign, tested with xdg-desktop-portal-gnome dialogs.
1. Expose an `XdgShellHandler::parent_changed()` callback so compositors can react to parent changes.

    The primary use is to restack surfaces as the protocol intends:

    > This surface should be stacked above the parent surface and all other ancestor surfaces.

    Another use is to signal parent changes to wlr-foreign-toplevel-management.

    This makes xdg-foreign depend on `XdgShellHandler`. Not sure if it's doable otherwise; realistically it's required because there's no other toplevel-equivalent yet.
1. Remove `ToplevelSurface::set_parent()`. This was used back in zxdg-shell days, not any more. Anyhow, exposing this publicly seems a bit strange because it can cause client protocol errors down the line from compositor actions.

### DIFF
--- a/src/wayland/shell/mod.rs
+++ b/src/wayland/shell/mod.rs
@@ -18,6 +18,7 @@
 use crate::{utils::Serial, wayland::compositor};
 use thiserror::Error;
 use wayland_server::protocol::wl_surface::WlSurface;
+use xdg::XdgToplevelSurfaceData;
 
 pub mod kde;
 pub mod wlr_layer;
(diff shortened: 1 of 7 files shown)


PolyMeilex:
> there's no other toplevel-equivalent yet.

XWayland is such a equivalent, but either way I don't mind this depending on xdg shell.

Drakulix (approved):
Seems all very sensible!

PR #1611, opened 2024-12-14
input: Treat Mouse `BTN_SIDE` and `BTN_EXTRA` as `MouseButton::Back` and `MouseButton::Forward` Respectively

bbb651:
These are the keycodes my own mouse produces for these actions (Logitech G102, `046d:c092`), although it might be influenced by `libratbag`.

This also matches winit and chromium.


Drakulix (approved):
Thanks!

PR #1612, opened 2024-12-14
use weak surface ref inside transaction

cmeissl:
use weak surface ref inside transaction to prevent keeping them alive longer as needed

this also resolves a deadlock when trying to send events in a drop impl stored in the surface state of a blocked transaction

see https://github.com/Smithay/wayland-rs/pull/772 for a different way to solve the same deadlock


Drakulix (approved):
Regardless of fixing this in wayland-rs, this seems like a good idea.

PR #1617, opened 2024-12-21
xdg: Add getters for pending configures and current state serial

YaLTeR:
The pending configures are useful for when you need to check the state that you sent to the client, but that it hadn't acknowledged yet, along with their serials. I only expose read access so that you can't mess with the internal commit tracking.

The serials in general (like the current state serial) are useful in case you want to synchronize something to surface commits. The case I have in niri is that I want to know if the surface had already committed the state that it had last acked, or not yet. Specifically, for the surface size, I want to differentiate between when the surface acked a new size and hasn't committed in response yet, and when the surface acked a new size but committed a different size in response.


Drakulix (approved):
LGTM

PR #1621, opened 2024-12-27
Improve geometry APIs, to be closer to `euclid`

ids1024:
I want to update https://github.com/Smithay/smithay/pull/1136, make any needed improvements upstream to `euclid`, and get that merged. But it requires a lot of changes both in Smithay and compositors using it. Perhaps adding some APIs and uses of the `#[deprecated]` attribute first could make that a little less painful to deal with.

Even if Smithay didn't move to `euclid`, it would be worth borrowing some ideas to make geometry handling more strongly typed, as well as a couple API conveniences.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -1114,6 +1114,7 @@ impl<Kind> Rectangle<f64, Kind> {
 
 impl<N: Coordinate, Kind> Rectangle<N, Kind> {
     /// Create a new [`Rectangle`] from the coordinates of its top-left corner and its dimensions
+    #[deprecated(note = "use new or from_size")]
     #[inline]
     pub fn from_loc_and_size(loc: impl Into<Point<N, Kind>>, size: impl Into<Size<N, Kind>>) -> Self {
         Rectangle {
@@ -1312,32 +1313,35 @@ impl<N: Coordinate, Kind> Rectangle<N, Kind> {
                     continue;
                 }
 
-                let top_rect = Rectangle::from_loc_and_size(
+                let top_rect = Rectangle::new(
                     item.loc,
-                    (item.size.w, intersection.loc.y.saturating_sub(item.loc.y)),
+                    (item.size.w, intersection.loc.y.saturating_sub(item.loc.y)).into(),
                 );
-                let left_rect: Rectangle<N, Kind> = Rectangle::from_loc_and_size(
-                    (item.loc.x, intersection.loc.y),
-                    (intersection.loc.x.saturating_sub(item.loc.x), intersection.size.h),
+                let left_rect: Rectangle<N, Kind> = Rectangle::new(
+                    (item.loc.x, intersection.loc.y).into(),
+                    (intersection.loc.x.saturating_sub(item.loc.x), intersection.size.h).into(),
                 );
-                let right_rect: Rectangle<N, Kind> = Rectangle::from_loc_and_size(
+                let right_rect: Rectangle<N, Kind> = Rectangle::new(
                     (
                         intersection.loc.x.saturating_add(intersection.size.w),
                         intersection.loc.y,
-                    ),
+                    )
+                        .into(),
                     (
                         (item.loc.x.saturating_add(item.size.w))
                             .saturating_sub(intersection.loc.x.saturating_add(intersection.size.w)),
                         intersection.size.h,
-                    ),
+                    )
+                        .into(),
                 );
-                let bottom_rect: Rectangle<N, Kind> = Rectangle::from_loc_and_size(
-                    (item.loc.x, intersection.loc.y.saturating_add(intersection.size.h)),
+                let bottom_rect: Rectangle<N, Kind> = Rectangle::new(
+                    (item.loc.x, intersection.loc.y.saturating_add(intersection.size.h)).into(),
                     (
                         item.size.w,
                         (item.loc.y.saturating_add(item.size.h))
                             .saturating_sub(intersection.loc.y.saturating_add(intersection.size.h)),
-                    ),
+                    )
+                        .into(),
                 );
 
                 if !top_rect.is_empty() {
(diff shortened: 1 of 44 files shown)


Drakulix:
> This matches `euclid::Rect`'s API, but it also seems a little better to
have strong typing without implicit conversion like this.

I am not sure I agree with this, when we are just adding `into`-calls everywhere. It just makes everything more busy. What is the future goal here to actually make use of that type-safety?

src/utils/geometry.rs
```diff
@@ -1338,6 +1350,26 @@ impl<N: Coordinate, Kind> Rectangle<N, Kind> {
...
+            loc: (Zero::zero(), Zero::zero()).into(),
+            size,
+        }
+    }
+
+    /// Create a new [`Rectangle`] with location and size zero
+    #[inline]
+    pub fn zero() -> Self {
+        Rectangle {
+            loc: (Zero::zero(), Zero::zero()).into(),
```
m4rch3n1ng, line 1367: i already added an associated `const ZERO` to the `Coordinate` trait in #1586, so you could drop the num-traits dependency and just use `N::ZERO` here.

ids1024:
I've added a `Mul` impl for `Rectangle` (which `euclid` has). This is useful in at least a couple places in cosmic-comp.

Where possible it seems best to use high-level operations like that on these strongly-typed geometry types instead of calls to things like `Rectange::new`, though I don't know how many of the use cases can be better handled with something like that.

> I am not sure I agree with this, when we are just adding into-calls everywhere. It just makes everything more busy. What is the future goal here to actually make use of that type-safety?

Where we really want to get type safety is by using more coordinate spaces, distinguishing between points and vectors (it shouldn't be possible to add two "points" in the same space to get an element of that space.) And things like [euclid::Translation2D](https://docs.rs/euclid/latest/euclid/struct.Translation2D.html). https://github.com/Smithay/smithay/pull/1620 happens to be a good example of something we'd want to make statically impossible.

If we want to move to `euclid`, this change to match `euclid::Rectangle::new` will be necessary. That and renaming `loc`/`w`/`h` to `origin`/`width`/`height` account for a lot of the changes needed to compile with the `euclid` types (plus extension traits for `to_physical`, etc).

If we don't move to `euclid`, avoiding implicit conversion generally seems like a best practice. It's not a big deal, but there are a couple places where it's not obvious that the argument is a tuple, so we're asserting what coordinate space it's part of. In general it doesn't make things much worse. Something like [`euclid::rect`](https://docs.rs/euclid/latest/euclid/fn.rect.html) could be handy when constructing a rectangle from four scalars (particularly in tests).

Though if that change is too controversial I can drop 5471a59b430c96a412da09578170af5a1136aeda from this PR. The other small additions here seem useful.

Drakulix:
> Where we really want to get type safety is by using more coordinate spaces, distinguishing between points and vectors (it shouldn't be possible to add two "points" in the same space to get an element of that space.) And things like [euclid::Translation2D](https://docs.rs/euclid/latest/euclid/struct.Translation2D.html). #1620 happens to be a good example of something we'd want to make statically impossible.

Agreed.

> If we want to move to `euclid`, this change to match `euclid::Rectangle::new` will be necessary. That and renaming `loc`/`w`/`h` to `origin`/`width`/`height` account for a lot of the changes needed to compile with the `euclid` types (plus extension traits for `to_physical`, etc).

Not sure yet, if we truly want to move to `euclid` though it seems to be quite a good option. Our own code has grown significantly more complex, so there might be more to be gained from euclid than just compatibility with other crates. But I also don't think the maintenance overhead is really making this necessary and in general we probably would want to limit the number of dependencies rather than increasing it.

> If we don't move to `euclid`, avoiding implicit conversion generally seems like a best practice. It's not a big deal, but there are a couple places where it's not obvious that the argument is a tuple, so we're asserting what coordinate space it's part of. In general it doesn't make things much worse. Something like [`euclid::rect`](https://docs.rs/euclid/latest/euclid/fn.rect.html) could be handy when constructing a rectangle from four scalars (particularly in tests).

Yeah agreed. Given that euclid also seems to be a very promising alternative, it is probably not a bad idea to align the apis. (Though we definitely should add public re-exports to make the transition less painful.)

Drakulix (approved):

Drakulix:
LGTM, but needs a rebase

ids1024:
> Not sure yet, if we truly want to move to euclid though it seems to be quite a good option. 

Yeah. If we don't move to euclid, we probably still want to copy various things from euclid, which has various things the smithay geometry code lacks, and generally seems a bit better thought out of an API. At which point it seems a bit unnecessary to duplicate it and not just use euclid (and contribute anything that's missing upstream).

Euclid could also be useful client-side, and if Smithay uses it as well, potentially in code that's shared. That's one benefit to having it in a different crate.

But if there are issues with moving to euclid, we can ultimately still stick with our own types, and improve things here.

src/utils/geometry.rs
```diff
@@ -1122,6 +1122,30 @@ impl<N: Coordinate, Kind> Rectangle<N, Kind> {
...
+    pub fn from_size(size: Size<N, Kind>) -> Self {
+        Rectangle {
+            loc: (N::ZERO, N::ZERO).into(),
+            size,
+        }
+    }
+
+    /// Create a new [`Rectangle`] with location and size zero
+    #[inline]
+    pub fn zero() -> Self {
```
Drakulix, line 1156: I feel like this function should be marked `const`. Any reason against that?

Drakulix (approved):
LGTM. Minor nitpick

src/utils/geometry.rs
ids1024: Good point. The `euclid` version isn't, I guess since https://docs.rs/num-traits/latest/num_traits/identities/trait.Zero.html#tymethod.zero isn't (which perhaps should be, though that may let it be used with heap-allocated bigints as scalars, or such).

If we want to move to `euclid`, it makes sense not to add `const` where `euclid` doesn't have it. If not, we may as well.

Though we also don't seem to have a single `const fn` in `geometry.rs` currently. And I guess no one has complained about that so far.

Drakulix: Fine to skip this then.

PR #1630, opened 2025-01-09
xdg: Add `get_anchor_point method` to `PositionerState`

ids1024:
These two changes are used for https://github.com/pop-os/cosmic-comp/pull/1122.

Cosmic-comp's implementation is a little unusual. It tries to fit the popup within the tile if it can do so without resizing. If this fails, it instead tries to fit it within the output, using resizes as well if still needed. I'm not sure if many compositors will need this, but this seems as good a way as any to expose a way to do this.

It also has a check to test which output the anchor point is in. It seems reasonable to expose this as a method.

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -569,6 +569,32 @@ impl PositionerState {
         }
     }
 
+    /// Get the anchor point for a popup as defined by this positioner.
+    ///
+    /// Defined by `xdg_positioner.set_anchor_rect` and
+    /// `xdg_positioner.set_anchor`.
+    pub fn get_anchor_point(&self) -> Point<i32, Logical> {
+        let mut point = self.anchor_rect.loc;
+
+        point.y += if self.anchor_has_edge(xdg_positioner::Anchor::Top) {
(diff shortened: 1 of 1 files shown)


YaLTeR:
FWIW you don't need an extra argument, just do this:

```rust
// Do not try to resize to fit the padded target rectangle.
let mut no_resize = positioner;
no_resize
    .constraint_adjustment
    .remove(ConstraintAdjustment::ResizeX);
no_resize
    .constraint_adjustment
    .remove(ConstraintAdjustment::ResizeY);

let geo = no_resize.get_unconstrained_geometry(padded.to_i32_round());
if padded.contains_rect(geo.to_f64()) {
    return geo;
}

// Could not unconstrain into the padded target, so resort to the regular one.
positioner.get_unconstrained_geometry(target.to_i32_round())
```

That's how I intended it to be used from the start.

ids1024:
Ah, I hadn't event noticed that `get_unconstrained_geometry` (unlike `get_geometry`) consumes `self` and is a method of a `Copy` type. Yeah, that makes sense.

Drakulix (approved):

PR #1635, opened 2025-01-10
drm/atomic: Better logging for atomic requests

Drakulix:
`Invalid screen configuration` issues are rare these days, but decrypting them from logs is very cumbersome whenever they occur. This is mostly due to the property handles not having any meaning, if you don't have the property mapping.

So we do log the mapping now as well.

Additionally in debug mode we construct the whole request differently to give us actually nice output. We need a bunch of collections to do so, so this is inefficient enough to only do in debug builds, long term it might be better to find a solution how to deconstruct drm-rs's `AtomicModeReq`. But for now this should be a significant improvement.

### DIFF
--- a/src/backend/drm/surface/atomic.rs
+++ b/src/backend/drm/surface/atomic.rs
@@ -750,7 +750,6 @@ impl AtomicDrmSurface {
         let current_conns = current.connectors.clone();
         let pending_conns = pending.connectors.clone();
         let mut removed = current_conns.difference(&pending_conns);
-        let mut added = pending_conns.difference(&current_conns);
 
         for conn in removed.clone() {
             if let Ok(info) = self.fd.get_connector(*conn, false) {
@@ -760,7 +759,7 @@ impl AtomicDrmSurface {
             }
         }
 
-        for conn in added.clone() {
+        for conn in &pending_conns {
             if let Ok(info) = self.fd.get_connector(*conn, false) {
                 info!("Adding connector: {:?}", info.interface());
             } else {
@@ -782,7 +781,7 @@ impl AtomicDrmSurface {
                 self.crtc,
                 Some(pending.blob),
                 pending.vrr,
-                &mut added,
+                &mut pending_conns.iter(),
                 &mut removed,
                 &*planes,
             )?;
--- a/src/backend/drm/surface/atomic.rs
+++ b/src/backend/drm/surface/atomic.rs
@@ -6,7 +6,11 @@ use drm::control::{
     connector, crtc, dumbbuffer::DumbBuffer, framebuffer, plane, property, AtomicCommitFlags, Mode, PlaneType,
 };
 
+#[cfg(debug_assertions)]
+use std::collections::HashMap;
 use std::collections::HashSet;
+#[cfg(debug_assertions)]
+use std::fmt;
 use std::os::unix::io::AsRawFd;
 use std::sync::{
     atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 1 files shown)


src/backend/drm/surface/atomic.rs
```diff
@@ -440,31 +452,35 @@ impl AtomicDrmSurface {
...
+                fence: None,
+            }),
+        };
+        let req = AtomicRequest::build_request(
+            &prop_mapping,
+            self.crtc,
             Some(pending.blob),
             pending.vrr,
+            &mut added,
+            &mut removed,
```
cmeissl, line 476:
```suggestion
            added,
            removed,
```

src/backend/drm/surface/atomic.rs
```diff
@@ -492,29 +508,33 @@ impl AtomicDrmSurface {
...
+                fence: None,
+            }),
+        };
+        let req = AtomicRequest::build_request(
+            &prop_mapping,
+            self.crtc,
             Some(new_blob),
             pending.vrr,
+            &mut pending.connectors.iter(),
+            &mut [].iter(),
```
cmeissl, line 530:
```suggestion
            pending.connectors.iter(),
            [],
```

src/backend/drm/surface/atomic.rs
```diff
@@ -682,21 +702,24 @@ impl AtomicDrmSurface {
         let pending_conns = pending.connectors.clone();
         let mut removed = current_conns.difference(&pending_conns);
         let mut added = pending_conns.difference(&current_conns);
```
cmeissl, line 704:
```suggestion
        let removed = current_conns.difference(&pending_conns);
        let added = pending_conns.difference(&current_conns);
```

src/backend/drm/surface/atomic.rs
```diff
@@ -682,21 +702,24 @@ impl AtomicDrmSurface {
...
+        let prop_mapping = self.prop_mapping.read().unwrap();
 
-        let req = self.build_request(
+        let req = AtomicRequest::build_request(
+            &prop_mapping,
+            self.crtc,
+            Some(pending.blob),
+            pending.vrr,
             &mut added,
             &mut removed,
```
cmeissl, line 713:
```suggestion
            added,
            removed,
```

src/backend/drm/surface/atomic.rs
```diff
@@ -727,7 +750,6 @@ impl AtomicDrmSurface {
         let current_conns = current.connectors.clone();
         let pending_conns = pending.connectors.clone();
         let mut removed = current_conns.difference(&pending_conns);
```
cmeissl, line 752:
```suggestion
        let removed = current_conns.difference(&pending_conns);
```

src/backend/drm/surface/atomic.rs
```diff
@@ -752,20 +774,23 @@ impl AtomicDrmSurface {
...
-                &mut added,
-                &mut removed,
-                &*planes,
+            let req = AtomicRequest::build_request(
+                &prop_mapping,
+                self.crtc,
                 Some(pending.blob),
                 pending.vrr,
+                &mut pending_conns.iter(),
+                &mut removed,
```
cmeissl, line 785:
```suggestion
                &pending_conns,
                removed,
```

src/backend/drm/surface/atomic.rs
```diff
@@ -837,12 +862,15 @@ impl AtomicDrmSurface {
...
         // page flips work just like commits with fewer parameters..
-        let req = self.build_request(
+        let prop_mapping = self.prop_mapping.read().unwrap();
+        let req = AtomicRequest::build_request(
+            &prop_mapping,
+            self.crtc,
+            None,
+            self.state.read().unwrap().vrr,
             &mut [].iter(),
             &mut [].iter(),
```
cmeissl, line 872:
```suggestion
            [],
            [],
```

cmeissl (changes requested):
No idea why we had `&dyn Iterator` here, but imo this is the perfect opportunity to get rid of it.

Drakulix (commented):
drm/atomic: Get rid of `&mut dyn Iterator`

cmeissl (approved):

cmeissl (approved):
LGTM

Issue #1615, opened 2024-12-19
`GlesRenderer::import_shm_buffer` needs synchronization for shared EGL contexts in different threads


ids1024:
@Drakulix and I noticed some apparent issues here debugging https://github.com/pop-os/cosmic-comp/issues/1078. I'm not sure they fully explain what's going on there, but it does seem like something that needs to be fixed.

* If two threads call `GlesRenderer::import_shm_buffer` with shared EGL contexts (`EGLContext::new_shared`), it looks like both could try to draw to the same GLES texture. There should be some kind of synchronization to prevent parallel writes (and to avoid unncessary copy of the same damage regions from the shm buffer).
* If one context renders to the GLES texture, and another context tries to read it, it looks like that also needs synchronization, and this doesn't happen implicitly.
   - See section "C.2 Propagating Changes to Objects" in https://registry.khronos.org/OpenGL/specs/es/2.0/es_full_spec_2.0.pdf
     * Looks like we need to wait on the fence (or use glFinish) before a `glBindTexture` in a different context
     
Not using shared EGL contexts, or importing a seperate GLES texture for each shared context, could fix this. But it would be good to avoid unnecessary duplication of copies from shm buffers to GPU textures if possible, even though it's not normally the most important thing.

(I guess if there are major changes to the logic for mirroring shm buffers to GPU buffers, we should also consider how best to manage the reverse, which is needed for `ext-image-copy-capture-v1`.)

ids1024:
To synchronize, we should presumably use `eglCreateSync` after the draw calls, then another renderer using a shared context can `eglWaitSync` on the sync point stored alongside the texture ide.

To avoid copying the same damage multiple times, maybe `import_shm_buffer` shouldn't be taking damage rectangles, but should handle the damage since the last commit internally. (So it doesn't get the same damage rectangles in two different threads without knowing it needs to copy the same thing.)

Then if there's any additional damage, it can wait on the existing sync point (if any), draw the addition damage regions, and then export a new one.

I guess something should also block rendering to a texture when it's still being read elsewhere? Though ultimately that shouldn't happen with correct clients, since they won't be modifying a buffer and committing new damage until the buffer is released, at which point nothing in the compositor should be using it. So maybe not needed as long as it's not undefined behavior (in the sense of being unsound).

PR #1637, opened 2025-01-14
feature: Renderer Framebuffer type

Drakulix:
Superseeds https://github.com/Smithay/smithay/pull/1616.

Original description:
> Attempt to solve https://github.com/Smithay/smithay/issues/1615.
>
> This solution isn't perfect, most notably it doesn't synchronize across the access of `Bind<GlesTexture>`. To do that we would need to be able to store the lock, which means moving the `GlesTarget` into the `GlesFrame` and making `Bind` return a `Frame`. I think we talked about this, so we should fix this at some point, but I didn't want to do that refactor now.
>
> Fixing `Bind` would automatically solve this for exporting and other things as well, as all of these rely to binding a framebuffer container object to our texture.
>
> (Also binding a texture to write to it and sampling from it in a separate thread is probably a operation not as common.)

I lied, I did the refactor. The changes to `Renderer` and `Bind` are fairly small, however the fallout was huge, but since none of this compiles without all the other changes, I wasn't sure how to split this apart further. Please excuse the giant PR/commit everybody.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ encoding_rs = { version = "0.8.33", optional = true }
 profiling = "1.0.13"
 smallvec = "1.11"
 pixman = { version = "0.2.1", features = ["drm-fourcc", "sync"], optional = true }
+aliasable = { version = "0.1.3", optional = true }
 
 
 [dev-dependencies]
@@ -99,7 +100,7 @@ backend_session_libseat = ["backend_session", "libseat"]
 desktop = []
 renderer_gl = ["gl_generator", "backend_egl"]
 renderer_glow = ["renderer_gl", "glow"]
-renderer_multi = ["backend_drm"]
+renderer_multi = ["backend_drm", "aliasable"]
 renderer_pixman = ["pixman"]
 renderer_test = []
 use_system_lib = ["wayland_frontend", "wayland-backend/server_system", "wayland-sys", "gbm?/import-wayland"]
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -34,8 +34,8 @@ pub use uniform::*;
 use self::version::GlVersion;
 
 use super::{
-    sync::SyncPoint, Bind, Blit, Color32F, DebugFlags, ExportMem, Frame, ImportDma, ImportMem, Offscreen,
-    Renderer, RendererSuper, Texture, TextureFilter, TextureMapping,
+    sync::SyncPoint, Bind, Blit, BlitFrame, Color32F, DebugFlags, ExportMem, Frame, ImportDma, ImportMem,
+    Offscreen, Renderer, RendererSuper, Texture, TextureFilter, TextureMapping,
 };
 use crate::backend::egl::{
     ffi::egl::{self as ffi_egl, types::EGLImage},
@@ -1605,6 +1605,36 @@ impl Offscreen<GlesRenderbuffer> for GlesRenderer {
                 destruction_callback_sender: self.destruction_callback_sender.clone(),
             })))
         }
+    }
+}
+
+impl<'buffer> BlitFrame<GlesTarget<'buffer>> for GlesFrame<'_, 'buffer> {
+    fn blit_to(
+        &mut self,
+        to: &mut GlesTarget<'buffer>,
+        src: Rectangle<i32, Physical>,
+        dst: Rectangle<i32, Physical>,
+        filter: TextureFilter,
+    ) -> Result<(), Self::Error> {
+        let res = self.renderer.blit(self.target, to, src, dst, filter);
+        self.target
+            .0
+            .make_current(&self.renderer.gl, &self.renderer.egl)?;
+        res
+    }
+
+    fn blit_from(
+        &mut self,
+        from: &GlesTarget<'buffer>,
+        src: Rectangle<i32, Physical>,
+        dst: Rectangle<i32, Physical>,
+        filter: TextureFilter,
+    ) -> Result<(), Self::Error> {
+        let res = self.renderer.blit(from, self.target, src, dst, filter);
+        self.target
+            .0
+            .make_current(&self.renderer.gl, &self.renderer.egl)?;
+        res
     }
 }
 
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -12,7 +12,7 @@ use std::{
     sync::{
         atomic::{AtomicBool, AtomicPtr, Ordering},
         mpsc::{channel, Receiver, Sender},
-        Arc, Mutex, RwLock,
+        Arc, Mutex, RwLock, RwLockWriteGuard,
     },
 };
 use tracing::{debug, error, info, info_span, instrument, span, span::EnteredSpan, trace, warn, Level};
@@ -35,7 +35,7 @@ use self::version::GlVersion;
 
 use super::{
     sync::SyncPoint, Bind, Blit, Color32F, DebugFlags, ExportMem, Frame, ImportDma, ImportMem, Offscreen,
-    Renderer, Texture, TextureFilter, TextureMapping, Unbind,
+    Renderer, RendererSuper, Texture, TextureFilter, TextureMapping,
 };
 use crate::backend::egl::{
     ffi::egl::{self as ffi_egl, types::EGLImage},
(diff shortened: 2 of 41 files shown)


ids1024:
Have you seen this fix https://github.com/pop-os/cosmic-comp/issues/1078, or any similar issues, or are the synchronization issues this fixes currently theoretical and not reproducible?

Drakulix:
Yes I could reproduce it and the first commit already fixed that, given that this synchronizes the imports.

The second commit fixes synchronization across the board for shared contexts, but we haven't run into this in cosmic-comp as we don't bind shared-textures, but just render them.

ids1024:
Not having an `Rc<EGLSurface>` in the `target` field will be helpful if we want to make `GlesRenderer` implement `Send` (other than that, it has a pointer in `gl_debug_span`, `GlesBuffer::image`, and the `_not_send` member.) Not that I have a specific use for that, but it could be useful, if there are no issues with doing that.

ids1024:
Needing a `RendererSuper` is annoying, but I guess not that bad as workarounds go.

Cargo.toml
```diff
@@ -69,6 +69,7 @@ encoding_rs = { version = "0.8.33", optional = true }
 profiling = "1.0.13"
 smallvec = "1.11"
 pixman = { version = "0.2.1", features = ["drm-fourcc", "sync"], optional = true }
+aliasable = "0.1.3"
```
ids1024, line 72: If this is only used by the multigpu renderer, it should probably be an optional dependency enabled by `renderer_multi`.

Drakulix: done

Drakulix:
> I guess I might need to think about it for a while before I have any sense if the AliasableBox/transmute part is sound... (Not sure if this will also make abstracting gles/pixman more complicated too)

https://morestina.net/blog/1868/self-referential-types-for-fun-and-profit was the inspiration for this once I realized that this is simply self-referential.

Though I'll leave seeing though those lifetime to you. I have stared long enough at the code to be confident, that it is sound, but I would really like somebody to come to the same conclusion. :sweat_smile:

nferhat:
I think `RenderContext::draw` should get a `FnOnce(&mut T)` instead of `FnOnce(&T)`
https://github.com/Smithay/smithay/blob/fe31867e3afac2543c4016fb8ed99df3e11eb6da/src/backend/renderer/element/texture.rs#L594

src/backend/renderer/multigpu/mod.rs
```diff
@@ -753,28 +755,50 @@ impl<R: GraphicsApi, T: GraphicsApi> AsMut<<R::Device as ApiDevice>::Renderer>
...
-    <<R::Device as ApiDevice>::Renderer as Renderer>::Error: 'static,
-    <<T::Device as ApiDevice>::Renderer as Renderer>::Error: 'static,
+    <<R::Device as ApiDevice>::Renderer as RendererSuper>::Error: 'static,
+    <<T::Device as ApiDevice>::Renderer as RendererSuper>::Error: 'static,
 {
     node: DrmNode,
-    frame: Option<<<R::Device as ApiDevice>::Renderer as Renderer>::Frame<'frame>>,
+    frame: Option<<<R::Device as ApiDevice>::Renderer as RendererSuper>::Frame<'frame, 'buffer>>,
+    framebuffer:
+        Option<AliasableBox<<<R::Device as ApiDevice>::Renderer as RendererSuper>::Framebuffer<'frame>>>,
```
ids1024, line 799: There should probably be a comment mentioning that `frame` borrows `framebuffer`, and the order of the fields should be preserved to maintain the drop order.

ids1024:
Okay, after reading through https://morestina.net/blog/1868/self-referential-types-for-fun-and-profit, and looking through that part of the code here, I understand why this makes the struct self-referential, and it seems like a relatively straightforward case of that. As the article mentions, `transmute` isn't *as* scary as usual when it's just extending the lifetime.

So I think that part of the code looks correct, as does everything else here.

Alternatives:
- It should be possible to use `ouroboros`, but that's not especially more convenient or understandable, and would add more compile time in macros, only to generate something similar use `aliasable`. (As much as I'd like to just trust a widely used crate with the `unsafe` part.)
- I guess we could avoid this problem by making `Renderer::render` take a callback instead of returning a `Frame<'_, '_>`. But the existing API is somewhat more convenient.

So overall, this seems like a good approach. Hopefully Rust will have a better way to deal with self-referential types at some point.

src/backend/renderer/pixman/mod.rs
```diff
@@ -1216,26 +1204,25 @@ impl Bind<Dmabuf> for PixmanRenderer {
...
+        Ok(image)
     }
 }
 
-impl Bind<PixmanRenderBuffer> for PixmanRenderer {
-    #[profiling::function]
-    fn bind(&mut self, target: PixmanRenderBuffer) -> Result<(), <Self as Renderer>::Error> {
-        self.target = Some(PixmanTarget::RenderBuffer(target));
-        Ok(())
+impl Bind<Image<'static, 'static>> for PixmanRenderer {
```
cmeissl, line 1222: We should allow to bind an `Image` with non `'static` data lifetime. This would allow callers to bind `byte[]` slices in a safe way without having to rely on the ugly hack we do in the `DrmCompositor` for binding the dumb buffer slice.

src/backend/renderer/mod.rs
```diff
@@ -103,27 +105,23 @@ impl From<wayland_server::protocol::wl_output::Transform> for Transform {
...
-    /// Binding to target, while another one is already bound, is rendering defined.
-    /// Some renderers might happily replace the current target, while other might drop the call
-    /// or throw an error.
-    fn bind(&mut self, target: Target) -> Result<(), <Self as Renderer>::Error>;
+    /// The `output_size` specifies the dimensions of the display **before** the `dst_transform` is
+    /// applied.
+    ///
+    /// This function *may* error, if:
+    /// - The given dimensions are unsupported (too large) for this renderer
+    /// - The given Transformation is not supported by the renderer (`Transform::Normal` is always supported).
```
cmeissl, line 116: Is this a left-over from the `render` refactor? If we have some constraints between `bind` and `render` we should describe them, but `output_size` is not part of this function signature.

src/backend/renderer/multigpu/mod.rs
Drakulix: done

src/backend/renderer/gles/mod.rs
```diff
@@ -1334,53 +1394,36 @@ impl ExportMem for GlesRenderer {
...
 }
 
 impl Bind<Dmabuf> for GlesRenderer {
-    #[instrument(level = "trace", parent = &self.span, skip(self))]
-    #[profiling::function]
-    fn bind(&mut self, dmabuf: Dmabuf) -> Result<(), GlesError> {
+    fn bind<'a>(&mut self, dmabuf: &'a mut Dmabuf) -> Result<GlesTarget<'a>, GlesError> {
         unsafe {
             self.egl.make_current()?;
         }
```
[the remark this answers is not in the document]
Drakulix: done

src/backend/renderer/pixman/mod.rs
Drakulix: I couldn't find a way to introduce the necessary bound of `Image<'a, 'a>` outlives `PixmanTarget<'b>`. I want to add a `where 'a: 'b`-bound to `bind`, but that isn't compatible with the trait definition.

Adding `'b` to `Bind<'b>` doesn't help, as we still can't express the bound in the trait, `Target: 'b` doesn't help.

Nor does `where Target: 'a` even though that should imply, that `bind` should only be available if `Image<'b, 'b>` outlives `PixmanTarget<'a>`. The compiler doesn't seem smart enough.

Adding another lifetime to `PixmanTarget` makes it impossible to implement `RendererSuper`.

I couldn't find a way to specify the additional requirement with an higher-order bound (`for<'b>`), but that feels like what is necessary here.

src/backend/renderer/mod.rs
Drakulix: fixed

src/backend/renderer/pixman/mod.rs
cmeissl: I see, that's quite unfortunate. I also unsuccessfully tried to find a safe way to express this.
Would be really great if we could use this to basically bind a scoped slice  :/

Theoretically is should just work as the `'bits` inside of the `Image` are already guaranteed to outlive any reference to the `Image`. In `PixmanTarget`/`PixmanTargetInternal` we only care that the `'bits` of the `Image` live at least as long as the reference of the image itself. So the only thing blocking this is that we can not prove it to the compiler. I wonder if it would be safe to just cast the lifetime in bind.

cmeissl: Quick test of the change: https://github.com/cmeissl/smithay/commit/0e5eee5f59693907f87becfc71c6dba32db8e999

cmeissl (approved):

PR #1641, opened 2025-01-22
Add primary-selection filter

notpeelz:
I'm working on a PR for niri to disable the primary selection. I've had to make some changes in Smithay in order to support disabling of the primary selection at runtime (live-reloading of the config).

Niri PR: https://github.com/YaLTeR/niri/pull/1011

@kchibisov I was told you were the last person who reworked clipboard handling in Smithay, and that you might want to take a look at these changes.

### DIFF
--- a/src/wayland/selection/primary_selection/mod.rs
+++ b/src/wayland/selection/primary_selection/mod.rs
@@ -103,10 +103,25 @@ impl PrimarySelectionState {
     /// Register new [`PrimaryDeviceManager`] global
     pub fn new<D>(display: &DisplayHandle) -> Self
     where
-        D: GlobalDispatch<PrimaryDeviceManager, ()> + 'static,
+        D: GlobalDispatch<PrimaryDeviceManager, PrimaryDeviceManagerGlobalData> + 'static,
         D: PrimarySelectionHandler,
     {
-        let manager_global = display.create_global::<D, PrimaryDeviceManager, _>(1, ());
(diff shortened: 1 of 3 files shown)


kchibisov:
If the filter is done the same way as any other filter, I'm not sure there's anything needed from my end, since it doesn't touch anything in clipboard handling itself.

src/wayland/selection/primary_selection/mod.rs
```diff
@@ -94,21 +94,32 @@ pub trait PrimarySelectionHandler: Sized + SeatHandler + SelectionHandler {
...
+impl std::fmt::Debug for PrimarySelectionState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PrimarySelectionState").field("manager_global", &self.manager_global).finish()
+    }
 }
 
 impl PrimarySelectionState {
     /// Register new [`PrimaryDeviceManager`] global
-    pub fn new<D>(display: &DisplayHandle) -> Self
+    pub fn new<D, F>(display: &DisplayHandle, filter: F) -> Self
```
PolyMeilex, line 110: No need to brake and complicate the API for such an esoteric use case. Let's just add another function `with_filter` like so: https://github.com/Smithay/smithay/blob/953959e6069b3e14dba96fdaa46c65990c21d5c9/src/wayland/shell/xdg/decoration.rs#L117-L155

Drakulix (approved):
Thanks!

Issue #1604, opened 2024-12-11
Double input with fcitx5 on chrome dev under some cases


CnTeng:
This was happend for me in niri and cosmic.

Chrome which utilizes text-input-v3 has the issue of letters get doubly input for certain input boxes, even when no ime is in use (but no issues when the process of fcitx5 is being killed).

Related issue: https://github.com/YaLTeR/niri/issues/620#issue-2481371388
https://issues.chromium.org/issues/40113488#comment127

merrkry:
I can reproduce this, and here's a tldr:

When having multiple chromium / electron (e.g. multiples windows from the same electron application) window running, every character typed will be repeated in current input focus field, as many times as the number of chromium/electron instance.

These chromium flags are used:

```
--enable-wayland-ime --wayland-text-input-version=3
```

hayandev:
Although it doesn’t fully adhere to the text input v3 specification, it appears to work reasonably well with minimal code changes.
The addition of a pending status required by the protocol specification is mentioned in another existing issue(https://github.com/Smithay/smithay/issues/1614), so presumably whoever is looking at that issue will address it separately.

PR #1642, opened 2025-01-22
Fix repeated key input issue in Chrome with multiple windows

hayandev:
This PR fixes issue https://github.com/Smithay/smithay/issues/1604

According to the text input v3 specification, the enable request should activate at most one text input on the compositor and ignore any subsequent enable requests until a disable request is received.

However, currently, Smithay does not guarantee this behavior and sends events to all text input objects created by the same client.

We will modify it to send events only to the text input object that has received the enable request and update the management of the enabled text input objects.

### DIFF
--- a/src/wayland/text_input/text_input_handle.rs
+++ b/src/wayland/text_input/text_input_handle.rs
@@ -2,7 +2,7 @@ use std::sync::{Arc, Mutex};
 
 use tracing::debug;
 use wayland_protocols::wp::text_input::zv3::server::zwp_text_input_v3::{self, ZwpTextInputV3};
-use wayland_server::backend::ClientId;
+use wayland_server::backend::{ClientId, ObjectId};
 use wayland_server::{protocol::wl_surface::WlSurface, Dispatch, Resource};
 
 use crate::input::SeatHandler;
@@ -21,6 +21,7 @@ struct Instance {
 pub(crate) struct TextInput {
     instances: Vec<Instance>,
     focus: Option<WlSurface>,
+    enabled_resource_id: Option<ObjectId>,
 }
 
 impl TextInput {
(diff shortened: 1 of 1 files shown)


src/wayland/text_input/text_input_handle.rs
```diff
@@ -77,6 +81,18 @@ impl TextInputHandle {
...
+    fn set_enabled_resource_id(&self, resource_id: Option<ObjectId>) {
+        let mut inner = self.inner.lock().unwrap();
+
+        match (&inner.enabled_resource_id, &resource_id) {
+            (Some(_), Some(_)) => {}
+            (None, None) => {}
+            (_, _) => {
+                inner.enabled_resource_id = resource_id;
+            }
+        }
```
Drakulix, line 93: Maybe that is just me, but I find this easier to read:

```suggestion
        if inner.enabled_resource_id.is_some() != resource_id.is_some() {
            inner.enabled_resource_id = resource_id;
        }
```

Drakulix (approved):
Looks good to me, thanks for fixing this.
Found one minor nitpick.

Drakulix:
thanks!

kchibisov:
After this commit fcitx text_input is straight doesn't work at all, juts gets instantly disabled and that's about it.

src/wayland/text_input/text_input_handle.rs
```diff
@@ -77,6 +81,13 @@ impl TextInputHandle {
         self.inner.lock().unwrap().focus = surface;
     }
 
+    fn set_enabled_resource_id(&self, resource_id: Option<ObjectId>) {
+        let mut inner = self.inner.lock().unwrap();
+        if inner.enabled_resource_id.is_some() != resource_id.is_some() {
+            inner.enabled_resource_id = resource_id;
+        }
+    }
+
```
kchibisov, line 90: It works the other way around, the IME should state that it's ready meaning that it'll send the `Entered` to the client and it'll react with `enabled` and only one of them must be honored IIRC, however with this check it won't ever advertise since it won't be enabled until you actually call `enabled`, which means with this PR it won't ever happen, thus it will never work at all, thus not sure how it could ever worked, until the client just enables it because it _can_.

hayandev: I'm sorry, you are right. I was focused on the behavior of enable and the input bug related to Chrome. I tested it once with Qt and Gtk apps (excluding Chrome), and it worked well, so I thought it was working correctly (as shown in the attached image).

image

I just tested it several times with Gtk apps, and while it works in some cases, it fails more often, and as you mentioned, it's unclear why it works at all.

I would appreciate it if you could revert the commit. I apologize again for not testing more thoroughly.

kchibisov: I'll fix myself, don't worry.

kchibisov: Qt doesn't support text-input v3, so it worked just via dbus, I guess, and maybe some gtk apps worked that way as well, on a pure text-input-v3 system it couldn't possible work, just for the reference.

PR #1646, opened 2025-01-24
rename Rectangle::from_extemities to Rectangle::from_extremeties

m4rch3n1ng:
fixes a typo in the function name: "extremities" is the correct name, "extemities" is just a misspelling

the function is not that common to call from outside of smithay, anvil don't use it at all, cosmic uses it exactly once and niri only three times, but it is still a breaking change, so i just deprecated it and hid it on the docs instead of removing it.


PolyMeilex (approved):

PolyMeilex:
Thanks!

YaLTeR:
The PR title itself contains a typo in the new name which threw me off, but the diff seems to be correct

PR #1648, opened 2025-01-24
text-input: fix active instance tracking

### LINKED ISSUE #1604: Double input with fcitx5 on chrome dev under some cases

kchibisov:
Only one text-input must be enabled at a time, by checking the text-input id when issuing a request and comparing to the active one, however the regular `Enabled` events must be send for all of them, and as for the leave, we also send it for all the events to make it symmetrical with the `enter.

Fixes: 475072d410df (text-input: Ensure only one enabled...)
Fixes #1604.

--

Should be backported to patch release, otherwise text-input doesn't work at all.

### DIFF
--- a/src/wayland/input_method/input_method_handle.rs
+++ b/src/wayland/input_method/input_method_handle.rs
@@ -211,7 +211,7 @@ where
     ) {
         match request {
             zwp_input_method_v2::Request::CommitString { text } => {
-                data.text_input_handle.with_focused_text_input(|ti, _surface| {
+                data.text_input_handle.with_active_text_input(|ti, _surface| {
                     ti.commit_string(Some(text.clone()));
                 });
             }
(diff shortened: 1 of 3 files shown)


kchibisov:
Could you test this @hayandev with chrome?

hayandev:
> Could you test this @hayandev with chrome?

It works great!
Thank you for fixing the code I made a mistake in. I’ve just started developing a Linux desktop environment.
so, I would appreciate your understanding and forgiveness. Thank you so much for creating such a fantastic project.

Drakulix (approved):
Thanks!

PR #1657, opened 2025-02-08
utils: Fix `HookId` recycle logic

Paraworker:
Currently, the `usize` ID is recycled in `Hook`’s destructor. However, since `Hook` also implements `Clone`, the `usize` ID might be recycled more than once. And `HookId` might still be used even after the `usize` ID has already been recycled.

This PR moves the ID recycling to `HookId`’s destructor. The trade-off is that `HookId` can't be `Copy`.

Actually, I’m not sure if we even need a recyclable ID implementation, it just adds extra complexity. If we’re using a `u64` for the ID, running out of values shouldn’t really be a concern.

### DIFF
--- a/src/utils/hook.rs
+++ b/src/utils/hook.rs
@@ -3,8 +3,8 @@ use std::sync::Arc;
 crate::utils::ids::id_gen!(hooks_id);
 
 /// Unique hook identifier used to unregister commit/descruction hooks
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
-pub struct HookId(usize);
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct HookId(Arc<InnerId>);
 
 pub(crate) struct Hook<T: ?Sized> {
     pub id: HookId,
@@ -22,7 +22,7 @@ impl<T: ?Sized> std::fmt::Debug for Hook<T> {
 impl<T: ?Sized> Clone for Hook<T> {
     fn clone(&self) -> Self {
         Self {
-            id: self.id,
+            id: self.id.clone(),
             cb: self.cb.clone(),
         }
     }
@@ -31,14 +31,23 @@ impl<T: ?Sized> Clone for Hook<T> {
 impl<T: ?Sized> Hook<T> {
     pub fn new(cb: Arc<T>) -> Self {
         Self {
-            id: HookId(hooks_id::next()),
+            id: HookId(Arc::new(InnerId::new())),
             cb,
         }
     }
 }
 
-impl<T: ?Sized> Drop for Hook<T> {
+#[derive(Debug, Eq, PartialEq)]
+struct InnerId(usize);
+
+impl InnerId {
+    fn new() -> Self {
+        Self(hooks_id::next())
+    }
+}
+
+impl Drop for InnerId {
     fn drop(&mut self) {
-        hooks_id::remove(self.id.0);
+        hooks_id::remove(self.0);
     }
 }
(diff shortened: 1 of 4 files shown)


PolyMeilex (approved):

jjramsey:
I tried compiling Niri (particularly this fork of it: https://github.com/jjramsey/niri/tree/update-smithay-for-idle-notify-fix) against an updated Smithay and got the following error messages that may be relevant to this MR, especially to the part where you mention "The trade-off is that `HookId` can't be `Copy`.":
```
error[E0505]: cannot move out of `self.pre_commit_hook` because it is borrowed
   --> src/window/mapped.rs:469:62
    |
468 |     fn drop(&mut self) {
    |             --------- binding `self` declared here
469 |         remove_pre_commit_hook(self.toplevel().wl_surface(), self.pre_commit_hook);
    |         ---------------------- ----                          ^^^^^^^^^^^^^^^^^^^^ move out of `self.pre_commit_hook` occurs here
    |         |                      |
    |         |                      borrow of `*self` occurs here
    |         borrow later used by call

error[E0507]: cannot move out of `self.pre_commit_hook` which is behind a mutable reference
   --> src/window/mapped.rs:469:62
    |
469 |         remove_pre_commit_hook(self.toplevel().wl_surface(), self.pre_commit_hook);
    |                                                              ^^^^^^^^^^^^^^^^^^^^ move occurs because `self.pre_commit_hook` has type `HookId`, which does not implement the `Copy` trait
    |
help: consider cloning the value if the performance cost is acceptable
    |
469 |         remove_pre_commit_hook(self.toplevel().wl_surface(), self.pre_commit_hook.clone());
    |                                                                                  ++++++++
```
I don't know enough about either the HookId or Niri to know if changing `self.pre_commit_hook` to `self.pre_commit_hook.clone()` is the right choice here, or if that would introduce a bug in Niri.

Tagging @YaLTeR because this should be relevant to him, and he might be able to answer some questions.

YaLTeR:
Seems like we need to update niri accordingly, yes

Paraworker:
> I tried compiling Niri (particularly this fork of it: https://github.com/jjramsey/niri/tree/update-smithay-for-idle-notify-fix) against an updated Smithay and got the following error messages that may be relevant to this MR, especially to the part where you mention "The trade-off is that `HookId` can't be `Copy`.":
> 
> ```
> error[E0505]: cannot move out of `self.pre_commit_hook` because it is borrowed
>    --> src/window/mapped.rs:469:62
>     |
> 468 |     fn drop(&mut self) {
>     |             --------- binding `self` declared here
> 469 |         remove_pre_commit_hook(self.toplevel().wl_surface(), self.pre_commit_hook);
>     |         ---------------------- ----                          ^^^^^^^^^^^^^^^^^^^^ move out of `self.pre_commit_hook` occurs here
>     |         |                      |
>     |         |                      borrow of `*self` occurs here
>     |         borrow later used by call
> 
> error[E0507]: cannot move out of `self.pre_commit_hook` which is behind a mutable reference
>    --> src/window/mapped.rs:469:62
>     |
> 469 |         remove_pre_commit_hook(self.toplevel().wl_surface(), self.pre_commit_hook);
>     |                                                              ^^^^^^^^^^^^^^^^^^^^ move occurs because `self.pre_commit_hook` has type `HookId`, which does not implement the `Copy` trait
>     |
> help: consider cloning the value if the performance cost is acceptable
>     |
> 469 |         remove_pre_commit_hook(self.toplevel().wl_surface(), self.pre_commit_hook.clone());
>     |                                                                                  ++++++++
> ```
> 
> I don't know enough about either the HookId or Niri to know if changing `self.pre_commit_hook` to `self.pre_commit_hook.clone()` is the right choice here, or if that would introduce a bug in Niri.
> 
> Tagging @YaLTeR because this should be relevant to him, and he might be able to answer some questions.
> 
> ETA: After digging a bit further into the Smithay code, it looks like using `.clone()` would be okay. Before the change to the recycle logic, a copy of `self.pre_commit_hook`, which was more-or-less an integer (technically, a tuple struct whose only field was an integer), was passed to `remove_pre_commit_hook()`. After the change, `self.pre_commit_hook.clone()` would essentially be a reference to that integer (again, glossing over some technicalities). I could be wrong, of course, since I'm pretty new to Rust.

Using `clone()` should always work where passing by value is required.

Issue #1422, opened 2024-05-07
`KbdRc` and `SeatRc` are leaked


ids1024:
Looking at `valgrind` logs of anvil, it seems allocations that are part of `KbdRc` are never freed when exiting cleanly. It seems `Drop` is never called on `KbdRc` or `SeatRc`.

I think `known_kbds` needs to use `Weak` to avoid a reference cycle. I'm still unsure what is preventing `SeatRc` from being freed...

ids1024:
Looks like `SeatGlobalData` and `SeatUserData` never get dropped.

ids1024:
Or I guess `SeatUserData` is freed when a client exist before the compositor, but not when the compositor exits with a window open?

PolyMeilex:
This might have been fixed by: #1516

PR #1516, opened 2024-08-27
Fixup Seat ref count cycles

PolyMeilex:
Before:
```mermaid
flowchart TD
    WlKeyboard ---> KeyboardUserData --> KeyboardHandle --> KbdRc --> WlKeyboard
    WlPointer ---> PointerUserData --> PointerHandle --> WlPointerHandle  --> WlPointer
    WlTouch ---> TouchUserData --> TouchHandle ---> WlTouch
```
After:
```mermaid
flowchart TD
    WlKeyboard ---> KeyboardUserData --> KeyboardHandle --> KbdRc -.-> |Weak| WlKeyboard
    WlPointer ---> PointerUserData --> PointerHandle --> WlPointerHandle  -.-> |Weak| WlPointer
    WlTouch ---> TouchUserData --> TouchHandle -.-> |Weak| WlTouch
```

PR #1663, opened 2025-02-25
Fix leaks of `SeatRc` and `KbdRc`, but using `Weak` in user data

ids1024:
This fixes https://github.com/Smithay/smithay/issues/1422.

This can be tested by adding `Drop` impls that print to those two types, or using tools that detect leaks.

It's probably worth examining more to see why this created a cycle or otherwise didn't get freed, but if we don't anticipate issues, maybe it's best to use `Weak` in every possible place. I also need to look into some other reported leaks.

### DIFF
--- a/src/wayland/seat/keyboard.rs
+++ b/src/wayland/seat/keyboard.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{fmt, sync::Weak};
 
 use tracing::{error, instrument, trace, warn};
 use wayland_server::{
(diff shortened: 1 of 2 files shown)


ids1024:
Updated now that calloop is released. LeakSanitizer in Anvil shows fewer leaks now.

Drakulix:
Any particular reason why this only addresses the `KeyboardHandle` and not `PointerHandle` and `TouchHandle` as well?

ids1024:
Good point.

Actually, that reference cycle may have been fixed a different way by https://github.com/Smithay/smithay/pull/1516 (I may do a bit more testing).

MemorySanitizer does show the calloop change addresses some leaks appearing there.

ids1024:
Ah, I see. So https://github.com/Smithay/smithay/pull/1516 changes `known_kbds` and the like to use `wayland_server::Weak`, while the change here uses a `std::sync::Weak` in `KeyboardUserData` and `SeatUserData`. Either prevents a leak, and both also works (using weak references for both doesn't cause early drops of anything since Wayland objects aren't destroyed on drop).

May be best to still make this change, as well as a similar change for pointer and touch. Avoiding strong references like this in user datas seems good in general.

PR #1664, opened 2025-02-26
fix: return the geometry of the layer surface instead of bbox

wash2:
I'm not too familiar with the smithay code base, but I believe the geometry should not include the subsurfaces. It causes layer surfaces to be arranged as if they have a larger size than they do, if they have a subsurface that extends beyond the layer shell surfaces. If the subsurface has a negative x or y position, the issue becomes worse, because when the drawn size is increased, the subsurface is not even drawn in the allotted extra space.


wash2:
> > It causes layer surfaces to be arranged as if they have a larger size than they do, if they have a subsurface that extends beyond the layer shell surfaces.
> 
> I believe that's the correct behavior though? Subsurfaces are intended to be an indivisible part of the "window" (in this case, layer). For toplevel windows, if they don't explicitly set the geometry, they also assume a bbox across the surface tree. Layer surfaces don't have a protocol to set the geometry, so they always have to assume a bbox.

Ok, if that is the correct behavior according to the protocol, I get it. For what it is worth, the use case is tooltips that may extend beyond the bounds of the layer shell surface.

The second issue I listed is still a problem in my opinion. If it is increasing the size of the surface, but placing the subsurface such that it is outside the expanded bounds anyway, then that should be fixed, though, it needs to be done differently than how I've done it.

YaLTeR:
> the use case is tooltips

They should use xdg-popup, which is not a subsurface.

Drakulix:
After reading through all the relevant protocols here again, I think I'll have to agree with @YaLTeR. The full bbox seems to be what is implied by the protocol to be used for arrangement.

Regarding sub-surfaces in general, with xdg-shell the compositor is allowed to clip contents outside of the geometry. But if we assume the bbox includes all subsurfaces as well, then negative subsurfaces should shift the whole layer-shell surface origin.
In any case smithay shouldn't do any clipping, but leave that to an actual compositor.

So these things need fixing.

As for tooltip, yeah they should probably be xdg-popups.

YaLTeR:
> Regarding sub-surfaces in general, with xdg-shell the compositor is allowed to clip contents outside of the geometry. But if we assume the bbox includes all subsurfaces as well, then negative subsurfaces should shift the whole layer-shell surface origin.

Yeah, that is supposed to happen from my understanding. If you don't set geometry on an xdg-toplevel, the same thing will happen.

What needs fixing here is layer-shell should get a geometry request ideally..

Drakulix:
> What needs fixing here is layer-shell should get a geometry request ideally..

Well... https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/28/diffs#note_1341317 (I'd encourage you commenting)

PR #1667, opened 2025-03-01
gles/texture: Add is_unique_reference()

YaLTeR:
Intended for usecases like `TextureRenderBuffer` where the texture contents can change. If someone else references the same texture (for example, someone stored the texture render element for later use), the texture contents would change under the hood for them. This method lets the compositor check that nobody else has a reference to the same texture, making it safe to change its contents.

This doesn't (cannot) track extra graphics API references to the same texture object, but those are harder to make and less common; the compositor can figure out the uniqueness on its own then.

Questions for review:
1. Do I understand correctly that the TTY/winit backends don't keep other references to the textures? E.g. they don't create raw OpenGL references that go around this `Arc`.
2. Does the same function make sense for `MultiTexture` and/or `PixmanTexture`? I wasn't sure because those involve multiple texture objects across different graphics APIs.

### DIFF
--- a/src/backend/renderer/gles/texture.rs
+++ b/src/backend/renderer/gles/texture.rs
@@ -48,6 +48,15 @@ impl GlesTexture {
     pub fn is_y_inverted(&self) -> bool {
         self.0.y_inverted
     }
+
+    /// Whether this is the only reference to this texture (strong or weak)
+    ///
+    /// Note that this tracks only references to this Smithay object (that you can get by cloning
+    /// it). If you make a reference via OpenGL directly somehow, you need to keep track of it on
+    /// your own.
+    pub fn is_unique_reference(&mut self) -> bool {
+        Arc::get_mut(&mut self.0).is_some()
+    }
 }
 
 #[derive(Debug)]


cmeissl:
>     1. Do I understand correctly that the TTY/winit backends don't keep other references to the textures? E.g. they don't create raw OpenGL references that go around this `Arc`.

For textures created through the `Offscreen` Trait this is true, yes. Textures created though the `Import` Traits might be referenced be the `Renderer` internally for purposed like caching.
 
>     2. Does the same function make sense for `MultiTexture` and/or `PixmanTexture`? I wasn't sure because those involve multiple texture objects across different graphics APIs.

This is a bit harder to answer imo. I had a superficial look at the offscreen branch in niri and I think we should provide abstractions in smithay to make this use-case easier. What I had in mind for some time already is a swapchain based element similar to the memory element we already provide. 

Outline of the idea:
- Element internally hosting a `Swapchain` based on `DmabufAllocator`.
- Providing a function to aquire and render to a swapchain `Slot`, returning a render element holding on the `Slot` and importing the `Dmabuf`.
- RenderElement can use the imported dmabuf for sampling, but could also return the `Slot<Dmabuf>` as the backing store for direct scan-out
(Note: This will require making `Slot` clone or wrapping it into something that can be cloned e.g. Arc<Slot<Dmabuf>>)

So imo adding this function to more types is not necessary

cmeissl (approved):

PR #1669, opened 2025-03-07
renderer/memory: Fix resize not updating internal size

Drakulix:


cmeissl (approved):
I wonder how that got unnoticed for so long, thanks!

One thing I noticed is that we might miss some possible optimization for the unchanged case.  We unconditionally use Arc::make_mut which could end in clone if there is some strong reference active.

Drakulix:
> One thing I noticed is that we might miss some possible optimization for the unchanged case. We unconditionally use Arc::make_mut which could end in clone if there is some strong reference active.

Fixed

cmeissl (approved):
Thanks 😊

PR #1670, opened 2025-03-09
add get client pid to x11 surface

PapyElGringo:
In order to be able to match a Surface to the `.desktop` file that started it retrieving the proper surface PID is mandatory.

Currently the `X11Surface.pid()` return the underlying `_NET_WM_PID` artifact which is not robust enough as the `Signal` Flatpak App return "2" for exemple while the new `X11Surface.get_client_pid` return the proper PID.

I took inspiration of the following [mutter code](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/x11/window-x11.c#L1859-1893)
```
static pid_t
meta_window_x11_get_client_pid (MetaWindow *window)
{
  MetaX11Display *x11_display = window->display->x11_display;
  xcb_connection_t *xcb = XGetXCBConnection (x11_display->xdisplay);
  xcb_res_client_id_spec_t spec = { 0 };
  xcb_res_query_client_ids_cookie_t cookie;
  xcb_res_query_client_ids_reply_t *reply = NULL;

  spec.client = meta_window_x11_get_xwindow (window);
  spec.mask = XCB_RES_CLIENT_ID_MASK_LOCAL_CLIENT_PID;

  cookie = xcb_res_query_client_ids (xcb, 1, &spec);
  reply = xcb_res_query_client_ids_reply (xcb, cookie, NULL);

  if (reply == NULL)
    return 0;

  uint32_t pid = 0, *value;
  xcb_res_client_id_value_iterator_t it;
  for (it = xcb_res_query_client_ids_ids_iterator (reply);
       it.rem;
       xcb_res_client_id_value_next (&it))
    {
      spec = it.data->spec;
      if (spec.mask & XCB_RES_CLIENT_ID_MASK_LOCAL_CLIENT_PID)
        {
          value = xcb_res_client_id_value_value (it.data);
          pid = *value;
          break;
        }
    }

  free (reply);
  return (pid_t) pid;
}
```

The code use the X11 `Res` protocol to query the `LOCAL_CLIENT_PID` for a X11 window.

### DIFF
--- a/src/utils/sealed_file.rs
+++ b/src/utils/sealed_file.rs
@@ -56,7 +56,7 @@ impl SealedFile {
         use rand::{distributions::Alphanumeric, Rng};
         use rustix::{
             io::Errno,
-            shm::{Mode, ShmOFlags},
+            shm::{self, Mode},
         };
 
         let mut rng = rand::thread_rng();
@@ -68,9 +68,9 @@ impl SealedFile {
             let mut shm_name = name.to_bytes().to_owned();
             shm_name.push(b'-');
             shm_name.extend((0..7).map(|_| rng.sample(Alphanumeric)));
-            let fd = rustix::shm::shm_open(
+            let fd = shm::open(
                 shm_name.as_slice(),
-                ShmOFlags::RDWR | ShmOFlags::CREATE | ShmOFlags::EXCL,
+                shm::OFlags::RDWR | shm::OFlags::CREATE | shm::OFlags::EXCL,
                 Mode::RWXU,
             );
             if !matches!(fd, Err(Errno::EXIST)) || n > 3 {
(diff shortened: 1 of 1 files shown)


Drakulix:
> The code use the X11 `Res` protocol to query the `LOCAL_CLIENT_PID` for a X11 window.

Did you find some SPEC that lists what exactly the `LOCAL_CLIENT_PID` referrs to?

I assume the `LOCAL`-part is concerned with x11 over network, which we do not have to worry about. However if the PID is queried and returned by the client, this smells like it could go wrong and many ways with process namespaces.

PapyElGringo:
Here is the protocol description https://www.x.org/releases/X11R7.7/doc/resourceproto/resproto.txt but I don't know if it holds the informations you are looking for.

But this protocol is used by Mutter to address the same need.

What's could go wrong exactly?

PapyElGringo:
Sorry I probably made a wrong manipulation trying to rebase and I had to make a new PR https://github.com/Smithay/smithay/pull/1741

PR #1683, opened 2025-03-18
renderer: Require `Texture` for framebuffer types

Drakulix:
I think it is pretty reasonable to be able to query the format and size of a bound framebuffer.

This is particularly useful in code that might now get passed in an opaque framebuffer handle, but needs to know the format/size to allocate an offscreen texture for post-processing. Sending this data out of band seems entirely unnecessary.

Bike-shedding: Should we rename the `Texture` trait at this point?


ids1024:
This seems like a good idea.

> Bike-shedding: Should we rename the Texture trait at this point?

I don't really know what name would be better. Calling the framebuffer a type of "texture" isn't really wrong in general (whether or not it's a "texture" in the terminology of the specific render backend). `Buffer` could describe this trait as well, but we already have two `struct Buffer` and one `trait Buffer` in other smithay modules.

cmeissl (approved):
I agree with @ids1024 that `Texture` is suitable for frambuffers too, so LGTM

PR #1685, opened 2025-03-19
fix: Catch underflow in `saturating_sub_timespec`

Drakulix:
Turns out `Timespec` stores it's internal representation in signed integers.

This caused `secs.checked_sub(1)` to return `-1`, which then gets cast into `u64::MAX`.

See: https://github.com/pop-os/cosmic-comp/issues/1062#issuecomment-2735450019

This should fix it.


PolyMeilex (approved):
Nice catch

cmeissl:
I will do a quick check that this does not break on arm or aarch64.

cmeissl (approved):

PR #1690, opened 2025-03-22
handlers: add sanity check for neg width / height

### LINKED ISSUE #1689: panics on negative size request from wayland client
[trace.txt](https://github.com/user-attachments/files/19405563/trace.txt)

nnyyxxxx:
fixes #1689

### DIFF
--- a/src/wayland/compositor/handlers.rs
+++ b/src/wayland/compositor/handlers.rs
@@ -301,6 +301,10 @@ where
                 }
             }
             wl_surface::Request::DamageBuffer { x, y, width, height } => {
+                if width < 0 || height < 0 {
+                    return;
+                }
+
                 PrivateSurfaceData::with_states(surface, |states| {
                     states
                         .cached_state


cmeissl:
I am not sure just ignoring the complete request is the right approach. Clients sending invalid data shouldn't bring down the compositor, but normally would receive a protocol error.

Like said in the linked issue I would also like to take a look at how this happens.

cmeissl:
There is no error defined for this and wlroots also seems to just ignore the whole damage request for negative sizes. Nothing bad should happen, but I would still like to know where this originates so that it can be reported and fixed client side.

nnyyxxxx:
[log.txt](https://github.com/user-attachments/files/19405762/log.txt)

cmeissl:
Thanks, the relevant part of the log file is

```
[1791838.112] {mesa egl surface queue}  -> wl_surface#39.damage_buffer(13, 74, -5, 50)
```

mesa does not send damage rects by itself afaik. A normal eglSwapBuffers would just result in `i32::MAX` for `width` and `height`. so someone is explicitly passing these values to either `eglSwapBuffersWithDamage` or queue present in vulkan.
from the log it looks like `Alacritty` is the client in question?

nnyyxxxx:
>  from the log it looks like Alacritty is the client in question?

yes

cmeissl:
Tried to reproduce this by resizing `Alacritty`, but had no luck during the quick test (version 0.14.0, 0.15.0, 0.15.1 and master). Maybe @kchibisov has an idea on how we end up with a negative width in the damage rects?

cmeissl (approved):
Thanks!

PR #1694, opened 2025-03-27
seat: Allow accessing `WlKeyboard`/`WlPointer` instances

Drakulix:
This allows implementing for examples Xwayland eavesdropping like KDE implements for global shortcuts: https://invent.kde.org/plasma/kwin/-/merge_requests/1950

### DIFF
--- a/src/utils/iter.rs
+++ b/src/utils/iter.rs
@@ -1,0 +1,71 @@
+/// Common iterator types
+use wayland_server::{backend::ClientId, Resource, Weak};
+
+use std::{fmt, sync::MutexGuard};
+
+/// Iterator helper over a mutex of client objects
+pub struct LockedClientObjsIter<'a, T: 'static, G, F> {
+    pub iterator: std::iter::FilterMap<std::slice::Iter<'static, Weak<T>>, F>,
+    pub guard: MutexGuard<'a, G>,
+}
+
+pub(crate) fn new_locked_obj_iter_from_vec<T: Resource + 'static>(
+    guard: MutexGuard<'_, Vec<Weak<T>>>,
+    client: ClientId,
+) -> impl Iterator<Item = T> + '_ {
+    new_locked_obj_iter(guard, client, |guard| guard.iter())
+}
+
+pub(crate) fn new_locked_obj_iter<
+    'a,
+    T: Resource + 'static,
+    G,
+    F: for<'b> FnOnce(&'b G) -> std::slice::Iter<'b, Weak<T>>,
+>(
+    guard: MutexGuard<'a, G>,
+    client: ClientId,
+    iterator_fn: F,
+) -> impl Iterator<Item = T> + 'a {
+    let iterator = unsafe {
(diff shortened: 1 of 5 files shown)


YaLTeR:
How frequently are these supposed to run? Because if frequently, then allocating a new Vec sounds not ideal

Drakulix:
> How frequently are these supposed to run? Because if frequently, then allocating a new Vec sounds not ideal

Yeah, we have the same issue with `client_outputs` for `Output`, because of the internal `Mutex`. I was hoping to just ignore this until [`MutexGuard::map`](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html#method.map) is stablized, but I guess we could create our own `Iterator`-type, that contains the `MutexGuard`?

Drakulix:
Okay, I added a commit with an entirely too complicated iterator type to avoid allocating the vector. Unfortunately we end up allocating a bunch of boxes instead, which still might be cheaper?

src/utils/iter.rs
```diff
@@ -0,0 +1,71 @@
+/// Common iterator types
+use wayland_server::{backend::ClientId, Resource, Weak};
+
+use std::{fmt, sync::MutexGuard};
+
+/// Iterator helper over a mutex of client objects
+pub struct LockedClientObjsIter<'a, T: 'static, G, F> {
+    pub iterator: std::iter::FilterMap<std::slice::Iter<'static, Weak<T>>, F>,
+    pub guard: MutexGuard<'a, G>,
+}
```
ids1024, line 10: So, `iterator` includes a borrow into `guard`? Is this sound under aliasing rules?

ids1024: (And presumably the fields shouldn't be `pub` here.)

Drakulix: Yes, because it actually has a pointer into the Mutex, which is inside the guard behind a reference. And because we don't modify the guard in any way the compiler won't allow moving the original mutex while it is borrowed.

Drakulix: > (And presumably the fields shouldn't be `pub` here.)

Doesn't make a difference due to this not being part of the public api, but good catch.

ids1024: Right, I guess the fields aren't exposed since this is only used in the API as `Into<Iterator>` (so, both the fields and the struct itself don't need to be `pub`.)

Yeah, I guess this isn't an aliasing issue, looking at how libstd defines `Mutex` and `MutexGuard`. Technically this may be an implementation detail (if there's no explicit guarantee that this is safe to do), but should be fine. Hopefully `MutexGuard::map` gets stabilized before *too* long.

Drakulix:
See https://github.com/pop-os/cosmic-comp/pull/1328/commits/235edd7b509bb544146c64e24f062e5f060ec486 for a usage example

PolyMeilex (approved):
Not a huge fan of all the hidden locks in the API, but there is not much we can do about that so LGTM

Issue #1695, opened 2025-03-28
Periodic 'Failed to initialize semaphore for plane fence' nvidia driver crash


brandoncazander:
Hello. I use smithay through niri and I've been encountering a very disruptive crash in the nvidia drivers that I haven't seen with any other desktop environments (wayland or X). The crash is not recoverable without rebooting through SysRq.

```
[drm:nv_drm_atomic_apply_modeset_config [nvidia_drm]] *ERROR* [nvidia-drm] [GPU ID 0x00000100] Failed to initialize semaphore for plane fence
[drm:nv_drm_atomic_commit [nvidia_drm]] *ERROR* [nvidia-drm] [GPU ID 0x00000100] Failed to apply atomic modeset.  Error code: -11
```

You can see the full crash and more details in my report to niri: https://github.com/YaLTeR/niri/issues/1339. It was suggested that I upstream this report to smithay.

It also seems very relevant to #1489, but I'm running a more modern nvidia driver that shouldn't need this fix. I built niri locally and bumped the smithay version check referenced in this PR to disable IN_FENCE but the crashes continued.

**System information**
* smithay commit: 0cd3345c59f7cb139521f267956a1a4e33248393
* niri version: niri 25.02 (v25.02)
* Distro: Ubuntu 24.04.2
* Kernel: 6.8.0-55-generic
* GPU: NVIDIA RTX A1000 Laptop
* Driver: nvidia-dkms-570 (570.124.04)
* CPU: Intel Core i7-12850HX
```
$ cat /etc/modprobe.d/nvidia-drm-modeset.conf 
options nvidia_drm modeset=1 fbdev=1
```

**Relevant issues**
https://github.com/YaLTeR/niri/issues/1339
https://github.com/NVIDIA/open-gpu-kernel-modules/issues/622
https://github.com/hyprwm/Hyprland/pull/7151

---

I understand these nvidia drivers are pretty problematic and it seems more like an issue with them than smithay, but am hoping I can maybe help contribute to understand the cause. I'm very motivated to fix this crash so that I can migrate over to niri, so if you have any debugging suggestions or suggestions for ways that I can help, please let me know!

Drakulix:
So given that this doesn't look like something crashing inside smithay, I am afraid there isn't much we can do here.

[...]

PR #1697, opened 2025-03-29
drm/compositor: allow to explicitly enable/disable explicit sync

cmeissl:
this moves the private decision for enabling explicit sync to a public capability enum, allowing to override it.

related to #1695

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -657,7 +657,7 @@ impl<B: Buffer, F: Framebuffer> FrameState<B, F> {
     fn test_state(
         &mut self,
         surface: &DrmSurface,
-        supports_fencing: bool,
+        capabilities: Capabilities,
         plane: plane::Handle,
         state: PlaneState<B, F>,
         allow_modeset: bool,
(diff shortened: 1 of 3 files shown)


src/backend/drm/compositor/mod.rs
```diff
@@ -1035,6 +1035,59 @@ bitflags::bitflags! {
...
+        })?;
+
+        // `IN_FENCE_FD` makes commit fail on Nvidia driver
+        // https://github.com/NVIDIA/open-gpu-kernel-modules/issues/622
+        let is_nvidia = driver.name().to_string_lossy().to_lowercase().contains("nvidia")
+            || driver
+                .description()
+                .to_string_lossy()
+                .to_lowercase()
+                .contains("nvidia");
```
orowith2os, line 1067: why not make this a debug flag or environment variable instead, like `SMITHAY_DEBUG_DISABLE_EXPLICIT_SYNC`? I don't think it should be on the compositor to deal with misbehaving drivers in this case.

i509VCB: I find env vars have their own ux issues since many people don't know of the probably dozen that exist which you could tune.

Also an explicit type (with a little more code) means you could switch at runtime between modes.

I could see the env var make sense for configuring the default value.

orowith2os: Both sounds good to me, if it can be shimmied in here.

cmeissl: I am not a big fan of environment variables when they are avoidable. Giving explicit control allows projects like niri to implement thst config option in a way more visible way.

The reason for keeping the nvidia check is simply to not break existing code.

But I am not opposed to adding an environment variable that overrides the default.

PR #1699, opened 2025-03-30
Omit type conversion in `AsRenderElements::render_elements()` #1626

kenoss:
This patch omits type conversion in
`AsRenderElements::render_elements()` and simplifies it.

Closes: #1626


kenoss:
See also #1700

Drakulix:
imo, this is the worst of both worlds, as we now have a guaranteed allocation. I think the issue of #1700 are fixable and the overall approach is simply better.

kenoss:
Thanks! Let's continue discussion on #1700.

PR #1700, opened 2025-03-30
[DO NOT MERGE] Alternative of #1699, option (C) in #1626

kenoss:
This shows how (C) in #1626 looks like.

### DIFF
--- a/src/desktop/space/element/mod.rs
+++ b/src/desktop/space/element/mod.rs
@@ -1,11 +1,7 @@
 use std::hash::Hash;
 
 #[cfg(feature = "wayland_frontend")]
-use crate::{
-    backend::renderer::{element::surface::WaylandSurfaceRenderElement, ImportAll},
-    desktop::LayerSurface,
-    wayland::shell::wlr_layer::Layer,
-};
+use crate::{backend::renderer::ImportAll, desktop::LayerSurface, wayland::shell::wlr_layer::Layer};
 use crate::{
     backend::renderer::{
         element::{AsRenderElements, Wrap},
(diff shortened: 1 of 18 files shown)


src/desktop/space/element/mod.rs
```diff
@@ -171,37 +167,30 @@ impl<
...
-            >(surface, renderer, location, scale, alpha)
-            .into_iter()
-            .map(SpaceRenderElements::Surface)
-            .map(C::from)
-            .collect(),
+            SpaceElements::Layer { surface, .. } => surface
+                .render_elements(renderer, location, scale, alpha)
+                .into_iter()
+                .map(SpaceRenderElements::Surface)
+                .collect::<Vec<_>>(),
```
kenoss, line 187: Still we need `.collect::<Vec<_>>()` for monomorphization; we need a common type with the arm below.

src/desktop/space/mod.rs
```diff
@@ -399,15 +399,30 @@ impl<E: SpaceElement + PartialEq> Space<E> {
...
-                    .render_elements::<<E as AsRenderElements<R>>::RenderElement>(
-                        renderer,
-                        location.to_physical_precise_round(scale),
-                        scale,
-                        alpha,
-                    )
+                    .render_elements(renderer, location.to_physical_precise_round(scale), scale, alpha)
+                    // So, we need to collect.
+                    .into_iter()
+                    .collect::<Vec<_>>()
```
kenoss, line 425: troublesome point 2

src/desktop/space/element/mod.rs
Drakulix: We don't need a `Vec` and eagerly evaluate the vector.

We could either
- `Box` it (still an allocation)
- create a dispatching `enum`, which imo the macro should just create?

src/desktop/space/mod.rs
Drakulix: Adding a `move` to the closure should fix this, given we collect afterwards anyway.

src/desktop/space/element/mod.rs
kenoss: Reg. `enum`

> create a dispatching enum, which imo the macro should just create?

This is one of what [`either`](https://crates.io/crates/either) crate provides ([doc](https://docs.rs/either/1.15.0/either/enum.Either.html#method.into_iter).

I thought generalizing it won't work well. Consider the case that there are, say, five types. For example, 1. The `enum` that implements `AsRenderElements` has five arms; or 2. there are four `.chain()` calls and early returns. We have to return different types and a dispatching `enum` needs lots of variants. A problem is API. In `Either`, a user specifies variants explicitly.

```
return Either::Left(std::iter::empty());
...
return Either::Right(vec![1, 2, 3, 4, 5]);
```

So, we need something like this:

```
return CushioningMaterial::Variant0(...);
return CushioningMaterial::Variant1(...);
return CushioningMaterial::Variant2(...);
...
```

It's feasible to write a proc macro that helps this pattern, but I feel it's not a good idea as it is too magical and hinders user's intuition.

Note: Here's (one of) my (bad) idea:

```
#[cushioning_material::auto_convert_return]
fn render_elements() -> ... {
    ...
    // Automatically convert it to `return CushioningMaterial::Variant0(...);`.
    return std::iter::empty();
    ...
    // Automatically convert it to `return CushioningMaterial::Variant1(...);`.
    return ;
    ...
    // Or, alternatively, we can use explicit marker. (I guess it's unavoidable as we have a pattern that doesn't encourage `return`.)
    match ... {
        ... => {
            #[foo]
            ...
        }
        ... => {
            #[foo]
            ...
        }
    }
}
```

---

Reg. `Box`

I think this type of discussion should be based on a benchmark, but here's my opinion: Using `Box<dyn IntoIterator<...>>` or `Box<dys Iterator<...>>` introduces vtable calls (`.next()`) per render element per frame. I feel `.collect<Vec<_>>()` sohuld be cheaper, as the cost of it is memcpy, roughly. (Ignoring `map` etc. as it is common in these options.)

Let me know if you have a way of benchimarking.

src/desktop/space/mod.rs
kenoss: No.

```
error: captured variable cannot escape `FnMut` closure body
   --> src/desktop/space/mod.rs:421:17
    |
419 |               .flat_map(move |e: &InnerElement<E>| {
    |                                                  - inferred to be a `FnMut` closure
420 |                   let location = e.render_location() - region.loc;
421 | /                 e.element
422 | |                     .render_elements(renderer, location.to_physical_precise_round(scale), scale, alpha)
    | |_______________________________________________________________________________________________________^ returns a reference to a captured variable which escapes the closure body
    |
    = note: `FnMut` closures only have access to their captured variables while they are executing...
    = note: ...therefore, they cannot allow references to captured variables to escape
```

The followings are my understanding.

1. `T` has a member `xs` and implements `AsRenderElements`, by returning `self.xs.iter().map(|x| ...)`. [`Iter`](https://doc.rust-lang.org/std/slice/struct.Iter.html) grabs `self.xs`, [`Map`](https://doc.rust-lang.org/std/iter/struct.Map.html) can grab other members. So, in general, opaque `impl IntoIterator`/`impl Iterator` implicitly grab references to arguments. (Like `+ use<...>` in this PR.)
2. `renderer: &mut R` is caputuerd to the `FnMut`. This `FnMut` can use `renderer` as lifetime `'b`, which is the scope of this `FnMut`. (While `renderer` outlives outer `collect()`.)
3. 1 and 2 contradict.

PR #1702, opened 2025-04-02
compositor: Allow fractional client scales

Drakulix:
I don't know, if we want to merge this, but this seems to match what Kwin does.


Drakulix:
Okay, this seems to work unreasonably well with Xwayland and matches what kwin does, I don't see a point to not allow client_scale overrides of fractional values, no matter if a compositor is actually going to use this or not.

Unfortunately this means we need an f64, as some common fractional scale values cannot be represented correctly by f32.

cmeissl (approved):
LGTM

PR #1704, opened 2025-04-07
renderer: Introduce `ContextId`

Paraworker:
Introduce `ContextId` to represent a renderer context, replacing the previous use of `TypeId` + `usize` ID.

This also aligns with the best practice of using the newtype pattern for IDs.

### DIFF
--- a/src/backend/renderer/damage/mod.rs
+++ b/src/backend/renderer/damage/mod.rs
@@ -47,7 +47,7 @@
 //!
 //! const WIDTH: i32 = 10;
 //! const HEIGHT: i32 = 10;
-//! # let mut renderer = DummyRenderer;
+//! # let mut renderer = DummyRenderer::default();
 //! # let mut framebuffer = DummyFramebuffer;
 //! # let buffer_age = 0;
 //!
--- a/src/backend/renderer/element/memory.rs
+++ b/src/backend/renderer/element/memory.rs
@@ -71,7 +71,7 @@
 //!
 //! // Initialize a static damage tracker
 //! let mut damage_tracker = OutputDamageTracker::new((800, 600), 1.0, Transform::Normal);
-//! # let mut renderer = DummyRenderer;
+//! # let mut renderer = DummyRenderer::default();
 //! # let mut framebuffer = DummyFramebuffer;
 //!
 //! let mut last_update = Instant::now();
--- a/src/backend/renderer/element/memory.rs
+++ b/src/backend/renderer/element/memory.rs
@@ -104,7 +104,7 @@
 //! ```
 
 use std::{
-    any::TypeId,
+    any::Any,
     collections::{hash_map::Entry, HashMap},
     sync::{Arc, Mutex, MutexGuard},
 };
@@ -116,7 +116,7 @@ use crate::{
         allocator::{format::get_bpp, Fourcc},
         renderer::{
             utils::{CommitCounter, DamageBag, DamageSet, DamageSnapshot, OpaqueRegions},
-            Frame, ImportMem, Renderer,
+            ContextId, Frame, ImportMem, Renderer,
         },
     },
     utils::{Buffer, Logical, Physical, Point, Rectangle, Scale, Size, Transform},
@@ -223,8 +223,8 @@ struct MemoryRenderBufferInner {
     transform: Transform,
     opaque_regions: Option<Vec<Rectangle<i32, Buffer>>>,
     damage_bag: DamageBag<i32, Buffer>,
-    textures: HashMap<(TypeId, usize), Box<dyn std::any::Any + Send>>,
-    renderer_seen: HashMap<(TypeId, usize), CommitCounter>,
+    textures: HashMap<ContextId, Box<dyn Any + Send>>,
+    renderer_seen: HashMap<ContextId, CommitCounter>,
 }
 
 impl Default for MemoryRenderBufferInner {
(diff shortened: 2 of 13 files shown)


Paraworker:
> > Use the `RendererId` type to uniquely identify a `Renderer`, so there's no need to store the `TypeId` of a `TextureId` as a key to indicate which `Renderer` the texture is originating from.
> 
> Thanks for working on this, however I don't think this is 100% true.
> 
> For one we don't uniquely identify renderers.
> 
> `GlesRenderer`s might share contexts and thus `RendererId`s. So the Id really identifies `EGLContext`s.
> 
> And there is no reason a different renderer of the same type might not be able to import a texture from any given renderer, so storing the originating renderer isn't as important as is storing the type sometimes.
> 
> E.g. the `PixmanRenderer`s textures are just memory slices. I don't see anything in the API of pixman preventing textures of one renderer to be used in the other. (This is probably why the `PixmanRenderer` has a static renderer id.)
> 
> However I like the API changes and cleanup this PR does, if we want to merge it we just need to make sure we know what the ID represents and what we want to achieve.
> 
> I think the Id needs to represent renderers, that can import the same textures. So for the `GlesRenderer` (and thus the `GlowRenderer`) these proposed changes are correct, as is for the `MultiRenderer`, where this property depends on the internal `render`-gpu's renderer.
> 
> However the `PixmanRenderer` and `DummyRenderer` should both have a static Id that is shared by all their respective renderer instances.
> 
> The documentation of the `RendererId` type should also explain this (in general, not naming any specific renderers) and we might want to think about a different name to better represent this property?

Thanks for the clarification. So in this case, I also think the names `id()` and `RendererId` are a bit too generic. Maybe something like `ContextId` would be a better fit?

cmeissl:
+1 for something including context in the name

Paraworker:
@Drakulix I've updated this PR to use `ContextId`. If you find anything unclear or inaccurate in the documentation, feel free to make changes.

cmeissl (approved):
LGTM thanks!

small nit, I would have preferred to have separate commits for the context stuff and the unrelated format changes.

Paraworker:
> LGTM thanks!
> 
> small nit, I would have preferred to have separate commits for the context stuff and the unrelated format changes.

Yup, I’ve split things up.

cmeissl (approved):
Thanks!

Drakulix:
thanks!

Paraworker:
I'm wondering whether it's reasonable for `MultiRenderer` to return the `ContextId` of its internal renderer, given that it defines its own `MultiTexture` type. Would that still be texture-compatible?

Drakulix:
> I'm wondering whether it's reasonable for `MultiRenderer` to return the `ContextId` of its internal renderer, given that it defines its own `MultiTexture` type. Would that still be texture-compatible?

Oh damn, no it isn't, we need to revert this.

It should absolutely be the Id of the context (because the MultiRenderer type is short-lived), but it needs to be it's own type in the cache not overriding other texture types from the internal renderer.

Paraworker:
> > I'm wondering whether it's reasonable for `MultiRenderer` to return the `ContextId` of its internal renderer, given that it defines its own `MultiTexture` type. Would that still be texture-compatible?
> 
> Oh damn, no it isn't, we need to revert this.
> 
> It should absolutely be the Id of the context (because the MultiRenderer type is short-lived), but it needs to be it's own type in the cache not overriding other texture types from the internal renderer.

But the previous logic doesn't seem entirely correct either.
Imagine that in the future we have `VulkanRenderer`, and both `VulkanRenderer` and `GlesRenderer` have their `usize` IDs starting from 0.
When they're used in combination with `MultiRenderer`, their keys would both be (TypeId of MultiTexture, 0).

Paraworker:
I'm thinking about assigning `MultiRenderer` its own `ContextId` by mapping the internal renderer's `ContextId` to it.

Something like:

```rust
    fn context_id(&self) -> ContextId {
        static ID_CACHE: LazyLock<Mutex<HashMap<ContextId, ContextId>>> =
            LazyLock::new(|| Mutex::new(HashMap::new()));

        ID_CACHE
            .lock()
            .unwrap()
            .entry(self.render.renderer().context_id())
            .or_insert_with(ContextId::next)
            .clone()
    }
```

Drakulix:
Hmm, that makes it a little more annoying to get the context id of the underlying renderer, which is useful given `MultiTexture::get` exists, but I guess that would work.

Drakulix:
Also we would add a mutex to a hot path. This is called very frequently.

Paraworker:
> Also we would add a mutex to a hot path. This is called very frequently.

This can be easily optimized, just add a `context_id` field to `MultiRenderer` and store it when the instance is created.

```rust
fn map_context_id(source: ContextId) -> ContextId {
    static ID_CACHE: LazyLock<Mutex<HashMap<ContextId, ContextId>>> =
        LazyLock::new(|| Mutex::new(HashMap::new()));

    ID_CACHE
        .lock()
        .unwrap()
        .entry(source)
        .or_insert_with(ContextId::next)
        .clone()
}

MultiRenderer {
    context_id: map_context_id(render.renderer().context_id()),
    render,
    target: None,
    other_renderers: others,
    span: tracing::Span::current(),
}
```

Paraworker:
> Hmm, that makes it a little more annoying to get the context id of the underlying renderer, which is useful given `MultiTexture::get` exists, but I guess that would work.

I think using `as_ref().context_id()` to get the underlying renderer’s `ContextId` is acceptable, it’s just a bit unintuitive.

PR #1706, opened 2025-04-09
xwm: Add events/method for randr primary output management

Drakulix:
A lot of games expect to be able to query a primary output via xrandr. This adds a way for compositors to manage this state.

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -92,6 +92,7 @@
 //! ```
 
 use crate::{
+    output::Output,
     utils::{x11rb::X11Source, Client, Logical, Point, Rectangle, Size},
     wayland::{
         selection::SelectionTarget,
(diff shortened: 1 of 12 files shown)


PolyMeilex (approved):

src/xwayland/xwm/mod.rs
```diff
@@ -1284,6 +1313,75 @@ impl X11Wm {
...
+            .conn
+            .randr_get_screen_resources_current(self.screen.root)?
+            .reply()?;
+        let info = self
+            .conn
+            .randr_get_output_info(current_primary, res.config_timestamp)?
+            .reply()?;
+        Ok(Some(
+            std::str::from_utf8(&info.name)
+                .expect("X11 protocol violation, output name not utf-8")
```
cmeissl, line 1338: I don't know how likely this is, but a malicious client could bring the compositor down with this expect. Is there an alternative like ignoring the output in this case?

Drakulix: I honestly think this is a non-issue, as this value comes from the xorg server and the protocol explicitly states that this string is utf-8.

cmeissl (approved):

PR #1718, opened 2025-04-18
xdg: Resize opposing edges when unconstraining popups

Ottatop:
This PR makes `PositionerState` resize the top and left edges when unconstraining popups in addition to the bottom and right edges that are already getting resized.

A user ran into an issue where a popup would overflow past the top of the screen (pinnacle-comp/pinnacle#305). This issue reproduces in Niri and cosmic-comp.

I believe the current positioner code assumes that a slide will always take place before a resize and thus only cuts off the left and bottom sides of the popup rectangle when resizing. In the issue above, gnome-builder creates the popup with `resize_y` but not `slide_y` constraint adjustments which leaves the top side constrained, so we need to resize the opposing edges as well.

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -749,7 +749,7 @@ impl PositionerState {
                 geo.loc.x -= min(off_right, -off_left);
             }
 
-            (_, off_right, _, _) = compute_offsets(target, geo);
+            (off_left, off_right, _, _) = compute_offsets(target, geo);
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
LGTM.

@YaLTeR any opinion on this? You added these abstractions, right?

YaLTeR:
Assuming this fixes the issue as intended, looks ok.

> You added these abstractions, right?

IIRC I largely adapted them from cosmic-comp

PR #1720, opened 2025-04-19
support delayed blocker

cmeissl:
This PR extends the existing `Blocker` API with a `BlockerKind`. Allowing to put blockers into two groups, `Immediate` and `Delayed`. `Blocker`s with `BlockerKind::Delayed` will only be evaluated once all  `Immediate` have been cleared.
After all  `Immediate` blockers have been cleared all `Blocker`s will be notified, allowing delayed blockers to make a decision based on that.
All blockers still need to be cleared for the transaction to be unblocked.

Additionally the PR adds a reference implementation of a delayed blocker, `SurfaceBarrier`.

One example for this is implementing animation transactions spanning multiple otherwise unrelated surfaces.
While this can be implemented without the `SurfaceBarrier`, it might help to enhance synchronization when some other blockers further delay the buffer.

A compositor could place a `Immediate` blocker, like for example a simple `Barrier`, on each surface participating in the animation transaction and also register it in a `SurfaceBarrier` with the same scope as the animation transaction.
Once the `Immediate` blocker representing the transaction is cleared, the blocker in the `SurfaceBarrier` will try to resolve and ping the compositor once it completed.

cc @YaLTeR

### DIFF
--- a/src/wayland/compositor/transaction.rs
+++ b/src/wayland/compositor/transaction.rs
@@ -50,10 +50,31 @@ use crate::utils::Serial;
 
 use super::{tree::PrivateSurfaceData, CompositorHandler};
 
+/// Kind for a [`Blocker`]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum BlockerKind {
+    /// A immediate blocker which needs to be cleared before all delayed blockers.
+    Immediate,
+    /// Defines a delayed blocker which will be evaluated after
+    /// all immediate blockers are cleared.
+    Delayed,
+}
+
 /// Types potentially blocking state changes
 pub trait Blocker {
     /// Retrieve the current state of the blocker
     fn state(&self) -> BlockerState;
+
+    /// Retrieve the kind of the blocker
+    fn kind(&self) -> BlockerKind {
+        BlockerKind::Immediate
+    }
+
+    /// Notifies the blocker that all immediate blockers have been cleared
+    /// for a particular surface
+    fn notify(&self, surface: &WlSurface) {
+        let _ = surface;
+    }
 }
 
 /// States of a [`Blocker`]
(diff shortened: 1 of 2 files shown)


src/wayland/compositor/transaction.rs
```diff
@@ -108,6 +130,114 @@ impl Blocker for Barrier {
...
+        surfaces
+            .entry(surface.downgrade())
+            .and_modify(|state| *state = true);
+        if surfaces
+            .iter()
+            .all(|(surface, state)| *state || !surface.is_alive())
+        {
+            let signaled = self.ready.swap(true, std::sync::atomic::Ordering::Acquire);
+            if !signaled {
+                self.ping.ping();
```
YaLTeR, line 235: IIRC in niri I didn't use the Ping source for the actual ready notification because it would arrive only on the next event loop iteration. I'm not sure if this is a problem here.

YaLTeR (commented):
Only a cursory glance for now; will have to see how it looks in niri to give proper feedback.

src/wayland/compositor/transaction.rs
```diff
@@ -342,7 +376,30 @@ impl TransactionQueue {
                 i += 1;
             } else {
                 // this transaction is to be applied, yay!
-                ready_transactions.push(self.transactions.remove(i));
+                let transaction = &mut self.transactions[i];
+
+                transaction.notify_blockers();
```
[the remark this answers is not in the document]
cmeissl: Depends on the impl of the delayed blocker and the precense of other delayed blockers. It could unblock from within notify and if it is the last one it would not be called again. But yes, it could get notified multiple times until the transaction completed. Maybe something worth documenting on the notify function. Just FYI, I also evaluated a design based on a callback which could place additional blockers but that had its own shortcomings.

src/wayland/compositor/transaction.rs
cmeissl: Interesting. I am also not sure this is a problem in this case. We could use a channel instead maybe?

ewgdg:
`SurfaceBarrier` waits until after Immediate blockers, but it still cannot guarantee grouped atomic application because one `Transaction` can contain multiple delayed blockers, for example after `PendingTransaction::merge_into()`. One delayed blocker can release while another delayed blocker in the same Transaction is still Pending.

PR #1721, opened 2025-04-20
feat: implement KeyboardHandle::set_modifier_state

erdii:
to enable compositor side control of modifiers such as num lock.

This PR is an implementation of a [recommendation](https://github.com/YaLTeR/niri/issues/291#issuecomment-2657328970) made by @kylewillmon in a Niri issue requesting support to enable num lock at startup: https://github.com/YaLTeR/niri/issues/291

Disclaimer: I'm new to rust so I'm extra happy for all feedback that points out code smells.

### DIFF
--- a/src/input/keyboard/modifiers_state.rs
+++ b/src/input/keyboard/modifiers_state.rs
@@ -48,6 +48,72 @@ impl ModifiersState {
         self.iso_level5_shift = state.mod_name_is_active(&xkb::MOD_NAME_MOD3, xkb::STATE_MODS_EFFECTIVE);
         self.serialized = serialize_modifiers(state);
     }
+
+    /// Serialize modifier state back to be sent to xkb.
+    pub fn serialize_back(&self, state: &xkb::State) -> SerializedMods {
+        let keymap = state.get_keymap();
+
+        let mut locked: u32 = 0;
+        let mut depressed: u32 = 0;
+
+        if self.caps_lock {
+            let index = keymap.mod_get_index(&xkb::MOD_NAME_CAPS);
+            if index != xkb::MOD_INVALID {
+                locked |= 1 << index;
+            }
+        }
+        if self.num_lock {
+            let index = keymap.mod_get_index(&xkb::MOD_NAME_NUM);
+            if index != xkb::MOD_INVALID {
+                locked |= 1 << index;
+            }
+        }
+        if self.ctrl {
+            let index = keymap.mod_get_index(&xkb::MOD_NAME_CTRL);
+            if index != xkb::MOD_INVALID {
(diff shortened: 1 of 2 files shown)


ids1024:
Currently cosmic-comp has a couple places it just sends a key press and key release to toggle things like numlock. Probably not ideal, but it works.

https://github.com/Smithay/smithay/pull/1597 is also something like this, though the intended use case is for a somewhat different and more complicated use case (changing how all key presses are handled, in a nested compositor where the parent compositor tracks the modifier state).

Drakulix:
> Currently cosmic-comp has a couple places it just sends a key press and key release to toggle things like numlock. Probably not ideal, but it works.

Right, though I think a "proper" way to do this would be preferrable.

> #1597 is also something like this, though the intended use case is for a somewhat different and more complicated use case (changing how all key presses are handled, in a nested compositor where the parent compositor tracks the modifier state).

Given 1597 is more involved and the api more complex, I don't think it would hurt to have this method as well, no?

src/input/keyboard/modifiers_state.rs
```diff
@@ -48,6 +48,75 @@ impl ModifiersState {
...
+            assert!(
+                index != xkb::MOD_INVALID,
+                "capslock is invalid modifier in keymap"
+            );
+            locked |= 1 << index;
+        }
+
+        if self.num_lock {
+            let index = keymap.mod_get_index(&xkb::MOD_NAME_NUM);
+            assert!(index != xkb::MOD_INVALID, "numlock is invalid modifier in keymap");
```
YaLTeR, line 70: Are these asserts guaranteed to pass even for weird custom keymaps? If not then I'd convert them to `if`s?

ids1024:
> Given 1597 is more involved and the api more complex, I don't think it would hurt to have this method as well, no?

Yeah. Also for https://github.com/Smithay/smithay/pull/1597 I've been trying to think of how to overhaul the keyboard API instead of complicating it like that PR currently does. So if we likely want to break the API but haven't come up with a better design yet, this addition seems fine in the mean time, even if it might be superseded by something else.

src/input/keyboard/modifiers_state.rs
erdii: I don't know tbh.

I'm not a fan of silently swallowing explicit requests by default, but I'm happy to remove the asserts if wanted.

I'll do more research and try do dig up relevant parts of the xkb spec.

YaLTeR: You *never* want to crash the compositor from something user-configurable (an xkb keymap)

Drakulix (approved):
LGTM. Thanks!

PR #1728, opened 2025-04-25
wayland: XDG toplevel tag protocol

PolyMeilex:


ids1024:
This should be unblocked now that the new `wayland-rs` release is up on crates.io.

src/wayland/xdg_toplevel_tag.rs
```diff
@@ -0,0 +1,195 @@
...
+pub trait XdgToplevelTagHandler:
+    GlobalDispatch<XdgToplevelTagManagerV1, ()> + Dispatch<XdgToplevelTagManagerV1, ()> + 'static
+{
+    /// Toplevel tag was set/updated.
+    #[allow(unused)]
+    fn on_set_tag(&mut self, toplevel: XdgToplevel, tag: String) {}
+
+    /// Toplevel description was set/updated.
+    #[allow(unused)]
+    fn on_set_description(&mut self, toplevel: XdgToplevel, description: String) {}
```
ids1024, line 98: It doesn't look like any trait in smithay currently has methods named with `on_*`. So it might be more consistent to just call them `set_tag` and `set_description`?

Not sure about having empty default implementations, vs just making the functions mandatory. At least for the intended purpose of the protocol, every implementer should care about the tag, though maybe not the description.

src/wayland/xdg_toplevel_tag.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        match request {
+            Request::SetToplevelTag { toplevel, tag } => {
+                let data = toplevel.data::<XdgShellSurfaceUserData>().unwrap();
+
+                compositor::with_states(&data.wl_surface, |states| {
+                    states
+                        .data_map
+                        .insert_if_missing_threadsafe(XdgToplevelTagSurfaceData::default);
+                    let data = states.data_map.get::<XdgToplevelTagSurfaceData>().unwrap();
```
ids1024, line 153: The `insert_if_missing_threadsafe` and `get` could be combined into a `get_or_insert_threadsafe`.

src/wayland/xdg_toplevel_tag.rs
PolyMeilex: Yeah right, we don't do `on_*`. We have `*_changed` in xdg shell tho, so perhaps that's  a more consistent choice. `set_*` is good as well as that does not indicate any deduplication logic which I don't do in this implementation.

---

The callbacks are optional, as you can still access the data without listening to those.

src/wayland/xdg_toplevel_tag.rs
PolyMeilex: Oh, that's a nice addition, replaced.

Drakulix (approved):
LGTM! Thanks :)

PR #1730, opened 2025-05-14
Fix incorrect tablet pen tilt handling in `libinput` backend

eepysillycat:
Hi! I noticed that `TabletToolEvent<LibinputInputBackend>` incorrectly returns `dx` and `dy` properties of libinput event as `tilt_x` and `tilt_y` so pen tilt in drawing programs like Krita gets broken and I can't use brushes relying on it

Seems like everything is working correctly after my changes


cmeissl (approved):
Thanks!

PR #1732, opened 2025-05-15
implement a Point::new and a Size::new constructor

m4rch3n1ng:
the main reason is that i was very annoyed at having to unnecessary double parenthesis when calling `Size::from((w, h))`, which i don't have to do with `Size::new(w, h)`.

another benefit is, that `Point::new` is const so you can create a `Point` in const (though i haven't ever needed that). since `Size::new` has a `debug_assert!` that calls a trait method, i cannot make that `const` without removing the assertion.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -341,6 +341,17 @@ pub struct Point<N, Kind> {
     _kind: std::marker::PhantomData<Kind>,
 }
 
+impl<N, Kind> Point<N, Kind> {
+    /// Create a new Point
+    pub const fn new(x: N, y: N) -> Point<N, Kind> {
+        Point {
+            x,
+            y,
+            _kind: std::marker::PhantomData,
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


ids1024:
Euclid has `const fn new()` methods for `Size2D` and `Point2D`, so this seems to be no problem for moving to Euclid later.

https://docs.rs/euclid/latest/euclid/struct.Size2D.html#method.new
https://docs.rs/euclid/latest/euclid/struct.Point2D.html#method.new

It also has standalone `size2` and `point2` functions, but those are just shorter aliases for `::new` (unlike the `rect` function, which takes 4 scalars, while `Rect::new` takes a point and size).

Drakulix (approved):

Drakulix:
> Euclid has `const fn new()` methods for `Size2D` and `Point2D`, so this seems to be no problem for moving to Euclid later.

Then lets merge! Thanks @m4rch3n1ng!

PR #1733, opened 2025-05-15
impl Div<N> for Size<N, Kind>

m4rch3n1ng:
there already is an `impl Mul<N> for Size<N, Kind>`, so i think it just makes sense to also have a function that does the opposite. also saves me one `Size::from((size.w * 3 / 4, size.h * 3 / 4))` where i can now do a `size * 3 / 4`. you could also do this with `.downscale` (and `.upscale` respectively), but those are a little less convenient to use and (in my opinion) less clear.


Drakulix (approved):
LGTM, thanks!

PR #1739, opened 2025-05-21
wayland/drm_syncobj: Allow somewhat graceful migration/destruction

Drakulix:
While the drm_syncobj protocol does not assume any specific drm device given that syncpoint kinda life outside of that, as a compositor we still need a kms to perform the necessary ioctls and thus need to deal with the fact that our device could disappear.

While most compositors don't really handle they primary gpu disappearing, we can at least add some api to help compositors to gracefully recover. (Any client using that gpu will most likely crash anyway, but might possibly recover with robustness and reconnects.)

We need to handle two cases here:
1. The device we used for drm_syncobj disappeared, but we have another with the necessary functionality: Allow updating the device. (The event source of any pending timelines *should* receive an event as far as I know and thus resolve. This is to be tested, potentially we need to unblock/protocol-error on an ioctl returning `Err`.)
2. All drm devices with syncobj capabilities are gone. We need to disable the global so no new clients bind it and let old timelines error out / resolve (same as above).

This PR provides two new methods on `DrmSyncobjState` to facilitate that.


ids1024:
So if the import device has been detached, existing `DrmTimeline` and `DrmSyncPoint` instances will no longer work, because they contain the device fd and syncobj handle they were imported on? Even for clients using a different GPU? And there's no way to communicate to the client that it needs to create new ones?

So synchronization is broken in some way for all clients using explicit sync?

But presumably we could keep the timeline fd and import it on a different device, then have things keep working.

(In theory this is also an issue if there is no longer a GPU that supports syncobjs, but that shouldn't really be an issue, because clients shouldn't be trying to use syncobjs then.)

Drakulix:
> So if the import device has been detached, existing `DrmTimeline` and `DrmSyncPoint` instances will no longer work, because they contain the device fd and syncobj handle they were imported on? Even for clients using a different GPU? And there's no way to communicate to the client that it needs to create new ones?

So in my testing with an eGPU, we can detach the GPU, if we call `state.device_removed` manually before actually unloading the device. (Be that by detaching or unloading the module.) Otherwise we get a segfault in `libgallium`, which breaks the compositor, if the device wasn't lazily unloaded before. (This even works if the device was deemed the primary gpu previously and cosmic-comp select a new one.)

The second we call `device_removed` surfaces of clients on that GPU appear blank (expected). Once we disconnect the GPU they crash/exit, which is also expected, but cosmic-comp survives.

> But presumably we could keep the timeline fd and import it on a different device, then have things keep working.

For clients running on different gpus (tested with our extra sockets and `weston-simple-egl` on egl-wayland) they freeze. Which indicates to me that the eventfd is never triggered.

Drakulix:
> > Otherwise we get a segfault in `libgallium`
> 
> Of course. I guess technically the driver shouldn't segfault here? But we can't really rely on it.

Perhaps we need robustness support here to have that working. (We should really get around to that at some point even if just to handle gpu resets.)

> > Which indicates to me that the eventfd is never triggered.
> 
> Well, even if the eventfd is triggered, for the next frame we should get an error trying to call `DrmSyncPoint::eventfd` or other methods, since we imported the syncobj on a GPU that is now disconnected and are trying to use ioctls on that GPU's dev node.

So fixing this ended up being more elaborate then I hoped, but this works now.

ids1024:
I suppose when `.invalidate()` is called due to the primary GPU removal, an output on another GPU with an explicit synchronized surface may show a corrupted frame due to the early eventfd signalling, but the next frame should be fine?

Wayland's goal that "every frame is perfect" doesn't exactly apply well to unexpected primary GPU removal anyway. Though I wonder if that could corrupt later frames for an image-copy session due to damage tracking.

ids1024:
Anyway, this looks good overall.

I think signalling the eventfd in `Drop for DrmTimelineDeviceSpecific` shouldn't occur when the frame is supposed to be on screen since the `DrmSyncPoint` will still be in the surface cached state?

`wp_linux_drm_syncobj_surface_v1::destroy` does state that "Any timeline point set by this object before the last commit will not be affected." So technically we should make sure this holds..

Drakulix:
> I think signalling the eventfd in `Drop for DrmTimelineDeviceSpecific` shouldn't occur when the frame is supposed to be on screen since the `DrmSyncPoint` will still be in the surface cached state?
> 
> `wp_linux_drm_syncobj_surface_v1::destroy` does state that "Any timeline point set by this object before the last commit will not be affected." So technically we should make sure this holds..

Good catch! Removed the `Drop`-implementation and added a call to `invalidate()` on `update_device` instead.



> I suppose when `.invalidate()` is called due to the primary GPU removal, an output on another GPU with an explicit synchronized surface may show a corrupted frame due to the early eventfd signalling, but the next frame should be fine?

Yeah I don't think there is anything else we can do here.

 > Wayland's goal that "every frame is perfect" doesn't exactly apply well to unexpected primary GPU removal anyway. Though I wonder if that could corrupt later frames for an image-copy session due to damage tracking.

That is a good point as we potentially copy from the surface too early and leave the broken contents inside the clients copy buffers. But given how rare this event should be in the first place, I think that is acceptable for now. If we really wanted to work around this, we can try to damage any capture of surface on next commit in cosmic-comp.

PR #1740, opened 2025-05-21
input/keyboard: Use hash for ID of `KeymapFile`

ids1024:
This fixes excessive keymap events with `virtual-keyboard-unstable-v1` clients like fcitx5.

This depends on `sha2`. Adding a new dependency for this is annoying, but this is a fairly widely used crate. `blake3` could also work but is less widely used. Perhaps a dependency couple be avoided if `SipHasher` is good enough, but as I understand it's not really meant to guarantee an absence of collisions (but may be good enough in practice here?).

I expect computing a hash won't impact performance much, compared to clients having to process excessive `keymap` events


Drakulix (approved):
Seems fine to me. :+1:

PR #1742, opened 2025-05-22
chore: clippy fixes

Drakulix:


dcz-self:
I can't place a review, but in my code I would leave a comment about why "large_enum_variant" doesn't matter. Is that not performance-sensitive code, is the differences small in absolute terms, or something like that. Otherwise I'd disable it in the project scope.

Drakulix:
It doesn't matter because:
a) we can do anything about it in the macro code
b) there will always be the "GenericCatcher" variant, which is purely cosmetic to work around some generics limitations and will always be empty. So triggering a size difference here is very likely.

dcz-self:
What I meant to say is that while I'm sure that there are good reasons, I follow the rule that the "why" should be near the code that's not super clear in the form of comments for newcomers like me who might not understand it. But it's just a suggestion based on the way I work.
The changes look good to me overall.

PolyMeilex (approved):
Mentioned code comment would be nice, but LGTM either way.

Drakulix:
done

PR #1743, opened 2025-05-22
fix: explicitly send empty pre-edit string when backspacing after IME input only one letter

The-Lost-Light:
Fix a bug in chromium and electron that  Fcitx5 can't delete first letter in IME pre-edit region.

ref:[Ozone-wayland: support text_input_v3 protocol ](https://issues.chromium.org/issues/40113488#comment127)

### DIFF
--- a/src/wayland/input_method/input_method_handle.rs
+++ b/src/wayland/input_method/input_method_handle.rs
@@ -38,6 +38,7 @@ pub(crate) struct InputMethod {
 pub(crate) struct Instance {
     pub object: ZwpInputMethodV2,
     pub serial: u32,
+    pub preedit_sent: bool,
 }
 
 impl Instance {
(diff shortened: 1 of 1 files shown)


kchibisov:
As per this sounds like a chrome bug https://bugs.kde.org/show_bug.cgi?id=492225 to me.

The-Lost-Light:
Reasonable! Thanks for your reply.

ilsubyeega:
@The-Lost-Light Hello, did this patch resolve your issue?
- https://issues.chromium.org/issues/409716545
- https://chromium-review.googlesource.com/c/chromium/src/+/6626557

The-Lost-Light:
> @The-Lost-Light Hello, did this patch resolve your issue?
> 
>     * https://issues.chromium.org/issues/409716545
> 
>     * https://chromium-review.googlesource.com/c/chromium/src/+/6626557

Yes! I try and it fixed in chromium-snapshot

PR #1745, opened 2025-05-23
Add xx_input_method_v1 support.

dcz-self:
Notably missing: repositioning, reactivity of the popup.


dcz-self:
To test:
- stiwri from "popup" branch https://codeberg.org/dcz/stiwri/src/branch/popup

```
# in anvil
cargo run --release -- --winit &
WAYLAND_DISPLAY=wayland-1 yad &
# in stiwri
WAYLAND_DISPLAY=wayland-1 cargo run --release --bin im_popup
```

dcz-self:
This is an implementation of https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/407

dcz-self:
Lookng at CI failures, I need some help understanding what they are.

For example, this code wasn't even touched by my change.
```
  error: fields `popup_geometry_callback`, `new_popup`, and `popup_repositioned` are never read
     --> src/wayland/input_method/input_method_handle.rs:172:16
```

There's also the ack_configure callback which is for some reason exposed to the compositor, but I don't really understand why, so I couldn't come up with a sensible doc comment.

ids1024:
> For example, this code wasn't even touched by my change.

I think that's a result of the earlier warning. Because certain methods are never used, the variables that are read only in those methods are never read.

dcz-self:
Added repositioning:

```
WAYLAND_DISPLAY=wayland-1 cargo run --release --bin im_popup_reposition
```

[Screencast_20250524_202116.webm](https://github.com/user-attachments/assets/8e343f45-28bf-4d7a-a30e-95db773db244)

dcz-self:
I did some cleanups:
- lints, except the ack_configure which I don't understand
- both input methods are supported and working

Is there anything I can do to make reviewing this easier?

anvil/src/state.rs
```diff
@@ -339,6 +344,64 @@ impl<BackendData: Backend> InputMethodHandler for AnvilState<BackendData> {
...
 delegate_input_method_manager!(@<BackendData: Backend + 'static> AnvilState<BackendData>);
 
+impl<BackendData: Backend> InputMethodHandlerV3 for AnvilState<BackendData> {
+    fn new_popup(&mut self, surface: PopupSurfaceV3) {
+        if let Err(err) = self.popups.track_popup(PopupKind::from(surface)) {
+            warn!("Failed to track popup: {}", err);
+        }
+    }
+
+    fn popup_repositioned(&mut self, _: PopupSurfaceV3) {}
```
rano-oss, line 354: What happens when you move the parent, does the popup follow the parent?

src/desktop/wayland/popup/mod.rs
```diff
@@ -70,13 +75,15 @@ impl PopupKind {
...
             PopupKind::InputMethod(ref t) => t.get_parent().map(|parent| parent.location).unwrap_or_default(),
+            PopupKind::InputMethodV3(ref t) => t.get_parent().location,
         }
     }
 
     fn send_done(&self) {
         match *self {
             PopupKind::Xdg(ref t) => t.send_popup_done(),
             PopupKind::InputMethod(_) => {} //Nothing to do the IME takes care of this itself
+            PopupKind::InputMethodV3(_) => {} //Nothing to do the IME takes care of this itself
```
rano-oss: This might be useful as a way to communicate that the compositor closed the popup? Or is it redundant since normally that would follow from a deactivate?

rano-oss: There was nothing to do previously since there was no concept of how the popup should function. But you are free to add that now if it is needed

rano-oss:
run these on your branch and correct them btw:
`cargo clippy --features "test_all_features" -- -D warnings`
`cargo fmt --all -- --check`
`cargo hack check --each-feature --no-dev-deps --exclude-features use_bindgen`

rano-oss:
LGTM otherwise, someone that knows xdg positioner should probably have a look at the positioner and popup part

src/desktop/wayland/popup/mod.rs
dcz-self: Yes, the only time the compositor closes the popup is in deactivate, so I decided there's no need for another notification. Would one be useful?

dcz-self:
The only thing left that breaks ci is the ack_configure which I have no idea what to do with.

rano-oss:
Just add documentation to this function: 
```
error: missing documentation for a method
   --> src/wayland/input_method_v3/mod.rs:131:5
    |
131 | /     fn popup_ack_configure(
132 | |         &mut self,
133 | |         _surface: &WlSurface,
134 | |         _serial: Serial,
135 | |         _client_state: PopupSurfaceState,
136 | |     ) {
    | |_____^
    |
    = note: `-D missing-docs` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(missing_docs)]`
```

dcz-self:
Well that's the problem, I don't have a clue what it's meant to do :P I copy-pasted it without understanding and I hope a reviewer can enlighten me.

dcz-self:
This now relies only on published protocols now that experimental is out.

src/wayland/text_input/text_input_handle.rs
```diff
@@ -186,13 +187,15 @@ impl TextInputHandle {
 #[derive(Debug)]
 pub struct TextInputUserData {
     pub(super) handle: TextInputHandle,
-    pub(crate) input_method_handle: InputMethodHandle,
+    pub(crate) input_method_handle: input_method::InputMethodHandle,
+    pub(crate) input_method_v3_handle: input_method_v3::InputMethodHandle,
```
ids1024: We shouldn't have two input methods at once. Including using different protocols. Maybe it would make sense if these were grouped together into a single enum?

dcz-self: I don't think this works, because those are slots for input methods that may be filled in at any time. See a couple lines below:

```
if !data.input_method_handle.has_instance() && !data.input_method_v3_handle.has_instance() {
```

The instance for either may not be present, and because the `Option` is rather deeply nested (even behind a mutex each), there's no way to turn it into an enum.

I don't remember why it is so, but I do remember originally not liking it and exposing the Option directly. Eventually I just had to copy the same design in v3.

rano-oss: Actually having two at once should be just fine, it depends on how it is used but functionally there really isn't any problem. Though I have only tested it locally, so take my word for it ;)

rano-oss: I would even go so far as to say that several input method instances is a wanted feature.

ids1024: If you had to input methods clients though, won't both try to create a popup at the same time, and potentially insert text in response to the same input?

I don't know if there's really a way with the design of this protocol (any version) to have reasonable behavior with multiple input method engines running at once.

dcz-self: IMO multiple input methods only work with handover of the protocol. But handover is outside of scope for this, so this MR doesn't allow multiple input method clients.

dcz-self:
Rebased.

src/wayland/text_input/text_input_handle.rs
rano-oss: Should not really be a problem, as long as one has a one-to-one relationship between text-input client and input method client. Input method 1 would have a list [ti1, ti2, t3] and input method 2 [t3, t4, t5]. We can make sure they never overlap in the compositor. Also about the popup, each IME should just own their popup. There is nothing really needed in the protocol to support this, just remove the part that allows only one. I would make a POC if life wasn't in the way currently, maybe in 6 months.

rano-oss: But I get the problem, if v2 and v3 don't have a way to separate which IME own which text input then double input methods are definitely a problem.

dcz-self:
Rebased again.

dcz-self:
Fixed the last problem, fixed conflict.

src/wayland/text_input/text_input_handle.rs
dcz-self: That is correct, the first one acquires the input methos and locks the following ones.

PR #1746, opened 2025-05-25
wayland/cursor_shape: update to version 2

m4rch3n1ng:
updates the cursor shape protocol to version 2. released in [wayland-protocols version 1.42](https://lists.freedesktop.org/archives/wayland-devel/2025-March/044027.html) it contains two new cursors: `dnd-ask` and `all-resize`.

needs `cursor-icon` v1.2.0, as that is the version with https://github.com/rust-windowing/cursor-icon/pull/17 included.

i tested it by patching winit and the smithay client toolkit (see https://github.com/Smithay/client-toolkit/pull/496) and using a cursor theme that has the `dnd-ask` cursor (one of the new ones, [bibata has them for example](https://github.com/ful1e5/Bibata_Cursor/blob/main/svg/groups/hand/dnd-ask.svg)) and it works when testing it with my own compositor:

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ appendlist = "1.4"
 ash = { version = "0.38.0", optional = true }
 bitflags = "2.2.1"
 calloop = "0.14.0"
-cursor-icon = "1.0.0"
+cursor-icon = "1.2.0"
 cgmath = "0.18.0"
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
Looks good! thanks :)

PR #1748, opened 2025-06-02
gles: Don't call `GenTextures` before `make_current()`

ids1024:
The `unwrap_or_else()` call here in `import_shm_buffer()` has a call to `self.gl.GenTextures()`, but `self.egl.make_current()` was only being called in the next block.

It appears this was the cause of some issues I saw on a multi-GPU system, where `GenTextures` ended up being called on the last current context, which if it was on the other GPU, caused this `GlesTexture` to be using the same texture ID as whatever other texture was assigned the same ID.


Drakulix (approved):
nice catch!

PR #1750, opened 2025-06-02
gles: Define a `GlesContext`, with a `CurrentGlesContext` guard

ids1024:
Still working through issues to use this everywhere. The goal is to prevent things like https://github.com/Smithay/smithay/pull/1748.

In particular, `GlesFrame` needs to be updated to hold a `CurrentGlesContext`. Things like `blit_to`/`bllit_from` are a little funny here since they call a function that makes a different context current, then restore the context.

### DIFF
--- a/src/backend/renderer/gles/context.rs
+++ b/src/backend/renderer/gles/context.rs
@@ -1,7 +1,10 @@
 use std::{fmt, ops};
 
-use super::ffi;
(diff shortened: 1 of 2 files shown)


ids1024:
I guess `EGLContext::wait()` should also require a current context.

Edit: For modularity the `egl` module could define an `ActiveContext` trait, and require that for `create`/`wait` instead of the requirement being in the doc comment but not enforced in any way.

PR #1755, opened 2025-06-11
compositor/gbm: Don't scanout client buffers without explicit node matches

Drakulix:
See https://gitlab.freedesktop.org/drm/amd/-/issues/2075#note_2582039

### DIFF
--- a/src/backend/drm/exporter/gbm.rs
+++ b/src/backend/drm/exporter/gbm.rs
@@ -34,13 +34,23 @@ pub enum Error {
 pub struct GbmFramebufferExporter<A: AsFd + 'static> {
     gbm: gbm::Device<A>,
     drm_node: Option<DrmNode>,
+    #[cfg_attr(not(feature = "wayland_frontend"), allow(unused))]
+    import_node: Option<DrmNode>,
 }
 
 impl<A: AsFd + 'static> GbmFramebufferExporter<A> {
-    /// Initialize a new framebuffer exporter
-    pub fn new(gbm: gbm::Device<A>) -> Self {
+    /// Initialize a new framebuffer exporter.
+    ///
+    /// `import_node` will be used to filter dmabufs to originate from a particular
+    /// device before considering them for direct scanout.
+    /// If `import_node` is `None` direct-scanout of client-buffers won't be used.
+    pub fn new(gbm: gbm::Device<A>, import_node: Option<DrmNode>) -> Self {
         let drm_node = DrmNode::from_file(gbm.as_fd()).ok();
-        Self { gbm, drm_node }
+        Self {
+            gbm,
+            drm_node,
+            import_node,
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/backend/drm/exporter/gbm.rs
```diff
@@ -83,16 +83,25 @@ impl<A: AsFd + 'static> ExportFramebuffer<GbmBuffer> for GbmFramebufferExporter<
     fn can_add_framebuffer(&self, buffer: &ExportBuffer<'_, GbmBuffer>) -> bool {
         match buffer {
             #[cfg(not(all(feature = "backend_egl", feature = "use_system_lib")))]
-            ExportBuffer::Wayland(buffer) => matches!(
-                crate::backend::renderer::buffer_type(buffer),
-                Some(crate::backend::renderer::BufferType::Dma)
-            ),
+            ExportBuffer::Wayland(buffer) => crate::wayland::dmabuf::get_dmabuf(buffer)
+                .ok()
+                .is_some_and(|buf| buf.node().is_some_and(|node| Some(node) == self.drm_node)),
```
cmeissl: Could this also affect split kms devices? Not sure what will be reported as the buf node in this case.

Drakulix: The old code didn't even work for non-split gpus, given the gbm device tracks the primary node.

Drakulix:
Can't think of a better way to introduce the render_node for this check.

If we are satisfied with this approach, I'll add some more docs to `GbmFramebufferExporter::new` before merging this.

src/backend/drm/exporter/gbm.rs
```diff
@@ -83,16 +88,25 @@ impl<A: AsFd + 'static> ExportFramebuffer<GbmBuffer> for GbmFramebufferExporter<
     fn can_add_framebuffer(&self, buffer: &ExportBuffer<'_, GbmBuffer>) -> bool {
         match buffer {
             #[cfg(not(all(feature = "backend_egl", feature = "use_system_lib")))]
-            ExportBuffer::Wayland(buffer) => matches!(
-                crate::backend::renderer::buffer_type(buffer),
-                Some(crate::backend::renderer::BufferType::Dma)
-            ),
+            ExportBuffer::Wayland(buffer) => crate::wayland::dmabuf::get_dmabuf(buffer)
+                .ok()
+                .is_some_and(|buf| buf.node().is_some_and(|node| Some(node) == self.render_node)),
```
cmeissl: Not sure I like the name `render_node` in this context, maybe `import_node` or something like this?

Drakulix: sure.

cmeissl:
> Can't think of a better way to introduce the render_node for this check.
> 
> If we are satisfied with this approach, I'll add some more docs to `GbmFramebufferExporter::new` before merging this.

imo it looks fine, so beside docs LGTM

Drakulix:
> > Can't think of a better way to introduce the render_node for this check.
> > If we are satisfied with this approach, I'll add some more docs to `GbmFramebufferExporter::new` before merging this.
> 
> imo it looks fine, so beside docs LGTM

Already added some.

> /// `import_node` will be used to filter dmabufs to originate from a particular
> /// device before considering them for direct scanout.
> /// If `import_node` is `None` direct-scanout of client-buffers won't be used.

You think that covers it?

cmeissl (approved):

PR #1757, opened 2025-06-12
Implement `pointer-warp-v1` protocol

PolyMeilex:


ids1024:
Wayland-rs is released with the necessary protocol now.

src/wayland/pointer_warp.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Whether or not the compositor honors the request is implementation defined, but it should
+    /// - honor it if the surface has pointer focus, including when it has an implicit pointer grab
+    /// - reject it if the enter serial is incorrect
+    /// - reject it if the requested position is outside of the surface
+    ///
+    /// Note that the enter serial is valid for any surface of the client, and does not have to be from the surface the pointer is warped to.
+    ///
+    /// * `serial` - serial number of the surface enter event
+    #[allow(unused)]
+    fn warp_pointer(&mut self, surface: WlSurface, pointer: WlPointer, x: f64, y: f64, serial: Serial) {}
```
ids1024: Would `(x, y)` here be accurately typed as a `Point<f64, Logical>`?

(Still would be good to break up coordinate spaces so that encodes the fact it's surface-local, but I think this is of the coordinate spaces we use that type for, anyway.)

PolyMeilex: I'm not sure to be honest, I would assume it's `Point<f64, Client>` (analogously to how `wl_pointer::motion` works in Smithay), and one would have to do:
```rs
let client_scale = wl_pointer
    .data::<PointerUserData<D>>()
    .unwrap()
    .client_scale
    .load(Ordering::Acquire);
let location: Point<f64, Client> = todo!();
let location = location.to_logical(client_scale);
```
To get the logical position out of it.

Drakulix: Well then maybe we should do the calculation inside the abstraction like we do with `client_scale` everywhere else.

PolyMeilex: done, callback now accepts `Point<f64, Logical>`

PR #1759, opened 2025-06-15
Fix integer overflows in `LayerMap::arrange()`

YaLTeR:
Stumbled upon this due to a use-after-free in quickshell, but it's quite easy to reproduce by just requesting i32::MAX margins over Wayland.

Sample test case in niri: https://github.com/YaLTeR/niri/commit/776c9bd2cda500bfa755a4a9149e6ab0300f95a5#diff-c9d5077797e52564389913b095b8def52b21223fe33660897c9d181354881a80R35

Initially I started converting all math to `.saturating_add()` but that got unwieldy so I decided to just wrap it in a `Saturating<>`.

### DIFF
--- a/src/desktop/wayland/layer.rs
+++ b/src/desktop/wayland/layer.rs
@@ -1,7 +1,7 @@
 use crate::{
     desktop::{utils::*, PopupManager},
     output::{Output, WeakOutput},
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
Didn't know about Saturating! Looks good

PR #1763, opened 2025-06-24
wayland/output: Add an `Output::leave_all` function, and use in anvil before removing output global

ids1024:
This is meant to address things like https://github.com/pop-os/cosmic-comp/issues/1463. I haven't tested with anything that has been crashing yet, but `foot` no longer produces a warning.

Without the `space.refrash` after `space.unmap_output` in anvil, I see it `enter` and `leave` the output again when it shouldn't. Perhaps `output_enter` being called, and resulting in `output_update`, but it's not until `Space::refresh` that the element has it's `output_overlap` synced with the space. Not sure if there's a better way to manage this...

Alternately to `leave_all`, compositors could make sure things like `Space` all send leave before the global is removed. Though that needs to be carefully handled for every type of output. (But making sure unexpected uses of `enter` don't happen is needed either way).

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -624,6 +624,7 @@ struct SurfaceData {
     dh: DisplayHandle,
     device_id: DrmNode,
     render_node: Option<DrmNode>,
+    output: Output,
     global: Option<GlobalId>,
     drm_output: DrmOutput<
         GbmAllocator<DrmDeviceFd>,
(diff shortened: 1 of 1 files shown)


Drakulix:
It seems a bit hacky, but I really don't have a better idea. So imo this looks good.

ids1024:
I suppose there's nothing really unreasonable about a `leave_all` function when `Output` tracks what surfaces are entered on it. It's a bit messy when things like the space abstraction have their own tracking, but I guess that's true already.

We can always overhaul the way these APIs work later if anyone has better ideas.

Drakulix (approved):

PR #1768, opened 2025-07-02
Add option to accept any node for direct scanout

chrisduerr:
This patch introduces a breaking change which allows restoring the pre-0.7.0 functionality of acccepting any node for direct scanout for the `GbmFramebufferExporter` constructor.

While it would be possible to do the same in a non-breaking manner by introducing a separate constructor, this should be more elegant long-term while requiring minimal breaking changes.

### DIFF
--- a/src/backend/drm/exporter/gbm.rs
+++ b/src/backend/drm/exporter/gbm.rs
@@ -35,7 +35,7 @@ pub struct GbmFramebufferExporter<A: AsFd + 'static> {
     gbm: gbm::Device<A>,
     drm_node: Option<DrmNode>,
     #[cfg_attr(not(feature = "wayland_frontend"), allow(unused))]
-    import_node: Option<DrmNode>,
+    import_node: NodeFilter,
 }
 
 impl<A: AsFd + 'static> GbmFramebufferExporter<A> {
@@ -43,8 +43,10 @@ impl<A: AsFd + 'static> GbmFramebufferExporter<A> {
     ///
     /// `import_node` will be used to filter dmabufs to originate from a particular
     /// device before considering them for direct scanout.
-    /// If `import_node` is `None` direct-scanout of client-buffers won't be used.
-    pub fn new(gbm: gbm::Device<A>, import_node: Option<DrmNode>) -> Self {
+    ///
+    /// If `import_node` is [`NodeFilter::None`], direct-scanout of client-buffers
+    /// won't be used.
+    pub fn new(gbm: gbm::Device<A>, import_node: NodeFilter) -> Self {
         let drm_node = DrmNode::from_file(gbm.as_fd()).ok();
         Self {
             gbm,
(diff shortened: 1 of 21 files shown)


src/backend/drm/exporter/gbm.rs
```diff
@@ -95,7 +97,7 @@ impl<A: AsFd + 'static> ExportFramebuffer<GbmBuffer> for GbmFramebufferExporter<
             #[cfg(not(all(feature = "backend_egl", feature = "use_system_lib")))]
             ExportBuffer::Wayland(buffer) => crate::wayland::dmabuf::get_dmabuf(buffer)
                 .ok()
-                .is_some_and(|buf| buf.node().is_some_and(|node| Some(node) == self.import_node)),
+                .is_some_and(|buf| buf.node().is_some_and(|node| self.import_node == node)),
```
chrisduerr, line 100: I'm not fully aware of the context in which this method is used, but for implementing this patch I've assumed that not having any node will never allow adding a framebuffer, even if `NodeFilter::All` is used.

chrisduerr:
I'd recommend rebasing this, should be all separated nicely.

src/backend/drm/exporter/gbm.rs
cmeissl: Yes, that reflects the current behavior, so keeping it like this is fine imo.

chrisduerr:
> Are those two unrelated commits still needed?

If you take a look at the individual commits, you can see that with the latest Smithay master they don't really contain much anymore.

The clippy lint stuff I'd just squash because imo they're still improvements it's just not going to be an error yet with Smithay's pipeline (but will in the future). The links stuff is now just a formatting change which imo is an improvement, but could also be considered unnecessary noise.

If you want me to remove any of this, let me know, but it seems fine to me? I'd definitely squash though.

src/backend/drm/exporter/gbm.rs
```diff
@@ -35,16 +35,18 @@ pub struct GbmFramebufferExporter<A: AsFd + 'static> {
...
     /// Initialize a new framebuffer exporter.
     ///
     /// `import_node` will be used to filter dmabufs to originate from a particular
     /// device before considering them for direct scanout.
-    /// If `import_node` is `None` direct-scanout of client-buffers won't be used.
-    pub fn new(gbm: gbm::Device<A>, import_node: Option<DrmNode>) -> Self {
+    ///
+    /// If `import_node` is [`NodeFilter::None`], direct-scanout of client-buffers
+    /// won't be used.
+    pub fn new(gbm: gbm::Device<A>, import_node: NodeFilter) -> Self {
```
Drakulix, line 49: We could write this as `import_node: impl Into<NodeFilter>` to ease the api change, but perhaps the more explicit notion is better here.

cmeissl: Yeah, I also noticed this but did not mention it because it would not help for the `None` case and being explicit here is fine.

cmeissl:
> If you want me to remove any of this, let me know, but it seems fine to me? I'd definitely squash though.

No, I am fine with including these commits. I was just not sure if these were meant to just make the CI happy because of some nightly changes.
Lets keep them.

PR #1771, opened 2025-07-08
xwm: Handle `_NET_WM_WINDOW_OPACITY` window property

ids1024:
As defined by https://specifications.freedesktop.org/wm-spec/latest/ar01s05.html.

Not sure how often this is used, but `xwaylandvideobridge` uses it (among other properties to try to hide its window from view).


Drakulix (approved):
LGTM

PR #1773, opened 2025-07-09
wl_seat v10 with compositor key repeat

dcz-self:
This adds compositor-side key repeat from the just released wl_keyboard v10.

There are some problems:
1. ~~Repeat events have the same serial~~
2. Filtering may be broken
3. Arming the repeat timeout is ugly

With the above in mind, I'm not sure if I plugged it in the correct place, so I'm marking this as a draft.

### DIFF
--- a/src/backend/input/mod.rs
+++ b/src/backend/input/mod.rs
@@ -96,6 +96,26 @@ pub enum KeyState {
     Pressed,
 }
 
+/// Keyboard key event. Modelled on the wl_keyboard KeyState.
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum KeyEvent {
+    /// Key was released
+    Released,
+    /// Key was pressed
+    Pressed,
+    /// Key is being held and repetition was triggered.
+    Repeated,
+}
+
+impl From<KeyState> for KeyEvent {
+    fn from(k: KeyState) -> Self {
+        match k {
+            KeyState::Released => Self::Released,
+            KeyState::Pressed => Self::Pressed,
+        }
+    }
+}
+
 /// Trait for keyboard event
 pub trait KeyboardKeyEvent<B: InputBackend>: Event<B> {
     /// Returns the numerical button code of the keyboard button.
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -1,18 +1,23 @@
 //! Keyboard-related types for smithay's input abstraction
 
-use crate::backend::input::KeyState;
+use crate::backend::input::{Event, InputBackend, KeyEvent, KeyState, KeyboardKeyEvent};
+use crate::reexports::calloop::LoopHandle;
 use crate::utils::{IsAlive, Serial, SERIAL_COUNTER};
+use calloop::RegistrationToken;
 use downcast_rs::{impl_downcast, Downcast};
 use std::collections::HashSet;
 #[cfg(feature = "wayland_frontend")]
 use std::sync::RwLock;
+use std::time::Duration;
 use std::{
     default::Default,
     fmt, io,
     sync::{Arc, Mutex},
 };
 use thiserror::Error;
-use tracing::{debug, info, info_span, instrument, trace};
+#[cfg(feature = "wayland_frontend")]
+use tracing::error;
+use tracing::{debug, info, info_span, instrument, trace, warn};
 
 use xkbcommon::xkb::ffi::XKB_STATE_LAYOUT_EFFECTIVE;
 pub use xkbcommon::xkb::{self, keysyms, Keycode, Keysym};
@@ -47,7 +52,7 @@ where
         seat: &Seat<D>,
         data: &mut D,
         key: KeysymHandle<'_>,
-        state: KeyState,
+        event: KeyEvent,
         serial: Serial,
         time: u32,
     );
@@ -215,6 +220,19 @@ pub(crate) struct KbdInternal<D: SeatHandler> {
     led_mapping: LedMapping,
     pub(crate) led_state: LedState,
     grab: GrabStatus<dyn KeyboardGrab<D>>,
+    /// Holds the token to cancel key repeat.
+    /// The token gets cleared when the keyboard is dropped, to neutralize the repeat callback.
+    pub(crate) key_repeat_timer: Arc<Mutex<Option<RegistrationToken>>>,
+}
+
+#[cfg(feature = "wayland_frontend")]
+impl<D: SeatHandler> Drop for KbdInternal<D> {
+    fn drop(&mut self) {
+        let timer = self.key_repeat_timer.lock().unwrap().take();
+        if timer.is_some() {
+            error!("A keyboard was dropped without unregistering a repeat handler. This is a bug in smithay or in the compositor.");
+        }
+    }
 }
 
 // focus_hook does not implement debug, so we have to impl Debug manually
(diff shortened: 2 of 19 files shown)


dcz-self:
I fixed the serial generation by pushing the logic up a bit, close to where the backend receives the event.

The events go now like:
1. comp: Backend
2. smithay: Repeat registration -> transforms backend event into generic event, adds Repeated
3. comp (via callback): event handler
4. comp: filtering -> adds Serial
5. comp: forwarding
6. smithay: normal client handling

This means the repeating is global to the seat: switch from one client to the other and repeating continues. If repeating was per-client, then the compositor itself would not receive repeat events without making it look like a client (and I'm not ready for that level of rearchitecting). The good news is I think feels more natural :) ~~The bad news is corner cases when switching because a Repeated event may not appear without a Pressed event according to wl_keyboard:~~

> The key may only enter the repeated state after entering the pressed
>	state and before entering the released state.

~~I haven't tried to add those corner cases yet.~~

EDIT: I read the spec more carefully and this is not needed.

Does this design look right?

src/input/keyboard/mod.rs
```diff
@@ -215,6 +218,17 @@ pub(crate) struct KbdInternal<D: SeatHandler> {
...
+    /// Warning: the token cannot unregister itself when the object is dropped, because it can't hold a reference to the event loop. This object is Send, but the event loop is not.
+    pub(crate) key_repeat_timer: Option<RegistrationToken>,
+}
+
+#[cfg(feature = "wayland_frontend")]
+impl<D: SeatHandler> Drop for KbdInternal<D> {
+    fn drop(&mut self) {
+        if self.key_repeat_timer.is_some() {
+            error!("A keyboard was dropped without unregistering a repeat handler. This is a bug in smithay or in the compositor.");
+        }
```
PolyMeilex: This is kinda hacky, could we perhaps create a `is_alive: Arc<AtomicBool>` and pass it to the timer callback, so that if the keyboard is dead the last timer call is simply ignored, and the timer is unregistered on its own by doing `if is_dead { return TimeoutAction::Drop }`?

Not saying that this is a good idea tho, just thinking out loud.

anvil/src/state.rs
```diff
@@ -301,6 +301,27 @@ impl<BackendData: Backend> SeatHandler for AnvilState<BackendData> {
...
+        mut callback: impl FnMut(&mut Self) + 'static,
+    ) -> RegistrationToken {
+        self.handle.insert_source(
+            calloop::timer::Timer::from_duration(duration),
+            move |_, _, state| {
+                callback(state);
+                calloop::timer::TimeoutAction::ToDuration(duration)
+            },
+        ).unwrap()
+    }
```
PolyMeilex: Is this used anywhere?

src/input/keyboard/mod.rs
```diff
@@ -931,6 +950,91 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
...
+                let handle = get_handle(data);
+                guard.key_repeat_timer = Some(handle.insert_source(
+                    calloop::timer::Timer::from_duration(duration),
+                    move |_, _, data| {
+                        time_ms += delay as u32;
+                        on_event(data, KeyEvent::Repeated, time_ms, keycode);
+                        let mut guard = kbd.internal.lock().unwrap();
+
+                        match guard.key_repeat_timer {
+                            Some(token) => data.clear_timeout(token),
```
PolyMeilex: Could you explain a little bit what is the use case of `clear_timeout` (and `set_timeout`), could this just be a `handle.remove(token)`?

src/input/mod.rs
```diff
@@ -161,6 +161,19 @@ pub trait SeatHandler: Sized {
...
     fn led_state_changed(&mut self, _seat: &Seat<Self>, _led_state: LedState) {}
+    
+    /// Register a timeout for key repeat.
+    fn set_timeout(
+        &mut self,
+        duration: std::time::Duration,
+        callback: impl FnMut(&mut Self) + 'static,
+    ) -> RegistrationToken;
+    
+    /// Cleare a nonrepeating timeout for key repeat.
```
PolyMeilex:
```suggestion
    /// Clear a nonrepeating timeout for key repeat.
```

src/input/keyboard/mod.rs
dcz-self: Yes, it's handle.remove(token) now. It was a leftover from a previous attempt that I haven't removed from the code or apparently from my head.

src/input/keyboard/mod.rs
dcz-self: I think this is a good idea when the consequences of a small oversight by the compositor is a stuck key.
I managed to come up with something like you suggested.

dcz-self:
This still relies on unreleased wayland-rs, but I copleted and undrafted it to reflect my confidence.

PolyMeilex:
~`wayland_protocols_experimental` is now up on crates.io~
`wayland_rs` is now up on crates.io

dcz-self:
Updated, now using all upstream.

dcz-self:
Bump - this is pretty solid I think.
It would make Smithay the first library supporting the key repeat, as far as I can tell.

dcz-self:
Is there anything I can do to push this forward?

dcz-self:
Rebased.

dcz-self:
Updated to fix a failing pipeline.
The other failing pipelines seem to be the same on master, and I don't even know how to start fixing them.

dcz-self:
Copying from Matrix the eplanation about why there are compositor-side API changes:

TL;DR: I don't think this should be transparent. That would mean forcing a design choice on downstreams

EventLoop is not the main reason for changing the downstream code. The main reason is the relationship between repeating and consuming events.
Right now, without repeat, when the compositor eats a buttondown event because it triggers some action, that event will not trigger repeating. That means two things: holding a button doesn't retrigger a compositor action, and after triggering and finishing a compositor action, the active application won't get repeats.
An alternative is to plug in repeats earlier, and have the compositor be able to trigger internal actions also on repeats.
Because there is no automatically right answer here, I decided to not hardcode it in smithay, but let the compositor author choose the behaviour. In anvil, I implemented the alternative pipeline with repeats before internal events.
Choosing behaviour is not transparent, so you get to call `key_register_repeat` and `key_stop_repeat` explicitly.
This is the real reason stuff shifted around: if I can reconstruct this correctly, `process_key_event_windowed` had to be separated from general event processing because it's now a callback that gets triggered on repeat.

dcz-self:
Is there anything I can do to push this forward?

dcz-self:
Bump.

src/backend/input/mod.rs
```diff
@@ -96,6 +96,26 @@ pub enum KeyState {
...
+    Pressed,
+    /// Key is being held and repetition was triggered.
+    Repeated,
+}
+
+impl From<KeyState> for KeyEvent {
+    fn from(k: KeyState) -> Self {
+        match k {
+            KeyState::Released => Self::Released,
+            KeyState::Pressed => Self::Pressed,
```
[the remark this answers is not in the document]
dcz-self: What do you think could be done here?

For some background, this exists only to support the virtual-keyboard protocol which can issue Pressed or Released but not Repeated. This conversion happens to feed the event into the rest of the processing.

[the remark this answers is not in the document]
dcz-self: No, the conversion goes in the other direction: from 2 states to 3.
Pressed or Released → Pressed or Released or (unused Repeated).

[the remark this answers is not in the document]
dcz-self: In practice, this means that when someone uses the virtual-keyboard protocol, they won't be able to trigger key repeat if the client supports server-side key repeat. The virtual-keyboard protocol would need changes.

PR #1778, opened 2025-07-15
drm/manager: Provide explicit lock api

Drakulix:
Needs some testing in cosmic-comp, but I believe this is necessary to provide race-free state changes.

Note: We could potentially provide two variants of this api, the old one and the new one. I don't see a particular advantage to that approach other than preserving a relatively new and barely used api.


Drakulix:
Confirmed this does end up fixing our issues in cosmic.

Especially changing the `Output`, when used as an `OutputModeSource` can be particularly brittle, as that is used for `CRTC_W` and `CRTC_H`, which might cause the commit to be rejected, if the hardware cannot do the necessary scaling with a new mode.

So without locks and careful order of the operations any thread might use the wrong resolution temporarily or any tests by the `DrmOutput`/`DrmOutputManager` using the standard code paths can be rejected causing the new valid configuration to fail being applied.

This is sufficient to get rid of the races, though this is still easier to shoot yourself in the foot than I would like it to be.

cmeissl:
> This is sufficient to get rid of the races, though this is still easier to shoot yourself in the foot than I would like it to be.

Yeah, not sure how to resolve this. I guess at some point it would be great to have some kind of output abstraction across backends.

PR #1781, opened 2025-07-19
Reapply "selection: instantly notify clients when source dies"

kchibisov:
With the fixes and adjustments that were missed during the review. The new protocol errors were also relaxed to follow `wlroots` at this time, since firefox does violate protocol a bit and thus crashing it is not a great thing to do.

This reverts commit d34af7bfe0f940be58e0d030207092e361ddd7da.


kchibisov:
@YaLTeR wrt primary selection there was one check missing in place, which wasn't noticed. As for the firefox, I've relaxed the restrictions to follow wlroots here. It does have the multiple source errors, but only for `data_control` family, even though, the same wording is present inside the other two protocols.

YaLTeR:
Gave this some testing, seems to work, including on the problematic cases that I found the last time.

PR #1783, opened 2025-07-21
fix: the point containment to an `input_region`

Bqrry4:
This fixes a bug where defining an input region on the bottom or right edges of a surface causes pointer input to be skipped when moving the pointer to those edges. Especially noticeable and annoying when the surface is fullscreen or positioned so the region aligns with the screen edges.

### DIFF
--- a/src/desktop/wayland/utils.rs
+++ b/src/desktop/wayland/utils.rs
@@ -47,7 +47,7 @@ impl RendererSurfaceState {
             .input_region
             .as_ref()
             .unwrap()
-            .contains(point.to_i32_round())
+            .contains(point.to_i32_floor())
     }
 }
 


Drakulix:
Do you have a good test case for this? I wonder if using `floor` here would make the situation worse for the left/upper edge of input regions.

phisch:
@Drakulix

> [!<>] 
> Epilepsy warning. This video contains flashing.

https://github.com/user-attachments/assets/9c4fb10c-d8d9-4ebf-bc31-cc1c88e0a1f3

phisch:
@Drakulix

You can use this quickshell configuration:
```qml
import Quickshell
import QtQuick

PanelWindow {
    anchors {
        top: true
        left: true
        right: true
        bottom: true
    }

    mask: Region {
        item: mask_rectangle
    }

    Rectangle {
        id: mask_rectangle
        anchors.fill: parent
        color: hover_handler.hovered ? "#000000" : "#808080"
    }

    HoverHandler {
        id: hover_handler
        target: mask_rectangle
    }
}
```

It creates a layer shell surface that is anchored to all sides, and then adds a mask that fills the entire surface. It'll have a black background on hover, and a grey one when not hovered. So technically it shouldn't be possible to not hover it, but it flashes grey on the edges sometimes.

Here is a video showing it:

> [!caution] 
> Epilepsy warning. This video contains flashing.

https://github.com/user-attachments/assets/9c4fb10c-d8d9-4ebf-bc31-cc1c88e0a1f3

Bqrry4:
> I wonder if using `floor` here would make the situation worse for the left/upper edge of input regions.

Internally the `input_region.contains(...)` just calls the `rect.contains(point)`, which is defined as
```rust
(p.x >= self.loc.x)
  && (p.x < self.loc.x.saturating_add(self.size.w))
  && (p.y >= self.loc.y)
  && (p.y < self.loc.y.saturating_add(self.size.h))
```
As you can see, a point in between [size.wh - 0.5, size.wh] will be discarded when round is used instead of floor. So this won't affect the left/upper edges.

Drakulix (approved):
Makes sense, thx.

PR #1785, opened 2025-07-22
element/surface: Allow filtering of scanout candidates

Drakulix:
This adds an API to allow filtering, which elements should be considered for scanout and which shouldn't.

Example usage and motivation in cosmic: https://github.com/pop-os/cosmic-comp/pull/1542

Draft because I am not sure, if this is a particular good api for this feature. Ping @cmeissl on that.

Essentially since the `DrmCompositor` is the only piece of code using `underlying_storage` at the moment I opted for the lazy way of just returning `None`, if we don't want the target to be considered for scanout. This is a bit hacky, maybe another method would be more appropriate.

The bigger issue however is the change to `render_elements_from_surface_tree`.

I didn't want to modify `AsRenderElements` to pass through a closure barely used by any other element. This ends up still working in cosmic-comp, as we happen to use `render_elements_from_surface_tree` almost everywhere wayland-surfaces are involved anyway to separate popup-elements and toplevel-elements (which the default implementation for `desktop::Window` doesn't do).

However this is arguably annoying as this makes it impossible to do such filtering, while relying on `Space` or other `desktop`-abstractions for rendering.

So I am asking for ideas to make this easily controllable from compositors without trashing our current api (maybe that is a pipe-dream).

Alternatively non of this needs to be in smithay, a compositor could simply wrap the elements to do the filtering. But given the elements don't contain any references to the original surface except for their id, compositors would likely end up not using `AsRenderElements` again and we would just make everything even more annoying by forcing compositors to add another wrapper class to do this.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -3027,7 +3027,7 @@ where
             return None;
         };
 
-        // only try to assgin elements on a cursor plane that indicate so
+        // only try to assign elements on a cursor plane that indicate so
         if element.kind() != Kind::Cursor {
             trace!(
                 "skipping element {:?} on cursor plane(s), element kind not cursor",
@@ -3712,6 +3712,15 @@ where
         E: RenderElement<R>,
     {
         if !frame_flags.contains(FrameFlags::ALLOW_OVERLAY_PLANE_SCANOUT) {
+            return Err(None);
+        }
+
+        // only try to assign elements on an overlay plane that indicate so
+        if element.kind() != Kind::ScanoutCandidate && element.kind() != Kind::Cursor {
+            trace!(
+                "skipping element {:?} on overlay plane(s), element kind not scanout-candidate/cursor",
+                element.id(),
+            );
             return Err(None);
         }
 
--- a/src/backend/renderer/element/mod.rs
+++ b/src/backend/renderer/element/mod.rs
@@ -334,6 +334,23 @@ pub enum Kind {
     /// Not marking a cursor element as `Cursor` may result in lower performance and increased power usage.
     /// In contrast, marking elements that change frequently as `Cursor` can degrade performance significantly.
     Cursor,
+    /// The element is a good candidate for scanout
+    ///
+    /// An element marked as a scanout candidate is expected to refresh frequently or with a constant rate.
+    /// As such it makes a good candidate to save resources by placing it on an overlay plane, when using the drm backend
+    /// (see [`DrmCompositor`](crate::backend::drm::compositor::DrmCompositor)).
+    ///
+    /// Examples of applications like this include video players or games. To detect good candidates compositor should
+    /// employ heuristics. For example, it might keep track of the surface commit rate and mark surfaces hitting thresholds
+    /// as scanout candidates.
+    ///
+    /// Note: You can select a different `Kind` every time you create the render element from e.g. a wayland surface.
+    /// This means it is expected that compositors dynamically choose the appropriate `Kind` at runtime instead
+    /// of relying on static configuration. E.g. it can make sense to switch back to `Unspecified` during very dynamic
+    /// scenes like animations, as moving planes can be slow on certain hardware.
+    ///
+    /// Elements not marked as scanout candidates will never be considered, when selecting potential buffers for overlay planes.
+    ScanoutCandidate,
     /// The element kind is unspecified
     #[default]
     Unspecified,
--- a/src/backend/renderer/element/surface.rs
+++ b/src/backend/renderer/element/surface.rs
@@ -84,8 +84,66 @@ use crate::{
 
 use super::{CommitCounter, Element, Id, Kind, RenderElement, UnderlyingStorage};
 
+/// Ways to evaluation the [`Kind`] of a surface.
+pub enum KindEvaluation {
+    /// Static evaluation, which will always return the same [`Kind`]
+    Static(Kind),
+    /// Dynamic evaluation, which will return a [`Kind`] based on the [`SurfaceData`]
+    Dynamic(fn(&SurfaceData) -> Kind),
+    /// Dyanmic evaluation, which will return a [`Kind`] based on the [`SurfaceData`] and potentially additional captured context.
+    Closure(Box<dyn Fn(&SurfaceData) -> Kind>),
+}
+
+impl fmt::Debug for KindEvaluation {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            KindEvaluation::Static(kind) => kind.fmt(f),
+            _ => f.write_str("Kind::<dynamic>"),
+        }
+    }
+}
+
+impl From<Kind> for KindEvaluation {
+    fn from(value: Kind) -> Self {
+        KindEvaluation::Static(value)
+    }
+}
+
+impl<F: Fn(&SurfaceData) -> Kind + 'static> From<F> for KindEvaluation {
+    fn from(value: F) -> Self {
+        KindEvaluation::Closure(Box::new(value))
+    }
+}
+
+// TODO: specialization
+/*
+impl From<fn(&SurfaceData) -> Kind> for KindEvaluation {
+    fn from(value: fn(&SurfaceData) -> Kind) -> Self {
+        KindEvaluation::Dynamic(value)
+    }
+}
+*/
+
+impl KindEvaluation {
+    /// Create a `KindEvaluation` from a pure function.
+    ///
+    /// Using this instead of the `From`-implementation skips boxing
+    /// and thus can provide better performance as it is not relying on dynamic dispatching.
+    pub fn from_fn(func: fn(&SurfaceData) -> Kind) -> Self {
+        KindEvaluation::Dynamic(func)
+    }
+
+    fn eval(&self, data: &SurfaceData) -> Kind {
+        match self {
+            KindEvaluation::Static(res) => *res,
+            KindEvaluation::Dynamic(func) => func(data),
+            KindEvaluation::Closure(func) => func(data),
+        }
+    }
+}
+
 /// Retrieve the [`WaylandSurfaceRenderElement`]s for a surface tree
-#[instrument(level = "trace", skip(renderer, location, scale))]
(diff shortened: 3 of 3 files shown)


src/backend/renderer/element/surface.rs
```diff
@@ -94,6 +94,7 @@ pub fn render_elements_from_surface_tree<R, E>(
     scale: impl Into<Scale<f64>>,
     alpha: f32,
     kind: Kind,
+    consider_for_scanout: impl Fn(&SurfaceData) -> bool,
```
YaLTeR: `kind` was originally introduced to skip cursor surfaces for scanout, would be unfortunate to add an extra parameter that is a strict superset of the functionality

Drakulix: `Kind::Cursor` is sadly a bit special as well, as it specifically denotes surfaces that should be considered to be scanned out on the cursor-plane, while this affects scanout on any plane.

But I guess, if we added the frametime tracking to smithay as well, we could introduce something like `Kind::AboveRefreshRate(u32)`, which really breaks the name though.

Alternatively we could have `Kind::ConstantRate` (or a more explicit `ScanoutCandidate`) vs `Kind::Unspecified` and use that for filtering in `DrmCompositor`, while compositors use whichever heuristic to figure out if they want to use the new variant?

EDIT: But the latter wouldn't be granular enough as the closure can filter on surface-basis, while a new `Kind`-variant would mark the whole surface-tree, which is definitely not what we want. E.g. for browsers rendering a video in a sub-surface.

Drakulix:
Okay I added a new variant of this, which (almost) preserves the current api and re-uses `Kind` for the filtering.

The new api is best viewed over [all commits](https://github.com/Smithay/smithay/pull/1785/files), but is pushed a second one to preserve the old attempt for now. (I intend to squash before merging.)

If we were to add the [per-surface frame-time estimation](https://github.com/pop-os/cosmic-comp/pull/1542/commits/f82864a5539235de3e945beba18793b294dcadd3) from cosmic directly to smithay as well (which doesn't need to be the default, users could choose to add the post-commit-hook), we could even have `Window` and `LayerSurface` make use of that to set the new `Kind::ScanoutCandidate` (bike-shedding over the name is welcome) automatically.

src/backend/renderer/element/mod.rs
```diff
@@ -334,6 +334,12 @@ pub enum Kind {
     /// Not marking a cursor element as `Cursor` may result in lower performance and increased power usage.
     /// In contrast, marking elements that change frequently as `Cursor` can degrade performance significantly.
     Cursor,
+    /// The element is a good candidate for scanout
+    ///
+    /// An element marked as a scanout candidate is expected to refresh frequently or with a constant rate.
```
YaLTeR, line 339: I guess I'd clarify in the docs here that:
- you can change Kind every frame. The way this is currently worded, emphasizing frequent refreshes, makes my brain drift towards some persistent surface flag
- given that you can change Kind every frame, you want to reset it back to Unspecified during movement animations because those are usually slow (and whatever else is slow)
- would be good to mention examples like video players
- would be good to mention how the compositor is actually supposed to know that a surface is a good scanout candidate. Something along "For example, the compositor can keep track of the surface commit rate and mark surfaces with consistent commits as scanout candidates." I guess, after you do more experimenting in cosmic-comp with how to best use this

src/backend/drm/compositor/mod.rs
```diff
@@ -3715,6 +3715,15 @@ where
             return Err(None);
         }
 
+        // only try to assgin elements on a cursor plane that indicate so
+        if element.kind() != Kind::ScanoutCandidate && element.kind() != Kind::Cursor {
```
cmeissl, line 3719: What is the Intention of including the check for Cursor here?

Small nit: the comment above does reference the cursor plane

This does only not apply to the primary plane, if this is intentional the docs should also mention it imo.

cmeissl: Just noticed it does mention the overlay plane, but maybe we should still explicitly mention that it does not apply to the primary plane.

Drakulix: > What is the Intention of including the check for Cursor here?

There is modern hardware, that doesn't expose a Cursor plane anymore. E.g. macbooks running asahi-linux don't, as what mac os uses as a "cursor"-plane does violate user-space assumptions as it can't overlap with the screen edge. Given the plane is otherwise very performant it is simply exposed as an overlay plane.

AMD might also not expose their weird cursor plane with the upcoming color-pipeline api, since it's color-space is determined by the top-most plane below the cursor. In fact it internally already re-uses free overlay planes for the cursor if available.

`kwin` thus gained the ability to simply put the cursor on an overlay plane to enable hardware cursors on these machines, which we most likely also want to allow.

> This does only not apply to the primary plane, if this is intentional the docs should also mention it imo.

Yes, the idea was that there isn't really a reason to not do primary-scanout, if we can. If there is reason to do this in general because of buggy hardware or similar, there are also the `FrameFlags`. (Which e.g. cosmic-comps branch is using to disable overlay-scanout in general, when animations are running instead of individually filtering all elements.)

cmeissl: > `kwin` thus gained the ability to simply put the cursor on an overlay plane to enable hardware cursors on these machines, which we most likely also want to allow.

I see, thanks for the explanation. Then only the comment should be changed from
> // only try to assgin elements on a cursor plane that indicate so

to

> // only try to assgin elements on a overlay planes that indicate so

We can keep the part relaxing the check so that we try to assign a cursor element to an overlay plane, but we should create a follow up issue for that. The reason is that I expect the majority of cursor elements (including the compositor handled cursors) to not support scan-out at all. So in this case we still have to somehow copy the pixel data to some scan-out buffer.

Drakulix: Good catch. We probably can't even re-use the existing code, as we don't want to use a DumbBuffer on anything that is not a cursor plane, so we likely need to render with the user-provided renderer into a new swapchain?

If we do this correctly, this might enable a path to compositing on something that isn't the overlay plane.

src/backend/renderer/element/mod.rs
Drakulix: Expanded the doc comment. Does this seem good now?

Drakulix:
Marking as ready to review, though I still intend to fix up the commit history, if the approach looks good to everybody as it stands.

It works very well from my current testing in cosmic. Is there interest from e.g. niri (@YaLTeR) to add the frame-time estimation code directly to smithay?

src/backend/drm/compositor/mod.rs
```diff
@@ -3715,6 +3715,15 @@ where
             return Err(None);
         }
 
+        // only try to assgin elements on an overlay plane that indicate so
```
YaLTeR:
```suggestion
        // only try to assign elements on an overlay plane that indicate so
```
while we're at it

src/backend/renderer/element/mod.rs
YaLTeR: Yeah, much better

YaLTeR (commented):
> It works very well from my current testing in cosmic. Is there interest from e.g. niri (@YaLTeR) to add the frame-time estimation code directly to smithay?

Do you have some draft of how it will look so I can see how it will look on the niri side?

Logic-wise, it's straightforward in niri to see when animations are ongoing during rendering. Not sure how annoying it would be to thread this down to where I call render elements from surface tree though.

Other than that, I don't mind enabling it and giving it some testing, but I don't know how to measure battery or energy drain, so I'm limited to just seeing if a surface gets scanned out and then hoping that it does something good.

Drakulix:
> > It works very well from my current testing in cosmic. Is there interest from e.g. niri (@YaLTeR) to add the frame-time estimation code directly to smithay?
> 
> Do you have some draft of how it will look so I can see how it will look on the niri side?

Here is the cosmic code, neatly split into commits: https://github.com/pop-os/cosmic-comp/pull/1542.

The minimum for supporting this in niri would be the equivalent of https://github.com/pop-os/cosmic-comp/pull/1542/commits/8f7c9a3807d4f1ab7d5ac0fd5f4e2b0ba8d786bc and https://github.com/pop-os/cosmic-comp/pull/1542/commits/e3dc775c6f751f1a6c03fc0538d3076a1433ac24 .

The former of these would be well suited to be added to smithay directly, which then niri could just re-use.

That could then also be used by the `Window` and `LayerSurface` abstractions as well, when the post-commit hook is enabled. (But I guess that wouldn't concern you, as niri doesn't use `Space` or `Window` as well, right?)

> Logic-wise, it's straightforward in niri to see when animations are ongoing during rendering. Not sure how annoying it would be to thread this down to where I call render elements from surface tree though.

Note that you don't need to do that. You can simply not pass the overlay-plane `FrameFlag`, when animations are running in the drm-backend. I do that in cosmic: https://github.com/pop-os/cosmic-comp/pull/1542/commits/8db29b794b8ac0c7f5a5ef6659a6ef19b6d308fc

The only exception to that is the `MoveGrab`, which directly passes `Unspecified` to the attached window, when rendering, as that doesn't count as an animation in the global sense, but still very much affects the grabbed window.
 
> Other than that, I don't mind enabling it and giving it some testing, but I don't know how to measure battery or energy drain, so I'm limited to just seeing if a surface gets scanned out and then hoping that it does something good.

If you have an amd gpu there is a neat visual confirmation debugfs flag: https://dri.freedesktop.org/docs/drm/gpu/amdgpu/display/dc-debug.html#dc-visual-confirmation

With that you can easily see when you compositor is using the plane or not and spot if it is disabled during animations, etc. Good checks are windowed games and video players like mpv, as well as video elements in browsers. If those trigger correctly you filter out all the uninteresting targets.

src/backend/drm/compositor/mod.rs
cmeissl: For compositor provided cursors we could do that outside of the drm abstraction. Would be a good opportunity to allow scan-out for all custom compositor provided elements. Not sure how to handle the case for client provided shm buffers. It could also be done like for the custom elements in case no cursor plane is available or implicit in the DrmCompositor.

cmeissl (approved):
beside https://github.com/Smithay/smithay/pull/1785/files#r2231543758 LGTM

PR #1788, opened 2025-07-31
Fix gles without instancing

cmeissl:
@chrisduerr can you please check if this fixes the issue on the PP?


chrisduerr:
> @chrisduerr can you please check if this fixes the issue on the PP?

Will do later today and report back here.

chrisduerr:
I can confirm that I can no longer reproduce the issue with this patch applied.

PR #1793, opened 2025-08-06
xwayland: Fix potentially missed keyboard-events

Drakulix:
This PR tries to address multiple short-comings of our current focus tracking code specifically around Xwayland (but not necessarily requiring Xwayland for reproduction, just weird usage of our public api).

1. We want to send `wl_keyboard::leave` events for focused surfaces, which are about to be destroyed. On the client-side these show up as targeting a `nil`-surface, as the client-side drops the surface immediately after sending the destruction request. However longer-lived clients might still rely on these events and omitting them isn't valid from the protocol perspective. E.g. Xwayland uses these to clear any still "pressed" keys.

However if we remove the focus after processing the `unmap_notify`, the `X11Surface` already has no `wl_surface` any more to forward the leave-event to. Keeping a weak-variant doesn't help - as the surface would already be destroyed - and keeping a `WlSurface` alive sounds like a good opportunity for leaks.

So lets fix this with an destruction_hook, which means wrapping types like `X11Surface` have one less event to worry about.

2. `X11Surface` currently might completely ignore any keyboard events, if the surface in question isn't mapped and we don't have a `wl_surface` to target. This is simply a bad API, as downstream needs to verify it targets a mapped surface before treating it like any other `KeyboardTarget`. We also need to intercept these events to send additional messages to the X11 socket...

cosmic-comp is doing this wrong currently, mapping x11 windows on `XwaylandHandler::map_notify`, which is intuitive, but frequently misses the `wl_keyboard::enter`-event of all things.

This is theoretically an issue with pointer/touch events as well, but we don't have anything to render without a surface and no valid input-regions anyway, so this doesn't manifest in any serious errors, while keyboard related shenanigans cause a variety of fun issues like surfaces never getting focus or "stuck" keys.

To combat this we could split the `X11Surface` type into a `MappedX11Surface`-variant or .... simply keep track of any keyboard events we should have delivered already. That is what the second commit tries to do (not so elegantly, but it works just fine in testing).

Anyway... feedback is very welcome.

### DIFF
--- a/src/wayland/seat/keyboard.rs
+++ b/src/wayland/seat/keyboard.rs
@@ -1,4 +1,4 @@
-use std::{cell::Cell, fmt};
+use std::{cell::RefCell, fmt};
 
 use tracing::{error, instrument, trace, warn};
 use wayland_server::{
@@ -15,7 +15,7 @@ use crate::{
     backend::input::{KeyState, Keycode},
     input::{
         keyboard::{KeyboardHandle, KeyboardTarget, KeysymHandle, ModifiersState},
-        Seat, SeatHandler, SeatState,
+        Seat, SeatHandler, SeatState, WeakSeat,
     },
     utils::{iter::new_locked_obj_iter_from_vec, HookId, Serial},
     wayland::{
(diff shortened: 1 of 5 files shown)


Drakulix:
For context an alternative downstream fix, which can be avoided by this PR and I would consider an API issue (debatable): https://github.com/pop-os/cosmic-comp/pull/1567

src/wayland/seat/keyboard.rs
```diff
@@ -164,37 +168,77 @@ pub fn serialize_pressed_keys(keys: impl Iterator<Item = Keycode>) -> Vec<u8> {
...
+        enter_internal(self, seat, state, keys.iter().map(|h| h.raw_code()), serial)
     }
 
     fn leave(&self, seat: &Seat<D>, state: &mut D, serial: Serial) {
         *seat.get_keyboard().unwrap().arc.last_enter.lock().unwrap() = None;
         for_each_focused_kbds(seat, self, |kbd| kbd.leave(serial.into(), self));
+        with_states(self, |states| {
+            if let Some(hook) = states.data_map.get::<FocusDestroyHook>() {
+                let _ = hook.0.take();
+            }
```
ids1024: Shouldn't this also call `remove_destruction_hook`?

Drakulix: Yes, of course. Forgot somehow!

src/wayland/seat/keyboard.rs
```diff
@@ -164,37 +168,79 @@ pub fn serialize_pressed_keys(keys: impl Iterator<Item = Keycode>) -> Vec<u8> {
...
+                }
+            }
         }
+    });
+    with_states(surface, |states| {
+        states
+            .data_map
+            .get_or_insert::<FocusDestroyHook, _>(Default::default)
+            .0
+            .set(Some(hook_id))
```
cmeissl: Do we prevent calling enter multiple times with the same target? If not this could lead to dangling destruction hooks. It would be possible to either replace the hook here removing any present hook or check before creating a new one.

Drakulix: Good call

Drakulix: done

cmeissl (approved):

PR #1798, opened 2025-08-11
Fix `NodeFilter::All` without render node attached

chrisduerr:
This changes the behavior of the node filter when not having any node specified on the dmabuf. Previously this would always cause direct scanout to fail, regardless of the specified `NodeFilter`. This patch changes it to always accept scanout with `NodeFilter::All`.


Drakulix:
Nah, seems fine to me.

PR #1799, opened 2025-08-12
Add #[must_use] annotations to geometry module

chrisduerr:
This patch liberally sprinkles #[must_use] annotations on all geometry functions which return `Self` without modifying `self`.

---

I just wasted 3 seconds testing a patch just to realize I called `downscale` without assigning it to anything. This patch certainly takes a bit of a lazy approach just copy-pasting the annotations everywhere, but I do think that it's likely to save at least a couple minutes of developer headache in the long term.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -33,14 +33,18 @@ pub trait Coordinate:
     /// A Coordinate that is 0
     const ZERO: Self;
     /// Downscale the coordinate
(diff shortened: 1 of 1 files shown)


PolyMeilex:
> This patch certainly takes a bit of a lazy approach just copy-pasting the annotations everywhere, but I do think that it's likely to save at least a couple minutes of developer headache in the long term.

Yeah, it's better than nothing, thanks!

PR #1802, opened 2025-08-14
fix(test): correct typo "falliable" -> "fallible"

Olexandr88:


PolyMeilex (approved):
Thanks! This will have to be adjusted in all the places the enum is used in the codebase.

PR #1805, opened 2025-08-19
Return `SyncPoint` from blit methods of renderer

ids1024:
These methods are not part of a frame, so if an explicit sync point is needed, they need to return one.

I assume creating a sync point will have quite a low cost. When we're not exporting to an fd or waiting on a different renderer. Otherwise we might need a more cumbersome API; like an `&mut Option<SyncPoint>` method that is filled only if the caller needs a sync point. The `glFinish` invoked if EGL fences aren't supported would be more costly though; but that generally shouldn't be needed.

I think some of the synchronization may get more complicated with a Vulkan renderer, since Vulkan allows commands to complete out of order from when they are started and be synchronized separately, as I understand? I don't have experience with Vulkan synchronization though.

### DIFF
--- a/src/backend/drm/compositor/frame_result.rs
+++ b/src/backend/drm/compositor/frame_result.rs
@@ -347,7 +347,7 @@ where
         }
 
         // first do the potential blit
-        if let Some((sync, mut dmabuf, geometry)) = primary_dmabuf {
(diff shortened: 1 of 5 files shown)


Drakulix (approved):
Looks correct to me.

Drakulix (approved):

PR #1807, opened 2025-08-26
anvil: Support `ToggleTint` on winit/X11

ids1024:
Also makes a small (breaking) change in `WinitGraphicsBackend` to have separate `renderer`/`renderer_mut` methods.


ids1024:
Updated with a couple fixes so the CI build with different anvil features passes.

PR #1811, opened 2025-09-03
multigpu: Don't attempt cross-device imports on software renderers

Drakulix:
llvmpipe segfaults currently, when attempting to import a dmabuf it cannot handle: https://gitlab.freedesktop.org/mesa/mesa/-/issues/6438

While it would be nice to fix this upstream, realistically old mesa versions will ship for quite some time. In general it doesn't make sense to do dma-copies with software renderers anyway.

We can figure out from EGL, if we have llvmpipe as a renderer, but we don't know the node it uses. (We could figure it out via `EGLDevice`, but that might not be the node, we set on the dmabuf, especially in the software renderer case where we most likely don't have a render-node and only a primary one, if it was created from a `GbmDevice` and not via enumerating `EGLDevice`s.)

So instead we assume downstream users that do cross-gpu imports on purpose via the `GlesRenderer` or `EGLDisplay` know what they are doing and check for llvmpipe.

In the `MultiRenderer` this doesn't work, but here we have a node to match the dmabuf, that is associated with the renderer. So skip any (potential) cross-device imports in the `MultiRenderer`.

(This also expects compositors that make use of the `MultiRenderer` to either set a node on every dmabuf **or** to never ever advertise the `GbmDevice`s node of a software renderer as a main-device on the dmabuf protocol. Otherwise you might end up with failing to import every client dmabuf. Realistically you shouldn't advertise the dmabuf protocol, if you only have a software renderer anyway.)

### DIFF
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -49,7 +49,7 @@ use crate::{
         egl::{
             fence::EGLFence,
             ffi::egl::{self as ffi_egl, types::EGLImage},
-            EGLContext, EGLSurface, MakeCurrentError,
+            EGLContext, EGLDevice, EGLSurface, MakeCurrentError,
         },
     },
     utils::{Buffer as BufferCoord, Physical, Rectangle, Size, Transform},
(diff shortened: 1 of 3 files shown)


cmeissl (approved):
small nit: the gles part could be split into a separate commit

Drakulix:
> small nit: the gles part could be split into a separate commit

done!

PR #1812, opened 2025-09-03
pixman: Add framebuffer accessor

Drakulix:
Little commit I coded up attempting something else that never came to fruition.

I do think this might be useful and/or more consistent with e.g. `with_context` in `GlesRenderer`.


src/backend/renderer/pixman/mod.rs
```diff
@@ -349,6 +349,24 @@ impl PixmanFrame<'_, '_> {
...
     }
+
+    /// Run custom drawing code on the framebuffer used by this frame.
+    ///
+    /// You are given mutable access to the underlying [`pixman::Image`].
+    /// The state of the image is considered an implementation detail, some modifications may have undesired side effects.
+    ///
+    /// # Safety
+    ///
+    /// - The image must still referr to the same memory region after `func` was executed.
```
PolyMeilex, line 360:
```suggestion
    /// - The image must still refer to the same memory region after `func` was executed.
```

PolyMeilex (approved):

Issue #1761, opened 2025-06-21
xdg_popup.grab() and parent vs. commit ordering


YaLTeR:
Smithay calls the compositor's `grab()` handler right away, wherein the compositor is supposed to check xdg-popup's parent for validity:

> The parent of a grabbing popup must either be an xdg_toplevel surface or another xdg_popup with an explicit grab.

However, Qt layer-shell seems to grab the popup right before assigning it a parent, breaking this logic:

```
[ 392910.890] {Default Queue}  -> xdg_wm_base#3.get_xdg_surface(new id xdg_surface#52, wl_surface#45)
[ 392910.986] {Default Queue}  -> xdg_surface#52.get_popup(new id xdg_popup#54, nil, xdg_positioner#53)
[ 392911.014] {Default Queue}  -> xdg_popup#54.grab(wl_seat#15, 168583)
[ 392911.030] {Default Queue}  -> zwlr_layer_surface_v1#43.get_popup(xdg_popup#54)
[ 392911.089] {Default Queue}  -> wl_surface#45.commit()

[ 392911.209] {Default Queue}  -> xdg_popup#54.reposition(xdg_positioner#55, 1)
[ 392911.934] {Default Queue} xdg_popup#54.configure(2121, 25, 173, 129)
[ 392911.947] {Default Queue} xdg_surface#52.configure(168585)
[ 392911.959] {Default Queue}  -> xdg_surface#52.ack_configure(168585)
[ 392911.999] {Default Queue} xdg_popup#54.repositioned(1)
[ 392912.012] {Default Queue} xdg_popup#54.configure(2121, 25, 173, 129)
[ 392912.023] {Default Queue} xdg_surface#52.configure(168586)
[ 392912.082] {Default Queue}  -> xdg_surface#52.ack_configure(168586)
[ 392912.119] {Default Queue}  -> xdg_surface#52.set_window_geometry(0, 0, 173, 129)
[ 392912.150] {Default Queue}  -> xdg_surface#52.set_window_geometry(0, 0, 173, 129)
[ 392913.165] {Default Queue}  -> wl_surface#45.commit()
```

Therefore, perhaps Smithay should delay calling the `grab()` handler until the commit?

Apparently, this ordering works fine in kwin. If not, then this would be a Qt bug instead. I'm not sure the protocol is clear on this.

cc @cmeissl @ids1024

outfoxxed:
>However, Qt layer-shell seems to

[...]

Issue #1717, opened 2025-04-17
Map before initial configure does not disconnect the client


YaLTeR:
[xdg-surface](https://wayland.app/protocols/xdg-shell#xdg_surface):

> any attempts by a client to attach or manipulate a buffer prior to the first xdg_surface.configure call must also be treated as errors

Smithay doesn't check this currently.

YaLTeR:
Fixed in #1817

PR #1817, opened 2025-09-05
xdg-shell, layer-shell, session-lock: Track last acked configure in MultiCache

YaLTeR:
Corresponding niri update: https://github.com/YaLTeR/niri/commit/5a4d794b17791386f4103415de95b7557cb369ad

Fixes my fullscreen anim bugs that I found with mpv and adwaita-1-demo.

----

The last-acked configure is part of the committed state, and should be tracked as such. Before this commit, it would be possible to observe a newer "current state" in a commit handler if the surface acked the next state while the compositor was waiting on a blocker set in a pre-commit handler.

As part of the change, the current_state() getter was removed, because it's confusing and likely not what you actually want. For instance, several uses in Anvil actually needed to access a different state.

The WmCapabilities checks in Anvil were removed because they are serverside and don't change; not sure what the intention was with those.

Also, added forced configures in unmaximize() and unfullscreen() because they too are required by the protocol.

----

If this is good I'll do a similar refactor for layer-shell and session-lock.

One kinda weird part is `current_server_state()` that I had to change to owned return, tbh I would change it to return an `Option<&State>`, but that would involve changing a bunch of callers. I can do that though if that's better.

Also, there's a question of whether any other state needs to be moved into Cached?

### DIFF
--- a/anvil/src/shell/xdg.rs
+++ b/anvil/src/shell/xdg.rs
@@ -319,14 +319,6 @@ impl<BackendData: Backend> XdgShellHandler for AnvilState<BackendData> {
     }
 
     fn unfullscreen_request(&mut self, surface: ToplevelSurface) {
-        if !surface
-            .current_state()
-            .states
-            .contains(xdg_toplevel::State::Fullscreen)
-        {
-            return;
-        }
-
         let ret = surface.with_pending_state(|state| {
             state.states.unset(xdg_toplevel::State::Fullscreen);
             state.size = None;
@@ -341,7 +333,13 @@ impl<BackendData: Backend> XdgShellHandler for AnvilState<BackendData> {
             }
         }
 
-        surface.send_pending_configure();
+        // The protocol demands us to always reply with a configure,
+        // regardless of we fulfilled the request or not
+        if surface.is_initial_configure_sent() {
+            surface.send_configure();
+        } else {
+            // Will be sent during initial configure
+        }
     }
 
     fn maximize_request(&mut self, surface: ToplevelSurface) {
@@ -379,19 +377,18 @@ impl<BackendData: Backend> XdgShellHandler for AnvilState<BackendData> {
     }
 
     fn unmaximize_request(&mut self, surface: ToplevelSurface) {
-        if !surface
-            .current_state()
-            .states
-            .contains(xdg_toplevel::State::Maximized)
-        {
-            return;
-        }
-
         surface.with_pending_state(|state| {
             state.states.unset(xdg_toplevel::State::Maximized);
             state.size = None;
         });
-        surface.send_pending_configure();
+
+        // The protocol demands us to always reply with a configure,
+        // regardless of we fulfilled the request or not
+        if surface.is_initial_configure_sent() {
+            surface.send_configure();
+        } else {
+            // Will be sent during initial configure
+        }
     }
 
     fn grab(&mut self, surface: PopupSurface, seat: wl_seat::WlSeat, serial: Serial) {
--- a/anvil/src/shell/xdg.rs
+++ b/anvil/src/shell/xdg.rs
@@ -264,49 +264,43 @@ impl<BackendData: Backend> XdgShellHandler for AnvilState<BackendData> {
     }
 
     fn fullscreen_request(&mut self, surface: ToplevelSurface, mut wl_output: Option<wl_output::WlOutput>) {
-        if surface
-            .current_state()
-            .capabilities
-            .contains(xdg_toplevel::WmCapabilities::Fullscreen)
-        {
-            // NOTE: This is only one part of the solution. We can set the
-            // location and configure size here, but the surface should be rendered fullscreen
-            // independently from its buffer size
-            let wl_surface = surface.wl_surface();
+        // NOTE: This is only one part of the solution. We can set the
+        // location and configure size here, but the surface should be rendered fullscreen
+        // independently from its buffer size
+        let wl_surface = surface.wl_surface();
 
-            let output_geometry = fullscreen_output_geometry(wl_surface, wl_output.as_ref(), &mut self.space);
+        let output_geometry = fullscreen_output_geometry(wl_surface, wl_output.as_ref(), &mut self.space);
 
-            if let Some(geometry) = output_geometry {
-                let output = wl_output
-                    .as_ref()
-                    .and_then(Output::from_resource)
-                    .unwrap_or_else(|| self.space.outputs().next().unwrap().clone());
-                let client = match self.display_handle.get_client(wl_surface.id()) {
-                    Ok(client) => client,
-                    Err(_) => return,
-                };
-                for output in output.client_outputs(&client) {
-                    wl_output = Some(output);
-                }
-                let window = self
-                    .space
-                    .elements()
-                    .find(|window| window.wl_surface().map(|s| &*s == wl_surface).unwrap_or(false))
-                    .unwrap();
-
-                surface.with_pending_state(|state| {
-                    state.states.set(xdg_toplevel::State::Fullscreen);
-                    state.size = Some(geometry.size);
-                    state.fullscreen_output = wl_output;
-                });
-                output.user_data().insert_if_missing(FullscreenSurface::default);
-                output
-                    .user_data()
-                    .get::<FullscreenSurface>()
-                    .unwrap()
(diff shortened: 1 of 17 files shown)


YaLTeR:
Another question: should this last_acked cached state be reset somewhere upon unmapping? I guess a pre-commit hook would be needed here? (The configured bool in the role should also be reset at that point. Is that bool needed, or is `ToplevelCachedState::last_acked.is_some()` the more correct version of it?)

anvil/src/shell/xdg.rs
```diff
@@ -221,14 +221,12 @@ impl<BackendData: Backend> XdgShellHandler for AnvilState<BackendData> {
...
-                        .unwrap()
-                        .current
-                        .states
-                        .contains(xdg_toplevel::State::Resizing)
+                        .cached_state
+                        .get::<ToplevelCachedState>()
+                        .current()
+                        .last_acked
+                        .as_ref()
+                        .is_some_and(|(_, state)| state.states.contains(xdg_toplevel::State::Resizing))
```
Drakulix: So could we not have a "current_state" function (like we had), that returns an `Option` and is the last_acked state? (Possibly named `last_acked_state`.)

Because this pattern seems to be recurring a lot.

Drakulix (commented):
> The WmCapabilities checks in Anvil were removed because they are serverside and don't change; not sure what the intention was with those.
> 
> Also, added forced configures in unmaximize() and unfullscreen() because they too are required by the protocol.

If that is not too cumbersome, I'd love to split off fixes like this into separate commits.

> One kinda weird part is current_server_state() that I had to change to owned return, tbh I would change it to return an Option<&State>, but that would involve changing a bunch of callers. I can do that though if that's better.

It seems like we already cloned it after requesting it in a bunch of places anyway? Seems fine to me honestly.

> Another question: should this last_acked cached state be reset somewhere upon unmapping? I guess a pre-commit hook would be needed here?

We had this logic introduced not too (in the grand scheme of things) recently. I guess that logic simply needs to be extended a bit? https://github.com/Smithay/smithay/pull/1571

anvil/src/shell/xdg.rs
YaLTeR: Added `with_committed_state()`, though this particular place couldn't use it because it doesn't have a `ToplevelSurface` handy. I used it in the niri commit, replacing one of the handrolled utils functions.

YaLTeR:
> If that is not too cumbersome, I'd love to split off fixes like this into separate commits.

Done.

YaLTeR:
Pushed a commit that also tracks the configured state through MultiCache last_acked.

Also:
- Removes dependency on renderer abstractions for tracking unmaps.
- Removes manual reset_initial_configure() because it's now always automatic.
- Makes the check for map before initial configure automatic: fixes https://github.com/Smithay/smithay/issues/1717.
- Adds missing role attributes reset on unmap (it only happened on role destroy previously).

Currently testing it on my machine with niri, so far so good.

YaLTeR:
I removed `popup_handle` because it wasn't actually set. It would be good to have `popup_handle` (and a corresponding `toplevel_handle` in the toplevel) for the error emitting path, but I'm not sure off the bat if setting them would cause a refcycle, so I opted for a loop over known toplevels for now.

YaLTeR:
Would be good to also move popup parent into MultiCache and get rid of `committed` (and along the way fix https://github.com/Smithay/smithay/issues/1761) but that sounds like a much deeper dive than I'm prepared to do at the moment :p

YaLTeR:
Uhm so apparently wlcs doesn't bother to ack the initial configure?
```
$ WAYLAND_DEBUG=client ./wlcs/wlcs ./target/debug/libwlcs_anvil.so --gtest_filter="FrameSubmission.post_one_frame_at_a_time"
(...)
[4128002.287] {Default Queue}  -> wl_compositor#4.create_surface(new id wl_surface#11)
[4128002.325] {Default Queue}  -> xdg_wm_base#7.get_xdg_surface(new id xdg_surface#9, wl_surface#11)
[4128002.352] {Default Queue}  -> xdg_surface#9.get_toplevel(new id xdg_toplevel#13)
[4128002.360] {Default Queue}  -> wl_surface#11.commit()
[4128002.378] {Default Queue}  -> wl_shm#6.create_pool(new id wl_shm_pool#14, fd 16, 160000)
[4128002.384] {Default Queue}  -> wl_shm_pool#14.create_buffer(new id wl_buffer#15, 0, 200, 200, 800, 0)
[4128002.388] {Default Queue}  -> wl_shm_pool#14.destroy()
[4128002.394] {Default Queue}  -> wl_surface#11.attach(wl_buffer#15, 0, 0)
[4128002.400] {Default Queue}  -> wl_surface#11.frame(new id wl_callback#16)
[4128002.405] {Default Queue}  -> wl_surface#11.commit()
error in client communication (pid 105972)
[4128002.829] {Display Queue} wl_display#1.delete_id(14)
[4128002.835] {Display Queue} wl_display#1.error(xdg_surface#9, 3, "must ack the initial configure before attaching buffer")
```

> After creating a role-specific object and setting it up (e.g. by sending the title, app ID, size constraints, parent, etc), the client must perform an initial commit without any buffer attached. The compositor will reply with initial wl_surface state such as wl_surface.preferred_buffer_scale followed by an xdg_surface.configure event. **The client must acknowledge it** and is then allowed to attach a buffer to map the surface.

(and the server debug for completeness)
```
[4251864.961] wl_compositor#4.create_surface(new id wl_surface#11)
[4251864.986] xdg_wm_base#7.get_xdg_surface(new id xdg_surface#9, wl_surface#11)
[4251865.020] xdg_surface#9.get_toplevel(new id xdg_toplevel#13)
[4251865.100] wl_surface#11.commit()
[4251865.187]  -> xdg_toplevel#13.configure_bounds(800, 600)
[4251865.201]  -> xdg_toplevel#13.wm_capabilities(array[16])
[4251865.211]  -> xdg_toplevel#13.configure(0, 0, array[4])
[4251865.220]  -> xdg_surface#9.configure(1)
[4251865.229] wl_shm#6.create_pool(new id wl_shm_pool#14, fd 15, 160000)
[4251865.251] wl_shm_pool#14.create_buffer(new id wl_buffer#15, 0, 200, 200, 800, 0)
[4251865.265] wl_shm_pool#14.destroy()
[4251865.272]  -> wl_display#1.delete_id(14)
[4251865.276] wl_surface#11.attach(wl_buffer#15, 0, 0)
[4251865.284] wl_surface#11.frame(new id wl_callback#16)
[4251865.291] wl_surface#11.commit()
[4251865.302]  -> wl_display#1.error(xdg_surface#9, 3, "must ack the initial configure before attaching buffer")
```

YaLTeR:
- Did the same refactor in layer-shell and session-lock.
- Exported more of the usual types from session-lock.
- Fixed an edge case (introduced in the refactor) with clients attaching a null buffer for the initial configure—Waybar does this.
- Added `with_cached_state()` to all three (layer-shell had a similar function so might as well).

Tested on:
- xdg: everything running on my desktop
- layer-shell: waybar, sfwbar, lxqt-panel, xfce4-panel, quickshell, xfdesktop, pcmanfm-qt --desktop
- session-lock: swaylock, hyprlock, gtklock

> Would be good to also move popup parent into MultiCache and get rid of committed

Added a FIXME comment mentioning this.

YaLTeR:
I think newer WLCS fixes the ack_configure issue? At least from a quick glance at its repo. How is it usually updated?

Drakulix (approved):
Looks all good by me. However I would like another review before merging this. @ids1024 @cmeissl one of you maybe up for the task?

YaLTeR:
Rebased, wrote changelog.

cmeissl (approved):
Looks good to me, thanks for working on this!

YaLTeR:
Added one commit on top that resolves the popup FIXME and delays the `grab()` handler until pre-commit. This fixes https://github.com/Smithay/smithay/issues/1761.

I kept the parent and the new field in the role attributes because they can only be set during the initial commit and cannot change afterward (not sure this is currently enforced in Smithay), so I figured there's no need to put them in the cached state.

Tested with: `pcmanfm-qt --desktop`, `lxqt-panel`, Quickshell with `Menu { popupType: Popup.Native`, a bunch of random apps and tools with popups that did not have issues before

Drakulix (approved):

PR #1819, opened 2025-09-09
wayland/layer: Exclusive edge fix, protocol version 5, `arrange` order

ids1024:
Iterate over surfaces that have an exclusive zone first, then other surfaces. This allows a surface that doesn't have an exclusive zone to be moved or resized for a later surface that does have an exclusive zone. As is probably desired.

This does help with a test client, but not with the client I was trying to fix; so I need to do a bit more testing here...

### DIFF
--- a/src/desktop/wayland/layer.rs
+++ b/src/desktop/wayland/layer.rs
@@ -272,7 +272,15 @@ impl LayerMap {
             );
             trace!("Arranging layers into {:?}", output_rect.size);
 
-            for layer in self.layers.iter() {
+            let exclusive_surfaces = self
+                .layers
+                .iter()
+                .filter(|l| matches!(l.effective_exclusive_zone(), ExclusiveZone::Exclusive(_)));
+            let non_exclusive_surfaces = self
+                .layers
+                .iter()
+                .filter(|l| !matches!(l.effective_exclusive_zone(), ExclusiveZone::Exclusive(_)));
+            for layer in exclusive_surfaces.chain(non_exclusive_surfaces) {
                 let surface = layer.wl_surface();
 
                 with_surface_tree_downward(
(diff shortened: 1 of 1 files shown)


src/desktop/wayland/layer.rs
```diff
@@ -272,7 +272,27 @@ impl LayerMap {
...
-            for layer in self.layers.iter() {
+            let exclusive_surfaces = self.layers.iter().filter(|l| {
+                matches!(
+                    with_states(l.wl_surface(), |states| states
+                        .cached_state
+                        .get::<LayerSurfaceCachedState>()
+                        .current()
+                        .exclusive_zone),
+                    ExclusiveZone::Exclusive(_)
+                )
```
YaLTeR: I'd extract this part into a `let is_exclusive = |layer| ...`, also perhaps this needs further anchor checks as described and done below?

> If the surface is:
> 
>    - not anchored
>    - anchored to only two perpendicular edges (a corner),
>    - anchored to only two parallel edges or anchored to all edges,
> 
> The exclusive value should be treated the same as ExclusiveZone::Neutral.

ids1024: I was thinking the exclusive zone check could be moved to just a (probably private) method of `struct LayerSurface`.

Good catch on those other cases that are equivalent to `ExclusiveZone::Neutral`. Though I guess iteration order doesn't matter for "neutral" surfaces, since they shouldn't be affected by exclusive zones, or cause effects on other surfaces.

Edit: Or rather than would be `ExclusiveZone::DontCare`, not `Neutral`; so this does matter.

ids1024:
Looks like the handling of the edge for exclusive zones there isn't quite right already.

```rust 
                      x if x.contains(Anchor::TOP) && x.contains(Anchor::BOTTOM) => {
                            zone.size.w -= amount;
                            if x.contains(Anchor::LEFT) {
                                zone.loc.x += amount;
                                zone.loc.x += data.margin.left;
                                zone.size.w -= data.margin.left;
                            }
                            if x.contains(Anchor::RIGHT) {
                                zone.size.w -= data.margin.right
                            }
                        }
 ````
 
 This case appears *before* the `x == Anchor::all()` case; so presumably this branch would be hit, and the `Anchor::Left` case inside it instead.

We could have a function to determine the exclusive edge, if any. And also add support for version 5 of the protocol supporting an explicit exclusive edge: https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/123

**Comment by ids1024** [comment]
The `arrange` fix can be opening a layer surface without an exclusive zone, then opening one with an exclusive zone. For instance on COMSIC, `sleep 5 && cosmic-panel`, then opening the workspaces view.

Before this change, the workspaces view didn't resize to make space for the panel. After this change, it does.

File: src/desktop/wayland/layer.rs [hunk 2]
```diff
@@ -473,6 +448,21 @@ impl LayerMap {
...
+        0 | 2 | 4 => None,
+        1 => Some(anchor),
+        3 => Some(match anchor.complement() {
+            Anchor::TOP => Anchor::BOTTOM,
+            Anchor::BOTTOM => Anchor::TOP,
+            Anchor::LEFT => Anchor::RIGHT,
+            Anchor::RIGHT => Anchor::LEFT,
+            _ => unreachable!(),
+        }),
+        _ => unreachable!(),
```
YaLTeR, line 475: Is this value guaranteed to be valid, or parsed from the client unchecked? If the former, perhaps add a comment here?

ids1024: Looking at this part again, I guess we want another check here, if the client tried to set multiple edges for the exclusive edge: https://github.com/Smithay/smithay/pull/1849

PR #1824, opened 2025-09-15
Fix PopupTree duplicate registration, implement NotTheTopmostPopup protocol error, some cleanup

Moulberry:
1. Various fixes for PopupTree being registered multiple times
2. Parent xdg_popup being destroyed while child is still alive now throws a NotTheTopmostPopup protocol error
3. Some other minor cleanup (retain_mut is happily now stable)


Moulberry:
Converted to a draft. Need to adjust the cleanup code to handle IME popups

Issue #1729, opened 2025-05-11
Popups parented to the same toplevel are not stacked properly


Supreeeme:
In both Niri and Anvil, if you have two popups that are parented to the same toplevel, the second one is stacked under the first one, but it should be stacked above. This works as expected on labwc (wlroots) and KDE.

According to the xdg_popup spec:
> A newly created xdg_popup will be stacked on top of all previously created xdg_popup surfaces associated with the same xdg_toplevel.

This is easily reproducible via [xwayland-satellite](https://github.com/Supreeeme/xwayland-satellite) and `gtk4-widget-factory`. See https://github.com/Supreeeme/xwayland-satellite/issues/153.

PR #1825, opened 2025-09-15
popup/manager: Fix toplevel popup iteration order

YaLTeR:
Protocol says:

> A newly created xdg_popup will be stacked on top of all previously created
> xdg_popup surfaces associated with the same xdg_toplevel.

New toplevel popups are pushed to the end of the vector, so iteration should go from the end to the start, since this function returns popups in topmost-first order.

Fixes various cases of tooltip opening after a popup menu, and ending up below. Also fixes multi-level popups in xwayland-satellite. Also fixes IME popups appearing below normal popups if there's an input field in a popup.

Issue-wise, fixes https://github.com/Smithay/smithay/issues/1729, https://github.com/Supreeeme/xwayland-satellite/issues/153, https://github.com/YaLTeR/niri/issues/125.

This fix is still not 100% sufficient I guess since it doesn't handle ordering across different popup trees, but I'm yet to encounter a situation where that would be needed.

### DIFF
--- a/src/desktop/wayland/popup/manager.rs
+++ b/src/desktop/wayland/popup/manager.rs
@@ -290,6 +290,10 @@ impl PopupTree {
             .lock()
             .unwrap()
             .iter()
+            // A newly created xdg_popup is stacked on top of all previously created xdg_popups. We
+            // push new ones to the end of the vector, so we should iterate in reverse, from newest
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Thx!

PR #1826, opened 2025-09-18
xdg_dialog: expose 'dialog' status

Moulberry:
Exposes the "dialog" status of a toplevel.

From xdg-dialog-v1:

A xdg_dialog_v1 object is an ancillary object tied to a xdg_toplevel. **Its purpose is hinting the compositor that the toplevel is a "dialog" (e.g. a temporary window) relative to another toplevel** (see xdg_toplevel.set_parent).

### DIFF
--- a/src/wayland/shell/xdg/dialog.rs
+++ b/src/wayland/shell/xdg/dialog.rs
@@ -10,7 +10,7 @@
 //! # use smithay::utils::Serial;
 //! # use smithay::wayland::shell::xdg::{XdgShellState, PopupSurface, PositionerState};
 //! # use smithay::reexports::wayland_server::protocol::{wl_seat, wl_surface};
-//! use smithay::wayland::shell::xdg::dialog::{XdgDialogState, XdgDialogHandler};
+//! use smithay::wayland::shell::xdg::dialog::{ToplevelDialogHint, XdgDialogState, XdgDialogHandler};
 //!
 //! # struct State { dialog_state: XdgDialogState, seat_state: SeatState<Self> }
 //! # let mut display = wayland_server::Display::<State>::new().unwrap();


src/wayland/shell/xdg/dialog.rs
```diff
@@ -243,27 +247,24 @@ impl<D: XdgDialogHandler> Dispatch<XdgDialogV1, ToplevelSurface, D> for XdgDialo
...
-            false
-        } else {
-            role.modal = true;
-            true
-        }
-    })
+/// Dialog hint assigned to toplevels by the xdg-dialog-v1 protocol
+#[derive(Default, Debug, Copy, Clone, PartialEq, Eq)]
+pub enum ToplevelDialogHint {
+    /// The toplevel hasn't been marked as a dialog by xdg-dialog-v1
```
YaLTeR:
```suggestion
    /// The toplevel is not associated with an xdg-dialog-v1 object.
```
Maybe like this? Makes it clearer that the presence of xdg-dialog-v1 is what matters. The original comment makes it sound like xdg-dialog-v1 could be present but not "marking" the surface as a dialog.

src/wayland/shell/xdg/dialog.rs
```diff
@@ -243,27 +247,24 @@ impl<D: XdgDialogHandler> Dispatch<XdgDialogV1, ToplevelSurface, D> for XdgDialo
...
+    #[default]
+    Unknown,
+    /// The toplevel is a "dialog" (e.g. a temporary window) relative
+    /// to its parent (see xdg_toplevel.set_parent)
```
YaLTeR:
```suggestion
    /// to its parent (see xdg_toplevel.set_parent).
```

Moulberry: [f31024e](https://github.com/Smithay/smithay/pull/1826/commits/f31024eb7cf9cbc9f9b44de9344c711ea2013283)

src/wayland/shell/xdg/dialog.rs
Moulberry: [f31024e](https://github.com/Smithay/smithay/pull/1826/commits/f31024eb7cf9cbc9f9b44de9344c711ea2013283)

Moulberry:
If merging, I'd recommend doing a squash merge to keep the commits clean, especially since there were resolved merge conflicts

Drakulix (approved):
LGTM, but CI disagrees

Moulberry:
Ah, the merge with upstream put the imports in the wrong order

PR #1829, opened 2025-09-19
Anvil: Fix wrong DnD offset

Moulberry:
The DnD should start at the hotspot aka the pointer location. The (0, 0) coordinate already corresponds to the pointer location. By offseting by -hotspot, the dnd is initiated at the top-left of the cursor's surface instead of the hotspot.

You can verify the behaviour is incorrect by running weston-dnd:
Without the patch, when initiating the dnd the element gets offset to the left.
With the patch, the element does not get offset.

(P.S. Even with this patch, there is still a very tiny offset when doing the DnD for me. I believe this is due to conversions between integer and floating point locations using round() instead of floor(). In my experience, floor() produces more correct results than round(). But that's an entirely separate and it might just be my setup.)

### DIFF
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -30,7 +30,7 @@ use smithay::{
     },
     input::{
         keyboard::{Keysym, LedState, XkbConfig},
-        pointer::{CursorImageStatus, CursorImageSurfaceData, PointerHandle},
+        pointer::{CursorImageStatus, PointerHandle},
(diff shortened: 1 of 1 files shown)


YaLTeR:
Tested the same change in niri, seems to work fine.

Drakulix (approved):
We had a few overall in regards to the hotspot code, so this might just be a left-over artifact. Thanks for cleaning it up!

PR #1830, opened 2025-09-19
fix clippy lints

cmeissl:


src/backend/renderer/element/mod.rs
```diff
@@ -164,6 +164,7 @@ impl Id {
     ///
     /// Note: The id will be re-used once all instances of this [`Id`]
     /// are dropped.
+    #[allow(clippy::new_without_default)]
```
cmeissl, line 167: I disagree with clippy here as we have some relevant comment on the new function.

Drakulix: works for me

Drakulix (approved):
thanks!

PR #1833, opened 2025-10-03
xwm: Use `hidden` in method names instead of `minimized`/`suspended`

ids1024:
https://github.com/Smithay/smithay/pull/1482 renamed `set_minimized()` to `set_suspended()`, but didn't rename `is_minimized()`.

But the name may not be ideal either.

https://specifications.freedesktop.org/wm-spec/latest/ar01s05.html says

> _NET_WM_STATE_HIDDEN should be set by the Window Manager to indicate that a window would not be visible on the screen if its desktop/viewport were active and its coordinates were within the screen bounds. The canonical example is that minimized windows should be in the _NET_WM_STATE_HIDDEN state. Pagers and similar applications should use _NET_WM_STATE_HIDDEN instead of WM_STATE to decide whether to display a window in miniature representations of the windows on a desktop.

While
https://wayland.app/protocols/xdg-shell#xdg_toplevel:enum:state:entry:suspended says:

> The surface is currently not ordinarily being repainted; for example because its content is occluded by another window, or its outputs are switched off due to screen locking.

So `_NET_WM_STATE_HIDDEN` shouldn't be set merely because a window is occluded, or on a different workspace? Both definitions could perhaps be a bit more precise.

It seems to me `_NET_WM_STATE_HIDDEN` is broader than "minimized", but more specific than the xdg-shell "suspended" state. So just call it "hidden" instead of using some other name.


Drakulix (approved):
Agreed. Thanks

PR #1835, opened 2025-10-05
Version bump 0.8.0

PolyMeilex:
I don’t have a proper changelog this time around (aside from what others already contributed), but it’s been three months since 0.7, so let's bump to 0.8.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 0.8.0
+
 ### Breaking changes
 
 `crate::wayland::selection::data_device::start_dnd` was removed in favor of exposing the
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "smithay"
-version = "0.7.0"
+version = "0.8.0"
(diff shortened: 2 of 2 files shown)


Drakulix:
I would prefer to fix CI first and then merge some already approved PRs.

hammerlink:
Hello! Do you have an idea when the updated version might be available?

jhupcs:
What's happening with this?

Drakulix:
> What's happening with this?

What is happening is that nobody working on and with smithay has/had the time and/or motivation/reason to bring forward a new release.

I am not trying to be impolite, but your comment doesn't really do anything, but add to the noise. Additionally you can clearly see from commits/other issues and PRs, that this repo is active.

So the issue tracker is probably not the best place to ask for this, as it represents work, not a Q/A forum. The matrix chat or the discussions (as you noticed yourself in https://github.com/Smithay/smithay/discussions/2077), is a better place for this.

But this isn't run like a company project with a fixed release schedule, so you might not get a clear answer, because there isn't one.

auyer:
This is currently using winit version = "0.31.0-beta.2".
Would it be a good idea to wait for 0.31 to come out?

PR #1842, opened 2025-10-25
egl: import linear dma buffers with the linear modifier

### LINKED ISSUE #1840: error importing dmabuf: Error::DeviceMissing for V4L2 M2M hw decoded video
(Moving from https://github.com/YaLTeR/niri/issues/2571)

When running niri directly on drm/kms output, using GStreamer's Wayland output that directly presents dmabufs from the hardware video decoder to the compositor (`gst-play-1.0 --videosink=waylandsink`) does not work, and `niri::backend::tty: error importing dmabuf: Error::DeviceMissing` appears in niri's console output. Gst says

```
ERROR buffer buffer: 0xffff8b5a8a30, pts 0:00:00.080000000, dts 99:99:99.999999999, dur 0:00:00.040000000, size 3133440, offset none, offset_end none, flags 0x40 cannot have a wl_buffer for file:///home/val/Downloads/Tractor_500kbps_x265.mp4
```

The hardware codec on non-AMD/Intel platforms is handled by a different driver than the GPU, so it's an import from a different device that's *not a DRM device at all* (the UAPI provided by those drivers is typically V4L2).

Seems like `MultiRenderer`'s `import_egl_buffer` somehow can't handle this..

cmeissl:
Use `EGL_EXT_image_dma_buf_import_modifiers` for importing linear buffers when available.
Omitting the modifiers results in the driver falling back to `invalid`, which might not represent
linear and can result in a modifier mismatch.

In addition also extend logging in multigpu and cache common extensions.

in case we do not know the dmabuf node we will
try all available nodes during import.

in case no node is found that can:
- allows cross device imports
- successfully imports the dmabuf the DeviceMissing error is returned hiding any
error during import attempts.

to make debugging such cases easier log
the errors returned from the renderer.

fixes #1840

### DIFF
--- a/src/backend/egl/display.rs
+++ b/src/backend/egl/display.rs
@@ -754,6 +754,7 @@ impl EGLDisplay {
             ]));
         };
 
+        let format = dmabuf.format();
         let mut out: Vec<c_int> = Vec::with_capacity(50);
 
         out.extend([
@@ -762,7 +763,7 @@ impl EGLDisplay {
             ffi::egl::HEIGHT as i32,
             dmabuf.height() as i32,
             ffi::egl::LINUX_DRM_FOURCC_EXT as i32,
-            dmabuf.format().code as u32 as i32,
+            format.code as u32 as i32,
         ]);
 
         let names = [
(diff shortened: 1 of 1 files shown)


valpackett:
This does, in fact, fix #1840 \o/

Yeah so it probably was the modifier thing. Video decoder buffers are compressed indeed

cmeissl:
> This does, in fact, fix #1840 \o/

awesome :)

> Yeah so it probably was the modifier thing. Video decoder buffers are compressed indeed

The buffer mentioned in #1840 uses NV12 with linear memory, there is no additional compression/tiling.
What was wrong is the handling of linear, which was handled by omitting the modifier. This will result in
the driver assuming `DRM_FORMAT_MOD_INVALID`. I did not check what the freedreno driver will actually
do in this case, but it either always fails with invalid modifiers or defaults to some driver specific tiling format (AFBC?).
On some drivers invalid does indeed correspond to linear, but this is not guaranteed. I have also seen drivers
just ignoring the format/modifier and using some other mechanism to detect the format (but only in case the resource
was actually created on the gpu, which is not the case with v4l2 memory).

Thanks for reporting and testing!

PR #1846, opened 2025-10-27
desktop/popup: fix focus not updating after popup dismiss

hammerlink:
`PopupGrab::has_ended` previously considered both active and dismissed grabs as keeping the grab alive. This caused the grab to remain "active" even after all popups were dismissed with `PopupGrab::ungrab(All)`.

According to the documentation, a grab has ended once all popups associated with it have been dismissed or destroyed. Including dismissed grabs in the active check prevented focus from updating after popup dismissal.

This change makes `has_ended` return true once all *active* grabs are gone, aligning the behavior with its documented intent.

This resolves a focus transfer issue observed in [pop-os/cosmic-panel#483](https://github.com/pop-os/cosmic-panel/issues/483)
(**tiling mode**, when an application is opened or focused via a popup it did not receive focus)

[src/shell/focus/mod.rs](https://github.com/pop-os/cosmic-comp/blob/master/src/shell/focus/mod.rs)
`popup_grab.ungrab(PopupUngrabStrategy::All);` is called in `append_focus_stack`.
`keyboard.set_focus(state, target.cloned(), serial);` is called in `update_focus_state` afterwards. At this point the focus should have been passed along. But because the popup was still considered active it was blocked.

### DIFF
--- a/src/desktop/wayland/popup/grab.rs
+++ b/src/desktop/wayland/popup/grab.rs
@@ -64,8 +64,12 @@ struct PopupGrabInternal {
 }
 
 impl PopupGrabInternal {
-    fn active(&self) -> bool {
+    fn has_any_grabs(&self) -> bool {
         !self.active_grabs.is_empty() || !self.dismissed_grabs.is_empty()
+    }
+
+    fn has_active_grabs(&self) -> bool {
+        !self.active_grabs.is_empty()
     }
 
     fn current_grab(&self) -> Option<&WlSurface> {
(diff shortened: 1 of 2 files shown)


src/desktop/wayland/popup/grab.rs
```diff
@@ -322,7 +322,14 @@ where
...
     pub fn has_ended(&self) -> bool {
-        !self.root.alive() || !self.toplevel_grab.active()
+        !self.root.alive()
+            || self
+                .toplevel_grab
+                .internal
+                .lock()
+                .unwrap()
+                .active_grabs
+                .is_empty()
```
Drakulix: Have you checked where else `toplevel_grab.active()` is used and if we should instead adjusts it's logic instead?

In general I'd prefer to not explicitly check `active_grabs`, but split the `active` method of `PopupGrabInner`, if necessary, into two variants, one considering dismissed grabs as well as active ones and one that doesn't.

hammerlink: It is also used in the `PopupManager` in the `cleanup` where I assume it needs the dismissed grabs as well before they can be removed. At least as long as they are alive.
I will split up the `active` method to cover both cases.

PR #1848, opened 2025-10-30
Wayland<->X11 Drag&Drop transfers

Drakulix:
WIP branch implementing drag&drop between wayland surfaces and xwayland surfaces.

Currently only dragging from wayland -> xwayland works, dnd exclusively between wayland surfaces or xwayland surfaces should not be impacted.

This goes away with the `ServerDndGrab`, which would've potentially allowed dropping from Xwayland->Wayland but not vise-versa. Instead it moves part of the DnD-code and most importantly the `DnDGrab` into `crate::input::dnd` and abstracts over both `Source`s and potential targets with the new `DndFocus`-trait. It needs to be implemented for `Seat::PointerFocus` if one wants to use the `DnDGrab`.

In theory the new `Source`-trait should allow to fully replace the `ServerDndGrab` (if that is even used by any downstream compositor), but I haven't tried that so far, as I have yet to write an implementation of that for Xwayland and not only the wayland-types. (Which is also why this is still a draft.)

However the `DnDFocus`-trait seems to work out as intended and I would love some feedback on the api design.

This still needs more validation, that it doesn't break clients using older wl_data_device versions and works with clients using older XDND versions than 5.

### DIFF
--- a/src/input/dnd/grab.rs
+++ b/src/input/dnd/grab.rs
@@ -277,11 +277,9 @@ where
 
     fn drop<'a>(&'a mut self, data: &mut D, into_target: impl Fn(&'a F) -> DndTarget<'a, D>) {
         // the user dropped, proceed to the drop
-        let validated = self.offer_data.as_ref().is_none_or(|data| data.validated());
+        let validated = self.offer_data.as_ref().is_some_and(|data| data.validated());
         if let Some(ref focus) = self.current_focus {
-            if validated {
-                focus.drop(data, self.offer_data.as_mut(), &self.seat);
-            }
+            focus.drop(data, self.offer_data.as_mut(), &self.seat);
         }
 
         if let Some(ref offer_data) = self.offer_data {


Drakulix:
Rebased onto https://github.com/Smithay/smithay/pull/1854

Drakulix:
Okay, all bugs should be squashed now, including:
- Quick mouse movements causing transfers from old windows to be aborted not handled correctly, which caused the source to hang
- Directly crossing between an X11 and a Wayland window not waiting for mime-types, causing the wayland window to reject the drop
- Fixing data-device v2 clients
- Fixing copy/paste from wayland->x11 being broken (and likely was broken for anything but text for a long time)

ids1024:
When I try to test drag and drop on `anvil` with `WAYLAND_DISPLAY= nautilus`, I see issues with the drag icon being offset fairly far away from the cursor. Though these appears to be an existing issue in `anvil` not affected by he changes here.

Drakulix:
> When I try to test drag and drop on `anvil` with `WAYLAND_DISPLAY= nautilus`, I see issues with the drag icon being offset fairly far away from the cursor. Though these appears to be an existing issue in `anvil` not affected by he changes here. 

Yes, anvil is not handling moving X11 windows correctly. This PR makes no attempt to fix that, but behaves correctly, if you don't move the window after spawn.

src/wayland/selection/data_device/source.rs
```diff
@@ -126,13 +118,61 @@ impl IsAlive for WlDataSource {
...
+    }
+
+    fn drop_performed(&self) {
+        if self.version() >= wl_data_source::EVT_DND_DROP_PERFORMED_SINCE {
+            self.dnd_drop_performed();
+        }
+    }
+
+    fn cancel(&self) {
+        if self.version() >= 3 {
```
cmeissl: Is there a const available for this since check?

src/wayland/selection/data_device/source.rs
```diff
@@ -126,13 +118,61 @@ impl IsAlive for WlDataSource {
...
+            self.cancelled();
+        }
+    }
+
+    fn finished(&self) {
+        if self.version() >= 3 {
```
cmeissl: same here

src/wayland/selection/data_device/mod.rs
```diff
@@ -113,84 +109,458 @@ use super::{
...
+
+    fn drop<S: Source>(&self, _data: &mut D, offer: Option<&mut WlOfferData<S>>, seat: &Seat<D>) {
+        let seat_data = seat
+            .user_data()
+            .get::<RefCell<SeatData<D::SelectionUserData>>>()
+            .unwrap()
+            .borrow();
+        for device in seat_data.known_data_devices() {
+            if let Some(ref offer) = offer {
+                if device.version() >= 3 && offer.state.lock().unwrap().chosen_action.is_empty() {
```
cmeissl: Is there a const available for this check?

src/wayland/selection/data_device/source.rs
Drakulix: Upon further investigation, I think it is actually correct to drop this check.

See: https://wayland.app/protocols/wayland#wl_data_source:event:cancelled

> For objects of version 2 or older, wl_data_source.cancelled will only be emitted if the data source was replaced by another data source.

What "replaced" means exactly in terms of a dnd drop isn't very clear, because any replacement requires a new dnd grab.

However from looking at toolkits, it doesn't seem that clients used to monitor their implicit grab to invalidate/free up the source themselves, even when binding version 2 or less.

The old implementation send `cancelled` always when the drop was "dropped". Without the check this piece of code does that again. The only other case, where smithay called "cancelled" was on "destroy", but only on version 3 or higher and if and only if the drop wasn't completed successfully before. I don't think emitting the event in that case for versions 2 and lower is wrong, given the drop didn't finish successfully, which is equal to the "cancelled" call on drop.

If we need to differentiate these cases, we need to add a new "CancelReason" or something as an argument to the `cancel`-call, but I don't think that is necessary given these are the only two cases.

src/wayland/selection/data_device/source.rs
Drakulix: Fixed

src/wayland/selection/data_device/mod.rs
Drakulix: Kinda. `chosen_action` can only be set, if `SetActions` is available. I thus used that request's constant.

YaLTeR:
Trying to update niri to this PR.

Some things I've hit:
1. In `WaylandDndGrabHandler::dnd_requested()`, when setting the pointer grab, should it be `Focus::Clear` or `Focus::Keep`? I'm getting very conflicting evidence 😄
    - the removed lines in the diff (they differ between client and server DnD grab)
    - the new anvil and smallvil examples (they differ) and the cosmic-comp PR
2. In `DndGrabHandler::dropped()`, the `target` is a generic `T: DndFocus<Self>`. In niri, this can only ever be a WlSurface, but how do I get that WlSurface out? Perhaps some downcasting support is needed?

Drakulix:
>     1. In `WaylandDndGrabHandler::dnd_requested()`, when setting the pointer grab, should it be `Focus::Clear` or `Focus::Keep`? I'm getting very conflicting evidence 😄
>        
>        * the removed lines in the diff (they differ between client and server DnD grab)
>        * the new anvil and smallvil examples (they differ) and the cosmic-comp PR

I don't think it makes any difference, as the `DnDGrab` will continue to send events to the target that started the implicit grab, that was validated for the request. As far as I understand the wayland protocol that is expected and for Xwayland it is necessary for correct behaviour. So the enum-value doesn't really do anything. I changed everything to `Focus::Keep` though for consistency.

>     2. In `DndGrabHandler::dropped()`, the `target` is a generic `T: DndFocus<Self>`. In niri, this can only ever be a WlSurface, but how do I get that WlSurface out? Perhaps some downcasting support is needed?

I changed the function a bit to get rid of the generic argument altogether. :)

src/input/dnd/grab.rs
```diff
@@ -123,27 +123,70 @@ where
     }
 }
 
+/// Enum over DndFocus candidates receiving a drop from `DnDGrab`
+#[derive(Debug, Clone)]
+pub enum DndTarget<'a, D: SeatHandler> {
```
YaLTeR: This derive Debug/Clone is automatically (incorrectly) gated on `D: Debug/Clone`. Need manual impl of Debug and Clone with the correct gate on `PointerFocus` and `TouchFocus`.

src/input/dnd/grab.rs
```diff
@@ -123,27 +123,70 @@ where
...
+        match self {
+            DndTarget::Pointer(_) => panic!("unwrap_pointer on pointer-based dnd grab target"),
+            DndTarget::Touch(t) => t,
+        }
+    }
+}
+
+impl<'a, F, D: SeatHandler<PointerFocus = F, TouchFocus = F>> DndTarget<'a, D> {
+    /// Returns the contained value consuming `self`.
+    pub fn unwrap(self) -> &'a F {
```
YaLTeR: `unwrap` is a scary name for a method that doesn't panic 😄 Maybe something like `into_inner`

src/wayland/selection/data_device/mod.rs
```diff
@@ -113,84 +109,460 @@ use super::{
...
-/// Events that are generated during client initiated drag'n'drop
+/// Handler for wayland client initiated drag'n'drop
 #[allow(unused_variables)]
-pub trait ClientDndGrabHandler: SeatHandler + Sized {
-    /// A client started a drag'n'drop as response to a user pointer action
+pub trait WaylandDndGrabHandler: SeatHandler + Sized {
+    /// A client requested a drag'n'drop as response to a user pointer action
     ///
     /// * `source` - The data source provided by the client.
     ///   If it is `None`, this means the DnD is restricted to surfaces of the
```
YaLTeR: This comment needs updating

src/input/dnd/grab.rs
Drakulix: Good call. Added manual-impl for `Debug` and dropped `Clone` since it actually doesn't make any sense with the borrowed value.

src/input/dnd/grab.rs
Drakulix: Yeah good suggestion. I initially wanted to implement `Into`, but apparently every impl for a reference overlaps with some stdlib implementation.

src/wayland/selection/data_device/mod.rs
Drakulix: Done.

Drakulix:
> After a DnD of a piece of text inside gtk4-demo, the window seemingly stops receiving pointer events until I unfocus and focus it again.
> [..]
> Actually, apparently I'm checking for pointer focus = None as a condition for triggering the niri hot corner. Previously, during DnD pointer focus was None, but with this PR, pointer focus isn't None during DnD, regardless if I set `Focus::Clear` or `Focus::Keep` when creating the `DnDGrab`. I can add a separate check for this so it's not a big deal, just a behavior change to be noted.

Looking at this in more detail, it seems that GTK is unsetting pointer focus on `wl_data_device.leave` and does not expect `wl_pointer` events during the drag.

Which leaves us a bit at a conundrum, because ideally it seems like we don't want to set the focus. However Xwayland needs the focus for the grabbing application to generate the necessary dnd-events. Sooo... I added a hack in https://github.com/Smithay/smithay/pull/1848/commits/eb301e5df33dab94d7ab1e02b72a9e1c025e6360 to force a re-enter on unset. I am not happy with this and suggestions are welcome.

> With the addition of new generic parameters, it becomes somewhat unwieldy, especially that S: Source parameter that can be several different things (at least WlSurface and WlDataSource).

Yeah, I am not sure, what would be a good way of solving this. Even with Xwayland not being deployed, what previously was the `ServerDndGrab` is now also handled by the `DndGrab` (and I think this unification is actually a good thing), meaning we cannot know, if downstream might add it's own Sources and Targets.

Any suggestions on how to improve this are welcome.

src/input/dnd/grab.rs
```diff
@@ -0,0 +1,551 @@
...
+            self.pointer_start_data.as_ref().unwrap().focus.clone(),
+            event,
+        );
+    }
+
+    fn button(&mut self, data: &mut D, handle: &mut PointerInnerHandle<'_, D>, event: &ButtonEvent) {
+        handle.button(data, event);
+
+        if handle.current_pressed().is_empty() {
+            handle.motion(
```
cmeissl: I also have to idea how to resolve this, but when we want to keep this as a workaround we should document the reason here.

Drakulix: I'll make sure to add a comment before merging.

Drakulix: Theoretically we could do this:
- Require `WaylandFocus` to be implemented for `SeatHandler::PointerFocus` (or `::TouchFocus` respectively)
- Get the underlying `wl_surface` and test the `client_data` for Xwayland.
- Either..
  - ..consistently never set `focus` for anything, but Xwayland.
  - ..don't set `focus` for non-xwayland WlSurfaces (which might be nicer for custom shells?)

That would however break this new behaviour, @YaLTeR seemed to prefer:
> Previously, during DnD pointer focus was None, but with this PR, pointer focus isn't None during DnD, regardless if I set Focus::Clear or Focus::Keep when creating the DnDGrab. 

In any case, we would not have consistency here and sometimes have a focus and sometimes none. It would however be more consistent with how other compositors behave, though I have yet to notice any issues arising from this hack.

YaLTeR: I checked `WAYLAND_DEBUG=1` and (as expected) on this PR, the client receives wl_pointer motion events even during the DnD, which is, well, unexpected. Would be unfortunate having to do technically incorrect Wayland behavior just for (the possibility of) built-in Xwayland (which niri doesn't even compile in).

I tested gtk4-demo again, and while the original issue seems resolved, I found more weirdness:
- When DnD-ing text over GtkNotebook tabs, on this PR the cursor reverts from text back to normal; on main Smithay and on Weston it doesn't.
- In Identity, on this PR, DnD-ing a video results in `ERROR g_object_unref: assertion 'G_IS_OBJECT (object)' failed` printed to the Identity output. This doesn't happen without this PR. I'm not sure what this is but it doesn't sound good.

This is to say, who knows what else will break from this.

> Theoretically we could do this:

I think this would be a good idea, to keep the Wayland part working properly.

> That would however break this new behaviour, @YaLTeR seemed to prefer:

No, I don't have preference here. I just meant that I can probably work around it. If anything, my preference is to preserve the main focus none behavior just because that means I don't have to change my code.

Drakulix: Pushed a [commit](https://github.com/Smithay/smithay/pull/1848/commits/266039e502edeb1916b3e5cd7ed40ee87a8dd6b2) that restores the previous behaviour and only special cases Xwayland. Since conditional where-clauses or trait-bounds aren't a thing, this was slightly annoying, but I feel like I found the shortest possible solution.

@YaLTeR Could you re-check the regressions you noticed?

YaLTeR: Thanks, those regressions are all fixed. Found a new one though: dragging a tab out of firefox into somewhere outside causes the DnD to get weirdly stuck inside firefox and not complete. Also, I couldn't find a client which uses a null dnd source.

Drakulix: weston-simple-dnd is able to use a null surface and I tested it (and pushed an upstream fix in the process).

I'll look at firefox though, thanks :)

Drakulix: Firefox is fixed. The bug was caused by a missing cancel event due to the offer being invalidated (correctly), when the window left the original firefox surface: https://github.com/Smithay/smithay/pull/1848/commits/f2609b8b9a9fbea5b26ad7dbbbbc2c6d9548df3e

(Also found and fixed a bug, where a wayland drop could be cancelled due to passing over a non-accepting X11 window.)

I also made sure to re-test:
- `weston-dnd --self-only`
- `cosmic-files` + `GDK_BACKEND=x11 nautilus` (in both directions)
- `GDK_BACKEND=x11 nautilus` between each other
- various wayland gtk apps (gtk4-demo)

I can't find any regressions any more.

PR #1849, opened 2025-10-30
wayland/layer: Add check that `exclusive_edge` has only one edge

ids1024:
We already have checks that the exclusive edge `Anchor` is valid, and contained in the anchors of the surface, but not a check that it doesn't contain multiple edges.

### DIFF
--- a/src/desktop/wayland/layer.rs
+++ b/src/desktop/wayland/layer.rs
@@ -383,6 +383,7 @@ impl LayerMap {
                         Some(Anchor::RIGHT) => {
                             zone.size.w -= amount + Saturating(data.margin.right);
                         }
+                        // Exclusive edge is always exactly one edge
                         Some(_) => unreachable!(),
                         None => {}
                     }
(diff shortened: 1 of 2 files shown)


Drakulix (approved):

YaLTeR:
Seems to break a bunch of LXQt components: https://github.com/YaLTeR/niri/issues/2758

stefonarch:
I suppose it's affecting only apps relying on `layer-shell-qt` - `xfdesktop` is not affected but `krunner` is.
`crystal-dock` is affected too.
Adding the "autohide" feature in `lxqt-panel` to the broken list.

ids1024:
Hm. I'll need to test layer-shell-qt, but maybe we also need to accept no bits set, and treat that as `None`, falling back to the implied exclusive edge. I assume more than one bit shouldn't be set, and my check for that presumably is otherwise correct.

The protocol spec isn't explicit on this. I probably should have looked at the other implementations to be sure about it.

ids1024:
Yeah, testing on a distro with a new enough version of lxqt/etc. I see this. It looks let it sets the exclusive edge to 0 when hiding the panel, along with setting the exclusive zone to 0.

ids1024:
https://github.com/Smithay/smithay/pull/1853 should match how KWin validates this.

PR #1851, opened 2025-11-02
tablet: Send pressure(0) before pen up events

Unk-now:
Send a pressure(0) event before sending the pen up event in both tip_up() and proximity_out() handlers. This improves compatibility with applications like Wine and MyPaint that expect the pressure to be explicitly set to zero before a pen up event.

While the tablet-v2 protocol does not strictly require this, major compositors like GNOME and Sway send this event for compatibility. Without it, some applications continue drawing strokes even after the pen is lifted.

Fixes compatibility issues reported in:
- https://bugs.winehq.org/show_bug.cgi?id=58463
- https://github.com/YaLTeR/niri/issues/2722
- https://github.com/YaLTeR/niri/issues/2576

**Note:** This patch has been tested manually on Arch Linux with WINE-git and Clip Studio Paint.
Automated CI cannot reproduce the exact environment (WINE + CSP), but the patch is low-risk
and only affects the pressure event sent before pen up.

### DIFF
--- a/src/wayland/tablet_manager/tablet_tool.rs
+++ b/src/wayland/tablet_manager/tablet_tool.rs
@@ -118,7 +118,7 @@ impl TabletTool {
             }
         }
 
-        self.is_down = false; 
+        self.is_down = false;
     }
 
     fn motion(


src/wayland/tablet_manager/tablet_tool.rs
```diff
@@ -110,13 +111,14 @@ impl TabletTool {
...
             {
                 if self.is_down {
+                    wl_tool.pressure(0);
                     wl_tool.up();
                     wl_tool.frame(time);
                 }
             }
         }
 
-        self.is_down = false;
```
PolyMeilex, line 119: Redundant trailing space

Unk-now:
@PolyMeilex,

Thank you for the review.

I fully agree with your point regarding the protocol strictness—it is likely an implementation bug on the client side that expects pressure to be explicitly zeroed out before the up event.

However, given that several major applications (Wine, MyPaint, CSP) rely on this behavior, and major compositors like GNOME and Sway are already sending this event for compatibility, I believe this PR is a necessary pragmatic step to ensure a smooth and functional user experience for smithay users.

I have addressed the "Redundant trailing space" issue and pushed the fix.

PolyMeilex (approved):
Yeah you are right, it's necessary evil I suppose.

PolyMeilex:
Thanks!

PR #1852, opened 2025-11-03
feat: support multiple input methods

rano-oss:


ids1024:
How is having multiple input method engines supposed to work? Only one input method can be handling input for a single seat/keyboard at a given moment, surely?

rano-oss:
I am working on a POC, and yes only one input method can handle input for a single seat/keyboard at a given moment. That is why there is an active input method field, anyway the thing is support for keyboard switching with global or per client would have to be added to the compositor so why not just get them all in one fell swoop. I will cook and post a video if I can make it work ;)

PR #1758, opened 2025-06-14
feat: add ext-background-effect-v1

edfloreshz:
This PR adds support for `ext-background-effect-v1`.

PR #1860, opened 2025-11-16
Introduce ext-background-effect-v1 protocol

PolyMeilex:
Based on the work from: #1758

### DIFF
--- a/src/wayland/background_effect/mod.rs
+++ b/src/wayland/background_effect/mod.rs
@@ -69,6 +69,11 @@ pub trait ExtBackgroundEffectHandler:
         let _ = wl_surface;
         let _ = region;
     }
+
+    /// Called when blur unset becomes pending, and awaits surface commit.
+    fn unset_blur_region(&mut self, wl_surface: WlSurface) {
+        let _ = wl_surface;
+    }
 }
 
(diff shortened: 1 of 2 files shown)


PolyMeilex:
Rebased, everything still seems to work fine.

src/wayland/background_effect/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+{
+    /// Getter for background-effect capabilities of the compositor
+    fn capabilities(&self) -> Capability {
+        // For now only blur is supported, so it's safe to assume it's supported by default
+        Capability::Blur
+    }
+
+    /// Called when blur becomes pending, and awaits surface commit.
+    /// Blur region is stored in wl_surface [BackgroundEffectSurfaceCachedState]
+    fn set_blur_region(&mut self, wl_surface: WlSurface, region: compositor::RegionAttributes) {
```
Drakulix, line 68: It looks like this function is never called? There is also no counter-part for when a client simply removes the blur region.

PolyMeilex: Oh yeah, my bad, not sure how I managed to forget about this. Fixed

PR #1861, opened 2025-11-16
Rewirte delegate macros to make them more readable

PolyMeilex:
The title is pretty self-explanatory. This is an attempt to make those macros a bit less dreadful.


Drakulix (approved):
This is very neat!

PR #1863, opened 2025-11-25
xwm: Add support for `_NET_ACTIVE_WINDOW` client message

ids1024:
I guess X11 clients use this essentially the same way `xdg-activation-v1` is used by native Wayland clients. A new method of `XwmHandler` is invoked so the compositor can handle the message.

In addition to the `window`, the client takes three 32-bit values in `data`. This ignores the "source indication" for now, since we also seem to ignore that for the `_NET_WM_MOVERESIZE` and `NET_WM_STATE` client messages. If we want to take that into account, we can define an enum and add it to all applicable trait methods.


Drakulix (approved):
LGTM

PR #1869, opened 2025-12-05
 backend/winit: Use winit `0.31.0-beta.2`

ids1024:
Uses hacky solution to handle change to need `ActiveEventLoop` to create a window. May be okay? But a different solution, and API change, will be needed if we want the backend to work on Android.

This also still needs changes to handle touch events.


ids1024:
I've adjusted to a somewhat better way to handle window creation. Though it's still a bit awkward.

In the future, we could maybe overhaul the API here to fit better the structure winit wants, possibly work on platforms like Android, support multiple windows, etc. But this keeps the current API and seems like it doesn't regress any functionality.

I don't think there's any reason we need to wait for a 0.31 stable release. I'll be using this to test the wayland-rs update.

Drakulix (approved):
LGTM

PR #1877, opened 2025-12-12
Dmabuf-feedback state adjustments

Drakulix:
First commit makes sure to not hold the `with_states`-lock, while calling into `DmabufHandler::new_surface_feedback`. This is a little ugly (and makes the factory a bit useless), but makes accidental deadlocks much less likely, while giving downstream more options to evaluate the surface-state and figure out the appropriate dmabuf-feedback. Alternatively we could pass `SurfaceState` to the handler as well, but that might make accessing the data harder for downstream. (E.g. the abstractions in cosmic-comp generally expect to be able to excess the `SurfaceState`.)

The second commit removes the code that clears the last feedback from the surface-state, once the last instance is cleared. This means any surface that had feedback at any point in it's lifetime will keep a reference, but it also means, that drivers that re-create the whole state don't cause us to loose that state potentially trapping them in a swapchain re-create loop.

I would argue it is pretty likely, that a surface that once used dmabufs will do so in the future and this is worth the potential very-small memory usage for an used feedback-struct.

### DIFF
--- a/src/wayland/dmabuf/dispatch.rs
+++ b/src/wayland/dmabuf/dispatch.rs
@@ -1,4 +1,4 @@
-use std::sync::{atomic::AtomicBool, Mutex};
+use std::sync::{atomic::AtomicBool, Arc, Mutex};
 
 use wayland_protocols::wp::linux_dmabuf::zv1::server::{
     zwp_linux_buffer_params_v1, zwp_linux_dmabuf_feedback_v1, zwp_linux_dmabuf_v1,
@@ -10,7 +10,7 @@ use wayland_server::{
 
 use crate::{
     backend::allocator::dmabuf::{Dmabuf, Plane, MAX_PLANES},
-    wayland::{buffer::BufferHandler, compositor},
+    wayland::{buffer::BufferHandler, compositor, dmabuf::SurfaceDmabufFeedbackStateInner},
 };
 
 use super::{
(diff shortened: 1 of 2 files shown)


src/wayland/dmabuf/dispatch.rs
```diff
@@ -109,18 +109,25 @@ where
...
                         .insert_if_missing_threadsafe(SurfaceDmabufFeedbackState::default);
                     let feedback_state = states.data_map.get::<SurfaceDmabufFeedbackState>().unwrap();
-                    feedback_state.add_instance(&feedback, || {
-                        state
-                            .new_surface_feedback(&surface, &DmabufGlobal { id: data.id })
-                            .unwrap_or_else(|| {
-                                data.default_feedback.as_ref().unwrap().lock().unwrap().clone()
-                            })
-                    })
+                    feedback_state.add_instance(&feedback, move || new_feedback)
```
cmeissl: So as you already wrote this makes the factory rather useless. 
But what bothers me more is that the initialization gets a lot more complicated than necessary.
`is_empty` in combination with the now useless factory makes following the code harder.

If we just want to keep the feedback alive forever we might also just remove the `Option` and `Default` impl all together and use `states.data_map.get::<SurfaceDmabufFeedbackState>().is_none()` and only in this case run the init logic and use `state.data_map.insert()` to fill the initial feedback.

So instead of

```
/// Feedback state for a surface
#[derive(Debug, Clone, Default)]
pub struct SurfaceDmabufFeedbackState {
    inner: Arc<Mutex<Option<SurfaceDmabufFeedbackStateInner>>>,
}
```

to use

```
/// Feedback state for a surface
#[derive(Debug, Clone)]
pub struct SurfaceDmabufFeedbackState {
    inner: Arc<Mutex<SurfaceDmabufFeedbackStateInner>>,
}
```

In this case the split into two commits would make no sense, so a single commit would be fine imo.

The docs from `new_surface_feedback` also needs to be updated in any case as this still includes the note about when this is called wrongly:

```
    /// This function allows to override the default [`DmabufFeedback`] for a surface
    ///
    /// Note: This will only be called if there is no alive surface feedback for the surface.
    /// Normally this will be the first time a surface requests feedback, but can also occur
    /// if all instances have been destroyed and a new surface request is sent by the client.
    ///
    /// Returning `None` will use the default [`DmabufFeedback`] from the global
    fn new_surface_feedback(
        &mut self,
        _surface: &WlSurface,
        _global: &DmabufGlobal,
    ) -> Option<DmabufFeedback> {
        None
    }
```

cmeissl (changes requested):
> I would argue it is pretty likely, that a surface that once used dmabufs will do so in the future and this is worth the potential very-small memory usage for an used feedback-struct.

Yes, and iirc most compositors will initialize a single scan-out feedback anyway and we allow to clone it.

src/wayland/dmabuf/dispatch.rs
Drakulix: yeah, good call. updated.

src/wayland/dmabuf/mod.rs
```diff
@@ -512,55 +512,32 @@ impl SurfaceDmabufFeedbackState {
...
         let mut guard = self.inner.lock().unwrap();
-        if let Some(inner) = guard.as_mut() {
-            if &inner.feedback == feedback {
-                return;
-            }
-
-            for instance in inner.known_instances.iter().filter_map(|i| i.upgrade().ok()) {
-                feedback.send(&instance);
-            }
+        if &guard.feedback == feedback {
```
cmeissl, line 514: The doc comment here is now also not completely correct.

Drakulix: fixed

PR #1878, opened 2025-12-17
xwm: Fixes for handling loss of connection X server

ids1024:


src/xwayland/xwm/surface.rs
```diff
@@ -1094,7 +1094,7 @@ impl X11Relatable for WlSurface {
 impl IsAlive for X11Surface {
     #[inline]
     fn alive(&self) -> bool {
-        self.state.lock().unwrap().alive
+        self.alive()
```
Drakulix: Isn't this recursive? I'd rather express it explicitly:

```suggestion
        X11Surface::alive(self)
```

ids1024: If this actually recursed, we'd get a `unconditional_recursion` warning from the compiler.

We can be more explicit like that, but I'm not sure it helps. It seems this compiles (with an `unconditional_recursion` error):

```rust
trait Foo {
    fn foo(&self);
}

struct Bar {}

impl Foo for Bar {
    fn foo(&self) {
        Bar::foo(self);
    }
}
```

So if the trait method was the only one by that name for the type, this would still recurse. But would warn.

Drakulix: If multiple implementations exist the fully-qualified syntax will always prefer the non-trait implementation though and I also like not having the mental overload of deducing, which implementation this is supposed to call.

ids1024: The point is that I think the fully-qualified and unqualified syntax will both favor the inherent method if it exits, while using the trait method if that's the only one that exists, and it's in scope.

So as I understand, any impression that it's more explicit is illusory.

Drakulix: Even if it doesn't make any difference on what the compiler does, the mental load argument still applies.

ids1024: Okay. I've changed it.

Drakulix (approved):

PR #1879, opened 2025-12-18
feat: improve primary gpu selection

yuxqiu:
This PR improves Smithay's primary GPU selection on systems where the `boot_vga` attribute is missing (for example, Asahi Linux). Previously, `udev::primary_gpu` would fall back to the `all_gpus` path and select a card that, on some setups, has no associated render node and is therefore unusable.

This change avoids that behavior and prefers GPUs that can actually be used. This PR was originally proposed in niri ([https://github.com/YaLTeR/niri/pull/3054](https://github.com/YaLTeR/niri/pull/3054)); @YaLTeR suggested it might also benefit Smithay.

### DIFF
--- a/src/backend/udev.rs
+++ b/src/backend/udev.rs
@@ -37,6 +37,8 @@
 //! See also `anvil/src/udev.rs` for pure hardware backed example of a compositor utilizing this
 //! backend.
 
+#[cfg(feature = "backend_drm")]
+use drm::node::{DrmNode, NodeType};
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
This seems like a good improvement, thanks!

yuxqiu:
@Drakulix I suddenly realized this might break something as I prolly should not use `from_syspath`? I push a new pr #1880 that should fix this.

PR #1885, opened 2025-12-21
wayland/text_input: fix enter/leave sending

### LINKED ISSUE #1884: [text-input-v3] Wayland compositors send `enter`/`leave` events only to one alive instance of `zwp_text_input_v3` even though there're several available
(Hello! I'm mostly forwarding issues discovered when investigating a user report with JetBrains IDEs under Niri: https://youtrack.jetbrains.com/issue/JBR-9767.)

If a client application obtains more than 1 instance of `zwp_text_input_v3`, e.g. Niri will be sending `zwp_text_input_v3::enter`/`leave` events to only 1 of them, although it's against [the specification](https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:event:enter):

>If client has created multiple text input objects, compositor must send this event to all of them.

It prevents JetBrains IDEs from enabling the proper instance on time, thus causing [JBR-9767](https://youtrack.jetbrains.com/issue/JBR-9767).

The issue is observable in the following screen recording (provided by an affected user) starting from 00:22: from the logs on the right side you can notice that the application has obtained a new `zwp_text_input_v3` instance `#66`, which then doesn't receive any `enter`/`leave` events, although `zwp_text_input_v3#35` does:

https://github.com/user-attachments/assets/c99b9ba0-7cfb-4586-9b5a-3389d8858b64

The logs from the right side of the screen: [niri-idea_wayland-debug-logs.txt](https://github.com/user-attachments/files/24244103/niri-idea_wayland-debug-logs.txt)

### Steps to reproduce

1. Install a 2025.3 release of any JetBrains IDE (it's important to use 2025.3 or higher, because earlier releases don't provide text-input-v3 support), e.g. the free IntelliJ IDEA Community Edition (https://www.jetbrains.com/idea/download/?section=linux).
2. Press Ctrl+Shift+A -> start typing "Edit Custom VM Options..." and choose this option when it appears.
3. In the opened dialog add the following line: `-Dawt.toolkit.name=WLToolkit`.
4. Restart the IDE.
5. Open a project that's managed by Git.
6. Try to use the input method in the editor (just to make sure it works).
7. Press Alt+0 to show the Git commit tool window, make sure its View Mode is set to Window:
Image

8. Try to use input method in a text field there.
9.
Expected behavior: you're able to type using the input method.
Actual behavior: you're not able to do so.

### System Information

* niri version: 25.11

* Distro: Arch Linux 6.17.9
* fcitx5: 5.1.16

* GPU: unknown (unlikely matters)

* CPU: unknown (unlikely matters)

kchibisov:
The enter/leave events were not sent for all the text_input objects created by the client, while it should have been.

Fixes #1884.

### DIFF
--- a/src/wayland/text_input/text_input_handle.rs
+++ b/src/wayland/text_input/text_input_handle.rs
@@ -31,7 +31,6 @@ impl TextInput {
                 let instance_id = text_input.instance.id();
                 if instance_id.same_client_as(&surface.id()) {
                     f(&text_input.instance, surface, text_input.serial);
-                    break;
                 }
             }
         };
@@ -142,7 +141,7 @@ impl TextInputHandle {
         });
     }
 
-    /// Access the text-input instance for the currently focused surface.
+    /// Access the text-input instances for the currently focused surface.
     pub fn with_focused_text_input<F>(&self, mut f: F)
     where
         F: FnMut(&ZwpTextInputV3, &WlSurface),


kchibisov:
I know, since I wrote the bug in the first place :- )

rano-oss (approved):
Looks sensible to me

PR #1889, opened 2025-12-27
Tracy GPU profiling v2

YaLTeR:
Supersedes https://github.com/Smithay/smithay/pull/1134, cc @cmeissl. Depends on https://github.com/nagisa/rust_tracy_client/pull/153.

Rebased things and reimplemented the timestamp query tracking. It now more or less matches what I do in https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/3391 and should be correct.

This is still WIP:
- ~~still tied to tracy-client; my plan is to try abstracting it away, so the downstream compositor can plug in Tracy on its own (the API will be Tracy shaped though)~~
- ~~someone needs to verify that I rebased stuff correctly (that the GPU spans make sense)~~
- ~~is there a better place to call `collect()`? During my testing I already hit a case where I needed to add `collect()` to `cleanup_texture_cache()` since otherwise no-damage frames build up timestamps that don't get collected. Can we avoid this situation somehow?~~
- ~~probably need to call `sync_gpu_time()` in more places, for example things like `import_shm_buffer` are sometimes called on their own, and without a sync close by, their GPU timestamps drift~~
- ~~need to figure out how to delete the timestamp queries~~
- ~~I set the max queries to 1024 matching Mutter; we can bump this if needed (and replace inline array with Vec in that case I suppose)~~
- The instrumentation needs to regularly call `collect()` (ideally before a batch of GPU work) and it needs to call `sync_gpu_time()` after a batch of GPU work. But ideally not in the middle of ongoing GPU work to avoid stalls. Currently I put these in `render()` and `blit()` -- the two places that create an EGL fence and flush GL at the end. And all GPU profiling spans are therefore also restricted to within `render()` and `blit()`, otherwise without `collect()` timestamps may build up and overflow, and without `sync_gpu_time()` they may drift. This misses some potentially interesting (?) places like buffer imports. Not sure what to do about this.

### DIFF
--- a/src/backend/renderer/gles/profiler.rs
+++ b/src/backend/renderer/gles/profiler.rs
@@ -62,7 +62,7 @@ pub struct GpuSpan {
 
 impl Drop for GpuSpan {
     fn drop(&mut self) {
-        assert!(!self.active, "GPU span must be properly exited");
+        debug_assert!(!self.active, "GPU span must be properly exited");
     }
 }
 
@@ -187,7 +187,7 @@ mod imp {
 
     impl Drop for QueryPool {
         fn drop(&mut self) {
-            assert!(self.pool.is_empty(), "QueryPool must be cleaned up before drop");
+            debug_assert!(self.pool.is_empty(), "QueryPool must be cleaned up before drop");
         }
     }
 


YaLTeR:
Oh yeah, obligatory screenshots. Winit with some offscreen rendering, then the final render:

Screenshot from 2025-12-27 13-42-01

TTY:

image

YaLTeR:
Hm, I notice that it likes to cause constant back and forth to some GL thread:

image

Possibly it's the `glFlush()` calls on `exit()`. Maybe need to bring it back from there to `finish_internal()`.

It's a similar problem to `sync_gpu_time()` I guess. We need to call `glFlush()` when we know there won't be GPU work immediately thereafter, and we don't want to call it if we know that there *will* be GPU work immediately.

YaLTeR:
I pushed a commit where I removed all profiler calls outside `render()` to avoid the `collect()` and `sync_gpu_time()` problem, also removed the `glFlush()` in exit. I think the GPU spans make much more sense this way? Here's an example when screencasting output with OBS:

image

YaLTeR:
I added `with_profiled_context()` and manual enter/exit span on `GlesFrame` to make it possible for compositors to profile their custom render elements.

Then, in the last commit, I took a stab at making the thing optional. There are two ways I imagine this can go:
1. Have a GpuProfiler trait with methods for enter span, exit span, upload GPU timestamp. Receive an instance of this type when creating `GlesRenderer`.
2. Depend on Tracy and have a bunch of cfgs to make it optional.

Option 1 sounds enticing, however there's one problem with it: I'm not sure it's possible to plug in tracy-client's `&'static` span locations this way. Or in general with Rust, at least until maybe we have `const fn` trait methods. So we'd have to use allocated spans which is a bit meh.

So here I tried Option 2. It's kinda wordy with all the cfgs, but I tried to make it manageable. My goals were:
- Make `tracy-client` dep optional. While it has its own enable/disable mechanism, it still links Tracy even when disabled, so it should be optional in Smithay.
- Even when the instrumentation is enabled, leave the actual `tracy-client` enablement flags to the compositor (since it may want to have it on-demand or something else).
- Keep the cfgs inside Smithay's profiler impl. It's just nice when you don't need to spam cfg in your compositor for profiling.

So, the profiler API is always present. If the new `tracy_gpu_profiling` feature is on, it calls into `tracy-client` and does the whole OpenGL timestamp querying. If the feature is off, it compiles into mostly empty stubs. The compositor can write GPU profiling instrumentation, then enable the features to make it active. Example here: https://github.com/YaLTeR/niri/commit/745f4a9eb4399d8968de62e640af6763c86c17e3

One new question: should the `GL_EXT_disjoint_timer_query` generation in build.rs also be disabled when the feature is disabled, or is it fine to leave that in?

YaLTeR:
And here's a screenshot showcasing multi-GPU rendering (niri always renders on the main GPU). Hopefully it looks right.

image

src/backend/renderer/gles/profiler.rs
```diff
@@ -0,0 +1,323 @@
...
+        fn drop(&mut self) {
+            assert!(self.pool.is_empty(), "QueryPool must be cleaned up before drop");
+        }
+    }
+
+    impl GpuProfiler {
+        pub fn new(gl: &ffi::Gles2, extensions: &[String]) -> Self {
+            let has_timer_query = extensions.iter().any(|ext| ext == "GL_EXT_disjoint_timer_query");
+            if !has_timer_query {
+                return Self { pool: None };
```
Drakulix, line 199: Should we maybe insert a (`Once`-guarded? maybe not, in case multiple drivers are in use.) `warn!`-message here? If the user has enabled gpu-profiling, but the driver/hardware doesn't support it, this might otherwise be surprising if they aren't aware?

build.rs
```diff
@@ -66,6 +66,7 @@ fn gl_generate() {
                 "GL_EXT_texture_format_BGRA8888",
                 "GL_EXT_unpack_subimage",
                 "GL_OES_EGL_sync",
+                "GL_EXT_disjoint_timer_query",
```
Drakulix, line 69: This can stay. While generating code for unused extensions slightly increases the compile times for non-profile builds as well, I don't think it does much in the grand scheme of things.

src/backend/renderer/gles/profiler.rs
YaLTeR: I guess considering that profiling is enabled, and that GlesRenderers aren't recreated often (?), a non-Once-gated warn!() should be fine

YaLTeR:
> Overall I like the api quite well. I am not sure however, if we want to force `tracy_gpu_profiling` to enable `renderer_gl`.
> 
> After all, once we gain a Vulkan renderer, that might be able to add profiling with tracy and a similar api?
> 
> Without `gl_renderer` the whole module shouldn't be compiled anyway, right?

Maybe we can postpone this change until then? Right now the thing is very tied to GL and I'm not sure how exactly it will look with a second renderer backend.

YaLTeR:
Alright, this should be done. Updated niri commit to the new `GlesFrame::with_gpu_span()` API: https://github.com/YaLTeR/niri/commit/e55807478890eb21251377b319cb7de8a079f452

Screenshot of the new GPU names commit on my laptop:

image

Please squash when merging if I don't.

YaLTeR:
My previous squash moved the Co-authored-by line to the middle of the message :facepalm:

src/backend/renderer/gles/profiler.rs
Drakulix: I mean cosmic-comp does re-create them, but our default builds won't enable profiling, so this is fine by me.

cmeissl:
> LGTM. @cmeissl Do you want to take a quick look at this as well?

Yes, will take a look tomorrow latest.

ids1024:
> The instrumentation needs to regularly call collect() (ideally before a batch of GPU work) and it needs to call sync_gpu_time() after a batch of GPU work. But ideally not in the middle of ongoing GPU work to avoid stalls. Currently I put these in render() and blit() -- the two places that create an EGL fence and flush GL at the end. And all GPU profiling spans are therefore also restricted to within render() and blit(), otherwise without collect() timestamps may build up and overflow, and without sync_gpu_time() they may drift. This misses some potentially interesting (?) places like buffer imports. Not sure what to do about this.

Yeah, doing this were we flush makes sense. Though it would be nice if Tracy could help us see what time is spent on buffer imports.

src/backend/renderer/gles/profiler.rs
```diff
@@ -0,0 +1,328 @@
...
+            let gpu_name = gpu_name.unwrap_or("GlesRenderer");
+
+            let client = tracy_client::Client::start();
+
+            let mut gpu_timestamp = 0;
+            unsafe { gl.GetInteger64v(ffi::TIMESTAMP_EXT, &mut gpu_timestamp) };
+
+            let context = client
+                .new_gpu_context(
+                    Some(gpu_name),
```
ids1024, line 212: `cosmic-comp` uses separate surface threads for each output being rendered, each with their own shared EGL context.

This seems to result in the contexts for all connectors of the same name being given the same name in `tracy`. (Why CPU activity is labeled with `profiling::register_thread!`).

YaLTeR: How would you suggest naming these instead?

ids1024: Although it looks like `tracy` does show the thread name as well when you hover over a GPU zone. So maybe that's okay.

YaLTeR: Yeah, hovering over GPU zones highlights the corresponding CPU zone

PR #1893, opened 2025-12-31
clamp the Size to 0 when constructing

### LINKED ISSUE #873: Crash in debug mode when receiving region with negative size
@kchibisov noticed that Catacomb crashes in debug mode and I managed to track it down to the `debug_assert` in `Size::from` which panics when a size is negative in debug mode.

The `WlRegion` `request` handler just does `Rectangle::from_loc_and_size`, which will call `into` on the `(width, height)` tuple. It should probably clamp the region rather than creating a bogus region when the size is negative?

m4rch3n1ng:
had to work around this semi-recently in https://github.com/m4rch3n1ng/mayland/commit/91b00c003d67ab132448466e2b44fd575f9f9472 (and a year ago in https://github.com/m4rch3n1ng/mayland/commit/9a37f29b6461d32cd4f6153ae4a22b7fe25a61c4) and wasn't sure if that was wanted, but as there seemingly is at least some demand (see #873), i thought why not just quickly make this (since it's not a lot of work) and wait for reviewer feedback so here we are.

this does make size creation slightly more expensive in release mode without debug assertions, but that should be a pretty negligible difference i assume.

similar in spirit to #1586, fixes #873.

### DIFF
--- a/src/utils/geometry.rs
+++ b/src/utils/geometry.rs
@@ -416,16 +416,7 @@ impl<N: Coordinate, Kind> Point<N, Kind> {
     #[inline]
     #[must_use = "this returns the result of the operation, without modifying the original"]
     pub fn to_size(self) -> Size<N, Kind> {
-        debug_assert!(
-            self.x.non_negative() && self.y.non_negative(),
-            "Attempting to create a `Size` of negative size: {:?}",
-            (self.x, self.y)
-        );
-        Size {
-            w: self.x,
-            h: self.y,
-            _kind: std::marker::PhantomData,
-        }
+        Size::new(self.x, self.y)
     }
 
     /// Convert this [`Point`] to a [`Size`] with the same coordinates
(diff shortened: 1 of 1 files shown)


YaLTeR:
Is it really a good idea to clamp automatically in Size? Sounds like potential to hide bugs.

emzywastaken:
> Is it really a good idea to clamp automatically in Size? Sounds like potential to hide bugs.

I do agree that this isn't a good idea, as my reason for engaging #873 was because some clients don't enforce min/max sizes themselves and as such you can resize them into a negative size, which I would argue is a bug in the compositor.

Kitty is an example of this, but only with client side decorations, as that changes the y position (relative to the toplevel surface) of the window geometry. Meaning when resizing, even if respecting the min/max size of the toplevel, its actually offset by the decorations, causing a panic in debug mode when constructing the new `Size`.

I can confirm this happens in smallvil, anvil and niri(when testing with `prefer-no-csd` unset) as well when running a debug build. This can be fixed by just taking account of decorations height and adding it to the minimum size.

*Not sure if I used the right terms here* :sweat_smile:

---

Video of niri crashing

https://github.com/user-attachments/assets/72ffb066-221c-4fec-ac87-167425c35ae7

YaLTeR:
My point is that it should be clamped at the call site, not in Size itself

emzywastaken:
> My point is that it should be clamped at the call site, not in Size itself

I meant to say i agree with it not being a good idea, sorry if i worded that poorly earlier

m4rch3n1ng:
ok fair enough, wasn't sure myself either but i thought fuck it, this is only like a 2 minute thing so i just did it.

i'm not entirely sure what the best way is to find all possible locations where this can happen (and it is then certainly no longer a 2 minute) so i'm just going to close this for now (and potentially open a separate pr some time in the future, no promises though).

sorry for kinda collectively wasting your time here.

Issue #1858, opened 2025-11-15
Overlay moves with its subsurfaces


nojvspine:
Basically this, in YaLTeR/niri#2787 I explained what happens, Niri and COSMIC have this issue, and I think other compositors that use Smithay also do

nojvspine:
Wanted to bump this issue cos I got no answer, is this being worked on?

nojvspine:
Got a recording of this bug.

https://github.com/user-attachments/assets/ceaa3b3d-483a-40a3-a037-4e4c2349a76d

PR #1897, opened 2026-01-06
layer: Only consider main surface for geometry

Drakulix:
Fixes https://github.com/Smithay/smithay/issues/1858.

The program in question (wl_shimeji) might position some negatively offset surfaces on it's main layer-shell surface.

This resulted in a bounding box with negative offset, which in turn shifted the render-location, which is obviously wrong.

I'd rather leave it to compositors to potentially clamp contents outside of the positioned layer-surface, as this obviously allows surfaces to overlap other layer-shell surfaces.

But in any way this is better than our current behaviour, which breaks the placement logic.

### DIFF
--- a/src/desktop/wayland/layer.rs
+++ b/src/desktop/wayland/layer.rs
@@ -1,4 +1,5 @@
 use crate::{
+    backend::renderer::utils::RendererSurfaceStateUserData,
     desktop::{utils::*, PopupManager},
     output::{Output, WeakOutput},
     utils::{user_data::UserDataMap, IsAlive, Logical, Point, Rectangle, Size},
@@ -146,10 +147,10 @@ impl LayerMap {
         if !self.layers.contains(layer) {
             return None;
         }
-        let mut bbox = layer.bbox();
+        let mut geo = layer.geometry();
         let state = layer_state(layer);
-        bbox.loc += state.location.unwrap_or_default();
-        Some(bbox)
+        geo.loc += state.location.unwrap_or_default();
+        Some(geo)
     }
 
     /// Returns a [`LayerSurface`] under a given point and on a given layer, if any.
@@ -576,6 +577,18 @@ impl LayerSurface {
     /// Returns the namespace of this surface
     pub fn namespace(&self) -> &str {
         &self.0.namespace
+    }
+
+    /// Returns the bounding box over this layer surface excluding its subsurfaces.
+    pub fn geometry(&self) -> Rectangle<i32, Logical> {
+        with_states(self.0.surface.wl_surface(), |states| {
+            let data = states.data_map.get::<RendererSurfaceStateUserData>();
+            data.and_then(|d| d.lock().unwrap().view()).map(|view| Rectangle {
+                loc: view.offset,
+                size: view.dst,
+            })
+        })
+        .unwrap_or_default()
     }
 
     /// Returns the bounding box over this layer surface and its subsurfaces.


src/desktop/wayland/layer.rs
```diff
@@ -578,6 +579,18 @@ impl LayerSurface {
         &self.0.namespace
     }
 
+    /// Returns the bounding box over this layer surface and its subsurfaces.
```
YaLTeR: Comment needs updating ig

Drakulix:
> I feel like this is just client working around the lack of set_geometry in layer-shell, and it works because of wrong handling in other compositors...

Yes absolutely and this has been brought up time and time again for both wlr-layer-shell and the proposed ext-layer-shell.

But this PR doesn't concern itself with that, it just fixes smithay's behaviour.

Since the bbox was used for the layer_geometry, e.g. the render-elements were computed with an already offset layer position and then applied the subsurface offset again, leading to the whole layer-surface shifting (without any re-arranging happening).

That cannot be intended in either case. The alternative would be to purposefully offset the render-location to counter act any negatively offset subsurfaces. But that would also break clients **and** need some additions to our arrange-method to consider the layer-surface size to include subsurfaces (which it doesn't currently).

So I'd rather leave the agency to offset the layer-surfaces manually or clip it to compositors.

src/desktop/wayland/layer.rs
Drakulix: thx, fixed.

Drakulix:
ping @cmeissl, @PolyMeilex, @chrisduerr, @kchibisov as YaLTeR wanted some more input and you spring to mind as potentially having an opinion on this. (No obligation and sorry for the ping, if you aren't interested at all.)

chrisduerr:
Is there a simpler client to test this? The one in the issue seems like something I'd rather avoid.

I'm curious if this is even an issue on catacomb, since I don't actually use Smithay's layer shell geometry calculation. So I guess I'd be in favor of just letting the compositors deal with it. :P

Drakulix:
> Is there a simpler client to test this? The one in the issue seems like something I'd rather avoid.
> 
> I'm curious if this is even an issue on catacomb, since I don't actually use Smithay's layer shell geometry calculation. So I guess I'd be in favor of just letting the compositors deal with it. :P

None I know of, but you basically just need something that sticks to either the top or left edge and then spawns a subsurface with a negative offset.

If you are affected depends on if you use bbox to determine the render-location before calling `render_element_from_surface_tree` on your LayerSurface. Then you also get the offset applied twice (once to the whole tree - which is wrong already - and once more to the subsurface), even without using LayerMap.

chrisduerr:
> If you are affected depends on if you use bbox to determine the render-location before calling render_element_from_surface_tree on your LayerSurface.

Well I don't call that either. :smile:

src/desktop/wayland/layer.rs
```diff
@@ -146,10 +147,10 @@ impl LayerMap {
         if !self.layers.contains(layer) {
             return None;
         }
-        let mut bbox = layer.bbox();
+        let mut geo = layer.geometry();
```
cmeissl, line 150: `layout_geometry` is also used to map input in anvil for example https://github.com/Smithay/smithay/blob/66c36c2820d3a23bbf3b634c23c963e417cecb9f/anvil/src/input_handler.rs#L292

From a quick look this should not break, but I am not 100 % sure.

Drakulix: I checked input and made sure they use the same offset.

PR #1901, opened 2026-01-08
wayland/shell/xdg: Reset toplevel state on surface reuse

mikairyuu:
Clients like Telegram (via Qt) reuse wl_surface objects. They destroy the xdg_toplevel role and immediately create a new one on the same wl_surface.

Currently, Smithay retains stale XdgToplevelSurfaceData and ToplevelCachedState attached to the wl_surface after the previous role is destroyed. When the client requests a new xdg_toplevel, Smithay assumes the surface is already in a specific state (e.g., configured), causing it to miss necessary setup events (like the initial configure).

This PR introduces reset_toplevel_surface_state to sanitize the surface state.
It is called in destroy (cleanup).
It is called in get_toplevel (initialization) with include_generic_state: false.
This ensures that creating a new toplevel role always results in a fresh state.

I verified this behavior against the XDG Shell protocol lifecycle. When a client destroys an xdg_toplevel object but keeps the wl_surface, the protocol implies the role-specific state is reset.
Currently, reference implementations (Weston) and wlroots-based compositors (Sway) treat a new get_toplevel request as generating a fresh logical window state.
This PR brings Smithay's behavior in line with these implementations.

Fixes https://github.com/pop-os/cosmic-comp/issues/1816

### DIFF
--- a/src/wayland/shell/xdg/handlers/surface/toplevel.rs
+++ b/src/wayland/shell/xdg/handlers/surface/toplevel.rs
@@ -184,22 +184,7 @@ where
             let toplevel = state.xdg_shell_state().known_toplevels.remove(index);
             let surface = toplevel.wl_surface().clone();
             XdgShellHandler::toplevel_destroyed(state, toplevel);
-            compositor::with_states(&surface, |states| {
-                *states
-                    .data_map
-                    .get::<XdgToplevelSurfaceData>()
-                    .unwrap()
-                    .lock()
-                    .unwrap() = Default::default();
-
-                let mut guard = states.cached_state.get::<SurfaceCachedState>();
-                *guard.pending() = Default::default();
-                *guard.current() = Default::default();
-
-                let mut guard = states.cached_state.get::<ToplevelCachedState>();
-                *guard.pending() = Default::default();
-                *guard.current() = Default::default();
-            })
+            reset_toplevel_surface_state(&surface, true);
         }
     }
 }
@@ -228,6 +213,29 @@ pub(super) fn make_toplevel_handle(
         wl_surface: data.wl_surface.clone(),
         shell_surface: resource.clone(),
     }
+}
+
+/// Resets the internal state of a surface to clear any residue from previous xdg_toplevel roles.
+///
+/// If `include_generic_state` is true, this will also clear generic wl_surface state
+/// (buffer scale, transform, etc). This should only be used when the role is being destroyed,
+/// not when it is being initialized/reused.
+pub fn reset_toplevel_surface_state(surface: &wl_surface::WlSurface, include_generic_state: bool) {
+    compositor::with_states(surface, |states| {
+        if let Some(role) = states.data_map.get::<XdgToplevelSurfaceData>() {
+            *role.lock().unwrap() = Default::default();
+        }
+
+        let mut toplevel_state = states.cached_state.get::<ToplevelCachedState>();
+        *toplevel_state.pending() = Default::default();
+        *toplevel_state.current() = Default::default();
+
+        if include_generic_state {
+            let mut surface_state = states.cached_state.get::<SurfaceCachedState>();
+            *surface_state.pending() = Default::default();
+            *surface_state.current() = Default::default();
+        }
+    });
 }
 
 pub fn get_parent(toplevel: &xdg_toplevel::XdgToplevel) -> Option<wl_surface::WlSurface> {
--- a/src/wayland/shell/xdg/handlers/surface.rs
+++ b/src/wayland/shell/xdg/handlers/surface.rs
@@ -31,7 +31,7 @@ use super::{
 
 mod toplevel;
 use toplevel::make_toplevel_handle;
-pub use toplevel::{get_parent, send_toplevel_configure};
+pub use toplevel::{get_parent, reset_toplevel_surface_state, send_toplevel_configure};
 
 mod popup;
 pub use popup::{make_popup_handle, send_popup_configure};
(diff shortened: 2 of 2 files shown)


src/wayland/shell/xdg/handlers/surface/toplevel.rs
```diff
@@ -184,22 +184,7 @@ where
             let toplevel = state.xdg_shell_state().known_toplevels.remove(index);
             let surface = toplevel.wl_surface().clone();
             XdgShellHandler::toplevel_destroyed(state, toplevel);
-            compositor::with_states(&surface, |states| {
```
YaLTeR, line 187: But doesn't this code already reset the xdg-toplevel state when the xdg-toplevel role is destroyed?

cmeissl: Yeah, was just about to comment the same.

cmeissl:
Can you take a look at the WAYLAND_DEBUG log and verify if the client destroys the toplevel?

smithay should already reset the state when the role is destroyed.

cmeissl:
see: https://github.com/Smithay/smithay/blob/1ca641f5bef0b09d36cec757bbb31b255c6b7b7f/src/wayland/shell/xdg/handlers/surface/toplevel.rs#L196

src/wayland/shell/xdg/handlers/surface/toplevel.rs
mikairyuu: The problematic Qt clients (I've encountered Telegram) in question keep commiting to the destroyed toplevel's surface (it sends one commit, probably for cleanup purposes), which means that on the next GetToplevel call from that surface it's already dirty and Smithay thinks it's already configured, so configure doesn't happen and window doesn't appear.

Reproduction is very easy - open media viewer on latest Telegram, then close, then open again.

YaLTeR: Media viewer works fine for me on Telegram 6.4.1 on niri

cmeissl: Can you upload an WAYLAND_DEBUG log containing the problematic situation? Does not need to be the whole log, but should at least show the destroy and re-create.

Having the cleanup logic in two places is not an option, if there is an issue with the existing cleanup we should fix that instead.

mikairyuu: [short_comp_log.log](https://github.com/user-attachments/files/24502077/short_comp_log.log)

Telegram/Qt continues to interact with the underlying wl_surface after destroying the xdg_toplevel and xdg_surface roles. specifically, it attaches a null buffer and commits.
(destroy)
`[2188567.179][rs] <- xdg_surface@55.destroy, ()`

(modification after being destroyed)
```
[2188567.204][rs] <- wl_surface@45.attach, (0, 0, 0)
[2188567.236][rs] <- wl_surface@45.commit, ()  <-- Surface state become dirty here
```

(re-create with the same dirty surface)
`[2189132.092][rs] <- xdg_wm_base@3.get_xdg_surface, (71, 45)`

Because that intervening commit happens after the destroyed handler has already run, the surface state is "dirty" again by the time get_toplevel is called.
This is why I believe get_toplevel is the only safe place to ensure the state is reset for the new role. I'm happy to remove the duplicate logic from destroyed if you prefer, relying solely on the sanitization during creation.

mikairyuu: > Media viewer works fine for me on Telegram 6.4.1 on niri

that is surely interesting... I'm reporting this on master cosmic-comp
I'd guess cosmic-comp checks against the state on wl_surface, as reported by smithay, while niri doesn't rely on smithay states

cmeissl: > Because that intervening commit happens after the destroyed handler has already run, the surface state is "dirty" again by the time get_toplevel is called. This is why I believe get_toplevel is the only safe place to ensure the state is reset for the new role. I'm happy to remove the duplicate logic from destroyed if you prefer, relying solely on the sanitization during creation.

No, the destroyed logic seems fine and we use the same logic in a lot of different places. Either we remove all of them or none, but I don't see anything wrong. Doing this during `get_toplevel` is the "big hammer" method and might prevent detecting other issues, so I am not in favor of this.

But before proceeding with any solution I want to understand what actually breaks.
initial_configure_sent is not set by smithay in the handlers automatically, it is only set on send_configure.
The compositor uses it to decide if it should send the initial configure when it receives a surface commit.
So atm I do not see what might go wrong in smithay, which makes it worth investigating this more.

I am not sure what you mean with "Surface state become dirty here". The commit should not touch initial configure as there is no toplevel alive on the surface. Maybe the toplevel is somehow kept around in cosmic and it falsely sends out a configure on the NULL commit? cc @Drakulix 

The only other technical reason that comes to my mind in smithay would be around transaction handling.

mikairyuu: @cmeissl 
You were absolutely right. I dug into cosmic-comp based on your hint, and the issue isn't a dirty state in Smithay, but rather how Cosmic handles window unmapping.
cosmic-comp currently treats all unmapped surfaces as "Pending Windows" (likely for X11 restoration), so when Telegram destroys the role but reuses the wl_surface, Cosmic resurrects the ghost window instead of creating a fresh one.
I have a clean fix for cosmic-comp that separates the "unmap" logic from the "save to pending" logic. I'm closing this PR to submit the fix there. Thanks for the pointer!

https://github.com/pop-os/cosmic-comp/pull/1975

cmeissl: nice, great find. I wonder if we could prevent such stuff in smithay, maybe a check in send_configure if the role object is actually alive.

PR #1902, opened 2026-01-10
feat(wayland): implement ext-image-capture-source-v1 and ext-image-copy-capture-v1 protocols

### LINKED ISSUE #1900: Implementing ext-image-capture-source-v1 and ext-image-copy-capture-v1
I'm developing a Smithay-based compositor that requires screen capture capabilities for client access. Per the protocol tracking in #781, neither `ext-image-capture-source-v1` nor `ext-image-copy-capture-v1` are currently implemented.

I have an immediate need for these protocols and am prepared to contribute an implementation. Before beginning work, I wanted to check:

1. Is there existing work on these protocols that I should be aware of?
2. Are there any design preferences or constraints from the maintainers' perspective?
3. Would a phased approach (ext-image-capture-source-v1 first, then ext-image-copy-capture-v1) be acceptable?

From my initial review, the existing infrastructure appears well-suited for this:
- `ExportMem` provides framebuffer readback
- `DmabufState` and `ForeignToplevelListState` offer relevant patterns
- The protocol definitions are available in wayland-protocols 0.32.4+

I can adapt my current development schedule to prioritize this if there's interest in an upstream contribution.

Greg Lamberson
Lamco Development
https://lamco.ai

glamberson:
Implements ext-image-capture-source-v1 and ext-image-copy-capture-v1 protocols for screen capture functionality.

Design uses callback-based extensibility with UserDataMap, allowing compositors to attach custom data to capture sources. This enables COSMIC-style workspace capture through custom globals without infecting the core API with generics.

Key changes:
- Full session/frame lifecycle with RAII cleanup
- Buffer validation (size, format, modifier) before frame callbacks
- Cursor session support for pointer capture
- Anvil integration demonstrating handler implementation

This implementation is adapted from cosmic-comp's screencopy module, originally authored by Victoria Brekenfeld (@Drakulix) and Ian Douglas Scott (@ids1024) at System76. Re-licensed to MIT with their approval per discussion in #1900.

Closes #1900

### DIFF
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -647,7 +647,7 @@ impl<BackendData: Backend> ImageCopyCaptureHandler for AnvilState<BackendData> {
                 smithay::reexports::wayland_server::protocol::wl_shm::Format::Argb8888,
                 smithay::reexports::wayland_server::protocol::wl_shm::Format::Xrgb8888,
             ],
-            #[cfg(any(feature = "udev", feature = "winit"))]
+            #[cfg(any(feature = "udev", feature = "winit", feature = "x11"))]
             dma: None,
         })
     }


src/wayland/image_capture_source/mod.rs
```diff
@@ -269,61 +226,36 @@ impl ImageCaptureSource {
...
+    /// Retrieve an [`ImageCaptureSource`] from an existing protocol resource.
     pub fn from_resource(resource: &ExtImageCaptureSourceV1) -> Option<Self> {
         resource
             .data::<ImageCaptureSourceData>()
             .map(|d| d.source.clone())
     }
 
     /// Add a protocol resource instance to this source.
-    fn add_instance(&self, resource: &ExtImageCaptureSourceV1) {
+    pub(crate) fn add_instance(&self, resource: &ExtImageCaptureSourceV1) {
```
Drakulix: Doesn't an externally implemented source also need access to this function?

HigherOrderLogic:
> For merging this still needs a more in-depth review to make sure this covers everything cosmic-comp and potentially niri need.

So is this PR in a "ready" state to implement the protocol in other compositor? I want to implement this protocol in Niri and maybe I'll give it a go with this PR.

glamberson:
> So is this PR in a "ready" state to implement the protocol in other compositor? I want to implement this protocol in Niri and maybe I'll give it a go with this PR.

Well, yes, presumably, although I"m testing still and can't probably get through all of it until morning.

ids1024:
So far this looks good. I'm working on updating `cosmic-comp`'s implementation to use this.

glamberson:
@ids1024 Great. I wanted to run this through testing but I just haven't had a minute to get to it. So I'm going to mark it as ready with the proviso it should undergo more testing.

ids1024:
https://github.com/pop-os/cosmic-comp/pull/1995

Seems to be working (though that's not surprising given this is based on the cosmic-comp version).

I was able to adopt the `image_capture_source` helpers here before the `image_copy_capture`, so the implementation of `image_capture_source` is usable even if not using the `image_copy_capture` support here, which is nice. It also works fine to add our custom workspace capture source on top of it.

So it seems to meet our requirements, and I think should suit the requirements of other compositors.

chrisduerr:
> It also works fine to add our custom workspace capture source on top of it.
> So it seems to meet our requirements, and I think should suit the requirements of other compositors.

I've actually been contemplating this for a bit since I'm using screencopy-v1 and I think that's an important feature. While I haven't tried using it there seem to be a lot of desktop abstractions built in by default, so I appreciate that there seems to be an option to bypass that without having to just write the protocol implementation myself.

glamberson:
> Seems to be working (though that's not surprising given this is based on the cosmic-comp version).
> So it seems to meet our requirements, and I think should suit the requirements of other compositors.

Yes this is going to save me a lot of time. Thanks @ids1024 and @Drakulix for allowing me to adapt your work!

src/wayland/image_capture_source/mod.rs
glamberson: Oops I totally missed this. I'm addressing this now.

YaLTeR:
> I'd like to hear some input on this from @YaLTeR, given niri seems to have some interest in implementing the protocol eventually. Just to hear if the overall design seems to be a good fit for niri now as well.

Looks alright at a glance. I was adding some weird code to my wlr-screencopy impl over the past days but it seems that none of that code is necessary for these ext protocols because of the much improved protocol design. So I expect I'll be able to use this Smithay impl. Can't say definitively without trying though.

src/wayland/image_copy_capture/mod.rs
```diff
@@ -0,0 +1,1400 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `transform` - The transform applied to the captured content
+    /// * `damage` - Optional damage regions, or `None` for full damage
+    /// * `presented` - The presentation timestamp
+    pub fn success(
+        self,
+        transform: impl Into<Transform>,
+        damage: impl Into<Option<Vec<Rectangle<i32, BufferCoords>>>>,
```
YaLTeR, line 694: This is damage compared to what? What's the intended way for the compositor to keep track of it?

Drakulix: Damage since the last time the compositor rendered into the buffer (assuming the client re-uses buffers).

cosmic-comp creates an `OutputDamageTracker` per session and has a `CommitCounter` per buffer it has encountered to match against and figure out the age of the buffer before rendering.

YaLTeR: Would it make sense to have some helper for this in Smithay?

Drakulix: Maybe, but I think how you exactly implement this depends on how you render your capture. Would have to double check though to figure out which parts of that logic are potentially useful.

glamberson: Hmm. OutputDamageTracker already handles damage accumulation - the missing piece is buffer       
  identity:                                                                                                                  
                                                                                                                             
 ```
 pub struct CaptureBufferAges {                                                                                             
      buffer_frames: HashMap<u32, usize>,                                                                                    
      frame_counter: usize,                                                                                                  
  }                                                                                                                          
                                                                                                                             
  impl CaptureBufferAges {                                                                                                   
      pub fn age_of(&self, buffer: &WlBuffer) -> usize {                                                                     
          self.buffer_frames                                                                                                 
              .get(&buffer.id().protocol_id())                                                                               
              .map(|last| self.frame_counter.saturating_sub(*last))                                                          
              .unwrap_or(0)                                                                                                  
      }                                                                                                                      
                                                                                                                             
      pub fn commit(&mut self, buffer: &WlBuffer) {                                                                          
          self.buffer_frames.insert(buffer.id().protocol_id(), self.frame_counter);                                          
          self.frame_counter = self.frame_counter.wrapping_add(1);                                                           
      }                                                                                                                      
  }                                                                                                                          

```                                                                                                                             
  But I share @Drakulix's caution - this only fits the "re-render into client buffer" case. Video codecs or encoder-state    
  pipelines need something different. Maybe better as documented pattern than API?  I can go either way.

ids1024: I don't think the server needs to track the age of a `WlBuffer`; at least if the `wlroots`/`sway` implementation is correct, it seems the client needs to track that and call `ext_image_copy_capture_frame_v1::damage_buffer` with appropriate accumulated damage.

I think this part of the protocol spec implies this:

https://wayland.app/protocols/ext-image-copy-capture-v1#ext_image_copy_capture_frame_v1:request:damage_buffer

> When a wl_buffer is captured for the first time, or when the client doesn't track damage, the client must damage the whole buffer.

Making it clear that the client has to "track damage", or damage the whole buffer.

(I guess https://github.com/pop-os/cosmic-comp/pull/1651 hasn't been merged in cosmic-comp; maybe I'll add that to the PR moving to this image-copy abstraction.)

YaLTeR: That's a different damage. There's damage on the client side, and there's damage on the compositor side.

ids1024: As I understand, a client using multiple buffers is expected to accumulate damage reported by the server, and pass it to `damage_buffer`. (Of course, if the client modifies the buffer itself, that damage should also be added.)

See https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/146/files  for how I did that in `xdg-desktop-portal-cosmic`, to get it working properly testing on Sway.

So the server only needs to track the damage since the last capture on the image-copy session, not since the last capture to a particular `wl_buffer`, and doesn't need to track the "age" of the `wl_buffer`.

YaLTeR: > The client indicates the accumulated damage since this wl_buffer was last captured.

Reads to me like per-buffer?

ids1024: It definitely wouldn't hurt if the wording of the protocol spec were clarified a bit, but "accumulated damage since this wl_buffer was last captured" is (assuming this buffer was previously captured on this session and not modified since then) the combination of the damage the server reported on the captures to other buffers since this buffer was last captured.

The wording of the protocol seems a little confusing partly because it seems to be mixing descriptions of how the protocol is defined to work with how it expects a client to use the protocol. (The compositor doesn't really care if the `wl_buffer` was previously captured. Only that the buffer contents are reflect the last frame captured on the session, other than the `buffer_damage` rects.)

glamberson: Ah, I had this backwards. Looking at the protocol spec and @ids1024's PRs more carefully:                                  
                                                                                                                             
  Server responsibility: Report damage since last ready on this session (not per-buffer)                                     
                                                                                                                             
  Client responsibility: Accumulate that damage per-buffer and send via damage_buffer                                        
                                                                                                                             
  So the server just needs age=1 semantics - "what changed since last frame." The CaptureBufferAges pattern I suggested would be client-side logic, not server-side.                                                                                    
                                                                                                                             
  The damage parameter in Frame::success() is simply "what changed in the source since the last capture on this session" . First frame is full damage, subsequent frames are incremental. No buffer tracking needed server-side.

YaLTeR: The protocol is very unclear on the serverside damage imo.

> Subsequent frames' damaged regions describe which parts of the buffer have changed since the last ready event.

"which parts of the buffer" implies that it's per-buffer. But what if the last ready event had a different buffer? You could also plausibly read this as "since the last ready event [in which this buffer was used]".

Drakulix: This sounds more like an upstream PR for the protocol would be in order rather than any changes to this API?

ids1024: Yeah, I don't believe we need an API change here. It would be good to get clearer wording upstream in `wayland-protocols`.

ids1024:
Are we waiting on another review from @YaLTeR here, or is this satisfactory to merge now?

(Though separately, it would be good to think of some better wording to MR upstream to `wayland-protocols` to clarify how damage works...)

glamberson:
> Are we waiting on another review from @YaLTeR here, or is this satisfactory to merge now?
> 
> (Though separately, it would be good to think of some better wording to MR upstream to `wayland-protocols` to clarify how damage works...)

Yes I hope we can go ahead and merge this as I need to start working with it.

PR #1904, opened 2026-01-11
Fix missing Copy for `LayerSurfaceConfigure`

chrisduerr:
This fixes what I can only assume to be an oversight, where `LayerSurfaceCachedState` had its `Copy` implementation removed due to the addition of `LayerSurfaceConfigure`, which itself was fully capable of deriving `Copy`.

### DIFF
--- a/src/wayland/shell/wlr_layer/mod.rs
+++ b/src/wayland/shell/wlr_layer/mod.rs
@@ -128,7 +128,7 @@ impl LayerSurfaceAttributes {
             .find(|configure| configure.serial == serial)
             .cloned()?;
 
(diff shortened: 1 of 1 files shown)


YaLTeR:
I think I may have removed it for consistency with ToplevelConfigure. That one has some non-Copy fields. So not having Copy here lets us add non-Copy fields in the future in a less-breaking fashion. Not a big deal though.

chrisduerr:
> So not having Copy here lets us add non-Copy fields in the future in a less-breaking fashion. Not a big deal though.

Removing it was a breaking change though, so that doesn't really help. :smile:

PR #1907, opened 2026-01-18
feat(input): add pre-compiled keymap

my4ng:
This allows setting a pre-compiled keymap while ensuring the keymap was created under the same keyboard handle, and more importantly, the same xkbcommon context within the same thread.

This is related to https://github.com/stefanboca/niri/pull/1 which adds per-device keyboard layout. To support this, if the keymap of a keyboard is different from that of the last used one, then we need to set the new keymap. Currently, the only methods to achieve this are `set_keymap_from_string` and `set_xkb_config`, which both require compiling the keymap every time on switching. The new `get_keymap` and `set_keymap` methods thus allow caching the compiled keymaps, likely in a HashMap, to prevent redundant keymap compilations.

### DIFF
--- a/src/input/keyboard/mod.rs
+++ b/src/input/keyboard/mod.rs
@@ -203,6 +203,21 @@ impl fmt::Debug for Xkb {
 // same thread
 unsafe impl Send for Xkb {}
 
+/// A keymap that is bound to a specific keyboard handler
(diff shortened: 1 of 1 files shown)


my4ng:
Superseded by #1903

PR #1910, opened 2026-01-22
element/damage: Better support effects on framebuffer contents

Drakulix:
This is an attempt add changing `Element`/`RenderElement` alongside the damage algorithms to better support effects on the underlying framebuffer, more specifically blurring the background. But similar effects are imaginable and this tries to be generic towards whatever future versions of `ext-background-effects-v1` might allow.

An implementation of a `BlurElement`, that makes use of this code, can be found here: https://github.com/pop-os/cosmic-comp/blob/frosted-glass_noble/src/backend/render/blur.rs#L206

To facilitate this, we need two things:
- We need to inform the element, when it's underlying framebuffer is dirty, so that it can re-capture the framebuffer and doesn't have to do that on every draw-call as these effects can be quite costly.
  - Most importantly we don't want to update the effect, if the surface on top of it updates or the a cursor passes over. (Assuming the element caches it's results, which makes sense for blurring but might not for others, which likely can be trivially implemented today.)
- We need to make sure to damage the whole element, whenever we want to allow the element to capture the framebuffer, so that everything below it is re-draw on that same cycle.

Facilitating the latter is easy. `Element` gets a new `is_framebuffer_effect` method. We use that for both "damaging the whole element" and figuring out, if we need the more expensive logic in the first place.

For the former, I introduced a `damage_index` to our `OutputDamageTracker` per element. Since we iterate through the elements top-to-bottom, we can record the index of all damage collected so far when we hit a new element to have a clean cut-off of damage above and below our element. We can then use that later to figure out, if any damage below the object occurred, which prompts `capture` to be called.

### DIFF
--- a/src/backend/renderer/damage/mod.rs
+++ b/src/backend/renderer/damage/mod.rs
@@ -130,6 +130,7 @@ struct ElementInstanceState {
     last_transform: Transform,
     last_alpha: f32,
     last_z_index: usize,
+    last_is_framebuffer_effect: bool,
 }
 
 impl ElementInstanceState {
@@ -141,12 +142,14 @@ impl ElementInstanceState {
         transform: Transform,
         alpha: f32,
         z_index: usize,
+        is_framebuffer_effect: bool,
     ) -> bool {
         self.last_src == src
             && self.last_geometry == geometry
             && self.last_transform == transform
             && self.last_alpha == alpha
             && self.last_z_index == z_index
+            && self.last_is_framebuffer_effect == is_framebuffer_effect
     }
 }
 
@@ -165,10 +168,11 @@ impl ElementState {
         transform: Transform,
         alpha: f32,
         z_index: usize,
+        is_framebuffer_effect: bool,
     ) -> bool {
         self.last_instances
             .iter()
-            .any(|instance| instance.matches(src, geometry, transform, alpha, z_index))
+            .any(|instance| instance.matches(src, geometry, transform, alpha, z_index, is_framebuffer_effect))
     }
 }
 
@@ -192,6 +196,7 @@ pub struct OutputDamageTracker {
     element_damage: Vec<Rectangle<i32, Physical>>,
     opaque_regions: Vec<Rectangle<i32, Physical>>,
     opaque_regions_index: Vec<Range<usize>>,
+    element_damage_index: Vec<usize>,
     element_opaque_regions: Vec<Rectangle<i32, Physical>>,
     element_visible_area_workhouse: Vec<Rectangle<i32, Physical>>,
     span: tracing::Span,
@@ -257,6 +262,7 @@ impl OutputDamageTracker {
             element_damage: Default::default(),
             opaque_regions: Default::default(),
             opaque_regions_index: Default::default(),
+            element_damage_index: Default::default(),
             element_opaque_regions: Default::default(),
             element_visible_area_workhouse: Default::default(),
             span: info_span!("renderer_damage"),
@@ -276,6 +282,7 @@ impl OutputDamageTracker {
             element_damage: Default::default(),
             opaque_regions: Default::default(),
             opaque_regions_index: Default::default(),
+            element_damage_index: Default::default(),
             element_opaque_regions: Default::default(),
             element_visible_area_workhouse: Default::default(),
             last_state: Default::default(),
@@ -298,6 +305,7 @@ impl OutputDamageTracker {
             element_opaque_regions: Default::default(),
             opaque_regions: Default::default(),
             opaque_regions_index: Default::default(),
+            element_damage_index: Default::default(),
             element_visible_area_workhouse: Default::default(),
             last_state: Default::default(),
         }
@@ -426,6 +434,13 @@ impl OutputDamageTracker {
                     element_damage,
                 );
 
+                if states
+                    .element_render_state(element_id.clone())
+                    .is_some_and(|state| state.needs_capture)
+                {
+                    element.capture_framebuffer(&mut frame, element_geometry)?;
+                }
+
                 element.draw(
                     &mut frame,
                     element.src(),
@@ -516,6 +531,7 @@ impl OutputDamageTracker {
         self.damage.clear();
         self.opaque_regions.clear();
         self.opaque_regions_index.clear();
+        self.element_damage_index.clear();
 
         let mut element_render_states = RenderElementStates {
             states: HashMap::with_capacity(elements.len()),
@@ -525,6 +541,7 @@ impl OutputDamageTracker {
         let mut element_damage = std::mem::take(&mut self.element_damage);
 
         let mut element_visible_area_workhouse = std::mem::take(&mut self.element_visible_area_workhouse);
+        let mut z_index = 0;
         for element in elements.iter() {
             let element_id = element.id();
             let element_loc = element.geometry(output_scale).loc;
@@ -559,18 +576,52 @@ impl OutputDamageTracker {
                 continue;
             }
 
-            let element_output_damage = element
-                .damage_since(
-                    output_scale,
-                    self.last_state.elements.get(element_id).map(|s| s.last_commit),
-                )
-                .into_iter()
-                .map(|mut d| {
-                    d.loc += element_loc;
-                    d
+            let element_src = element.src();
+            let element_geometry = element.geometry(output_scale);
+            let element_transform = element.transform();
+            let element_alpha = element.alpha();
+            let element_last_state = self.last_state.elements.get(element.id());
+            let element_is_framebuffer_effect = element.is_framebuffer_effect();
+
+            self.element_damage_index.push(self.damage.len());
+            if element_last_state
+                .map(|s| {
+                    !s.instance_matches(
+                        element_src,
+                        element_geometry,
+                        element_transform,
+                        element_alpha,
+                        z_index,
+                        element_is_framebuffer_effect,
+                    )
                 })
-                .filter_map(|geo| geo.intersection(output_geo));
-            self.damage.extend(element_output_damage);
+                .unwrap_or(true)
+            {
+                if let Some(intersection) = element_geometry.intersection(output_geo) {
+                    self.damage.push(intersection);
+                }
+                if let Some(state) = element_last_state {
+                    self.damage.extend(
+                        state
+                            .last_instances
+                            .iter()
+                            .filter_map(|i| i.last_geometry.intersection(output_geo)),
+                    );
+                }
+            } else {
+                let element_output_damage = element
+                    .damage_since(
+                        output_scale,
+                        self.last_state.elements.get(element_id).map(|s| s.last_commit),
+                    )
+                    .into_iter()
+                    .map(|mut d| {
+                        d.loc += element_loc;
+                        d
+                    })
+                    .filter_map(|geo| geo.intersection(output_geo));
+                self.damage.extend(element_output_damage);
+            }
 
             let element_opaque_regions_start_index = self.opaque_regions.len();
             let element_opaque_regions = element
@@ -599,6 +650,7 @@ impl OutputDamageTracker {
                     RenderElementState::rendered(element_visible_area),
                 );
             }
+            z_index += 1;
         }
         std::mem::swap(
             &mut self.element_visible_area_workhouse,
@@ -621,40 +673,6 @@ impl OutputDamageTracker {
                     .iter()
                     .filter_map(|i| i.last_geometry.intersection(output_geo)),
             );
-        }
-
-        // if the element has been moved or it's alpha or z index changed, damage it
-        for (z_index, element) in render_elements.iter().enumerate() {
-            let element_src = element.src();
-            let element_geometry = element.geometry(output_scale);
-            let element_transform = element.transform();
-            let element_alpha = element.alpha();
-            let element_last_state = self.last_state.elements.get(element.id());
-
-            if element_last_state
-                .map(|s| {
-                    !s.instance_matches(
-                        element_src,
-                        element_geometry,
-                        element_transform,
-                        element_alpha,
-                        z_index,
-                    )
-                })
-                .unwrap_or(true)
-            {
-                if let Some(intersection) = element_geometry.intersection(output_geo) {
-                    self.damage.push(intersection);
-                }
-                if let Some(state) = element_last_state {
-                    self.damage.extend(
-                        state
-                            .last_instances
-                            .iter()
-                            .filter_map(|i| i.last_geometry.intersection(output_geo)),
-                    );
-                }
-            }
         }
 
         // damage regions no longer covered by opaque regions
@@ -683,6 +701,24 @@ impl OutputDamageTracker {
                 "Output geometry, transform or clear color changed, damaging whole output geometry");
             self.damage.clear();
             self.damage.push(output_geo);
+        }
+
+        // for backdrop elements check if anything below them changed and add full-damage, if it did.
+        for (z_index, element) in render_elements
+            .iter()
+            .enumerate()
+            .filter(|(_, e)| e.is_framebuffer_effect())
+        {
+            let damage_index = self.element_damage_index[z_index];
+            let element_geometry = element.geometry(output_scale);
+            let intersection = element_geometry.intersection(output_geo);
+
+            if intersection.is_some_and(|i| self.damage.iter().skip(damage_index).any(|d| d.overlaps(i))) {
+                if let Some(state) = element_render_states.states.get_mut(element.id()) {
+                    state.needs_capture = true;
+                    self.damage.push(intersection.unwrap());
+                }
+            }
         }
 
         // That is all completely new damage, which we need to store for subsequent renders
@@ -743,6 +779,7 @@ impl OutputDamageTracker {
                     let elem_alpha = elem.alpha();
                     let elem_geometry = elem.geometry(output_scale);
                     let elem_transform = elem.transform();
+                    let element_is_framebuffer_effect = elem.is_framebuffer_effect();
 
                     if let Some(state) = map.get_mut(id) {
                         state.last_instances.push(ElementInstanceState {
@@ -751,6 +788,7 @@ impl OutputDamageTracker {
                             last_transform: elem_transform,
                             last_alpha: elem_alpha,
                             last_z_index: z_index,
+                            last_is_framebuffer_effect: element_is_framebuffer_effect,
                         });
                     } else {
                         let current_commit = elem.current_commit();
@@ -764,6 +802,7 @@ impl OutputDamageTracker {
                                     last_transform: elem_transform,
                                     last_alpha: elem_alpha,
                                     last_z_index: z_index,
+                                    last_is_framebuffer_effect: element_is_framebuffer_effect,
                                 }],
                             },
                         );
--- a/src/backend/renderer/damage/mod.rs
+++ b/src/backend/renderer/damage/mod.rs
@@ -541,7 +541,7 @@ impl OutputDamageTracker {
         let mut element_damage = std::mem::take(&mut self.element_damage);
 
         let mut element_visible_area_workhouse = std::mem::take(&mut self.element_visible_area_workhouse);
-        let mut z_index = 0;
+        let mut render_element_z_index = 0;
         for element in elements.iter() {
             let element_id = element.id();
             let element_loc = element.geometry(output_scale).loc;
@@ -591,7 +591,7 @@ impl OutputDamageTracker {
                         element_geometry,
                         element_transform,
                         element_alpha,
-                        z_index,
+                        render_element_z_index,
                         element_is_framebuffer_effect,
                     )
                 })
@@ -636,6 +636,8 @@ impl OutputDamageTracker {
             let element_opaque_regions_end_index = self.opaque_regions.len();
             self.opaque_regions_index
                 .push(element_opaque_regions_start_index..element_opaque_regions_end_index);
+
+            render_element_z_index += 1;
             render_elements.push(element);
 
             if let Some(state) = element_render_states.states.get_mut(element_id) {
@@ -650,7 +652,6 @@ impl OutputDamageTracker {
                     RenderElementState::rendered(element_visible_area),
                 );
             }
-            z_index += 1;
         }
         std::mem::swap(
             &mut self.element_visible_area_workhouse,
@@ -721,6 +722,11 @@ impl OutputDamageTracker {
             } else {
                 self.element_damage_index[z_index]
             };
+            let opaque_regions_index = if force_effect_redraw {
+                self.opaque_regions.len()
+            } else {
+                self.opaque_regions_index[z_index].start
+            };
             let element_geometry = element.geometry(output_scale);
             let intersection = element_geometry.intersection(output_geo);
 
@@ -728,6 +734,16 @@ impl OutputDamageTracker {
                 if let Some(state) = element_render_states.states.get_mut(element.id()) {
                     state.needs_capture = true;
                     self.damage.push(intersection.unwrap());
+                    // also drop all opaque regions on top, so they don't block re-drawing below the blur element
+                    for region in self.opaque_regions.iter_mut().take(opaque_regions_index) {
+                        // we want to leave `self.opaque_regions_index` intact,
+                        // fixing it up would be very involved, so lets do the next best thing
+                        // and keep at least part of the opaque region, if possible.
+                        *region = Rectangle::subtract_rect(*region, intersection.unwrap())
+                            .into_iter()
+                            .next()
+                            .unwrap_or_default();
+                    }
                 }
             }
         }
--- a/src/backend/renderer/damage/mod.rs
+++ b/src/backend/renderer/damage/mod.rs
@@ -657,6 +657,8 @@ impl OutputDamageTracker {
             &mut element_visible_area_workhouse,
         );
 
+        let mut force_effect_redraw = false;
+
         // add the damage for elements gone that are not covered an opaque region
         let elements_gone = self.last_state.elements.iter().filter(|(id, _)| {
             element_render_states
@@ -667,6 +669,7 @@ impl OutputDamageTracker {
         });
 
         for (_, state) in elements_gone {
+            force_effect_redraw = true;
             self.damage.extend(
                 state
                     .last_instances
@@ -680,6 +683,9 @@ impl OutputDamageTracker {
         element_damage.extend_from_slice(&self.last_state.opaque_regions);
         element_damage =
             Rectangle::subtract_rects_many_in_place(element_damage, self.opaque_regions.iter().copied());
+        if !element_damage.is_empty() {
+            force_effect_redraw = true;
+        }
         self.damage.extend_from_slice(&element_damage);
 
         // we no longer need the element damage, return it so that we can
@@ -701,6 +707,7 @@ impl OutputDamageTracker {
                 "Output geometry, transform or clear color changed, damaging whole output geometry");
             self.damage.clear();
             self.damage.push(output_geo);
+            force_effect_redraw = true;
         }
 
         // for backdrop elements check if anything below them changed and add full-damage, if it did.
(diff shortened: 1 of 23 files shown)


YaLTeR:
Haven't looked at the code, but I've got a question from the description: does this help with the case where blur needs to track damage slightly further than directly below the blur element? Since blur samples pixels from slightly outside. Or is this solved by having the blur element be slightly bigger itself?

Drakulix:
> Haven't looked at the code, but I've got a question from the description: does this help with the case where blur needs to track damage slightly further than directly below the blur element? Since blur samples pixels from slightly outside. Or is this solved by having the blur element be slightly bigger itself?

Yes, you can just increase the size of the `BlurElement` to match the window-geometry + radius and this code will ensure that the area around the window is also repainted to give accurate results when blitting from the framebuffer. You then just need to make sure to only render the inner window-geometry from your blurred texture.

My 2 cents on the matter: The implementation linked above I doesn't do this. I noticed even MacOS, which has different blur effects all over the place, doesn't implement this correctly. E.g. moving a window right next (pixel perfect) to a colored line doesn't make the color show up in the blur until the window overlaps with said line at least partially.

This lead me to lazily implement kawase-blur with `CLAMP_TO_EDGE` set for the captured texture (to correctly sample "something" outside of the texture geometry). And that emitted *exactly* the same bug while looking just fine.

So while this can be solved with this implementation, I don't think it really is necessary. (And you need to use something like `CLAMP_TO_EDGE` anyway for the blur to not look weird, when you windows come up against the outputs edge.)

YaLTeR:
Thanks. 

> I noticed even MacOS, which has different blur effects all over the place, doesn't implement this correctly. E.g. moving a window right next (pixel perfect) to a colored line doesn't make the color show up in the blur until the window overlaps with said line at least partially.

That's an interesting detail. Definitely makes it easier yeah.

src/backend/renderer/damage/mod.rs
```diff
@@ -141,12 +142,14 @@ impl ElementInstanceState {
...
         alpha: f32,
         z_index: usize,
+        is_framebuffer_effect: bool,
     ) -> bool {
         self.last_src == src
             && self.last_geometry == geometry
             && self.last_transform == transform
             && self.last_alpha == alpha
             && self.last_z_index == z_index
+            && self.last_is_framebuffer_effect == is_framebuffer_effect
```
YaLTeR, line 153: This makes sure that changing the return value of `is_framebuffer_effect()` without changing the id or commit counter will work correctly wrt. damage and stuff, right?

Drakulix: yes, this will cause full damage to be emitted once (and in the case of `false`->`true` also a call to `capture_framebuffer`).

src/backend/renderer/element/mod.rs
```diff
@@ -515,6 +532,19 @@ pub trait RenderElement<R: Renderer>: Element {
...
+
+    /// Notification, that the underlying framebuffer has changed allowing the element
+    /// to blit the contents of the `frame`.
+    ///
+    /// Will only be called if `Element::is_framebuffer_effect` returns `true`.
+    fn capture_framebuffer(
+        &self,
+        frame: &mut R::Frame<'_, '_>,
+        dst: Rectangle<i32, Physical>,
+    ) -> Result<(), R::Error> {
```
YaLTeR, line 603: So since you only get a `frame` here and no renderer, that means you cannot create `GlesTexture`s with their automatic cleanup, right? And have to manage the destruction manually somehow (relevant for pooling/reusing textures).

Drakulix: Yes, that is one of the things I am not so happy with here.

As you can see the cosmic-comp blur implementation currently uses a lot of raw GL code to side-step that issue, but yeah you don't get access to the destruction-channel this way.

Given this feature most likely needs a `Blit` anyway, I thought we have `BlitFrame` at least, but that also needs a `Framebuffer` and now you have to deal with storing a framebuffer and the texture in the element, making it self-referencial...

But think this is a bigger issue with the whole abstraction not necessarily something that needs to be fixed in this PR.

As for possible solutions I am not sure. We could implement more renderer-traits for the `GlesFrame` or possibly allow sub-`Frames` and then treat the whole thing more like a guard and abandon the strict separatation between `Renderer` and `Frame`? (related: https://github.com/Smithay/smithay/pull/1750)

src/backend/renderer/damage/mod.rs
```diff
@@ -599,6 +650,7 @@ impl OutputDamageTracker {
                     RenderElementState::rendered(element_visible_area),
                 );
             }
+            z_index += 1;
```
cmeissl: Small nit, I would move this up to `render_elements.push(element)` and maybe rename it to something like `render_element_zindex` to make it clearer that this z-index is only meant for visible render elements.

cmeissl (approved):

YaLTeR:
One change that tripped me up a little is that since is_framebuffer_effect() has a default impl, nothing tells you that you need to update your custom wrapper elements to pass through this new function. Something for the release notes I guess

YaLTeR:
In the cosmic-comp BlurElement impl, how does capture_framebuffer() handle the output transform? In winit when the transform is 180, the glBlitFramebuffer call blits from wrong coordinates for me.

YaLTeR:
Perhaps current transform needs to be passed into capture_framebuffer() too?

YaLTeR:
Found a potential issue with framebuffer effect, though not sure if it might be a mistake in [my code](https://github.com/YaLTeR/niri/blob/740e11fd792c193571f2bc2da91ae18873ba68af/src/render_helpers/blur.rs#L529). When it's covered by some other opaque window, and I move the covering window outside, it doesn't get redrawn:

https://github.com/user-attachments/assets/6177c8fc-7c42-4d3a-ad76-de89cf9682c7

Drakulix:
> Found a potential issue with framebuffer effect, though not sure if it might be a mistake in [my code](https://github.com/YaLTeR/niri/blob/740e11fd792c193571f2bc2da91ae18873ba68af/src/render_helpers/blur.rs#L529). When it's covered by some other opaque window, and I move the covering window outside, it doesn't get redrawn.

Pushed two commits that should fix the issue you are seeing here, please test.
I am not too happy with the code however, suggestions welcome.

Drakulix:
> In the cosmic-comp BlurElement impl, how does capture_framebuffer() handle the output transform? In winit when the transform is 180, the glBlitFramebuffer call blits from wrong coordinates for me.

Currently it doesn't. We should probably pass the output-transform. You could also receive the projection matrix of the `GlesFrame`, but that is likely not as helpful, given you'd want to use blit.

src/backend/renderer/damage/mod.rs
Drakulix: done!

YaLTeR:
Example way to repro on current latest commit:

https://github.com/user-attachments/assets/df090494-5f12-4eb4-a4dd-2dd70d901f31

YaLTeR:
Could it be that the following causes the issue?

- I move opaque window on top towards outside
- Only the newly exposed part of the framebuffer effect element gets damaged
- It calls capture_framebuffer() with the new logic
- capture_framebuffer() operates on the entire area, not just the damaged region
- part of the area that is still covered by the opaque surface was never damaged or redrawn
- so the framebuffer contents of that area contain the opaque surface, even though we're drawing the framebuffer effect element "below" it
- they end up in the captured framebuffer

Drakulix:
> Could it be that the following causes the issue?
> 
>     * I move opaque window on top towards outside
> 
>     * Only the newly exposed part of the framebuffer effect element gets damaged

This overlaps with the framebuffer-effect element, so the whole intersection of the framebuffer-effect element and the output geometry gets added to the damage.

> 
>     * It calls capture_framebuffer() with the new logic
> 
>     * capture_framebuffer() operates on the entire area, not just the damaged region
> 
>     * part of the area that is still covered by the opaque surface was never damaged or redrawn

All below it should be re-drawn in the same cycle. The new logic also ignores opaque regions on top of the framebuffer-effect element now, if any part of it (and thus the whole thing) got damaged.

>     * so the framebuffer contents of that area contain the opaque surface, even though we're drawing the framebuffer effect element "below" it
> 
>     * they end up in the captured framebuffer

Yes, that is what happened before.

YaLTeR:
I just thought of another fun edge case where a framebuffer effect may end up with wrong contents: drawing to a different damage tracker (e.g. offscreens or screencasts). The intended use is for capture_framebuffer() to write into some texture stored in the Element. So imagine drawing to the main damage tracker, then drawing to a different damage tracker once (e.g. one window screencast frame). This other damage tracker will call capture_framebuffer() which will update the texture (stored on the Element). Then, we draw back to the main damage tracker, with something changed *on top* of the element. There were no changes below, as far as the main damage tracker is concerned, so capture_framebuffer() is not called. But draw() *is* called, and it ends up drawing texture contents from the window screencast frame, rather than the intended ones from the main scene.

Drakulix:
> I just thought of another fun edge case where a framebuffer effect may end up with wrong contents: drawing to a different damage tracker (e.g. offscreens or screencasts). The intended use is for capture_framebuffer() to write into some texture stored in the Element. So imagine drawing to the main damage tracker, then drawing to a different damage tracker once (e.g. one window screencast frame). This other damage tracker will call capture_framebuffer() which will update the texture (stored on the Element). Then, we draw back to the main damage tracker, with something changed _on top_ of the element. There were no changes below, as far as the main damage tracker is concerned, so capture_framebuffer() is not called. But draw() _is_ called, and it ends up drawing texture contents from the window screencast frame, rather than the intended ones from the main scene.

Yeah caching the texture in the Element becomes problematic, if you use the same element across different framebuffers/damage-trackers. I guess this specifically becomes and issue with niri's block-on-screencast feature, right?

I am not sure how the code could address this without more information. The issue isn't really the different damage tracker, but that it potentially captured different contents, due to rendering for different intents, no?

Ideally it could remember the id's of the elements it captured somehow, so on the next render it would know, that there are different contents below instead of relying on the damage-tracker to store that information, but that seems not easily achievable.

Any ideas?

Drakulix:
Since all issues in the damage-tracking algorithm seem to be addressed, I updated the PR again. I dropped the `transform`-argument, since this can be gathered from the frame and added a method to query the `output_size` as well.

YaLTeR:
> Yeah caching the texture in the Element becomes problematic, if you use the same element across different framebuffers/damage-trackers. I guess this specifically becomes and issue with niri's block-on-screencast feature, right?

Yeah. I think in niri I'll just make sure to use different copies of the element for different damage trackers, but it's still a bit error prone because you can't check that an issue is happening other than stumbling upon it.

> I am not sure how the code could address this without more information. The issue isn't really the different damage tracker, but that it potentially captured different contents, due to rendering for different intents, no?

I think the issue is specifically different damage trackers, because the logic and state on whether a recapture needs to happen lies inside the damage tracker, and if you use two different ones, they don't share this info among each other.

> Any ideas?

Not sure tbh, one workaround could be to have some kind of unique damage tracker id that the element could test (similarly to how elements can test the renderer context id), but idk how good an idea this is. And then again, the decision to recapture is done by the damage tracker and not by the element.

Drakulix:
> Not sure tbh, one workaround could be to have some kind of unique damage tracker id that the element could test (similarly to how elements can test the renderer context id), but idk how good an idea this is. And then again, the decision to recapture is done by the damage tracker and not by the element.

Not a bad idea. imo either the object holding the capture-state (the damage-tracker) needs to also cache the result (the texture) or the caching needs to be able to somehow reliably invalidate itself or split the cache into two. And we definitely want to encourage caching the effect result.

I guess the id could cause trouble, if the damage-tracker is destroyed later and the texture starts to leak potentially? Storing the texture in the damage-tracker also has it's problems, how is it passed to the draw call?

I wonder if we have the same issue somewhere without framebuffer-effect, but I guess just with damage this problem doesn't come up.

cmeissl:
What about extending the whole rendering including the damage tracker with some external state context where the cached texture could be stored in. The context could be managed on a per use-case base, like output rendering, sceencapture and so on. Dropping the context would clean-up all cached textures.

Drakulix:
> What about extending the whole rendering including the damage tracker with some external state context where the cached texture could be stored in. The context could be managed on a per use-case base, like output rendering, sceencapture and so on. Dropping the context would clean-up all cached textures.

While I don't dislike the idea, I am failing to imagine the exact API for this. It seems very intrusive with everything from the DamageTracker's damage *and* render methods, `RenderElement`, `AsRenderElements` and probably more needing extensions for this new cache context.

Do you have other uses for this in mind? Should we also refactor the imported wl_buffer textures to use that context? I guess that would make it easier to do the import asynchronously from rendering, but given the textures are renderer-specific it is not like you can suddenly rely on textures being available, just because you use the same context. Would the context thus be specific to a `ContextId`?

I am struggling with with potentially adding a somehow overengineered feature, if there isn't a use-case for it outside of optimizing blur.

YaLTeR:
Hmm, another point for consideration. Currently frame buffer effect elements won't properly work if several clones of an element are present in the render list, because all copies will overwrite the same framebuffer texture stored in them (assuming it's shared via Arc/Rc among all of them, e.g. a GlesTexture).

Where this comes up is e.g. niri's overview which draws background/bottom layer-shell multiple times (once per visible workspace).

cmeissl:
> > What about extending the whole rendering including the damage tracker with some external state context where the cached texture could be stored in. The context could be managed on a per use-case base, like output rendering, sceencapture and so on. Dropping the context would clean-up all cached textures.
> 
> While I don't dislike the idea, I am failing to imagine the exact API for this. It seems very intrusive with everything from the DamageTracker's damage _and_ render methods, `RenderElement`, `AsRenderElements` and probably more needing extensions for this new cache context.
> 
> Do you have other uses for this in mind? Should we also refactor the imported wl_buffer textures to use that context? I guess that would make it easier to do the import asynchronously from rendering, but given the textures are renderer-specific it is not like you can suddenly rely on textures being available, just because you use the same context. Would the context thus be specific to a `ContextId`?
> 
> I am struggling with with potentially adding a somehow overengineered feature, if there isn't a use-case for it outside of optimizing blur.

No, no other specific use-case, just brainstorming some idea. I agree that this would be quite intrusive, but maybe a limited version could still work.
The cache context could be put in the damage tracker, which would allow automatic clean-up when the damage tracker is dropped and implicit clean-up when the element is no longer present.
This could also address the last message from @YaLTeR by using the element id and z-index as the cache key like we already do for the element instances. Maybe it would be enough enough to hand in a slot of the cache key to the function generating the texture.

cmeissl:
Hm, no, the cache texture would still have to go into the element somehow. But maybe the element could store a weak reference to the cache slot somehow. Not sure how it could identify the correct one without changing the draw signature though.

Edit: Maybe it could use the cache slot to set temporarily store it for one render cycle or something like this. Though I don't like that approach.

YaLTeR:
One usecase to consider here is drawing outside the damage tracker. This comes up in screencast or wlr-screencopy with damage: you call the screencopy-instance-specific damage tracker to compute damage, then if there's any damage, you render all elements with full damage to a (potentially completely fresh) buffer. 

This would need either some way to pass info from the damage tracker to the elements correctly, or alternatively some new method on the damage tracker that allows for this workflow (where you compute if there's any damage, but then draw with full damage).

Drakulix:
Okay, I think all remaining bugs here are squashed with the approaches discussed previously:

### Multiple DamageTrackers

I added a `UserDataMap`-based cache to the `DamageTracker`. While the code on the smithay side is very straight-forward, the consumer code is... not. (see [here](https://github.com/pop-os/cosmic-comp/blob/frosted-glass_noble/src/backend/render/wayland/blur_effect.rs#L79-L107) and [here](https://github.com/pop-os/cosmic-comp/blob/frosted-glass_noble/src/backend/render/wayland/blur_effect.rs#L79-L107)).

I am out of ideas how to improve this and it works. So... if nobody has any ideas, I would leave it like this for now.

### Multiple Effect-Elements with the same Id in a single DamageTracker

I added code to force re-capture in those cases, so it won't cause any render-glitches. Cosmic-comp only had a single case where this condition was triggered, which were workspace animations, which could end up rendering the same `LayerShell`-surface twice.

In those cases where a complete re-render happens on every frame anyhow and the whole animation doesn't even last a full second, I'd even argue this is good enough.

But I added namespaced `Id`s and a new `NamespacedElement`-wrapper and deployed that in cosmic-comp by [adding the workspace index](https://github.com/pop-os/cosmic-comp/blob/frosted-glass_noble/src/backend/render/mod.rs#L879) as a namespace.

The solution is simple and is confirmed working here.

---

I'd really like to get this over the finish line. Ping @YaLTeR for more testing?

There is a temporary `drakulix/testing-effects-subframe`-branch, that combines this PR and https://github.com/Smithay/smithay/pull/1946, which can be used for testing. Sub-Frames simplified cosmic-comps render code [quite a lot](https://github.com/pop-os/cosmic-comp/blob/frosted-glass_noble/src/backend/render/wayland/blur_effect.rs#L402-L606).

YaLTeR:
I'm updating niri to use this code.

I think I found a problem: you need to pass `cache` to `draw()` in addition to `capture_framebuffer()`. As I understand, the cosmic code [sets the texture to be used for drawing inside `capture_framebuffer()`](https://github.com/pop-os/cosmic-comp/blob/cc9549d5a4d80913432352de857487bfa391bf04/src/backend/render/wayland/blur_effect.rs#L339)—this is intended to be different between different clones of the element. However, when we're drawing two clones, and the first one had both `capture_framebuffer()` and `draw()`, while the second one didn't need recapture and only had `draw()`, it will use the texture from the first clone (as set in `capture_framebuffer()`). Passing `cache` to `draw()` would let it fetch the correct texture always.

Regarding subframes: they seem to work, though I'm not yet replacing my blitting code with a `blit_to()` call: I'm iffy about the extra target switch and sync point export and `glFlush()` that `GlesFrame::blit_to()` entails. Not sure there's anything to be fixed here since the function is supposed to work for all cases, but for this specific case I don't think a sync point + flush is necessary, and it'll just cause unnecessary pipeline stalls?

Drakulix:
> I think I found a problem: you need to pass `cache` to `draw()` in addition to `capture_framebuffer()`. As I understand, the cosmic code [sets the texture to be used for drawing inside `capture_framebuffer()`](https://github.com/pop-os/cosmic-comp/blob/cc9549d5a4d80913432352de857487bfa391bf04/src/backend/render/wayland/blur_effect.rs#L339)—this is intended to be different between different clones of the element. However, when we're drawing two clones, and the first one had both `capture_framebuffer()` and `draw()`, while the second one didn't need recapture and only had `draw()`, it will use the texture from the first clone (as set in `capture_framebuffer()`). Passing `cache` to `draw()` would let it fetch the correct texture always.

I am not sure I am following. So is this a problem across multiple DamageTrackers? I guess given each have their own cache, this would indeed mean, `draw` only has access to the last one that also called `capture_framebuffer`. Which I am not sure is relevant in cosmic-comp, as the same id should only be drawn in exactly the same set of elements/layout. (Except for maybe window capture.) Which explains, why I didn't notice it.

Urgh, I really didn't want to modify `draw`, but I guess this will have to happen, maybe with an `Option` for the cache-parameter. At this point I'd really like to use a separate `FramebufferEffectElement` trait (instead of the `RenderElement`), but given Rust isn't object-oriented there is no clean way (I can think of) to mix these into the normal Element-stack and only require the trait for objects returning `true` on `is_framebuffer_effect`..

> Regarding subframes: they seem to work, though I'm not yet replacing my blitting code with a `blit_to()` call: I'm iffy about the extra target switch and sync point export and `glFlush()` that `GlesFrame::blit_to()` entails. Not sure there's anything to be fixed here since the function is supposed to work for all cases, but for this specific case I don't think a sync point + flush is necessary, and it'll just cause unnecessary pipeline stalls?

Looking at the code, it seems to be a bit wrong anyway, given that `blit_to` and `blit_from` don't return the sync point, but if it is exportable `blit` returns early.

The target swap is necessary for binding the framebuffers correctly, but that itself shouldn't be an issue. I agree on the sync though.

It is a render-operation so on old nvidia drivers this `glFlush` was necessary, but on anything modern this should short-circut to an export without any usage of the `EGLFence`. So no effect on the pipeline, similar to `Frame::finish`, which has the same semantics otherwise.

YaLTeR:
> So is this a problem across multiple DamageTrackers? I guess given each have their own cache, this would indeed mean, draw only has access to the last one that also called capture_framebuffer.

I was actually thinking with two clones of an element within a single damage tracker. But I guess across two damage trackers it's a problem too.

> At this point I'd really like to use a separate FramebufferEffectElement trait (instead of the RenderElement), but given Rust isn't object-oriented there is no clean way (I can think of) to mix these into the normal Element-stack and only require the trait for objects returning true on is_framebuffer_effect..

I guess one way is to have a separate draw() function for framebuffer effect elements, default impl-d to panic, like capture framebuffer.. But I don't think it can be a separate trait without requiring every element to impl it (even if both functions have default impls and can be omitted)

Drakulix:
> I was actually thinking with two clones of an element within a single damage tracker. But I guess across two damage trackers it's a problem too.

Yeah, I see it now. Even if you namespace them, they only have one internal reference to the texture.

I have conceded and added the cache to the `draw`-call as well.

YaLTeR:
I converted my code to use the cache now, seems to work fine.

I started converting my manual screencopy render loop to `damage_tracker.render_output(age = 0)` and found another issue: since `element_state.needs_capture = true;` is currently set before checking for `age > 0`, it somehow ends up never set in my screencopy, so the framebuffer effect is never recaptured.

To clarify, my code looks like this:
```rust
let elements = ...;
let buffer = screencopy.buffer();

let damage = damage_tracker.damage_output(1, elements);
if damage.is_some() {
    screencopy.damage(...);
    damage_tracker.render_output(buffer, 0, elements);
}
```

This two-part logic is necessary to ensure a full redraw of the wlr-screencopy buffer (since we don't know its real age). I suspect however that it results in `render_output()` never seeing damage for framebuffer effect elements (since they were just damaged in `damage_output()` and haven't changed). So they never call `capture_framebuffer()`.

Drakulix:
> I converted my code to use the cache now, seems to work fine.
> 
> I started converting my manual screencopy render loop to `damage_tracker.render_output(age = 0)` and found another issue: since `element_state.needs_capture = true;` is currently set before checking for `age > 0`, it somehow ends up never set in my screencopy, so the framebuffer effect is never recaptured.
> [...]

I added a commit, which adds `DamageTracker::render_output_with_states`, which you can use with age=0 in-place and feed it the `states` of the previous `DamageTracker::damage_output`-call. That should ensure the framebuffer-effect is redrawn correctly.

YaLTeR:
It works but not entirely, the old "capture weirdness when something changes on top" issue manifests:

https://github.com/user-attachments/assets/cb65ab0c-03c1-4a46-8d10-0f0c04ea7615

The logic:
```rust
let elements = ...;
let buffer = screencopy.buffer();

let (damage, states) = damage_tracker.damage_output(1, elements);
if damage.is_some() {
    screencopy.damage(...);
    damage_tracker.render_output_with_states(buffer, 0, elements, states);
}
```

Drakulix:
> It works but not entirely, the old "capture weirdness when something changes on top" issue manifests.

I am brute-forcing myself through this bug a bit, but (in hindsight) obviously the second internal `damage_output`-call of `render_output_with_states` will set the opaque-regions again and thus bypass the the fix for this bug based on the `needs_capture` variable. I added an additional check to also check `needs_capture` for the provided states variable, which should make sure the fix is applied again.

Drakulix:
Added a commit that makes it possible to use the `update_surface_primary_scanout_output` and `surface_presentation_feedback_flags_from_states` helpers with namespaced surface elements.

Paraworker:
I noticed that `Element` already has the `Kind` enum. Would it make sense to represent framebuffer effects as another variant in `Kind` instead of introducing separate methods like `is_framebuffer_effect`?

YaLTeR:
That Kind determines the scanout state (if the element should go on the Cursor plane, etc.) so it's orthogonal to is_framebuffer_effect.

Paraworker:
> That Kind determines the scanout state (if the element should go on the Cursor plane, etc.) so it's orthogonal to is_framebuffer_effect.

Ah, that makes sense. The name Kind misled me a bit at first.

src/backend/renderer/damage/mod.rs
```diff
@@ -615,6 +637,7 @@ impl OutputDamageTracker {
         });
 
         for (_, state) in elements_gone {
+            force_effect_redraw = true;
```
cmeissl: I guess the compiler could be smart enough to figure it out, but we could still move this out of the loop.

src/backend/renderer/damage/mod.rs
```diff
@@ -683,6 +675,49 @@ impl OutputDamageTracker {
...
+            let with_element_state = with_states
+                .as_ref()
+                .and_then(|states| states.states.get(element.id()));
+
+            if element_state.needs_capture
+                || with_element_state.is_some_and(|state| state.needs_capture)
+                || intersection.is_some_and(|i| self.damage.iter().skip(damage_index).any(|d| d.overlaps(i)))
+            {
+                element_state.needs_capture = true;
+                self.damage.push(intersection.unwrap());
```
cmeissl: Afaict render_elements should already only contain elements that overlap with the output geometry, so this should never panic.
Otherwise `with_element_state.is_some_and(|state| state.needs_capture)` could trigger the panic on `intersection.unwrap()`

src/backend/renderer/damage/mod.rs
```diff
@@ -785,4 +826,138 @@ impl OutputDamageTracker {
...
+                );
+
+                if states
+                    .element_render_state(element_id.clone())
+                    .is_some_and(|state| state.needs_capture)
+                {
+                    let cache = self
+                        .last_state
+                        .effects_cache
+                        .entry(element.id().clone())
```
cmeissl: Should not really make a difference, but we already store the id in `element_id`.

src/backend/renderer/damage/mod.rs
```diff
@@ -785,4 +826,138 @@ impl OutputDamageTracker {
...
+                    element.capture_framebuffer(&mut frame, element.src(), element_geometry, cache)?;
+                }
+
+                element.draw(
+                    &mut frame,
+                    element.src(),
+                    element_geometry,
+                    &element_damage,
+                    &element_opaque_regions,
+                    self.last_state.effects_cache.get(element.id()),
```
cmeissl: Same here

src/backend/renderer/damage/mod.rs
```diff
@@ -785,4 +826,138 @@ impl OutputDamageTracker {
...
+                if states
+                    .element_render_state(element_id.clone())
+                    .is_some_and(|state| state.needs_capture)
+                {
+                    let cache = self
+                        .last_state
+                        .effects_cache
+                        .entry(element.id().clone())
+                        .or_default();
+                    element.capture_framebuffer(&mut frame, element.src(), element_geometry, cache)?;
```
cmeissl: `element.src()` could be stored as we might use it twice and there is no guarantee that the call has to be cheap

src/backend/renderer/damage/mod.rs
```diff
@@ -785,4 +826,138 @@ impl OutputDamageTracker {
...
+                }
+
+                element.draw(
+                    &mut frame,
+                    element.src(),
```
cmeissl: Second src call

cmeissl (approved):

cmeissl (approved):
Just a few nits from my side

src/backend/renderer/damage/mod.rs
Drakulix: Right, I checked `damage_output_internal` and we definitely already filter out elements with no overlap.

As such I moved the unwrap up to only do that once and simplify the if-condition.

Drakulix:
@cmeissl Should all be addressed. Would appreciate a quick sanity check on the new commit before merging.

cmeissl (approved):
LGTM

YaLTeR:
For the "Allow querying RenderElementStates with namespace" commit, is the idea for the compositor to keep track of and pick the "preferred" namespace for updating from render element states? I don't see a way to tell it "find element regardless of namespace", as far as I can tell passing `None` will only match unnamespaced elements.

So multiple unnamespaced clones will find "some" (undefined) single clone, but with a namespace you always have to pass the correct namespace (or correct absence of one), is that correct?

Drakulix:
> So multiple unnamespaced clones will find "some" (undefined) single clone, but with a namespace you always have to pass the correct namespace (or correct absence of one), is that correct?

Yes

PR #1911, opened 2026-01-22
Add `alive()` method for grabs, and add `PopupTouchGrab`

ids1024:
https://wayland.app/protocols/xdg-shell#xdg_popup:request:grab says popups that take a grab should also take touch input.

### DIFF
--- a/anvil/src/shell/grabs.rs
+++ b/anvil/src/shell/grabs.rs
@@ -164,7 +164,12 @@ impl<BackendData: Backend> PointerGrab<AnvilState<BackendData>> for PointerMoveS
         &self.start_data
     }
 
-    fn unset(&mut self, _data: &mut AnvilState<BackendData>) {}
+    fn unset(
+        &mut self,
+        _data: &mut AnvilState<BackendData>,
+        _handle: &mut PointerInnerHandle<'_, AnvilState<BackendData>>,
+    ) {
+    }
 }
 
 pub struct TouchMoveSurfaceGrab<BackendData: Backend + 'static> {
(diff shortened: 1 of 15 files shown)


ids1024:
I guess we need an MSRV of 1.88 to use `if let` chains. No matter, though this is a little neater with that.

ids1024:
Okay, I think this is working now.

Previously, running two copies of `cosmic-term` in `anvil` or `cosmic-comp`, opening a menu in one then touching a menu in the second would make the second instance crash with a ` xdg_popup was not created on the topmost popup` protocol error.

Not sure if we should change anything else related to that, but having a touch grab prevents that. Now either the grab is removed, or touch input, like other input, can't go to the other process.

I've added a new `.alive()` mechanism to grabs, which seems like a better way to make sure removing one of these grabs removes them all.

ids1024:
Oh, and I forgot that I meant to use `has_ended()` for defining `alive()`, which should also make the separate `has_ended()` checks unnecessary.

This may warrant a bit more testing, but assuming no unforseen issues, it's a nice cleanup to the popup grab code.

ids1024:
I guess to change how popup grabs work, there are more details to get the behavior right.

Though on cosmic-comp (on the existing version), I see `Esc` dismisses both the pointer and keyboard grabs but still leaves the popup open, so some changes here could be appropriate regardless.

ids1024:
Well now that issue is fixed too, but closing a menu by touching elsewhere on the window is making the window not open again...

ids1024:
I guess not having a `refresh` method for this does still mean the grabs may not have `unset` called for a long time if no input of a particular category occurs. (So the pointer grab isn't being `unset` if only touch input has happened since `alive` started returning `false`)

PR #1912, opened 2026-01-23
wayland/xdg_foreign: Fix panic when exporting surface with invalid role

Ericky14:
Initialize the exported object before validating the surface role to prevent wayland-backend panic


Ericky14:
> The comments describing that the object has to be initialized are not true, the truth is the opposite, the object should not be initialized.
> 
> The real bug is a typo in the post_error call, it should be `resource.post_error` rather than `data_init.post_error`

Fixed it, sorry about that, I was confused about how it worked then

PR #1916, opened 2026-02-01
Add GlesRenderer::with_profiled_context()

YaLTeR:
Counterpart to the GlesFrame method when you want to profile some operation that doesn't involve a GlesFrame. This one has `profiler.collect()` and `profiler.sync_gpu()` because it's a standalone rendering operation where we can't tell for sure if there will be any rendering right before or right after.


Drakulix (approved):
LGTM

PR #1919, opened 2026-02-02
Typos fix

Decodetalkers:
I have found a lot of typo, so I want to do some fix.. can it be accepted?

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -1,8 +1,7 @@
 ---
-# yamllint disable rule:line-length
 name: check_typos
 
-on:  # yamllint disable-line rule:truthy
+on:
   push:
   pull_request:
     branches:
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -1,0 +1,19 @@
+---
+# yamllint disable rule:line-length
+name: check_typos
+
+on:  # yamllint disable-line rule:truthy
+  push:
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: typos-action
+        uses: crate-ci/typos@v1.42.1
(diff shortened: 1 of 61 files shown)


cmeissl (approved):
I looked through the changes and they look fine to me.
No code changes, only docs and comments.

As this also extends the CI I want to wait for another maintainer to also take a look.
cc @Drakulix @PolyMeilex 

Small nit: the name of the CI task does not follow the naming of the existing tasks.

.github/workflows/typos.yml
```diff
@@ -0,0 +1,19 @@
+---
+# yamllint disable rule:line-length
```
PolyMeilex: Why? At a glance, I don't think we are exceeding the line length lint.

Decodetalkers: Sorry,it is a copied file, I will fix it

Decodetalkers: done

PolyMeilex (approved):

.github/workflows/typos.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: typos-action
+        uses: crate-ci/typos@v1.42.1
```
Drakulix, line 18: This is more a question for the other maintainers (cc @cmeissl, @PolyMeilex), but do we want a typos-linter?

cmeissl: I don't have a strong opinion on this tbh, but I think we should give it a try. Having all those typos fixed is good, but without linting I will be the first one to re-introduce some new ones.

If this turns out to cause troubles we can always remove the lint task.

PolyMeilex: Let's try, if it turns out to be annoying we can just drop it. I use typos-cli as LSP locally, and I'm pretty happy with it.

Drakulix: :+1:

Drakulix:
Thanks!

kode54:
What a lovely script. It went off on my fork repo. It treats warnings as errors, too.

"Simon Ser" -> "Simon Set"
"max_cll" -> "max_call"

PolyMeilex:
https://github.com/crate-ci/typos#false-positives

kode54:
Excellent, thanks for notifying me there's a workarounds config. I'll employ that in my fork, and consider the use if/when I attempt upstreaming things.

Issue #1859, opened 2025-11-15
Smithay doesn't play nice when EDID drops due to video switchbox changing inputs


alaughlin:
Apologies is this is more of a general wayland problem, but I'm seeing an issue where if I have my computer plugged into a video switchbox that does _not_ support edid emulation, when I switch away from the computer and back to it, the display will not wake up back unless I unplug and replug the cable.

From my general understanding and some light research, this is possibly because the compositor sees the EDID drop and shuts off the display? I think Xorg and other display servers (like what Windows and MacOS each use) can generally handle this more gracefully.

Please excuse my ignorance, but Is it possible to get this working in Smithay? Or is this a more deep rooted issue with Wayland itself?

PR #1923, opened 2026-02-07
fix(drm-extras): re-emit Connected event when modes change on already-connected connector

### LINKED ISSUE #1922: ConnectorScanner does not re-emit Connected event when modes change on already-connected connector
## Problem

`ConnectorScanner::scan()` in `smithay-drm-extras` only tracks `connector::State` transitions (Connected ↔ Disconnected ↔ Unknown). When a connector remains `Connected` across two consecutive scans, the `(Connected, Connected) => {}` arm is a no-op — no event is emitted regardless of whether the connector's properties (specifically its mode list) have changed.

**File**: `smithay-drm-extras/src/drm_scanner/connector_scanner.rs`, line 55

```rust
(State::Connected, State::Connected) => {}
```

## Impact

This causes a permanent dead state for connectors on USB-C docks with DP MST/alt-mode (and potentially other hotplug scenarios):

1. Kernel DRM reports a connector as `Connected` **before** EDID data is available (firmware negotiation in progress)
2. `get_connector(handle, true)` returns the connector with `state() == Connected` but `modes()` returns an empty list
3. The compositor calls `pick_mode()` → gets `None` → skips activation
4. A subsequent udev `Changed` event triggers another scan
5. `get_connector()` now returns populated modes, **but** `ConnectorScanner` sees `(Connected, Connected)` and emits nothing
6. The connector is stuck in "connected but never activated" state permanently

## Affected Compositors

This affects any compositor using `ConnectorScanner` or `DrmScanner` from `smithay-drm-extras`:

- **niri** — [YaLTeR/niri#869](https://github.com/YaLTeR/niri/issues/869) (crash/freeze on USB-C dock), [YaLTeR/niri#2691](https://github.com/YaLTeR/niri/issues/2691) (inconsistent display detection)
- Potentially any smithay-based compositor with USB-C dock or DP MST users

## Cross-Compositor Analysis

For context, I analyzed how other compositors handle this:

| Compositor | Handling |
|------------|----------|
| **Mutter (GNOME)** | 3-second `CONNECTION_CHANGE_TIMEOUT` debounce timer per connector; detects mode list changes as `META_KMS_RESOURCE_CHANGE_FULL` triggering full reload |
| **wlroots** | No retry — accepts connected+empty-modes as-is, relies on subsequent udev events |
| **KWin** | `isConnected()` requires `!m_driverModes.empty()` — connected+no-modes treated as disconnected; no retry for this case |

Mutter is the only compositor that handles mode-list changes on already-connected connectors at the library level.

## Proposed Fix

Compare the old and new `connector::Info` mode lists in the `(Connected, Connected)` arm and re-emit a `Connected` event when they differ:

```rust
(State::Connected, State::Connected) => {
    if old.modes() != conn.modes() {
        added.push(conn);
    }
}
```

This is safe because:
- `connector::Info` derives `PartialEq`, and `modes()` returns `&[control::Mode]` which also implements `PartialEq`
- Compositors already handle duplicate `Connected` events gracefully (they check if a surface already exists for the CRTC)
- The comparison is cheap for the common case (both mode lists are the same)

## Reproduction

- **Hardware**: ThinkPad P16 Gen 3, NVIDIA RTX PRO 4000 (nvidia-drm 590.48.01), Intel iGPU (i915), Lenovo ThinkPad USB-C Dock Gen 2
- **Monitors**: Two LEN S27q-10 (2560x1440@60Hz) connected via HDMI + DP MST through dock
- **Symptoms**: Intermittent — sometimes both monitors activate on boot, sometimes one stays dark permanently until dock is re-plugged

## Related

- Smithay #1859 — "Smithay doesn't play nice when EDID drops due to video switchbox changing inputs" (same root cause, different trigger)

coleleavitt:
Fixes #1922

## Problem

`ConnectorScanner` treats `(Connected, Connected)` as a no-op, only tracking state transitions. When a connector's mode list changes while remaining `Connected` (e.g., EDID becomes available after initial probe returned empty modes), no event is emitted.

This causes USB-C dock monitors with DP MST/alt-mode to get stuck in a permanent "connected but never activated" state when the kernel reports the connector as `Connected` before EDID data is ready.

## Fix

Add a new `Changed` event variant to both `ConnectorScanEvent` and `DrmScanEvent`. When a connector stays `Connected` but its mode list differs from the previous scan, emit `Changed` instead of silently ignoring it:

```rust
// ConnectorScanEvent
pub enum ConnectorScanEvent {
    Connected(connector::Info),
    Disconnected(connector::Info),
    Changed(connector::Info),  // NEW: mode list changed while staying connected
}

// DrmScanEvent
pub enum DrmScanEvent {
    Connected { connector, crtc },
    Disconnected { connector, crtc },
    Changed { connector, crtc },  // NEW
}
```

The scanner logic:
```rust
(State::Connected, State::Connected) => {
    if old.modes() != conn.modes() {
        changed.push(conn);
    }
}
```

This is better than re-emitting `Connected` (original approach) because:
- Compositors get explicit control over how to handle mode-list updates vs initial connections
- No duplicate `Connected` semantics
- Covers #1859 (video switchbox) naturally — input change = mode list change on an already-connected connector

## Context

Discovered while debugging intermittent monitor activation failures on a ThinkPad P16 Gen 3 with USB-C Dock Gen 2 (DP alt-mode + MST). Verified against kernel DRM source (`drm_probe_helper.c` lines 598-653) — `drm_helper_probe_detect()` sets connector status *before* EDID/mode enumeration, so userspace can observe `Connected` with empty modes.

Related:
- Smithay #1859 — same root cause with video switchbox
- niri #869, #2691 — downstream symptoms
- [niri PR #3409](https://github.com/YaLTeR/niri/pull/3409) — compositor-level workaround (bounded rescan timer)

### DIFF
--- a/anvil/src/udev.rs
+++ b/anvil/src/udev.rs
@@ -1138,6 +1138,9 @@ impl AnvilState<UdevData> {
                 } => {
                     self.connector_disconnected(node, connector, crtc);
                 }
+                // The connector's mode list changed while it stayed connected (e.g. EDID
+                // arrived after the initial probe returned empty/fallback modes). Compositors
+                // should re-evaluate the output's mode selection and recreate the surface here.
                 DrmScanEvent::Changed { .. } => {}
                 _ => {}
             }
--- a/smithay-drm-extras/src/drm_scanner/connector_scanner.rs
+++ b/smithay-drm-extras/src/drm_scanner/connector_scanner.rs
@@ -51,11 +51,11 @@ impl ConnectorScanner {
                     (State::Connected, State::Disconnected) => removed.push(conn),
                     (State::Disconnected | State::Unknown, State::Connected) => added.push(conn),
                     //
-                    // Re-emit Connected when the mode list changes while
-                    // staying connected.  This covers the EDID race where
-                    // the kernel initially reports a connector as Connected
-                    // with an empty mode list (EDID not yet read) and a
-                    // later rescan finds the modes populated.
+                    // Emit Changed when the mode list changes while staying
+                    // connected. This covers the EDID race where the kernel
+                    // initially reports a connector as Connected with an
+                    // empty or fallback mode list (EDID not yet read) and a
+                    // later rescan finds the real modes populated.
                     (State::Connected, State::Connected) => {
                         if old.modes() != conn.modes() {
                             changed.push(conn);
@@ -124,15 +124,12 @@ impl IntoIterator for ConnectorScanResult {
     type IntoIter = std::vec::IntoIter<ConnectorScanEvent>;
 
     fn into_iter(self) -> Self::IntoIter {
-        let mut events =
-            Vec::with_capacity(self.disconnected.len() + self.connected.len() + self.changed.len());
-        events.extend(
-            self.disconnected
-                .into_iter()
-                .map(ConnectorScanEvent::Disconnected),
-        );
-        events.extend(self.connected.into_iter().map(ConnectorScanEvent::Connected));
-        events.extend(self.changed.into_iter().map(ConnectorScanEvent::Changed));
-        events.into_iter()
+        self.disconnected
+            .into_iter()
+            .map(ConnectorScanEvent::Disconnected)
+            .chain(self.connected.into_iter().map(ConnectorScanEvent::Connected))
+            .chain(self.changed.into_iter().map(ConnectorScanEvent::Changed))
+            .collect::<Vec<_>>()
+            .into_iter()
     }
 }
--- a/smithay-drm-extras/src/drm_scanner.rs
+++ b/smithay-drm-extras/src/drm_scanner.rs
@@ -233,11 +233,12 @@ impl IntoIterator for DrmScanResult {
     type IntoIter = std::vec::IntoIter<DrmScanEvent>;
 
     fn into_iter(self) -> Self::IntoIter {
-        let mut events =
-            Vec::with_capacity(self.disconnected.len() + self.connected.len() + self.changed.len());
-        events.extend(self.disconnected.into_iter().map(DrmScanEvent::disconnected));
-        events.extend(self.connected.into_iter().map(DrmScanEvent::connected));
-        events.extend(self.changed.into_iter().map(DrmScanEvent::changed));
-        events.into_iter()
+        self.disconnected
+            .into_iter()
+            .map(DrmScanEvent::disconnected)
+            .chain(self.connected.into_iter().map(DrmScanEvent::connected))
+            .chain(self.changed.into_iter().map(DrmScanEvent::changed))
+            .collect::<Vec<_>>()
+            .into_iter()
     }
 }


PolyMeilex:
Hi!

1. Have you debugged this to check if what LLM states about lazy mode loading aligns with reality?
2. If so, wouldn't it be better to not emit the first connected event (the one that is not yet ready), rather than send duplicated connected event?
3. Or if it turns out that the modes can change at any time and not only on startup, perhaps we need a new Changed/Updated event?

coleleavitt:
Hi! Thanks for the thoughtful review.

**1. Verification:**

The initial investigation was LLM-assisted, but I've verified the mechanism against kernel DRM source and have a hardware reproducer (ThinkPad P16 Gen 3 + USB-C Dock Gen 2, DP alt-mode + MST hub).

The relevant kernel codepath is `drm_helper_probe_single_connector_modes()` in [`drm_probe_helper.c`](https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/drm_probe_helper.c):

- **Lines 598-606**: `drm_helper_probe_detect()` sets `connector->status` to `connected` *before* EDID/mode enumeration
- **Line 649**: `drm_helper_probe_get_modes()` reads EDID *after* status — can return 0 for DP MST when link training is still in-progress
- **Lines 651-653**: Kernel falls back to `drm_add_modes_noedid(1024x768)` when `count == 0 && connected`

So what smithay sees via `DRM_IOCTL_MODE_GETCONNECTOR`: connector reports `Connected` with either zero modes or just the 1024x768 fallback. A subsequent probe (triggered by kernel hotplug event as EDID arrives) returns the real mode list. With the current `(Connected, Connected) => {}` arm, that second probe is silently ignored.

On my hardware, this manifests as intermittent monitor activation failure — niri's `pick_mode()` returns `None` on the first probe, logs `warn!("error connecting connector: no mode")`, and never retries because smithay never re-emits the event.

**2. Suppressing the first empty-modes Connected:**

I considered this, but `modes().is_empty()` on a `Connected` connector isn't a reliable "not ready yet" signal:

- The kernel already provides 1024x768 fallbacks for no-EDID connectors (lines 651-653 above), so `modes()` may not actually be empty — just wrong
- Suppressing would break legitimate no-EDID connectors (VGA without EDID, some KVM switches)
- It would also break the switchbox case from #1859, where suppressing the initial Connected means the output never activates at all

**3. New `Changed` event — I think this is the right approach.**

The current PR overloads `Connected` for something semantically different. A new variant like `ConnectorScanEvent::Changed(connector::Info)` would:

- Give compositors explicit control over how to handle mode-list updates vs initial connections
- Cover #1859 naturally (switchbox input change = mode list change on an already-connected connector)
- Avoid the duplicate `Connected` semantics concern you raised

Happy to revise the PR with a `Changed` variant. Should it fire on *any* `connector::Info` change (property changes, encoder reassignment, etc.), or specifically just mode-list changes?

smithay-drm-extras/src/drm_scanner/connector_scanner.rs
```diff
@@ -52,7 +51,16 @@ impl ConnectorScanner {
                     (State::Connected, State::Disconnected) => removed.push(conn),
                     (State::Disconnected | State::Unknown, State::Connected) => added.push(conn),
                     //
-                    (State::Connected, State::Connected) => {}
+                    // Re-emit Connected when the mode list changes while
```
PolyMeilex: We are no longer re-emitting the connected event.

anvil/src/udev.rs
```diff
@@ -1138,6 +1138,7 @@ impl AnvilState<UdevData> {
                 } => {
                     self.connector_disconnected(node, connector, crtc);
                 }
+                DrmScanEvent::Changed { .. } => {}
```
PolyMeilex, line 1144: A lot of people use anvil as an example, so it would be nice to place a comment here, describing that one should probably handle the case of modes appearing after the connect.

smithay-drm-extras/src/drm_scanner/connector_scanner.rs
coleleavitt: Fixed — updated the comment to say "Emit Changed" in f08f79e.

anvil/src/udev.rs
coleleavitt: Added a comment explaining compositors should re-evaluate mode selection and recreate the surface. f08f79e.

coleleavitt:
@PolyMeilex

PR #1924, opened 2026-02-08
wayland: document destruction order

cmeissl:
when a client exits without properly destroying
resources the destroy callbacks will run in
undefined order.
this can happen in case the client does not wait
for resource destruction, exits without explicitly destroying resources or is disconnected by a
protocol error

for example see: https://gitlab.freedesktop.org/wayland/wayland/-/blob/e12bbe496941e546165055f9520745b7ba4366ce/src/wayland-server.c#L310

### DIFF
--- a/src/wayland/compositor/mod.rs
+++ b/src/wayland/compositor/mod.rs
@@ -599,6 +599,10 @@ pub trait CompositorHandler {
     /// The surface was destroyed.
     ///
     /// This allows the compositor to clean up any uses of the surface.
+    ///
+    /// Note: Destruction might happen explicitly by the client, or implicitly
+    /// when the client quits. In case of implicit destruction the order the
+    /// callbacks are called in is undefined.
     fn destroyed(&mut self, _surface: &WlSurface) {}
 }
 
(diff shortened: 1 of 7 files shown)


src/wayland/compositor/mod.rs
```diff
@@ -599,6 +599,10 @@ pub trait CompositorHandler {
     /// The surface was destroyed.
     ///
     /// This allows the compositor to clean up any uses of the surface.
+    ///
+    /// Note: Destruction might happen explicit by the client calling destroy or
+    /// implicit when the client is destroyed. In case of implicit destruction
+    /// the order is undefined and might not follow explicit protocol definitions.
```
YaLTeR:
```suggestion
    /// Note: Destruction might happen explicitly by the client calling destroy(), or
    /// implicitly when the client quits. In case of implicit destruction,
    /// the order the callbacks are called in is undefined and might not follow the logical object creation order.
```

cmeissl: `might not follow the logical object creation order.` I would prefer to express that protocol in-variants might get violated. There is no guarantee of order, but libwayland will destory the resources in order of creation, which is exactly the issue in niri (the surface is created and destroyed before the toplevel)

cmeissl: Maybe this is good enough:

```
    /// Note: Destruction might happen explicitly by the client, or implicitly
    /// when the client quits. In case of implicit destruction the order the
    /// callbacks are called in is undefined.
```

cmeissl: done

PR #1925, opened 2026-02-08
fix: release buffer reference after image copy capture completes

mayakwd:
`FrameInner` retains a reference to the client's `WlBuffer` after `Frame::success()` and `Frame::fail()` complete. 

Over time, with clients that rapidly create and destroy capture sessions, this causes thousands of leaked memfd file descriptors and tens of gigabytes of unreclaimable shared memory.

**Reproduction**
From https://github.com/pop-os/cosmic-comp/issues/2073:

```bash
$ ls -la /proc/$(pgrep cosmic-comp)/fd/ | grep 'memfd.*deleted' | wc -l
7809

$ cat /proc/$(pgrep cosmic-comp)/smaps_rollup | grep Pss_Shmem
Pss_Shmem: 16882370 kB    # 16.1 GB
```

After the fix: memfd count drops to 0.

### DIFF
--- a/src/wayland/image_copy_capture/mod.rs
+++ b/src/wayland/image_copy_capture/mod.rs
@@ -610,6 +610,7 @@ impl FrameInner {
             return;
         }
         self.failed = Some(reason);
+        self.buffer.take();
         if self.capture_requested {
             frame.failed(reason);
         }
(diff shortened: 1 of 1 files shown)


ids1024:
Releasing the reference to the `WlBuffer` here sounds fine, but it looks like the root issue here is `FrameInner` being leaked: https://github.com/Smithay/smithay/pull/1928.

mayakwd:
Got it, closing the pull request, since proper fix is already in place.

PR #1927, opened 2026-02-09
fix: avoid timer churn in idle_notify on every input event

mayakwd:
Fixes: [pop-os/cosmic-comp#2062](https://github.com/pop-os/cosmic-comp/issues/2062) and potentially many other issues

`notify_activity()` removes and re-creates a calloop Timer event source on every input event. With a 1000 Hz mouse, this means ~1000 remove+insert cycles per second.

calloop's `TimerWheel::cancel()` does not remove entries from its BinaryHeap — it only nullifies the token, leaving half-dead entries to live indefinitely in the heap. Each `cancel()` performs a linear scan of the growing heap, causing degradation over time.

After a notable time of use, the heap grows to millions of entries, and `Timer::unregister` consumes 93%+ of CPU time: https://github.com/pop-os/cosmic-comp/issues/2062#issuecomment-3869127469

**Fix proposed in this PR**
- Store `last_activity` as an `AtomicU64` timestamp (milliseconds since a process-wide monotonic epoch), updated on every input event with a single atomic store — no locks, no timer operations.
- The timer self-reschedules via `TimeoutAction::ToDuration(remaining)` when activity is detected, avoiding heap churn.
- Timer removal/insertion now happens only on rare events: notification creation, resume from idle, and inhibitor state changes.
- Cancel the timer in the destroyed handler to prevent callbacks on dead notification resources.

### DIFF
--- a/src/wayland/idle_notify/mod.rs
+++ b/src/wayland/idle_notify/mod.rs
@@ -42,11 +42,16 @@
 use std::{
     collections::HashMap,
     sync::{
-        atomic::{self, AtomicBool},
-        Mutex,
+        atomic::{self, AtomicBool, AtomicU64},
+        Mutex, OnceLock,
     },
-    time::Duration,
+    time::{Duration, Instant},
 };
+
+fn elapsed_millis() -> u64 {
+    static EPOCH: OnceLock<Instant> = OnceLock::new();
+    EPOCH.get_or_init(Instant::now).elapsed().as_millis() as u64
+}
 
 use calloop::{timer::TimeoutAction, LoopHandle, RegistrationToken};
 use wayland_protocols::ext::idle_notify::v1::server::{
@@ -74,6 +79,7 @@ pub struct IdleNotificationUserData {
     is_idle: AtomicBool,
     timeout: Duration,
     timer_token: Mutex<Option<RegistrationToken>>,
+    last_activity: AtomicU64,
 
     /// If listener was created with `get_input_idle_notification`
     ignore_inhibitor: bool,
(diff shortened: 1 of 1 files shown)


YaLTeR:
Also have a bunch of issue reports about this in niri

cmeissl:
Thanks! While it is a good idea to optimize this in smithay I also want to take a look (again) at the calloop part.
I expect more use-cases to suffer from this. Since rust 1.70 `BinaryHeap` offers a `retain` function that could be used to clean-up all cancelled timers. The last time I had a look at the performance part of this my PR to use `retain` was rejected because it would bump the MSRV. But now that most distros support rust >= 1.70 it might be worth a try again.

I want to compare both options, but definitely think it is a good idea to also optimize the pattern in smithay.

cmeissl:
@mayakwd I issued https://github.com/Smithay/calloop/pull/238. Are you able to patch cosmic-comp to use the linked PR and give it a try?

mayakwd:
@cmeissl sure, I'll give it a try in the afternoon!

mayakwd:
@cmeissl, I've checked cosmic-comp with the https://github.com/Smithay/calloop/pull/238.
The leak is totally gone. It seems there is no need for my PR, since it addresses only the symptoms of `idle_notification`, and your approach is more sophisticated. :-)

cmeissl:
> @cmeissl, I've checked cosmic-comp with the [Smithay/calloop#238](https://github.com/Smithay/calloop/pull/238). The leak is totally gone. It seems there is no need for my PR, since it addresses only the symptoms of `idle_notification`, and your approach is more sophisticated. :-)

Thanks for testing! FYI the latest calloop release already includes the fix, so it should be enough to bump calloop in cosmic-comp.

Issue #1562, opened 2024-10-15
Opening and closing "kitty" terminal causes VRAM usage to permanently increase


GallowsDove:
When you open a new "kitty" window, nothing special happens (as you would expect), however when you close the window VRAM usage increases by varying amount, depending on the size of the window. 

I've noticed this issue when looking at [another issue](https://github.com/NVIDIA/egl-wayland/issues/126#issuecomment-2379945259) with increasing VRAM caused by NVIDIA.

The issue could be unrelated, since someone mentioned something similar happening on AMD, however since I don't have an AMD card, I can't test it.

Tested with (positively):
- Niri
- Cosmic

Tested and not impacted:
- River
- Hyprland
- Sway

GallowsDove:
possibly related issue: https://github.com/hyprwm/Hyprland/issues/357

TheCodedProf:
Can confirm the same issue with an AMD card on Niri (have not tested elsewhere atm, but will try to do so when I'm home).

GallowsDove:
Edit: Does not seem to happen in anvil, but still find it weird that it happens both in cosmic and niri. Should I just report it to the respective compositors?

n1flh31mur:
Have the same with nvidia latest drivers with Arch. Hyprland starts slowly increase VRAM usage up to 1.5GB.

**Not using** kitty terminal, using alacritty.

vfosterm:
I think this is a general issue with the NVIDIA drivers when explicit sync is enabled. I'm using Hyprland and it definitely repros there when enabling explicit sync.

Limiting the free buffer pool for the compositor on the nvidia side is a functioning workaround for me. Haven't noticed any side effects and I'm able to use explicit sync again.

Details here:
https://github.com/NVIDIA/egl-wayland/issues/126#issuecomment-2379945259

GallowsDove:
> I think this is a general issue with the NVIDIA drivers when explicit sync is enabled. I'm using Hyprland and it definitely repros there when enabling explicit sync.
> 
> Limiting the free buffer pool for the compositor on the nvidia side is a functioning workaround for me. Haven't noticed any side effects and I'm able to use explicit sync again.
> 
> Details here:
> https://github.com/NVIDIA/egl-wayland/issues/126#issuecomment-2379945259

Like I said, that is  probably a completly unrelated issue. Yes, NVIDIA as such has issues, however the VRAM leak happens for me even with the workaround active. And it was also reproduced on AMD gpu.

[...]

PR #1929, opened 2026-02-11
Add WeakWindow for preventing VRAM leaks from stale handles

MartinKavik:
Add `WeakWindow` and `Window::downgrade` to help compositors fix and prevent issues like *Closing windows causes VRAM memory leak* (https://github.com/Smithay/smithay/issues/1562)

More details in issue linked above and in https://github.com/pop-os/cosmic-comp/issues/2084

tl;dr:
`WeakWindow` holds a weak reference to the underlying window resources, avoiding the strong `Arc` reference that keeps GPU buffers alive after a window is closed. Call `upgrade()` to get back an `Option<Window>`.

### DIFF
--- a/src/desktop/wayland/window.rs
+++ b/src/desktop/wayland/window.rs
@@ -16,7 +16,7 @@ use std::{
     hash::{Hash, Hasher},
     sync::{
         atomic::{AtomicU8, Ordering},
-        Arc, Mutex,
+        Arc, Mutex, Weak,
     },
     time::Duration,
 };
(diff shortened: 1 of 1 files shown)


nickjj:
In the linked issue, in reference to this PR, @cmeissl mentioned:

> So imo the best option is you open a PR with your proposed but include this description so that possible alternatives can be discussed.

Given this leak existed and has been reported for a ~year in niri / cosmic, I thought it might be worth commenting here to raise this while the iron is hot around this issue, even if the PR is already merged.

This code is above my pay grade but here's a couple of (naive) questions to prompt a discussion:

- Is it addressing the root cause in common cases as well as edge cases?
- What would alternative solutions look like?
- What part of this code addresses the issue, how does it protect against leaks in compositors like niri?
- Does it protect against leaks if an app is natively run through Wayland as well as XWayland?
- Do compositor implementers like niri need to think about this and add additional code to protect against leaks?
- Is this a backwards compatible change?
- Are there any opportunities to write tests to prevent regressions?

PR #1933, opened 2026-02-14
gles: clamp damage regions before glTexSubImage2D uploads

coleleavitt:
## Summary

- Clamp damage rectangles to texture bounds before passing to `glTexSubImage2D` in both `import_shm_buffer` and `update_memory`, preventing `GL_INVALID_VALUE` errors on strict GL implementations (NVIDIA 590.48.01+)

Fixes #533

## Problem

Two `glTexSubImage2D` call sites in `src/backend/renderer/gles/mod.rs` pass damage regions without clamping to texture dimensions. The GL spec requires `xoffset + width <= texture_width` and `yoffset + height <= texture_height`, otherwise `GL_INVALID_VALUE` is generated.

Mesa (Intel/AMD) silently tolerates this, but NVIDIA's open kernel module driver enforces strict per-spec validation:

```
[GL] GL_INVALID_VALUE error generated. Size and/or offset out of range.
```

The `update_memory` path (`ImportMem` trait) is the primary concern — callers like `MemoryRenderBufferRenderElement` (`element/memory.rs:326`) and `TextureRenderBuffer::update_from_memory` (`element/texture.rs:355`) pass regions without clamping to texture bounds.

The `import_shm_buffer` path is already clamped upstream at `utils/wayland.rs:209`, so the clamping there is defense-in-depth.

## Fix

For both affected call sites, intersect each damage region with the texture bounds (`Rectangle::from_size(...)`.intersection(...)`) before the `glTexSubImage2D` call:

1. **`import_shm_buffer`** (line ~957): Intersect each damage rect with `tex_bounds` inside the loop, skip if the intersection is empty (defense-in-depth)
2. **`update_memory`** (line ~1121): Intersect `region` with `tex_bounds` before the data length check, early-return `Ok(())` if empty (necessary fix — no upstream clamping)

`GL_UNPACK_ROW_LENGTH` / `GL_UNPACK_SKIP_*` parameters remain correct because the pixel data pointer and row stride are unchanged — only the upload sub-rectangle shrinks.

## Testing

- Verified on NVIDIA RTX (open kernel module 590.48.01), niri 25.11
- GL errors eliminated, rendering correct
- No regressions on normal resize/damage paths

## Related

- #533 — original damage clamping issue
- #539 — fixed damage accumulation, but not `TexSubImage2D` clamping

### DIFF
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -954,7 +954,15 @@ impl ImportMemWl for GlesRenderer {
                         ptr.offset(offset as isize) as *const _,
                     );
                 } else {
+                    let tex_bounds: Rectangle<i32, BufferCoord> =
+                        Rectangle::from_size((width, height).into());
                     for region in damage.iter() {
+                        let Some(region) = region.intersection(tex_bounds) else {
+                            continue;
+                        };
+                        if region.is_empty() {
+                            continue;
+                        }
                         trace!("Uploading partial shm texture");
                         self.gl.PixelStorei(ffi::UNPACK_SKIP_PIXELS, region.loc.x);
                         self.gl.PixelStorei(ffi::UNPACK_SKIP_ROWS, region.loc.y);
@@ -1112,6 +1120,14 @@ impl ImportMem for GlesRenderer {
         }
         let (read_format, type_) = gl_read_for_internal(texture.0.format.expect("We check that before"))
             .ok_or(GlesError::UnknownPixelFormat)?;
+
+        // Clamp the update region to the texture dimensions to avoid
+        // GL_INVALID_VALUE from glTexSubImage2D on strict drivers (NVIDIA).
+        let tex_bounds = Rectangle::from_size(texture.0.size);
+        let region = match region.intersection(tex_bounds) {
+            Some(r) if !r.is_empty() => r,
+            _ => return Ok(()),
+        };
 
         if data.len()
             < (region.size.w * region.size.h) as usize


cmeissl:
Can you please explain how this resolves #1922?

The damage rects should be clamped to the buffer size already before the import functions are called. Did you check why this is not the case? It would be nice to fix this for all renderer and not only gles.

> The same file already has correct clamping at six other sites using constrain/clamp.

What does this refer to?

coleleavitt:
> Can you please explain how this resolves #1922?

You're right — `Fixes #1922` is incorrect. That's the `ConnectorScanner` issue I filed separately. This PR addresses the `glTexSubImage2D` clamping originally reported in #533. I'll update the description.

> The damage rects should be clamped to the buffer size already before the import functions are called. Did you check why this is not the case? It would be nice to fix this for all renderer and not only gles.

I traced the call paths:

**`import_shm_buffer`**: You're right that damage is already clamped. `RendererSurfaceState::commit()` at `utils/wayland.rs:209` does `.intersection(Rectangle::from_size(buffer_dimensions))` before storing damage, so by the time it flows through `import_surface()` → `import_buffer()` → `import_shm_buffer()`, it should be within bounds. The clamping in this hunk is defense-in-depth only.

**`update_memory`**: This path has **no upstream clamping**. Two callers pass unchecked regions:
- `MemoryRenderBufferRenderElement` (`element/memory.rs:326`) — `buffer_damage` comes from `damage_bag.damage_since()` which returns accumulated user-supplied damage from `.add()` calls with no intersection against buffer bounds
- `TextureRenderBuffer::update_from_memory` (`element/texture.rs:355`) — takes `region: Rectangle<i32, Buffer>` directly from the caller, passed straight to `update_memory()` with no validation

So the `update_memory` fix is necessary — the `import_shm_buffer` fix is belt-and-suspenders.

I can split this into two commits if you prefer, or drop the `import_shm_buffer` hunk entirely and trust the upstream clamp. Alternatively, if you'd rather clamp in the callers (memory.rs/texture.rs) instead of the renderer, I'm happy to move it there. Let me know what you prefer.

> What does this refer to?

That was poorly worded. The five `constrain`/`clamp` sites (lines 2564, 2581, 2720, 3044, 3061) are in the frame rendering paths (`draw_solid`, `render_texture_from_to`, `render_pixel_shader_to`) — they clamp **rendering damage to destination bounds** in Physical coordinates, not import damage to texture bounds in BufferCoord. Different concern. I'll correct the description.

PR #1939, opened 2026-02-18
Clean up selection devices in `destroyed()` callback

midgethoen:
Ive been troubleshooting an [issue with lag in niri](https://github.com/niri-wm/niri/issues/2430#issuecomment-3919569581) when selecting text in the browser. it is noticable after hours/days and becomes progessively worse. I believe i traced it down to smithay not cleaning up devices of clients that do not call Release before exiting. in one of my cases smithay seemed to iterate over ~700k dead devices causing very noticable lag.

i added some debug statements and vibe-coded a 'test_polite_destroy' client to demonstrate the 'impolite' devices (wl-copy) not decrementing:
image
(device 1&2 appear on init and devices 3-6 seem to belong to kitty-terminal)

then i implemented the destroyed() methods of all devices and repeated the test:
image
(note that retain_devices is called twice, since for polite clients both release and destroyed are invoked. maybe only imlementing cleanup in destroyed would suffice?)

note that im new to rust and this project. im intentionally creating this PR early, i will be testing some more. any feedback is welcome.

### DIFF
--- a/src/wayland/selection/data_device/device.rs
+++ b/src/wayland/selection/data_device/device.rs
@@ -196,4 +196,20 @@ where
             _ => unreachable!(),
         }
     }
+
+    fn destroyed(
+        _state: &mut D,
+        _client: wayland_server::backend::ClientId,
+        resource: &WlDataDevice,
+        data: &DataDeviceUserData,
+    ) {
+        if let Some(seat) = Seat::<D>::from_resource(&data.wl_seat) {
+            if let Some(seat_data) = seat.user_data().get::<RefCell<SeatData<D::SelectionUserData>>>() {
+                seat_data.borrow_mut().retain_devices(|ndd| match ndd {
+                    SelectionDevice::DataDevice(ndd) => ndd != resource,
+                    _ => true,
+                });
+            }
+        }
+    }
 }
--- a/src/wayland/selection/ext_data_control/device.rs
+++ b/src/wayland/selection/ext_data_control/device.rs
@@ -130,4 +130,20 @@ where
             _ => unreachable!(),
         }
     }
+
+    fn destroyed(
+        _state: &mut D,
+        _client: wayland_server::backend::ClientId,
+        resource: &ExtDataControlDeviceV1,
+        data: &ExtDataControlDeviceUserData,
+    ) {
+        if let Some(seat) = Seat::<D>::from_resource(&data.wl_seat) {
+            if let Some(seat_data) = seat.user_data().get::<RefCell<SeatData<D::SelectionUserData>>>() {
+                seat_data.borrow_mut().retain_devices(|ndd| match ndd {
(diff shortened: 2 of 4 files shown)


src/wayland/selection/seat_data.rs
```diff
@@ -25,11 +26,13 @@ impl<U: Clone + Send + Sync + 'static> SeatData<U> {
 
     pub fn retain_devices<F: FnMut(&SelectionDevice) -> bool>(&mut self, retain: F) {
         self.known_devices.retain(retain);
+        debug!(count = self.known_devices.len(), "devices retained");
```
Drakulix: Could you either drop these or change them to `trace!`. This is too verbose even for the debug-level.

midgethoen: I've dropped them.

midgethoen:
I've finished up by removing the debug-logs, removing the test_client since it was generated and squasing the commits.

Drakulix:
Thanks again!

eevmanu:
Hi!

I don't want to open a separate issue to avoid adding noise, so I'm leaving this here in case it's useful to someone.

After this PR landed in `cosmic-comp` (via the Smithay bump from `da42569` to `e84a4ca`, check diff [here](https://github.com/smithay/smithay/compare/da42569...e84a4ca.diff)) and I updated it on my local machine (from `1dc9c53` to `3e84968`, check diff [here](https://github.com/pop-os/cosmic-comp/compare/1dc9c53...3e84968.diff)), I experienced intermittent clipboard failures with `cosmic-screenshot`, screenshots would not appear in the clipboard despite the whole UI to take it and the dropdown option list is in `Save to Clipboard` option and I press enter key. The issue happened twice today and then resolved on its own without me changing anything.

**Versions involved:**
- `cosmic-comp`: upgraded from `0.1~1773158788~24.04~1dc9c53` -> `0.1~1775571292~24.04~3e84968`
- `cosmic-screenshot`: `0.1.0~1773071219~24.04~20a0371` (recently got updated from `0.1.0~1772469534~24.04~d4dc8d8`, check diff [here](https://github.com/pop-os/cosmic-screenshot/compare/d4dc8d8...20a0371.diff))
- OS: Pop!_OS 24.04

**My guess (could be wrong tbh 😅):** `cosmic-screenshot` is a short-lived client, it sets the clipboard selection and exits quickly. With the new `destroyed()` cleanup, there might be a small timing window where the device is removed before a clipboard manager finishes reading the data. This would explain the intermittent nature (race condition).

I haven't been able to reproduce it reliably yet, so I completely understand if this isn't actionable. I'm happy to do more debugging on my side if needed. Also, if you think it'd be more appropriate to open a separate issue (here or in [cosmic-screenshot](https://github.com/pop-os/cosmic-screenshot)), just let me know.

Just wanted to leave this here in case other COSMIC desktop users run into something similar.

PR #1940, opened 2026-02-19
Improve API ergonomics

chrisduerr:
This makes several minor changes which improve the ergonomics around implementing the ext_image_copy_capture-v1 protocol.

### DIFF
--- a/src/wayland/foreign_toplevel_list/mod.rs
+++ b/src/wayland/foreign_toplevel_list/mod.rs
@@ -130,16 +130,11 @@ pub struct ForeignToplevelHandle {
 }
 
 impl ForeignToplevelHandle {
-    fn new(
-        identifier: String,
-        title: String,
-        app_id: String,
-        instances: Vec<Weak<ExtForeignToplevelHandleV1>>,
-    ) -> Self {
+    fn new(title: String, app_id: String, instances: Vec<Weak<ExtForeignToplevelHandleV1>>) -> Self {
         Self {
             inner: Arc::new((
                 Mutex::new(ForeignToplevelHandleInner {
-                    identifier,
+                    identifier: Alphanumeric.sample_string(&mut rand::rng(), 32),
                     title,
                     app_id,
                     instances,
@@ -258,6 +253,12 @@ impl ForeignToplevelHandle {
         self.inner.0.lock().unwrap().closed
     }
 
+    /// Check if two toplevel handles belong to the same toplevel.
+    pub fn matches(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.inner, &other.inner)
+            || self.inner.0.lock().unwrap().identifier == other.inner.0.lock().unwrap().identifier
+    }
+
     fn init_new_instance(&self, toplevel: ExtForeignToplevelHandleV1) {
         debug_assert!(
             !self.is_closed(),
@@ -328,7 +329,6 @@ impl ForeignToplevelListState {
         app_id: impl Into<String>,
     ) -> ForeignToplevelHandle {
         let handle = ForeignToplevelHandle::new(
-            Alphanumeric.sample_string(&mut rand::rng(), 32),
             title.into(),
             app_id.into(),
             Vec::with_capacity(self.list_instances.len()),


src/wayland/foreign_toplevel_list/mod.rs
```diff
@@ -130,16 +130,11 @@ pub struct ForeignToplevelHandle {
...
-        title: String,
-        app_id: String,
-        instances: Vec<Weak<ExtForeignToplevelHandleV1>>,
-    ) -> Self {
+    fn new(title: String, app_id: String, instances: Vec<Weak<ExtForeignToplevelHandleV1>>) -> Self {
         Self {
             inner: Arc::new((
                 Mutex::new(ForeignToplevelHandleInner {
-                    identifier,
+                    identifier: Alphanumeric.sample_string(&mut rand::rng(), 32),
```
chrisduerr, line 137: Happy to revert this if it is not desired. I just moved this around while I was trying out some changes and ultimately didn't move it back because I found that this would make abusing this method (using it incorrectly) more difficult.

It's not a public API either way, so it doesn't matter that much I guess.

src/wayland/foreign_toplevel_list/mod.rs
```diff
@@ -258,6 +253,11 @@ impl ForeignToplevelHandle {
         self.inner.0.lock().unwrap().closed
     }
 
+    /// Check if two toplevel handles are for the same client.
+    pub fn matches(&self, other: &Self) -> bool {
+        self.inner.0.lock().unwrap().identifier == other.inner.0.lock().unwrap().identifier
+    }
```
chrisduerr: Originally I wanted to change `identifier` to use an `Arc`, but that would have required changes to the public API of the `identifier` function which might break downstream and annoy people.

Ultimately I don't think people really care about the identifier, but rather care about checking if the identifier of two distinct handles matches. So I decided to just add a function that compares them by identifier, rather than a deep comparison.

src/utils/serial.rs
```diff
@@ -7,18 +7,9 @@ pub static SERIAL_COUNTER: SerialCounter = SerialCounter::new();
...
 pub struct Serial(pub(crate) u32);
 
-impl PartialEq for Serial {
-    #[inline]
-    fn eq(&self, other: &Self) -> bool {
-        self.0 == other.0
-    }
-}
-
-impl Eq for Serial {}
```
chrisduerr, line 20: Clippy didn't like hash with this manual impl of `PartialEq`, but I don't see why this is a manual impl to begin with?

cmeissl: Yeah, seems to be fine to have this derived. I also checked the history for some reason, but it just has been like this since before getting moved from wayland to utils.

src/wayland/foreign_toplevel_list/mod.rs
cmeissl: imo it is fine to move this into the handle

src/wayland/foreign_toplevel_list/mod.rs
```diff
@@ -258,6 +253,11 @@ impl ForeignToplevelHandle {
         self.inner.0.lock().unwrap().closed
     }
 
+    /// Check if two toplevel handles are for the same client.
```
cmeissl:
```suggestion
    /// Check if two toplevel handles belong to the same toplevel
```

not sure what client means here. imo this better reflects the spec, which states:

> A stable identifier for a toplevel
>
> This identifier is used to check if two or more toplevel handles belong to the same toplevel.

src/wayland/foreign_toplevel_list/mod.rs
cmeissl: Personally I would be fine with breaking the API, but if we can avoid it sure. This looks like a reasonable addition.

cmeissl (approved):
LGTM

Can you rebase the changes into 3 commits?
- serial stuff
- toplevel handle
- copy

chrisduerr:
@cmeissl Actually something I just realized is that this might deadlock if `self` and `other` are the same (not just pointing to the same toplevel). While it doesn't matter where I use it, I don't think this is acceptable for Smithy.

chrisduerr:
I changed it to use `RwLock` instead. While this is a **significant** change as far as design goes (so not just adding one random method), I think it's overall a net-benefit considering the amount of functions that do not require write access to the Mutex.

cmeissl:
> @cmeissl Actually something I just realized is that this might deadlock if `self` and `other` are the same (not just pointing to the same toplevel). While it doesn't matter where I use it, I don't think this is acceptable for Smithy.

right, how did I miss that.. sorry.

cmeissl:
> I changed it to use `RwLock` instead. While this is a **significant** change as far as design goes (so not just adding one random method), I think it's overall a net-benefit considering the amount of functions that do not require write access to the Mutex.

Hm, as this is internal change it would be fine. Though, iirc this could still deadlock if another thread tries to aquire write access between the first and second read.

chrisduerr:
> Hm, as this is internal change it would be fine. Though, iirc this could still deadlock if another thread tries to aquire write access between the first and second read.

You can't acquire write access while someone is holding a read lock.

**Edit:** To expand on this. Yes, the writer would be blocked, but while one read lock is held you can acquire any amount of read locks, so acquiring the first read lock would guarantee that the second read lock also could be acquired. As a result this operation should always complete and while the write lock trying to access in-between the reads would get blocked, it would not *deadlock*.

If you have a better solution in mind, I'm happy to change it. This just seemed like the simplest fix to me since it should largely improve on the existing code, without making any major changes.

cmeissl:
> > Hm, as this is internal change it would be fine. Though, iirc this could still deadlock if another thread tries to aquire write access between the first and second read.
> 
> You can't acquire write access while someone is holding a read lock.
> 
> **Edit:** To expand on this. Yes, the writer would be blocked, but while one read lock is held you can acquire any amount of read locks, so acquiring the first read lock would guarantee that the second read lock also could be acquired. As a result this operation should always complete and while the write lock trying to access in-between the reads would get blocked, it would not _deadlock_.
> 
> If you have a better solution in mind, I'm happy to change it. This just seemed like the simplest fix to me since it should largely improve on the existing code, without making any major changes.

That's not true entirely true, rust does not provide that guarantee afaict. The second call to read might deadlock when another thread tried to aquire a write lock in between.

Independent of the concurrency primitive in use it might work to do a simple arc pointer comparison to avoid the deadlock. If the arc matches it should be possible to short cut the actual identifier check and avoid any potential deadlock.

chrisduerr:
> That's not true entirely true, rust does not provide that guarantee afaict. The second call to read might deadlock when another thread tried to aquire a write lock in between.

~~That's fair. I believe this to be the case on Linux, but I'm not sure about other operating systems.~~

I was actually curious so I decided to run some tests. Turns out my assumption was completely made up and this code can be trivially proven to deadlock (or at least code that is pretty much identical in functionality).

> Independent of the concurrency primitive in use it might work to do a simple arc pointer comparison to avoid the deadlock. If the arc matches it should be possible to short cut the actual identifier check and avoid any potential deadlock.

Yeah that's actually the implementation I had originally, but thought the `RwLock` could improve efficiency. I'll switch back to that.

cmeissl:
> Yeah that's actually the implementation I had originally, but thought the `RwLock` could improve efficiency. I'll switch back to that.

Thanks! Small last nit, the commit message still includes RwLock.

chrisduerr:
Done!

PR #1942, opened 2026-02-21
OpenBSD: Fix remaining build issues

tobhe:
There are two remaining issues which I currently carry out of tree patches for to build on OpenBSD.

# OpenBSD doesn't support eventfd

 Since eventfd is entirely optional and smithay already has a supports_syncobj_eventfd runtime switch it is easy enough to conditionally disable.
Currently I use `#[cfg(not(target_os = "openbsd"))]` but maybe this would better be a feature?

# libEGL loading is too strict

The libEGL backend hardcodes a libEGL.so.1 library name which doesn't work on OpenBSD where the current base libEGL is `/usr/X11R6/lib/libEGL.so.2.0`. Dropping the .1 suffix seems like the appropriate fix since that should generally find the right version automatically on both platforms. Alternatively we could just conditionally hardcode either version.

### DIFF
--- a/src/backend/egl/ffi.rs
+++ b/src/backend/egl/ffi.rs
@@ -145,7 +145,7 @@ pub mod egl {
     use std::sync::{LazyLock, Once};
 
     pub static LIB: LazyLock<Library> =
-        LazyLock::new(|| unsafe { Library::new("libEGL.so.1") }.expect("Failed to load LibEGL"));
+        LazyLock::new(|| unsafe { Library::new("libEGL.so") }.expect("Failed to load LibEGL"));
 
     pub static LOAD: Once = Once::new();
     pub static DEBUG: Once = Once::new();
--- a/src/wayland/drm_syncobj/mod.rs
+++ b/src/wayland/drm_syncobj/mod.rs
@@ -66,6 +66,7 @@ pub use sync_point::*;
 
 /// Test if DRM device supports `syncobj_eventfd`.
 // Similar to test used in Mutter
+#[cfg(not(target_os = "openbsd"))]
 pub fn supports_syncobj_eventfd(device: &DrmDeviceFd) -> bool {
     // Pass device as placeholder for eventfd as well, since `drm_ffi` requires
     // a valid fd.
@@ -73,6 +74,11 @@ pub fn supports_syncobj_eventfd(device: &DrmDeviceFd) -> bool {
         Ok(_) => unreachable!(),
         Err(err) => err.kind() == std::io::ErrorKind::NotFound,
     }
+}
+
+#[cfg(target_os = "openbsd")]
+pub fn supports_syncobj_eventfd(device: &DrmDeviceFd) -> bool {
+    false
 }
 
 /// Handler trait for DRM syncobj protocol.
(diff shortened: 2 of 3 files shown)


Drakulix:
> Currently I use #[cfg(not(target_os = "openbsd"))] but maybe this would better be a feature?

`linux-drm-syncobj-v1` (and `linux-dmabuf-v1`) are clearly intended as being OS-specific given their name, so I wouldn't even object to changing the condition to `target_os = "linux"`. I assume openbsd has support for dmabuf's though, right?

tobhe:
> linux-drm-syncobj-v1 (and linux-dmabuf-v1) are clearly intended as being OS-specific given their name, so I wouldn't even object to changing the condition to target_os = "linux"

That might actually make more sense. Things might be a little more difficult since the BSDs all use versions of the Linux DRM stack so there is a lot of overlap.  

> I assume openbsd has support for dmabuf's though, right?

I don't think we do. Not sure about the other BSD's though.

EDIT: OpenBSD does even have drm_syncobj, just eventfd isn't supported https://github.com/openbsd/src/blob/master/sys/dev/pci/drm/drm_syncobj.c#L1529 (not sure if that is strictly required for the protocol)

Drakulix:
> > linux-drm-syncobj-v1 (and linux-dmabuf-v1) are clearly intended as being OS-specific given their name, so I wouldn't even object to changing the condition to target_os = "linux"
> 
> That might actually make more sense. Things might be a little more difficult since the BSDs all use versions of the Linux DRM stack so there is a lot of overlap.

Seems like similar things might be true for drm_syncobj. So perhaps a feature is the best solution here?

tobhe:
I took another close look at rustix and they explicitly list supported OSs, so a feature seems kind of wrong since at the end of the day you would still not work on platforms rustix doesn't define, see https://github.com/bytecodealliance/rustix/blob/main/src/event/mod.rs#L5-L10

Maintaining the same list would work but also seems a bit wrong since it would mean they have to be kept in sync.
In C this is where I would use `ifdef` to propagate the conditional from the dependency. is there a rust equivalent?

Drakulix:
> Maintaining the same list would work but also seems a bit wrong since it would mean they have to be kept in sync. In C this is where I would use `ifdef` to propagate the conditional from the dependency. is there a rust equivalent?

No this sadly doesn't exist.

So either this needs to be an additive feature (something like `backend_eventfd`), which compositors need to explicitly enable depending on their target platforms ... or ... we simply copy the list from rustix and add a comment noting that this should be updated when updating rustix.

src/backend/egl/ffi.rs
```diff
@@ -145,7 +145,7 @@ pub mod egl {
     use std::sync::{LazyLock, Once};
 
     pub static LIB: LazyLock<Library> =
-        LazyLock::new(|| unsafe { Library::new("libEGL.so.1") }.expect("Failed to load LibEGL"));
+        LazyLock::new(|| unsafe { Library::new("libEGL.so") }.expect("Failed to load LibEGL"));
```
bjorn3, line 148: On Debian `libEGL.so` is part of the `libegl-dev` package, which end users do not normally have installed. You did probably have to make the filename conditional on OpenBSD.

VlkrS: Could we resolve this like this?

https://github.com/VlkrS/smithay/commit/c3fd033e0d9029611af5d85fc6ea9b60fe797252

PR #1945, opened 2026-02-26
xdg-toplevel-icon: allow replacing previously-set icon sizes

### LINKED ISSUE #1944: XdgToplevelIconManager's perhaps needs a way to clear icon sizes
I'm happy to submit a PR for this, but wanted to start a discussion first.

`XdgToplevelIconManager` has `add_icon_size()`, but no way to remove or clear the current set of icons sizes.  I have two needs for runtime size changes:

1. The server-side decoration theme can change, and with that, the size of the icon we display in the decorations can change.
2. We have a "tabwin" window that comes up on Alt+Tab that lists all windows, and allows the user to cycle through them.  The tabwin displays the window icon along with the window title.  The size of the window icon depends on the size of the `wl_output` the tabwin is displayed on, as well as the number of windows being shown in the grid (more windows -> smaller icons so they all fit).  And a higher-resolution output will be able to hold more icons and so the sizes will be a bit different.  This is also dependent on GTK theme, as the icon sizing code takes into account window padding/borders/etc., and we have GTK theme/style properties for setting the icon size that themes can change.

Setting all the icon sizes on startup is fine as-is.  But if the SSD or GTK theme changes, we potentially want a new set of icon sizes.  So it would be useful to be able to clear out `XdgToplevelIconManager`'s list of icon sizes and start anew.

(This also raises a question about the `xdg_toplevel_icon` protocol in general: the [protocol docs](https://wayland.app/protocols/xdg-toplevel-icon-v1#xdg_toplevel_icon_manager_v1:event:icon_size) state that the `icon_size` event is sent after the client binds to the manager, followed by `done`.  It doesn't expressly forbid the compositor from sending a new set of icon sizes later, but given that this isn't explicitly allowed, I wouldn't be surprised if existing clients wouldn't know what to do if it were to do so.  At any rate, allowing me to clear out the icon list and provide a new set of sizes would at least allow _new_ clients to get the "correct" sizes.)

Questions:

1. Is it ok if I add a `clear_icon_sizes()` method to `XdgToplevelIconManager`?
2. On top of that (perhaps a separate PR), how do we feel about the idea of sending updated icon size lists to clients even after they've bound to the manager?  I guess this wouldn't really work with the API as-is, since `add_icon_size()` just takes one size, so there's no reasonable way for the manager to know when you've added all the new sizes so it can batch them and follow it with a `done` event.  We'd need something like `replace_icon_sizes(HashSet<i32>)`.

kelnos:
## Description

The icon sizes are replaced atomically with the newly-provided icons. Clients that bind to the interface in the future will get the new set of icons.  However, existing client binds will not be notified, as it's unclear if this is allowed by the protocol.

Closes #1944

## Checklist

### DIFF
--- a/src/wayland/xdg_toplevel_icon.rs
+++ b/src/wayland/xdg_toplevel_icon.rs
@@ -231,6 +231,17 @@ impl XdgToplevelIconManager {
         self.data.lock().unwrap().sizes.insert(size);
     }
 
+    /// Replaces previously-set icon sizes with those in `sizes`
+    ///
+    /// The new icon sizes will only be sent to newly-bound globals, not to existing binds.
+    ///
+    /// - `sizes` an iterator over `i32` icon size values
+    pub fn replace_icon_sizes(&mut self, sizes: impl IntoIterator<Item = i32>) {
+        let mut guard = self.data.lock().unwrap();
+        guard.sizes.clear();
+        guard.sizes.extend(sizes);
+    }
+
     /// Returns the [XdgToplevelIconManagerV1] global id.
     pub fn global(&self) -> GlobalId {
         self.global.clone()


src/wayland/xdg_toplevel_icon.rs
```diff
@@ -231,6 +231,15 @@ impl XdgToplevelIconManager {
         self.data.lock().unwrap().sizes.insert(size);
     }
 
+    /// Replaces previously-set icon sizes with those in `sizes`
```
cmeissl, line 234: It might be worth mentioning here that this will only apply to newly bound globals.
We have something similar in another protocol handler.

kelnos: Good call; updated

PR #1946, opened 2026-02-26
feat: sub-frames

Drakulix:
## Description

Alternative to https://github.com/Smithay/smithay/pull/1935.

<details>
  <summary>Old PR</summary>

This PR implements `Renderer` for the `GlesFrame`. This allows "sub"-frames to be created, which temporarily borrow the original `GlesFrame` to render to a different framebuffer, before restoring the previous framebuffer. Additionally this allows other traits such as `Import*` to be implemented for the `GlesFrame` easily.

There is a few things missing before this PR could be considered for merge:
- Various TODOs
  - Try to avoid some unnecessary `make_current`-calls. Though even an active `GlesFrame` is technically no guarantee, that there isn't another `GlesFrame` from a different context...
  - Make texture-filter and debug-flags functions/variables local to a specific frame and not for the context
- Create some `GlesRendererLike`-trait, so that `RenderElements`, such as the `TextureShaderElement` or the `PixelShaderElement` can be implemented generically across both `GlesRenderer`s and `GlesFrame`s.
- Allow subframes of `GlowFrame`s just the same.
- Investigate if we can (and want to) implement this for `PixmanRenderer`, `MultiRenderer` and `TestRenderer` as well.
- Make sure this is actually sound (there are a few raw-pointers in there, to allow indefinitely nested `GlesFrame`s even though we need to borrow the internal `GlesTarget` as well to be able to restore it.

Last but not least, there might be options to implement this in a more straight-forward manner:
- We could potentially borrow a `GlesRenderer` from a `GlesFrame` with some kind of `Guard`-struct that restores the `target` again on drop. This would require fewer `impl`-blocks and adjusted implementations.
- If we don't actually need sub-frames, but just `Import*` or `Offscreen` (as https://github.com/Smithay/smithay/pull/1935 suggests), we could also change those to only depend on `RendererSuper` and implement *that* for `Frame` instead of `Renderer`-itself.

Feedback is very much wanted. (Ping @YaLTeR, @cmeissl, @ids1024) 

</details>

EDIT: New Approach

A new `Frame`-trait (please bikeshed on the name) to allow borrowing the old renderer and then restore the current frame on drop.

Implemented for all renderers, but needs testing.

## Checklist

- [x] I agree to smithay's [Developer Certificate of Origin](https://github.com/Smithay/smithay/blob/master/DCO.md).

### DIFF
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -1717,7 +1717,7 @@ impl Offscreen<GlesRenderbuffer> for GlesRenderer {
     }
 }
 
-impl<'buffer> BlitFrame<GlesTarget<'buffer>> for GlesFrame<'_, 'buffer> {
+impl<'buffer> BlitFrame<GlesTarget<'buffer>> for GlesFrame<'_, '_> {
     fn blit_to(
         &mut self,
         to: &mut GlesTarget<'buffer>,


YaLTeR:
> * If we don't actually need sub-frames, but just `Import*` or `Offscreen`

A case just those don't cover is deferred offscreen rendering. Currently if you want to render some `RenderElement`s into an offscreen texture, you have to do it before the main framebuffer render. But it's desirable to render offscreens during the main framebuffer render in some cases:
- To avoid rendering offscreens that won't end up drawn (i.e. fully obscured). As I understand, ideally during render list construction you'd still damage track inside the offscreen to update its commit counter, but you wouldn't actually render the offscreen contents until the offscreen itself is drawn.
- To let a framebuffer effect composite into the offscreen, i.e. in capture_framebuffer() capture the framebuffer then draw it as a background inside the offscreen, then draw the offscreen itself.
You can't do this currently because in draw() you only have a GlesFrame with no way to switch to a different target (the offscreen texture), so you can't call RenderElement::draw() on the offscreen contents.

Subframes let you do this. Idk if they are the simplest way to do it. Maybe some way to switch the frame's target temporarily? Though I guess that also kind of goes against what a frame is supposed to represent.

Drakulix:
New approach! (Needs a bit of cleanup for CI, but compiles with standard feature sets.)

Drakulix:
Based on https://github.com/Smithay/smithay/pull/1816 because that made the whole mess of the new `MultiRenderer`-code somewhat easier to reason about.

This is ready for testing & review. :)

PR #1951, opened 2026-03-04
Fix: wl_data_offer accept when the last call isn't a known mime-type

ideasman42:
The Request::Accept handler checks whether the accepted mime type is in the source's offered list, setting accepted=false when it isn't.

This causes DnD drops to silently fail when the target client sends accept("text/plain") but the source only offers  "text/plain;charset=utf-8" - the last accept() sets accepted=false, and validated() returns false at drop time.

---

This breaks dropping *.jpg images into Blender on cosmic desktop (for example).

## Description
Fixes dropping with Blender (tested with 5.0).

## Checklist


ideasman42:
Closing as this is an area which I'm not confident enough with to be submitting code.

PR #1952, opened 2026-03-04
xkbcommon: allow for user-specifed xkb::ContextFlags when adding keyboard

dotpyfe:
## Description
When using smithay in settings that e.g. do not have standard include paths, adding a keyboard to a seat can result in a crash. To prevent this, allow users to specify their own `xkb` context flags when adding a keyboard.

A side effect of this change is the `new` functions are now dead code internally, but I figured it would be better to not potentially break other projects by removing them, and so allow pragmas have been added.

## Checklist

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -572,11 +572,18 @@ impl<D: SeatHandler + 'static> Seat<D> {
         repeat_delay: i32,
         repeat_rate: i32,
     ) -> Result<KeyboardHandle<D>, KeyboardError> {
-        Self::add_keyboard_with_context_flags(self, xkb_config, repeat_delay, repeat_rate, xkbcommon::xkb::CONTEXT_NO_FLAGS)
+        Self::add_keyboard_with_context_flags(
+            self,
+            xkb_config,
+            repeat_delay,
+            repeat_rate,
+            xkbcommon::xkb::CONTEXT_NO_FLAGS,
+        )
     }
 
-    /// Inner helper function for `add_keyboard` which allowed for overwriting default context
-    /// flags passed to creation of xkb::Context.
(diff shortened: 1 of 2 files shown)


PolyMeilex:
> A side effect of this change is the new functions are now dead code internally, but I figured it would be better to not potentially break other projects by removing them, and so allow pragmas have been added.

They are not public, therefore no other project can use them, so it's fine to just remove them.

EDIT: Or keep them, and simply add a new arg to them, instead of adding the new constructor.

src/input/mod.rs
```diff
@@ -570,9 +571,22 @@ impl<D: SeatHandler + 'static> Seat<D> {
         xkb_config: keyboard::XkbConfig<'_>,
         repeat_delay: i32,
         repeat_rate: i32,
+    ) -> Result<KeyboardHandle<D>, KeyboardError> {
+        Self::add_keyboard_with_context_flags(self, xkb_config, repeat_delay, repeat_rate, xkbcommon::xkb::CONTEXT_NO_FLAGS)
+    }
+
+    /// Inner helper function for `add_keyboard` which allowed for overwriting default context
+    /// flags passed to creation of xkb::Context.
```
PolyMeilex: It's a public API so not really "Inner helper function".

I would try something along the line of:

```suggestion
    /// [`Self::add_keyboard`] equivalent, that allows for overwriting default xkb::Context flags.
    ///
    /// For more info read [`Self::add_keyboard`] docs.
```

dotpyfe: fair point - changed docstring

dotpyfe:
acknowledged and updated - now only public API has been split into 'default' and user-definable versions

PR #1957, opened 2026-03-08
Add missing WmWindowType variants

kelnos:
## Description

The [wm-spec](https://specifications.freedesktop.org/wm/latest/ar01s05.html#id-1.6.7) specifies four more (COMBO, DESKTOP, DOCK, and DND) that are not present in the enum.

Aside from COMBO, the utility of these types on a Wayland desktop are perhaps questionable, but I figured I'd add all of them for completeness' sake.

## Checklist


Drakulix:
Looks like the additional enum-variants might have increased the size of the underlying representation tripping over clippy.

Can you add an `#[allow(clippy::large-enum-variant)]`-annotation please?

kelnos:
Ah, strange that adding simple enum variants like that would change the size of any struct that uses it.  But anyhow, added!

Drakulix:
Thanks!

PR #1961, opened 2026-03-12
image-capture-source: Don't send error on closed output/toplevel

ids1024:
## Description

I believe these tests are inherently racy, so we can't be sending a protocol error.

Instead this should work, but creating a capture session should immediately result in a `stopped` event.

Should we have an alternative to `output_source_created` and `toplevel_source_created` that is called when a source is created with this? (To add any user data the compositor expects, etc.) Or should we just document that the compositor shouldn't assume either of these have been called?

## Checklist


Drakulix:
> Should we have an alternative to `output_source_created` and `toplevel_source_created` that is called when a source is created with this? (To add any user data the compositor expects, etc.) Or should we just document that the compositor shouldn't assume either of these have been called?

I think the later.

PR #1985, opened 2026-04-05
New feature to distinguish whether the `Windows`, `Control`, `Alt`, and `Shift` keys are the left or right keys

jilliss:
For use by downstream NIRI

## Description

I noticed that niri was developed based on this project.
Furthermore, niri has an outstanding issue: it does not support advanced keyboard modifiers
such as Super_L. This commit adds support for them.

Specifically, it adds a `keycode` parameter to keyboard input to identify the specific key
and introduces some boolean values to track key presses for use by downstream components.

see  niri  issues [Do not support Shift_R key.](https://github.com/niri-wm/niri/issues/2817)

## Checklist

### DIFF
--- a/src/input/keyboard/modifiers_state.rs
+++ b/src/input/keyboard/modifiers_state.rs
@@ -34,6 +34,30 @@ pub struct ModifiersState {
     /// The "ISO level 5 shift" key
     pub iso_level5_shift: bool,
 
+    /// windows left key
+    pub logo_left: bool,
+
+    /// windows right key
+    pub logo_right: bool,
+
+    /// control left key
+    pub ctrl_right: bool,
+
+    /// control right key
+    pub ctrl_left: bool,
+
+    /// alt left key
+    pub alt_left: bool,
+
+    /// alt right key
+    pub alt_right: bool,
+
+    ///  shift left key
+    pub shift_left: bool,
+
+    ///  shift right key
+    pub shift_right: bool,
+
     /// Cached serialized modifier state, e.g. for sending in `wl_keyboard.modifiers`.
     ///
     /// Note that this may have outdated information compared to the other fields, and that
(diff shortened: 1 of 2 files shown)


jilliss:
@PolyMeilex  @YaLTeR  could you review this?

src/input/keyboard/modifiers_state.rs
```diff
@@ -48,6 +72,16 @@ impl ModifiersState {
...
+        self.init_key_val(state);
+        self.serialized = serialize_modifiers(state);
+    }
+
+    /// Synchronizes the internal modifier flags with the current XKB state.
+    ///
+    /// This method checks if standard modifier names (e.g., Ctrl, Alt, Shift, Caps Lock)
+    /// are currently active in the "effective" state and updates the corresponding
+    /// boolean fields of the struct.
+    pub fn init_key_val(&mut self, state: &xkb::State) {
```
PolyMeilex: Any reasons for this being part of the public API?
Going of the vibe it seems to be an internal helper function.

src/input/keyboard/modifiers_state.rs
```diff
@@ -57,6 +91,31 @@ impl ModifiersState {
...
+            xkb::keysyms::KEY_Control_L => self.ctrl_left = true,
+            xkb::keysyms::KEY_Control_R => self.ctrl_right = true,
+            xkb::keysyms::KEY_Alt_L => self.alt_left = true,
+            xkb::keysyms::KEY_Alt_R => self.alt_right = true,
+            xkb::keysyms::KEY_Shift_L => self.shift_left = true,
+            xkb::keysyms::KEY_Shift_R => self.shift_right = true,
+            _ => {} // Ignore non-modifier keys
+        }
+        // 4. Synchronize the serialized string representation of active modifiers
         self.serialized = serialize_modifiers(state);
```
PolyMeilex: Let's drop the meaningless LLM-style comments, and keep only the `// Ignore non-modifier keys` one

PolyMeilex (commented):
Hi!
Could you explain when and how are those new fields cleared? At a first glance at the diff I only see them being set to `true`

src/input/keyboard/modifiers_state.rs
jilliss: Optimized the assignment logic as suggested. Thanks!

src/input/keyboard/modifiers_state.rs
jilliss: done

jilliss:
Could you please review my code again? @PolyMeilex  @Drakulix

jilliss:
Updated the code based on your previous feedback. Ready for another look when you're available! Thanks for your time.   @PolyMeilex

PR #1991, opened 2026-04-12
fix orphaned pre-commit hook and missing returns after post_error

infraflakes:
## Description

The ext_session_lock_surface_v1 pre-commit hook added in GetLockSurface is never removed on destroy. After the destroyed callback resets last_acked to None, any commit to the WlSurface hits post_error and kills the client connection. 

The fix is observed and comfirmed to be working with quickshell's lock module from [Dank Material Shell](https://github.com/AvengeMedia/DankMaterialShell), the connection dies on unlock. Breaking similar session lock clients.  

## Checklist

### DIFF
--- a/src/wayland/session_lock/lock.rs
+++ b/src/wayland/session_lock/lock.rs
@@ -75,8 +75,13 @@ where
                     return;
                 }
 
+                // Add pre-commit hook for updating surface state (before creating user data).
+                let commit_hook_id =
+                    compositor::add_pre_commit_hook::<D, _>(&surface, LockSurface::pre_commit_hook);
+
                 let data = ExtLockSurfaceUserData {
                     surface: surface.downgrade(),
+                    commit_hook_id,
                 };
                 let lock_surface = data_init.init(id, data);
 
@@ -97,9 +102,6 @@ where
                     }
                 });
 
-                // Add pre-commit hook for updating surface state.
-                compositor::add_pre_commit_hook::<D, _>(&surface, LockSurface::pre_commit_hook);
-
                 // Call compositor handler.
                 let lock_surface = LockSurface::new(surface, lock_surface);
                 state.new_surface(lock_surface.clone(), output);
@@ -111,6 +113,7 @@ where
                 // Ensure session is locked.
                 if !data.lock_status.load(Ordering::Relaxed) {
                     lock.post_error(Error::InvalidUnlock, "Session is not locked.");
+                    return;
                 }
 
                 state.lock_state().locked_outputs.clear();
@@ -120,6 +123,7 @@ where
                 // Ensure session is not locked.
                 if data.lock_status.load(Ordering::Relaxed) {
                     lock.post_error(Error::InvalidDestroy, "Cannot destroy session lock while locked.");
+                    return;
                 }
             }
             _ => unreachable!(),
--- a/src/wayland/session_lock/surface.rs
+++ b/src/wayland/session_lock/surface.rs
@@ -3,7 +3,7 @@
 use std::sync::Mutex;
 
 use crate::backend::renderer::buffer_dimensions;
-use crate::utils::{IsAlive, Logical, SERIAL_COUNTER, Serial, Size};
+use crate::utils::{IsAlive, Logical, SERIAL_COUNTER, Serial, Size, hook::HookId};
 use crate::wayland::compositor::{self, BufferAssignment, Cacheable, SurfaceAttributes};
 use crate::wayland::viewporter::{ViewportCachedState, ViewporterSurfaceState};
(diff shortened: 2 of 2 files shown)


YaLTeR:
> The fix is observed and comfirmed to be working with quickshell's lock module from [Dank Material Shell](https://github.com/AvengeMedia/DankMaterialShell), the connection dies on unlock. Breaking similar session lock clients.

But I use DMS and it never died on unlock for me on any of my machines?

infraflakes:
> > The fix is observed and comfirmed to be working with quickshell's lock module from [Dank Material Shell](https://github.com/AvengeMedia/DankMaterialShell), the connection dies on unlock. Breaking similar session lock clients.
> 
> But I use DMS and it never died on unlock for me on any of my machines?

That's weird, I'm currently using nixos-unstable and dms via NixOS module and screen locking dies on unlock. The same happened to dms imported via flake.
screenshot-2026-04-12-14-30-32

YaLTeR:
Try on niri?

infraflakes:
> Try on niri?

I actually did and it didn't die, but I'm testing out on my [wm](https://github.com/infraflakes/srwc). I tried to reference Niri's code but I couldn't resolve this. But I find that niri's commit handler has a comment acknowledging that orphaned lock surface commits can occur after unlock:
```rust
// This message can trigger for lock surfaces that had a commit right after we unlocked
        // the session, but that's ok, we don't need to handle them.
        trace!("commit on an unrecognized surface: {surface:?}, root: {root_surface:?}");
```

But the pre-commit hook runs before that handler, so the post_error would still fire. It appears that Niri also has this issue and this fix would be for every wayland compositor based on Smithay.

YaLTeR:
> But the pre-commit hook runs before that handler, so the post_error would still fire. It appears that Niri also has this issue

But it apparently doesn't?

infraflakes:
> > But the pre-commit hook runs before that handler, so the post_error would still fire. It appears that Niri also has this issue
> 
> But it apparently doesn't?

I'm not sure if we're on the same page but what I meant is that the "issue" I refer to is about the pre-commit hook, not the shell die after unlock. And Niri's life cycle happens to prevent this, but I still think this is a bug. Please correct me if I'm wrong.

Drakulix (commented):
Weird that other compositors don't run into this, but the fix seems very much correct.

I don't see anything in the protocol, that explicitly forbits re-use of a wl_surface for a new ext-lock-surface, though I guess the following paragraph could be read this way:

> On binding this interface the compositor will immediately send the first configure event. After making the ack_configure request in response to this event the client should attach and commit the first buffer. Committing the surface before acking the first configure is a protocol error. Committing the surface with a null buffer at any time is a protocol error.

If that is the conclusion, a better fix would be to not reset `lack_acked` on destroy. @YaLTeR any opinion on that? 
At least I don't think there is anything unsafe about allowing surface re-use.

YaLTeR:
I don't have the surrounding code context "paged in", which was why I was asking why niri had no issue (I don't have the niri handling "paged in" either).

I'd verify what sway does. If it allows ext-session-lock role reuse then we can do the same in Smithay. The fix should be consistent with others (layer-shell and xdg-shell) which should also allow role reuse.

Drakulix:
> I don't have the surrounding code context "paged in", which was why I was asking why niri had no issue (I don't have the niri handling "paged in" either).

Fair enough, I primarily pinged you, because you did introduce code for lock-surface reuse: https://github.com/Smithay/smithay/pull/1817 (or more specifically https://github.com/Smithay/smithay/commit/e6d3633168ef291001ba8c7b8eed6d616cfa10ed).

> I'd verify what sway does. If it allows ext-session-lock role reuse then we can do the same in Smithay. The fix should be consistent with others (layer-shell and xdg-shell) which should also allow role reuse.

I don't think it does. The destroy logic doesn't seem to unset the role-object: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_session_lock_v1.c?ref_type=heads#L155
Which would be required in wlroots for a full-reset. (Note that the role-object is different from the role, which cannot be reassigned.)

However it will short-circuit the commit logic: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_session_lock_v1.c?ref_type=heads#L155


I think this PR currently would leave the surface in a state, where a client could destroy the lock-surface, do a NULL-commit and then re-use the `wl_surface` with `get_lock_surface`.

If we want to prevent that, I think we should simply add an `AliveTracker` to the `ExtLockSurfaceUserData`, remove all the destroy-handling logic and instead set the tracker to `false` and change the commit-hook to return early, if the surface isn't alive anymore.

PR #1992, opened 2026-04-14
docs: add emskin to the list of compositors

loyalpartner:
Add [emskin](https://github.com/emskin/emskin) to the "Other compositors that use Smithay" list.

emskin is a nested Wayland compositor that wraps Emacs, allowing any Wayland or X11 application (browsers, terminals, video players, etc.) to be embedded inside Emacs windows as native buffers.


Drakulix (commented):
Sure thing. Interesting project 👍

PR #1993, opened 2026-04-15
Enable `unsafe_op_in_unsafe_fn` except in certain places

ids1024:
## Description
`unsafe_op_in_unsafe_fn` is enabled by default in the 2024 edition. Previously, we already used it in some specific areas.

If we disable this for the generated EGL bindings, and for the GLES code, it's fairly manageable.

## Checklist


Drakulix (approved):
Looks good. Thanks for cleaning this up!

PR #1998, opened 2026-04-17
chore: Clippy warning fixes

PolyMeilex:
## Description

Just clippy fiexes.
I have disabled the [clippy::collapsible_match](https://rust-lang.github.io/rust-clippy/stable/index.html?search=clippy%3A%3Acollapsible_match#collapsible_match) lint tho, even tho it's trivial to change the code to fulfill it, I believe it will just cause needles code churn. 
It disallows using single branch ifs in match statements like so:
```rs
match todo!() {
    1 => {
        if true {
           // Clippy will force us to move the if to the match level
        }
    }
    2 => {}
    _ => {
        // If anything is to be added here the lint will no longer apply,
        // and worst case scenario the changes it forced might need to be reverted
    }
}
```

## Checklist

### DIFF
--- a/anvil/src/lib.rs
+++ b/anvil/src/lib.rs
@@ -1,4 +1,5 @@
 #![warn(rust_2018_idioms)]
+#![allow(clippy::collapsible_match)]
 // If no backend is enabled, a large portion of the codebase is unused.
 // So silence this useless warning for the CI.
 #![cfg_attr(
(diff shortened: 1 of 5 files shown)


YaLTeR:
I think it's also giving false positives atm on top of resulting in less clear code

PolyMeilex:
Glad that I'm not imagining things. Usually I try to stick to default clippy lints, but this one is just weird.

ids1024:
I also noticed that `cargo clippy --fix` wanted to make some dubious changes for `clippy::collapsible_match`. So yeah, just disabling that one seems reasonable.

Drakulix (approved):
Thanks!

PR #2000, opened 2026-04-18
Add WaylandSurfaceRenderElement::buffer()

YaLTeR:
## Description

This is useful for example to get to `buffer_y_inverted()` as there's no other good way to do it from what I can find. The texture `is_y_inverted()` is only present on concrete `GlesTexture` and not on the generic `R::Texture`.

There's `WaylandSurfaceRenderElement::underlying_storage()` that gets painfully close, but it requires a renderer (that it doesn't even use).

## Checklist


Drakulix (approved):
yeah, I don't see an issue with that.

PR #2004, opened 2026-04-19
drm_syncobj: expose timeline fd and point accessors

endigma:
## Description

External renderers that don't go through smithay's built-in RendererSurfaceState path need access to:

- `DrmTimeline`'s underlying syncobj fd, so they can import the same timeline into a `VkSemaphore` via `VK_KHR_external_semaphore_fd` and chain timeline wait/signal values into `VkTimelineSemaphoreSubmitInfo`.
- `DrmSyncPoint`'s timeline value (u64), needed as the wait/signal value in the same submit chain.
- A reference to the `DrmTimeline` a `DrmSyncPoint` belongs to, so an external cache can key semaphore imports per-timeline.

Using these is probably not idiomatic usage of Smithay for a normal compositor, but it is useful nonetheless for things using Smithay for less typical use cases.

Might end up being useful if Smithay ever gets an official Vulkan render backend, I'm using these to manually set up explicit sync with my own renderer. 

I'm not sure if it's possible or useful to add a test for this, if you have some suggestion I'm happy to add it.

## Checklist

### DIFF
--- a/src/wayland/drm_syncobj/sync_point.rs
+++ b/src/wayland/drm_syncobj/sync_point.rs
@@ -81,6 +81,12 @@ impl PartialEq for DrmTimeline {
     }
 }
 
+impl AsFd for DrmTimeline {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        self.0.timeline_fd.as_fd()
+    }
+}
+
 impl DrmTimeline {
     /// Import DRM timeline from file descriptor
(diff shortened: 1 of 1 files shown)


endigma:
Closing for now as this isn't working quite right with NVIDIA drivers:
- https://github.com/KhronosGroup/Vulkan-Docs/issues/2473
- https://github.com/ValveSoftware/gamescope/issues/1662

PR #2005, opened 2026-04-20
Fix JetBrain IDEs popups not getting keyboard focus

### LINKED ISSUE #2003: Handling of ICCCM 4.1.7
This issue is related to https://github.com/pop-os/cosmic-comp/issues/874

It occurs, because smithay directly blocks input to the X11Surface if it has specified `WmInputModel::None`
without forwarding it to xwayland.
[Here is the relevant piece of code](https://github.com/Smithay/smithay/blob/dba255cd18c1a808c85a1d1144ad48e0b3f44b8f/src/xwayland/xwm/surface.rs#L1243)
This appears to be the cause of the described issue. I removed it and when I did the intellij popups started working normally.
I cross-referenced the implementation here with hyprland, since intellij popups work normally in hyprland,
and I couldn't find any such check in hyprland's source code (it's perfectly possible I missed it).
Does this check serve some specific purpose or was it a just a natural looking optimization with an overlooked consequence?

iexavl:
## Description
Resolves #2003

## Checklist

### DIFF
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -1303,10 +1303,6 @@ impl<D: SeatHandler + 'static> KeyboardTarget<D> for X11Surface {
         serial: Serial,
         time: u32,
     ) {
-        if self.input_model() == WmInputModel::None {
-            return;
-        }
-
         let mut xstate = self.state.lock().unwrap();
         if let Some(surface) = xstate.wl_surface.as_ref() {
             KeyboardTarget::key(surface, seat, data, key, state, serial, time)
@@ -1322,10 +1318,6 @@ impl<D: SeatHandler + 'static> KeyboardTarget<D> for X11Surface {
     }
 
     fn modifiers(&self, seat: &Seat<D>, data: &mut D, modifiers: ModifiersState, serial: Serial) {
-        if self.input_model() == WmInputModel::None {
-            return;
-        }
-
         let mut state = self.state.lock().unwrap();
         if let Some(surface) = state.wl_surface.as_ref() {
             KeyboardTarget::modifiers(surface, seat, data, modifiers, serial);


Drakulix:
Ping @ids1024, see discussion in linked issue. I kinda want a second pair of eyes on this issue. Do you agree?

ids1024:
Overall I suspect this is better than the current behavior whether or not it's "correct", so I think it's probably reasonable to merge regardless.

...Though ICCCM is a standard for window managers, and isn't used by the X server, right? So this isn't necessarily something we should be ignoring and just leaving up to `XWayland` to sort out. Do the X input events end up being sent to the window with `WmInputModel::None`? Or to the other window?

I wonder if the Wayland compositor and X server disagree about what X windows has input focus. When the [ICCCM](https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html) defines input models, it talks about the client using `SetInputFocus`. Is the client setting the focus to a window other than the one with `WmInputModel::None`? But are we treating that window as focused, and using `KeyboardTarget` on the wrong `X11Surface`?

If the issue is indeed that the compositor and X server disagree on which window has keyboard focus, then if we fixed that, this check presumably would be fine. Though not necessary.

That might be harder to fix though.

iexavl:
> If the issue is indeed that the compositor and X server disagree on which window has keyboard focus, then if we fixed that, this check presumably would be fine. Though not necessary.

I think you might be right. To be honest this has turned into a bit of a head scratcher for me.
I noticed that the active window hint in pop os 22.04 (under Xorg) is different from the one I see in hyprland.
In pop os 22.04 when I open one of these intellij popups I see that the active window hint stays on the intellij window and doesn't move to the popup.
On hyprland, however, the active window hint does move to the popup when I focus it.
I checked what X window was considered "in focus" in pop os 22.04 and hyprland using 
`xprop -root _NET_ACTIVE_WINDOW`
In pop os 22.04, the active window was the intellij window, not the popup.
In hyprland the active window  was the popup itself.
I also performed the same test in cosmic-comp both with and without this patch. In both cases, it gave the root 
window, which sounds like the outlier. 
At this point I can probably take a look at pop os 22.04 WM and hyprland and I'm pretty sure I'd see two different things and I wouldn't really know which one is correct.

iexavl:
>  I also performed the same test in cosmic-comp both with and without this patch. In both cases, it gave the root window, which sounds like the outlier. 

Okay, it seems like this returns the root window no matter which window is in focus.

ids1024:
Yeah, the `_NET_ACTIVE_WINDOW` is separate from the focus in `XSetInputFocus`/``XGetInputFocus`.

Apparently `xdpyinfo | grep focus` shows the latter.

ids1024 (approved):
It looks like I was basically right about what's going on, but actually I do think this is the correct solution, not just a workaround.

* I am able to reproduce this in IDEA Community Edition, with the "New Java Class" dialog, with `WAYLAND_DISPLAY` unset. (The latest release has native Wayland support, and that should be better than using XWayland where available.)
* `xdpyinfo | grep focus` does show that the window with input focus remains unchanged when the dialog is opened. So X does consider the main window to have input focus, while cosmic-comp doesn't.
* `keyboard_handle_enter()` in `XWayland` sets `keyboard_focus` to the surface entered, but that's currently **only** used to call `xdg_system_bell_v1_ring`. So this shouldn't cause problems. (Not sure how bells work in X11, but I guess if there's no way to get a more specific surface, that's as good as anything.)
  - https://gitlab.freedesktop.org/xorg/xserver/-/blob/main/hw/xwayland/xwayland-input.c

Meanwhile we probably don't want the compositor to treat the main window as focused in any other regard, just because X11 has it as the input focus. And we want to let XWayland handle input if and only if some X11 windows has focus. So this seems overall correct.

PR #2008, opened 2026-04-24
Prototype a `Dispatch2`/`GlobalDispatch2` trait to dispatch delegate macros to oblivion

ids1024:
## Description
See https://github.com/Smithay/client-toolkit/pull/519 for a bit more context.

At present, this replaces `delegate_compositor!` with a single generic `delegate_dispatch2!` macro. It should be possible to update all the dispatch implementations this way.

If there are no issues with this, since `smithay` is less strict about API breaks than `wayland-rs`, we could merge this in `smithay` (compositors mostly should just need to add `delegate_dispatch2!` and remove the other macro calls). Then the next wayland-rs version (https://github.com/Smithay/wayland-rs/issues/900) can incorporate the updated version of the trait. At which point the macro is no longer necessary, and more type bounds can be remove as implied.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,10 @@ The following methods are no longer needed as Smithay does them automatically no
 
 You also no longer need to manually set `LayerSurfaceAttributes::initial_configure_sent`, Smithay handles it automatically.
 
+The `delegate_*!` macros have been replaced with a single `delegate_dispatch2!` macro, which implements dispatch in terms
+of new `Dipsatch2` and `GlobalDispatch2` traits. These will replace `Dispatch` and `GlobalDispatch` in a future version of
+`wayland_server`.
+
 ### Additions
 
 - ExtBackgroundEffect protocol is now available in `smithay::wayland::background_effect` module.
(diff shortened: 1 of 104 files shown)


ids1024:
CI passes, this works with cosmic-comp, and https://github.com/Smithay/smithay/pull/2017 shows it also works with an updated `wayland-rs` providing these versions of the trait (which allows removing more extra trait bounds).

So I think this should be good to merge.

Drakulix (approved):
Looks great. I'd say we shouldn't let this bit-rot as it is an improvement all around. But this needs a CHANGELOG entry, no? 😅

ids1024:
Amended with a line in the changelog.

This is pretty easy to adapt to, though when wayland-rs releases that will be more involved. May want to revive https://github.com/Smithay/smithay.github.io with an announcement of the new release.

Drakulix (approved):

YaLTeR:
I'm updating niri, specifically protocols implemented in niri itself. Do I understand correctly that when implementing protocols from other crates, such as:

```
impl<D> Dispatch2<ZwlrGammaControlManagerV1, D> for GlobalData
```

I want to define my own `GlobalData` inside niri? Since otherwise the coherence rules don't allow this definition because all of `Dispatch2`, `ZwlrGammaControlManagerV1` and `GlobalData` come from other crates.

Also if the user data is itself an object from another crate, then I also want to wrap it into my own struct for the same reason?

```
impl<D> Dispatch2<ExtWorkspaceHandleV1, D> for ExtWorkspaceManagerV1
```

PR #2010, opened 2026-04-27
xwm: use the same depth/visual/colormap for the frame window

kelnos:
## Description

The XWayland root window uses a 24bpp visual, so when we copy the depth and visual from the parent for the frame window, we end up with a 24bpp frame window.  When XWayland creates the wl_buffer for a client, it ends up using the frame window as a template for the buffer format, which will never have an alpha channel.  So if the client wants to do partial transparency, such as for CSD drop shadows, they'll end up drawing as opaque rectangles around the window.

Instead, match the client window's visual and colormap when creating the frame window.  This also caches the colormap that goes along with the selected visual so we avoid creating a new one with each window (and then need to free it later).  In practice, clients will likely use one of a small number of possible visuals, so it'll mostly be cache hits, and the cache will only have a few entries.

## Checklist

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -139,6 +139,7 @@ use atomic_float::AtomicF64;
 use calloop::{Interest, LoopHandle, Mode, PostAction, generic::Generic};
 use rustix::fs::OFlags;
 use std::{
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Seems sensible. Thanks!

PR #2014, opened 2026-04-28
xwm: clean up by_serial hash map when surface is destroyed

kelnos:
## Description

In a compositor that restarts XWayland on a crash, the new XWayland instance will start off with a fresh serial counter.  As new X11 windows are created, they can have the same serial as a window created on the previous XWayland instance.  Without this cleanup, the lookup in surface_for_serial() will return a stale/dead wl_surface, and the xwayland machinery will be unable to properly associate new windows with a surface.

## Checklist


Drakulix (approved):
thanks!

PR #2016, opened 2026-04-29
xwm: support _NET_WM_OPAQUE_REGION

kelnos:
## Description

X11 clients can set this property on their window to indicate a single fully-opaque region for the window.

I'm not sure if this is actually the right way to propagate the opaque region through; would appreciate some guidance if not.

## Checklist

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -227,6 +227,7 @@ mod atoms {
             _NET_WM_WINDOW_TYPE_TOOLTIP,
             _NET_WM_WINDOW_TYPE_UTILITY,
             _NET_WM_STATE_MODAL,
+            _NET_WM_OPAQUE_REGION,
             _MOTIF_WM_HINTS,
             _NET_STARTUP_ID,
 
@@ -803,6 +804,7 @@ impl X11Wm {
                 atoms._NET_CLIENT_LIST,
                 atoms._NET_CLIENT_LIST_STACKING,
                 atoms._NET_SHOWING_DESKTOP,
+                atoms._NET_WM_OPAQUE_REGION,
             ],
         )?;
         conn.change_property32(
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -10,9 +10,9 @@ use crate::{
         },
         touch::TouchTarget,
     },
-    utils::{Client, IsAlive, Logical, Rectangle, Serial, Size, user_data::UserDataMap},
+    utils::{Client, HookId, IsAlive, Logical, Physical, Rectangle, Serial, Size, user_data::UserDataMap},
     wayland::{
-        compositor,
+        compositor::{self, RectangleKind, RegionAttributes, SurfaceAttributes},
         seat::{WaylandFocus, keyboard::enter_internal},
     },
 };
@@ -81,6 +81,7 @@ pub(crate) struct SharedSurfaceState {
 
     // The associated wl_surface.
     wl_surface: Option<WlSurface>,
+    pre_commit_hook_id: Option<HookId>,
 
     title: String,
     class: String,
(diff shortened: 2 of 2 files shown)


src/xwayland/xwm/mod.rs
```diff
@@ -2293,7 +2302,26 @@ where
...
+                                    xsurface.state.lock().unwrap().opaque_region.map(|rect| {
+                                        RegionAttributes {
+                                            rects: vec![(RectangleKind::Add, rect)],
+                                        }
+                                    });
+                                compositor::with_states(wl_surface, |states| {
+                                    let mut guard = states.cached_state.get::<SurfaceAttributes>();
+                                    let attrs = guard.pending();
+                                    attrs.opaque_region = opaque_region;
+                                });
```
Drakulix: Hmm, while it would be nice to lock any opaque-region updates to commits, I am not sure those are meant to be applied atomically.

Does the spec say anything about new contents and when to apply changes to `__NET_WM_OPAQUE_REGION`? Did you see any updates to this without content changes in your testing?

src/xwayland/xwm/mod.rs
```diff
@@ -2612,3 +2640,43 @@ fn atom_from_mime(
...
+                            .to_logical(xwm.client_scale.load(Ordering::Acquire))
+                            .to_i32_round::<i32>(),
+                    )
+                }
+            }
+            _ => None,
+        }
+    });
+
+    xsurface.state.lock().unwrap().opaque_region = opaque_region;
```
Drakulix: The client scale change definitely looks correct, but does the documentation say anything with regards to what this is specifically relative to? I guess otherwise it is safe to assume it is relative to the x11 window geometry, which this should apply to.

Drakulix (commented):
Looks mostly good, definitely a render optimization, that we want to have.

Just a few questions on the specifics of the spec on this property.

Also: Any particular clients, that use this feature and you tested with?

src/xwayland/xwm/mod.rs
kelnos: Here is the spec: https://specifications.freedesktop.org/wm/1.5/ar01s05.html#id-1.6.19

... which I appear to have misread; looks like it's a list of rectangles, not just a single one, so I'll update to fix that.

I guess in general on X11 we don't have the same sort of synchronization between window properties and painting as we do on Wayland, so the compositing manager presumably should apply it as soon as it "sees" the property.  That means the client has to have already painted the window in a way that matches the values it sets in this property.

src/xwayland/xwm/mod.rs
kelnos: Yes, the spec says, "representing a rectangle in window coordinates where the pixels of the window's contents have a fully opaque alpha value."

kelnos:
Ok, pushed a fix so it will collect as many rectangles as are in the property.  I also refactored it a bit: noticed that there was already a path in `X11Surface` that I missed yesterday for fetching initial properties and handling updates, so now it uses the same pattern.

For testing, I wrote a quick `x11rb` program that paints to a window, uniformly with 50% transparency, and sets two opaque regions in the property.  It sets them "incorrectly": I was hoping that I'd see smithay _fail_ to blend properly under the "incorrectly opaque" rects.

That didn't work (that is, it blended properly underneath the entire window), so I added debugging prints in `update_opaque_regions()` to ensue that the rects were being picked up, which worked.

Just now I added a few more debugging prints, and I've discovered that the pre-commit hook I added isn't getting called at all... realized I should have that in `X11Surface::set_wl_surface()`, as the surface can get set in multiple ways.  That fixed things up.

So back to the test program: now it works as expected: smithay doesn't blend below the opaque regions in the window, so even though they're painted as partially transparent, I can't see other window content beneath the area where those rects are set as the opaque region.  The test program also updates `_NET_WM_OPAQUE_REGION` on window resize, and that appears to work properly as well.

kelnos:
Screenshot of it working for posterity.  The two orange-bordered rectangles are painted exactly as the rest of the window is painted, partially transparent, but we can't see the text in the terminal below those areas, as they're in the regions specified in `_NET_WM_OPAQUE_REGION`.

Screenshot_2026-04-29_14-06-26

Drakulix (approved):
LGTM now. Thanks for taking the time!

PR #2018, opened 2026-05-02
xwm: add support for _NET_WM_SYNC_REQUEST

kelnos:
## Description

This handles the protocol details around _NET_WM_SYNC_REQUEST, and exposes API so the compositor can coordinate frame/decoration resize with app repaints.

See https://specifications.freedesktop.org/wm/1.5/ar01s06.html#id-1.7.3 and https://fishsoup.net/misc/wm-spec-synchronization.html for details. (Note that this does not implement `_NET_WM_FRAME_DRAWN` or `_NET_WM_FRAME_TIMINGS`; this only implements the core `_NET_WM_SYNC_REQUEST`.)

(Draft as I haven't tested this yet; time to build the compositor side of this.  But initial feedback welcome.)

## Checklist

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -161,6 +161,7 @@ use x11rb::{
         composite::{ConnectionExt as _, Redirect},
         randr::{ConnectionExt as _, Notify, NotifyMask},
         render::{ConnectionExt as _, CreatePictureAux, PictureWrapper},
+        sync::{ConnectionExt as _, Counter},
         xfixes::ConnectionExt as _,
         xproto::{
             AtomEnum, CONFIGURE_NOTIFY_EVENT, ChangeWindowAttributesAux, Colormap, ColormapAlloc,
@@ -253,9 +254,12 @@ mod atoms {
             _NET_WM_STATE_SKIP_TASKBAR,
             _NET_WM_STATE_SKIP_PAGER,
             _NET_WM_STATE_STICKY,
+            _NET_WM_SYNC_REQUEST,
+            _NET_WM_SYNC_REQUEST_COUNTER,
             _NET_SHOWING_DESKTOP,
             _NET_SUPPORTING_WM_CHECK,
             _XSETTINGS_SETTINGS,
+            _XWAYLAND_ALLOW_COMMITS,
 
             // selection
             _WL_SELECTION,
@@ -291,6 +295,9 @@ pub use self::atoms::Atoms;
 use super::XWaylandClientData;
 
 crate::utils::ids::id_gen!(xwm_id);
+
+const XSYNC_REQUIRED_MAJOR_VERSION: u8 = 3;
+const XSYNC_REQUIRED_MINOR_VERSION: u8 = 1;
 
 /// Id of an X11 WM
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
@@ -494,6 +501,22 @@ pub trait XwmHandler {
         let _ = (xwm, window, timestamp);
     }
 
+    /// A client has acknowledged a `_NET_WM_SYNC_REQUEST` request.
+    ///
+    /// This is informational only; a compositor need not implement this or take any particular
+    /// action.
+    fn sync_request_acked(&mut self, xwm: XwmId, window: X11Surface) {
+        let _ = (xwm, window);
+    }
+
+    /// A client's `_NET_WM_SYNC_REQUEST` has timed out.
+    ///
+    /// This is informational only; a compositor need not implement this or take any particular
+    /// action.
+    fn sync_request_timeout(&mut self, xwm: XwmId, window: X11Surface) {
+        let _ = (xwm, window);
+    }
+
     /// Window requests access to the given selection.
     fn allow_selection_access(&mut self, xwm: XwmId, selection: SelectionTarget) -> bool {
         let _ = (xwm, selection);
@@ -546,6 +569,7 @@ pub struct X11Wm {
     screen: Screen,
     wm_window: OwnedX11Window,
     atoms: Atoms,
+    servertime_counter: Option<Counter>,
     xsettings: XSettings,
     randr_primary: u32,
 
@@ -573,6 +597,11 @@ pub struct X11Wm {
 impl Drop for X11Wm {
     fn drop(&mut self) {
         xwm_id::remove(self.id.0);
+
+        // Break reference cycle caused by deferred_sync hook
+        for window in std::mem::take(&mut self.windows) {
+            window.handle_destroyed();
+        }
     }
 }
 
@@ -730,6 +759,24 @@ impl X11Wm {
         let screen = conn.setup().roots[0].clone();
         let randr_primary = conn.randr_get_output_primary(screen.root)?.reply()?.output;
 
+        let sync_init = conn
+            .sync_initialize(XSYNC_REQUIRED_MAJOR_VERSION, XSYNC_REQUIRED_MINOR_VERSION)?
+            .reply_unchecked();
+        let sync_supported = sync_init.ok().flatten().is_some();
+        let servertime_counter = if sync_supported {
+            conn.sync_list_system_counters()?
+                .reply_unchecked()?
+                .and_then(|reply| {
+                    reply
+                        .counters
+                        .into_iter()
+                        .find(|counter| counter.name == b"SERVERTIME" && counter.counter != x11rb::NONE)
+                })
+                .map(|counter| counter.counter)
+        } else {
+            None
+        };
+
         {
             let font = FontWrapper::open_font(&conn, "cursor".as_bytes())?;
             let cursor = CursorWrapper::create_glyph_cursor(
@@ -784,34 +831,43 @@ impl X11Wm {
         conn.composite_redirect_subwindows(screen.root, Redirect::MANUAL)?;
 
         // Set some EWMH properties
+        let net_supported_base = [
+            atoms._NET_WM_STATE,
+            atoms._NET_WM_STATE_MAXIMIZED_HORZ,
+            atoms._NET_WM_STATE_MAXIMIZED_VERT,
+            atoms._NET_WM_STATE_HIDDEN,
+            atoms._NET_WM_STATE_FULLSCREEN,
+            atoms._NET_WM_STATE_MODAL,
+            atoms._NET_WM_STATE_FOCUSED,
+            atoms._NET_WM_STATE_ABOVE,
+            atoms._NET_WM_STATE_BELOW,
+            atoms._NET_WM_STATE_SHADED,
+            atoms._NET_WM_STATE_SKIP_TASKBAR,
+            atoms._NET_WM_STATE_SKIP_PAGER,
+            atoms._NET_WM_STATE_STICKY,
+            atoms._NET_WM_STATE_DEMANDS_ATTENTION,
+            atoms._NET_ACTIVE_WINDOW,
+            atoms._NET_WM_MOVERESIZE,
+            atoms._NET_CLIENT_LIST,
+            atoms._NET_CLIENT_LIST_STACKING,
+            atoms._NET_SHOWING_DESKTOP,
+            atoms._NET_WM_OPAQUE_REGION,
+            atoms._NET_WM_PING,
+        ];
+        let net_supported = if sync_supported {
+            net_supported_base
+                .into_iter()
+                .chain([atoms._NET_WM_SYNC_REQUEST, atoms._NET_WM_SYNC_REQUEST_COUNTER])
+                .collect::<Vec<_>>()
+        } else {
+            net_supported_base.to_vec()
+        };
         conn.change_property32(
             PropMode::REPLACE,
             screen.root,
             atoms._NET_SUPPORTED,
             AtomEnum::ATOM,
-            &[
-                atoms._NET_WM_STATE,
-                atoms._NET_WM_STATE_MAXIMIZED_HORZ,
-                atoms._NET_WM_STATE_MAXIMIZED_VERT,
-                atoms._NET_WM_STATE_HIDDEN,
-                atoms._NET_WM_STATE_FULLSCREEN,
-                atoms._NET_WM_STATE_MODAL,
-                atoms._NET_WM_STATE_FOCUSED,
-                atoms._NET_WM_STATE_ABOVE,
-                atoms._NET_WM_STATE_BELOW,
-                atoms._NET_WM_STATE_SHADED,
-                atoms._NET_WM_STATE_SKIP_TASKBAR,
-                atoms._NET_WM_STATE_SKIP_PAGER,
-                atoms._NET_WM_STATE_STICKY,
-                atoms._NET_WM_STATE_DEMANDS_ATTENTION,
-                atoms._NET_ACTIVE_WINDOW,
-                atoms._NET_WM_MOVERESIZE,
-                atoms._NET_CLIENT_LIST,
-                atoms._NET_CLIENT_LIST_STACKING,
-                atoms._NET_SHOWING_DESKTOP,
-                atoms._NET_WM_OPAQUE_REGION,
-                atoms._NET_WM_PING,
-            ],
+            &net_supported,
         )?;
         conn.change_property32(
             PropMode::REPLACE,
@@ -897,6 +953,7 @@ impl X11Wm {
             client_scale,
             screen,
             atoms,
+            servertime_counter,
             xsettings,
             randr_primary,
             wm_window,
@@ -932,6 +989,13 @@ impl X11Wm {
     /// Id of this X11 WM
     pub fn id(&self) -> XwmId {
         self.id
+    }
+
+    /// Whether or not the XSYNC extension is present
+    ///
+    /// This can be used to tell if the `_NET_WM_SYNC_REQUEST` protocol can be used.
+    pub fn sync_supported(&self) -> bool {
+        self.servertime_counter.is_some()
     }
 
     /// Raises a window in the internal X11 state
@@ -1439,6 +1503,7 @@ where
                 n.override_redirect,
                 Arc::downgrade(&conn),
                 xwm.atoms,
+                xwm.servertime_counter,
                 geometry,
                 xwm.dnd.xdnd_active.clone(),
             );
@@ -1628,8 +1693,8 @@ where
                 );
                 // Synthetic event
                 surface.configure(None).map_err(|err| match err {
-                    X11SurfaceError::Connection(err) => err,
                     X11SurfaceError::UnsupportedForOverrideRedirect => unreachable!(),
+                    err => err,
                 })?;
             }
         }
@@ -1729,7 +1794,7 @@ where
 
             if let Some(pos) = xwm.windows.iter().position(|x| x.window_id() == n.window) {
                 let surface = xwm.windows.remove(pos);
-                surface.state.lock().unwrap().alive = false;
+                surface.handle_destroyed();
                 drop(_guard);
                 state.destroyed_window(xwm_id, surface);
             }
@@ -2156,6 +2221,10 @@ where
                             }
                         }
                     }
+                }
+            } else if n.atom == xwm.atoms._NET_WM_SYNC_REQUEST_COUNTER {
+                if let Some(surface) = xwm.windows.iter().find(|w| w.window_id() == n.window).cloned() {
+                    surface.init_net_wm_sync_request()?;
                 }
             }
 
@@ -2594,6 +2663,30 @@ where
             xwm.screen.height_in_millimeters = n.mheight;
 
             xwm.dnd.update_screen(&xwm.screen)?;
+        }
+        Event::SyncAlarmNotify(n) => {
+            if let Some(surface) = xwm
+                .windows
+                .iter()
+                .find(|x| {
+                    let state = x.state.lock().unwrap();
+                    state.sync_alarm == Some(n.alarm) || state.sync_timeout_alarm == Some(n.alarm)
+                })
+                .cloned()
+            {
+                let is_sync_ack = surface.state.lock().unwrap().sync_alarm == Some(n.alarm);
+
+                if is_sync_ack {
+                    if surface.handle_sync_alarm(n.counter_value) {
+                        drop(_guard);
+                        state.sync_request_acked(xwm_id, surface);
+                    }
+                } else {
+                    surface.handle_sync_timeout();
+                    drop(_guard);
+                    state.sync_request_timeout(xwm_id, surface);
+                }
+            }
         }
         Event::Error(err) => {
             info!(?err, "Got X11 Error");
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -424,48 +424,53 @@ impl X11Surface {
         }
         state.wl_surface = surface;
 
-        if let Some(wl_surface) = state.wl_surface.as_ref() {
-            let hook_id = {
-                let state = Arc::downgrade(&self.state);
-                let conn = self.conn.clone();
-                let window = self.window;
-                let property_atom = self.atoms._NET_WM_OPAQUE_REGION;
-                let client_scale = self.client_scale.clone();
+        if let Some(wl_surface) = state.wl_surface.clone() {
+            self.register_opaque_regions_hook::<D>(&mut state, &wl_surface);
+        }
+    }
 
-                compositor::add_pre_commit_hook::<D, _>(wl_surface, move |_, _, wl_surface| {
-                    if let Some(state) = state.upgrade() {
-                        let update_opaque_region_if_needed = || {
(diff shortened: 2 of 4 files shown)


src/xwayland/xwm/mod.rs
```diff
@@ -725,6 +739,9 @@ impl X11Wm {
         let screen = conn.setup().roots[0].clone();
         let randr_primary = conn.randr_get_output_primary(screen.root)?.reply()?.output;
 
+        let sync_init = conn.sync_initialize(3, 1)?.reply_unchecked();
```
Drakulix: Can we document what these values mean? Perhaps making them `const`s with some descriptive names?

src/xwayland/xwm/surface.rs
```diff
@@ -70,6 +73,32 @@ const MWM_HINTS_FLAGS_FIELD: usize = 0;
 const MWM_HINTS_DECORATIONS_FIELD: usize = 2;
 const MWM_HINTS_DECORATIONS: u32 = 1 << 1;
 
+// From http://fishsoup.net/misc/wm-spec-synchronization.html
+//   "If the client is continually redrawing, then the last seen value may be out of date when the
+//   window manager sends the message. Picking a number that is 240 later would allow for 1 second
+//   of frames at 60fps."
+const SYNC_REQUEST_INCREMENT: i64 = 240;
```
Drakulix, line 100: So is this value kinda fps-dependant? Or am I misinterpreting the comment?

Drakulix (commented):
Some initial comments.

I also feel like this should probably have a minimal reference implementation in anvil, since it isn't immediately obvious how to use this feature correctly.

src/xwayland/xwm/surface.rs
kelnos: Sort of, but not in a way that really matters, I don't think.  (That is, I don't think we actually need to care about the actual framerate and adjust this.)  It's just a *very* generous fudge-factor in case the client gets busy and misses some frames.  And in practice clients and compositors don't care so much about the value (only that it's monotonically increasing), and will continue to work just fine even in the scenario this large increment value was intended for.

(I actually don't get their math there: `240 / 60 = 4`, but even in extended counter mode, the protocol only "uses up" 2 counter values per frame, so I'd think that should be _two_ seconds at 60fps.  But maybe I'm misunderstanding something.)

At any rate, both mutter and kwin still use 240, despite supporting higher framerates, so I think leaving it as the spec says is the right move.

src/xwayland/xwm/mod.rs
kelnos: Done!

kelnos:
Ok, I've done an implementation in Anvil.  It's unfortunately not as minimal as I'd hoped, as it's not entirely simple to orchestrate.

I'm thinking about ways to push more of the logic inside smithay itself.  Ideally I could push it *entirely* inside smithay: while this is most useful for throttling configure notify events on resizes, it's actually generally applicable.  You can pair a sync request with every single configure notify if you want to.  (But we could instead have `start_sync_requests()` & `stop_sync_requests()` so the compositor can control at least the time periods in which they are sent.)

The problem is the timeout.  For every sync request, Anvil sets a 1 second timeout (xfwm4 does 500ms, mutter does 1s; mutter's impl is more modern so I went with that), after which it gives up if it hasn't gotten a sync ACK, and continues configuring/resizing.  The timeout needs to be registered inside `configure()`, and then removed after `X11Wm::handle_xevent()` gets the alarm-notify event.  But `LoopHandle` has a type param, and I assume we really don't want to put a type param on `X11Wm` and `X11Surface` for this.  That would be a huge breaking change, and also just kinda ugly.

I think we could do this with erasure... `X11Wm` already takes a `LoopHandle` on construction, so it could store it in a struct with a type-erased trait impl, in an `Arc<dyn T>`.  Then it could pass that to each `X11Surface`, which would be able to use it whenever.

I'm thinking this could actually work... then Anvil would have basically just a couple extra lines of code: `start_sync_requests()` at the beginning of a resize operation, and then `stop_sync_requests()` when the resize completes or is cancelled.  What do you think?  Is the type-erased trait just too much?

`start_sync_requests()` should probably take two more parameters: the timeout duration, and timeout "policy".  Right now Anvil has no real policy: after a timeout it just keeps doing sync requests on subsequent configures.  xfwm4 disables sync requests for that window forever once it gets a single timeout.  I think mutter disables sync requests for the rest of the resize operation after a timeout, but then enables them again on the next resize.

What do you think?  I really like the idea of smithay doing all the work, and the compositor only having to enable/disable and provide a few "configuration" parameters.

kelnos:
Oh, also: I'm now setting `_XWAYLAND_ALLOW_COMMITS=0` on the X window while sync requests are in flight.  I spent hours trying to eliminate the black bars appearing along the resize edge (caused by XWayland committing a buffer at the new size before the client finishes painting), including using smithay's commit blocker functionality during the sync request (which I'm still surprised didn't work), as well as a "pending configure" system that didn't work out, and some other things.  But then I found that mutter tells XWayland to stop committing buffers during the sync request, and *finally* that fixed it.

Resize is a bit jerkier than I'd like, but that might just be how it is.  It at least looks better than it did before, and is more responsive.  I don't have mutter installed, but I plan to install it and test there at some point to see if it has the same problem.  (Though if it doesn't, I'm not sure what to do about it.)

Drakulix:
> I'm thinking this could actually work... then Anvil would have basically just a couple extra lines of code: `start_sync_requests()` at the beginning of a resize operation, and then `stop_sync_requests()` when the resize completes or is cancelled. What do you think? Is the type-erased trait just too much?

Hmm we are trying to keep smithay somewhat event-loop agnostic, even though we obviously have a lot of integration with calloop. So a type-erased "scheduler" seems a little unfortunate (and sounds like one of few - so far - examples where async would actually be helpful).

I haven't looked to deep into the code, but could we potentially pass the `LoopHandle` on `start_sync_requests` and store it only for the sync-duration at least?

> Oh, also: I'm now setting _XWAYLAND_ALLOW_COMMITS=0 on the X window while sync requests are in flight. I spent hours trying to eliminate the black bars appearing along the resize edge (caused by XWayland committing a buffer at the new size before the client finishes painting), including using smithay's commit blocker functionality during the sync request (which I'm still surprised didn't work), as well as a "pending configure" system that didn't work out, and some other things.

That seems fine to me.

kelnos:
> I haven't looked to deep into the code, but could we potentially pass the LoopHandle on start_sync_requests and store it only for the sync-duration at least?

That should be possible.  It would still have to do the type-erasure thing, as it would otherwise need the type parameter in the X event handler, which it wouldn't have access to there.  But, yes, `X11Surface` could hold onto it just for the duration of when it's doing the syncs.

Drakulix:
> > I haven't looked to deep into the code, but could we potentially pass the LoopHandle on start_sync_requests and store it only for the sync-duration at least?
> 
> That should be possible. It would still have to do the type-erasure thing, as it would otherwise need the type parameter in the X event handler, which it wouldn't have access to there. But, yes, `X11Surface` could hold onto it just for the duration of when it's doing the syncs.

This seems preferrable to me. At least that means only the sync-logic is bound to calloop for now and other code isn't likely to accidentally use the loop-handle for other things. We can try to lift that requirement in the future should a need for it arise.

kelnos:
One more thing I can think of: instead of methods to start/stop using sync events, I could add `X11:Surface::configure_with_sync()`, and on success the compositor would be expected to set a timeout.  If the timeout fires, then the compositor would call `X11Surface::sync_timeout()`.  Instead of removing `XwmHandler::sync_request_acked()` as planned, I'd leave it, and the compositor would have to cancel the timeout in that handler.  Calling the regular `configure()` method when any of this is active would cancel any pending sync and immediately send out a configure notify event, and commit the passed geometry as the new geometry.

That wouldn't be *too* bad from the compositor's perspective.  `configure_with_sync()` would handle gating further configure events on a pending sync, and `X11Surface` could apply the pending configure geometry when it gets the sync alarm.

Well, there's one annoying part: when a sync request is ACKed, there may be a buffered configure geometry stored on the surface that requires a new configure-notify and sync request, e.g.:

1. configure_with_sync(g1) -> send configure-notify(g1), send sync request, pending_geometry=g1
2. configure_with_sync(g2) -> buffered_geometry=g2
3. sync alarm received -> geometry=g1, send configure-notify(g2), send sync request, pending_geometry=g2, buffered_geometry=None
4. sync alarm received -> geometry=g2, buffered_geometry=None

At step 1, the compositor registers a timeout, that's fine.  The issue is that at step 3, the compositor needs to register a *new* timeout.  I suppose one way to do it is:

```rust
pub trait XwmHandler {
    fn sync_request_acked(&mut self, value: i64, next_value: Option<i64>);
}
```

The compositor will remove the existing timeout, and if `next_value` is not `None`, the compositor is expected to register a *new* timeout.  Not sure if this is too confusing...

But with this method, smithay wouldn't have to take a `LoopHandle` at all.  The compositor would get full control over the duration of the timeout and can set whatever policy it wants around what happens after a timeout.  What do you think?

Drakulix:
> The compositor will remove the existing timeout, and if `next_value` is not `None`, the compositor is expected to register a _new_ timeout. Not sure if this is too confusing...
> 
> But with this method, smithay wouldn't have to take a `LoopHandle` at all. The compositor would get full control over the duration of the timeout and can set whatever policy it wants around what happens after a timeout. What do you think?

I think this could work, if we add enough documentation.

kelnos:
Ok, I'm feeling pretty good about this now.  I did manage to push most of the logic into smithay, while leaving the responsibility of the timeout to the compositor.  Resize is pretty smooth, and the black bar along the resize edge has been eliminated.

There's an annoying issue though:

I was checking my implementation against Mutter's, and found that when a sync request is in-flight, and Mutter needs to send a new configure to the client (because the user has continued resizing), it does actually send those configures to the client.  It doesn't start a new sync request, but it does something that seems clever: when the sync request is ACKed, it waits until the *next* render-paint cycle to finish before setting `_XWAYLAND_ALLOW_COMMITS=1` on the window.  That allows the client to catch up to the successive configures and paint at the new size[0].  That way, the final geometry known by the client matches what the compositor wants it to be.

(I actually don't understand how that works in practice, because presumably while Mutter is waiting for the next render-paint cycle to finish, the user is still continuing to resize, and so a new sync request has been started, with new configure events sent.  It seems reasonable to expect that sometimes the client will start painting at one of these new sizes, and that will be inconsistent with the size ACKed in the *first* sync request.  But it *does* work, empirically, so I'm clearly missing something.)

Anyhow, we can't do that with smithay.  `X11Surface` has no idea about the compositor's render loop, and can't implement that delay.  We could expose an API so that the compositor can do this fixup inside its render loop, but that feels like a very awkward API to use: easy to forget, easy to get wrong, and also just conceptually odd.

So instead, when the compositor calls `configure_with_sync()` and there's a sync request in progress, smithay drops the configure request, because it doesn't want the client to paint at a size that doesn't agree with the configure that triggered the sync request.  Once the sync request is ACKed, smithay immediately sets `_XWAYLAND_ALLOW_COMMITS=1`, and it all seems to work fine, with no jerkiness (assuming the client paints fast enough), and no black bars on the resize edge.

But there's one problem: if the compositor doesn't send another configure after the sync request is ACKed, and there were other configures that were dropped while the sync request was in progress, those configures just get lost.  Consider:

1. User drags pointer 1 pixel during resize to geometry `g1`: start sync request, `configure(g1)`.
2. User drags pointer 1 more pixel to resize to geometry `g2`: configure is dropped.
3. User releases mouse pointer to end resize, geometry still at `g2`: final configure is dropped.
4. Client ACKs sync request.  surface geometry=`g1`, client has only painted at `g1`.  User sees the window stop resizing 1 pixel short of where they stopped.

One possibility for eliminating the above issue that I tried was to store a `buffered_geometry` when `configure_with_sync()` is called while a sync request is in progress.  It won't send a new sync request or configure, but when the sync gets ACKed, it immediately starts a new sync request and configure at the buffered geometry.  The trait handler method had another param that would let the compositor know it needs to register a new timeout handler.  Unfortunately this just gave me super-jerky resizes.  I couldn't conclusively figure out why, but my best theory is that `_XWAYLAND_ALLOW_COMMITS` was getting set to `1` right after the ACK was processed, but then immediately being flipped back to `0` (at the start of the next auto-sync), and often it was `1` for too short a time for the XWayland server to have time to commit a new buffer.

So right now I'm pushing responsibility for fixing this into the compositor.  When Anvil finishes a resize operation, and `configure_with_sync()` returns `Ok(SyncedConfigureResult::SyncInProgress)`, Anvil stashes the last geometry.  When it gets the handler call to say the sync request was ACKed, it sees if it has that last geometry stored, and if so, starts a new `configure_with_sync()`.  This works, and it turns out isn't *all* that much more code, but it's annoying that the compositor has to know that it needs to do this.

In the end this isn't too bad, though, I think.  I don't love that the compositor has to handle some of what are essentially protocol details, but the doc comments tell the compositor author what to do (please let me know if there's anything about the docs I can make clearer), and the change to Anvil to implement it was only an extra 115 lines of code.  For reference, the older draft that had a more minimal protocol impl in smithay required about 250 lines of code in Anvil to implement, and quite a few of those lines were not straightforward.

Drakulix:
> One possibility for eliminating the above issue that I tried was to store a `buffered_geometry` when `configure_with_sync()` is called while a sync request is in progress. It won't send a new sync request or configure, but when the sync gets ACKed, it immediately starts a new sync request and configure at the buffered geometry. The trait handler method had another param that would let the compositor know it needs to register a new timeout handler. Unfortunately this just gave me super-jerky resizes. I couldn't conclusively figure out why, but my best theory is that `_XWAYLAND_ALLOW_COMMITS` was getting set to `1` right after the ACK was processed, but then immediately being flipped back to `0` (at the start of the next auto-sync), and often it was `1` for too short a time for the XWayland server to have time to commit a new buffer.

So what I am not understanding here is, that theoretically calling `sync_request_acked` and then letting anvil call `configure_with_sync` should be exactly the same as `X11Wm` calling `configure_with_sync` on it's own before calling into anvil (to set any new timeouts), right?

Anvil isn't multi-threaded, so there can't be anything being processed or happening in between, if we don't go through an idle-callback. Could it simply be, that this inserts a `flush` or something somewhere in between? This sounds like something one should be able to figure out by stepping through the two variants with a debugger. Or what am I missing?

Otherwise I definitely agree, this new approach looks much cleaner to me on first sight. Still going through the code to do a more in-depth review of everything and the new documentation.

src/xwayland/xwm/surface.rs
```diff
@@ -380,9 +434,265 @@ impl X11Surface {
...
+    /// it fires.
+    ///
+    /// If a pending sync is already in progress, this configure is dropped and
+    /// [`SyncedConfigureResult::SyncInProgress`] is returned. The caller should not cancel an
+    /// existing timeout handler, or register a new timeout.  However, if the caller needs the
+    /// window's geometry to "settle" on `rect` (for example, if this is the final configure being
+    /// sent at the end of a resize operation), then the caller should save the rect that was
+    /// passed, and call [`X11Surface::configure_with_sync`] one final time in
+    /// [`XwmHandler::sync_request_acked`](super::XwmHandler::sync_request_acked) with the saved
+    /// rect.
```
Drakulix: It is not super clear to me, what happens in case the last sync-configure times out. Would the compositor need to send the final configure in the timeout handler or with `sync_request_acked` still happen to be called?

src/xwayland/xwm/surface.rs
```diff
@@ -380,9 +434,265 @@ impl X11Surface {
...
+    ///
+    /// If the client does not support the `_NET_WM_SYNC_REQUEST` protocol, a regular configure
+    /// sequence is initiated (as if [`X11Surface::configure`] was called), and
+    /// [`SyncedConfigureResult::ConfigureSent`] will be returned.  This will also happen in the
+    /// special case where the passed `rect` has the same size as the current geometry, and there
+    /// is no sync request in flight.  The caller should not register a timeout.
+    pub fn configure_with_sync(
+        &self,
+        rect: impl Into<Rectangle<i32, Logical>>,
+    ) -> Result<SyncedConfigureResult, X11SurfaceError> {
```
Drakulix: Couldn't we let compositor's pass a `LoopHandle` here and register the timeout in this method for them?

Then have a `sync_request_acked` and `sync_request_timed_out` handler in `X11WmHandler`? The token could be stored without type-erasure in the `X11Surface`.

Drakulix (commented):
Looks mostly good. Just a few more conceptual questions and one slight documentation remark. Really liking how this is turning out!

kelnos:
> So what I am not understanding here is, that theoretically calling `sync_request_acked` and then letting anvil call `configure_with_sync` should be exactly the same as `X11Wm` calling `configure_with_sync `on it's own before calling into anvil (to set any new timeouts), right?

Ah, Anvil doesn't do that, though: the next `configure_with_sync` happens when the user drags the pointer to continue the resize.

> Could it simply be, that this inserts a `flush` or something somewhere in between? This sounds like something one should be able to figure out by stepping through the two variants with a debugger. Or what am I missing?

That's a good question.  I don't think it's specifically a missing `flush`, as `set_allow_commits()` does do a `flush`, as does sending the sync-request+configure.

But I do wonder if doing a full round-trip when setting `_XWAYLAND_ALLOW_COMMITS` could help.  Not just `flush`, but `conn.set_property(...)?.check()?`.  My gut feeling is no, it won't, but maybe it's worth a try.

src/xwayland/xwm/surface.rs
kelnos: Yes, and Anvil does indeed do this on timeout, but that should be called out in the docs.

src/xwayland/xwm/surface.rs
kelnos: Hmm, yes, we could register the timeout in `configure_with_sync` and store the token, but then we'd have to pass the token back to the compositor in `sync_request_acked` so they can remove it.  That feels kinda awkward.  But it would work.

However, if I *can* fix it so `X11Wm` or `X11Surface` can immediately start a new `configure_with_sync` if it has a buffered configure, then it won't have the `LoopHandle` at that point with which to register a new timeout, so that just wouldn't work.

And either way, `configure_with_sync` taking the `LoopHandle` would go against your desire not to add more coupling to calloop...

Drakulix: > That feels kinda awkward. But it would work.

Right, I forgot, that you cannot disarm the timer with just the token.

In that case the current api definitely feels like the best compromise.

Drakulix:
> > So what I am not understanding here is, that theoretically calling `sync_request_acked` and then letting anvil call `configure_with_sync` should be exactly the same as `X11Wm` calling `configure_with_sync `on it's own before calling into anvil (to set any new timeouts), right?
> 
> Ah, Anvil doesn't do that, though: the next `configure_with_sync` happens when the user drags the pointer to continue the resize.

Or when the user lets go of the button I suppose. Either way, this makes sense to me now.

kelnos:
I think I'm going to work on this a bit more.  I'm just really not happy with the need to "save" the final configure when there's already a sync in progress.

I want to try supporting `_NET_WM_FRAME_DRAWN`, ~~which is a message from the client that's effectively a buffer commit.  That can act as a trigger to set `_XWAYLAND_ALLOW_COMMITS` back to 1, and there's a good chance that will just work.  Then I can keep sending configure events while a sync is active, and wait until i see `_FRAME_DRAWN` with the appropriate geometry before releasing xwayland to commit buffers.~~

GTK3 and GTK4 implement `_FRAME_DRAWN`, though I don't think qt does.  Maybe that's ok: in the end I kinda consider a lot of this stuff "best effort".  If *some* X11 apps have artifacts during resize, maybe that's ok, as X11 is the legacy/back-compat path now.

(There's also `_NET_WM_FRAME_TIMINGS`, which is server->client and is the X11 analog of wp-presentation-time.  I don't think that's going to help with this particular problem, but it could be nice to implement that as well at some point.)

**Edit**: I've misunderstood `_FRAME_DRAWN`.  I thought it was client->WM, so the client can tell the WM when it's done painting.  but it's actually WM->client, so the WM can tell the client that it has seen the damage the client has done and queued the new window content for re-compositing/redraw.  When the WM supports `_FRAME_DRAWN`, GTK will freeze its frame clock after finishing a paint cycle, and won't unfreeze it until it gets a corresponding `_FRAME_DRAWN`.

`_FRAME_TIMINGS` is indeed what I thought: after the page flip, and the WM/compositor has presentation feedback for a frame, it can send that feedback to the client.

kelnos:
Oh jeez, I just remembered something: the XSync extension has a built-in counter called SERVERTIME.  xwayland implements this as a millisecond-precision clock.  we don't need the compositor/event loop to handle the sync timeout at all!  i can fetch the current value of the counter, add 1000 to it, and set an alarm for that value.

`configure_with_sync()` can take a `timeout: Option<Duration>` (where `None` will use a default of 1s) and handle everything itself.

And if I can also figure out how to automatically send buffered configures after an in-progress sync request finishes, then the compositor won't have to do *anything* except replace `configure()` with `configure_with_sync()`.  No timeouts, no need for `XwmHandler::sync_request_acked()`, even!  That would be really nice...

kelnos:
Neat, using the xsync `SERVERTIME` counter for the timeout has dropped the net new number of lines in anvil to 69.  Still want to try to figure out how to eliminate the need to store and finalize the configure sequence at the end of the resize...

kelnos:
Success!

The change to Anvil is now just four lines: replacing each `configure()` in the resize grab with `configure_with_sync()`.  That's it!

Notable stuff:

1. XSync `SERVERTIME` counter & sync alarms are used for the sync request timeout, so the compositor doesn't need to provide a `LoopHandle` or manage the timeout itself.
2. In the case where the compositor wants to send more configures while a sync request is in progress, we keep track of the latest configure rect inside `X11Surface` and do nothing.  After the client ACKs the sync,  we hold `_XWAYLAND_ALLOW_COMMITS` at 1 until xwayland commits, and in the pre-commit handler we start the next sync using the buffered rect, which sets `_XWAYLAND_ALLOW_COMMITS` to 0.  This way, the compositor doesn't have to check the return value of `configure_with_sync()`
3. I left in the `sync_request_acked()` and `sync_request_timeout()` items on `XwmHandler`.  The compositor doesn't actually need to implement them anymore.  I thought maybe they might be useful for logging/metrics, but I can remove them if you think they're not useful.
4. For the `SERVERTIME` counter, instead of fetching it over and over again inside the `X11Surface` constructor (or the first time it's needed, in each surface), I fetch it once in `X11Wm` and then pass it to `X11Surface::new()`.  That means the signature of `::new()` has changed.  Is that ok?
5. ~~I've changed the `SpaceElement::bbox()` impl for `X11Surface` to return the committed buffer size, not the geometry stored inside `X11Surface`.  The issue is that the client is free to reject the size the WM sends to it (size increments, min size, max size, etc.), and if it does, `::bbox()` will return something that doesn't actually match the size of the surface.  I ran into this a few times while testing, and it caused visual artifacts.  Since the purpose of this commit is "artifact-free resizing", I figured it would be appropriate to fix that here too.  I can put it in a separate PR if you'd like.~~ Moved to #2030.
6. ~~Related to the above, `<X11Surface as SpaceElement>::bbox()` now calls `compositor::with_states()`.  I was concerned that this wasn't safe to do there, but `Window`'s impl of `SpaceElement::geometry()` does the same thing, so hopefully that's ok?~~
7. Calling plain old `configure()` while a sync request is in progress cancels the sync request and immediately sends the configure.

 I'm feeling like this is ready for a proper review!

src/xwayland/xwm/mod.rs
```diff
@@ -494,6 +501,16 @@ pub trait XwmHandler {
         let _ = (xwm, window, timestamp);
     }
 
+    /// A client has acknowledged a `_NET_WM_SYNC_REQUEST` request.
```
Drakulix, line 504: If we keep these (which I am slightly hesitant to), we should definitely expand the docs to note that implementing these is strictly optional.

Drakulix (approved):
So this looks great, I'd be happy to merge this as is, but to really make it useful for cosmic-comp, I'd need to ask you for a little more api.

For wayland-surfaces we already make use of different "geometries" at a time, depending on what we need for specific window management: https://github.com/pop-os/cosmic-comp/blob/master/src/shell/element/surface.rs#L180-L207

Sometimes I need the size of a window currently visible (committed) vs. the last one acked vs. the last one we send it.

As such I think we want to add getters for `pending_geometry` and `buffered_geometry` to the `X11Surface`.

src/xwayland/xwm/mod.rs
kelnos: Done!  And standing ready to remove them entirely if that's what you decide :)

kelnos:
> As such I think we want to add getters for `pending_geometry` and `buffered_geometry` to the X11Surface.

Yup, makes sense.  Added.

Drakulix (approved):
Looks great! Thanks a bunch for being patient with the whole process. :)

kelnos:
> *hanks a bunch for being patient with the whole process. :)*

No problem at all!  Compared to many open source projects, your process is easy and quick :D

PR #2019, opened 2026-05-03
xwm: lazily update opaque regions when the property changes

kelnos:
## Description

During resize, a client with _NET_WM_OPAQUE_REGION set will send a flood of changes to this property as the window changes size and the opaque regions move and change size.  Profiling shows that processing all of these updates can take 10-15ms, causing missed vblanks.

Not only does it take a while to process the flood of updates, fetching the updated property value from the XWayland server can be very slow during this period (I've observed a median of 3ms and P90 of 12ms, with a P99 of 27ms!).  My theory is that x11rb's socket processing architecture appears to be working against us here: in this flood of updates during a resize, a lot of other things are also going on: configure requests, configure notify events, etc.  This causes x11rb's incoming event processing to back up, and with the property fetch reply far down the queue of events to process, it takes a while for it to get there.

Instead, when receiving a PropertyNotify event for _NET_WM_OPAQUE_REGION, simply None-out the stored value and mark it as dirty.  In the pre-commit hook, if the value is dirty, re-fetch the property.  This avoids several tens of slow property fetches per frame, coalescing it down to one per frame at most.

I've also taken pains to avoid holding the mutex on SharedSurfaceState across the XWayland server round-trip; since that call can take several milliseconds when there's a flood of events, a multi-threaded compositor might experience lock contention if it's trying to do anything else with the X11Surface.

This doesn't fully fix the problem: the property fetch in the pre-commit handler can still be slow to the tune of several milliseconds (and sometimes worse) during resize.  But this change does eliminate most of the missed vblanks and skipped frames.  (If x11rb were to expose poll_for_reply(), we could do this in a non-blocking manner, and completely eliminate this issue, but it's a private function.)

In normal non-resize situations, the opaque region will change only rarely (if ever), and even when it does, it should presumably be at a time when there isn't a flood of traffic with the XWayland server, so the lazy re-fetch in the pre-commit handler should complete in microseconds.

## Checklist

### DIFF
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -97,6 +97,7 @@ pub(crate) struct SharedSurfaceState {
     window_type: Vec<Atom>,
     pub(crate) opacity: Option<u32>,
     opaque_region: Option<RegionAttributes>,
+    opaque_region_dirty: bool,
     pending_enter: Option<(
         Box<dyn std::any::Any + Send + 'static>,
         Vec<Keycode>,
@@ -244,6 +245,7 @@ impl X11Surface {
                 window_type: Vec::new(),
                 opacity: None,
                 opaque_region: None,
+                opaque_region_dirty: true,
                 pending_enter: None,
             })),
             xdnd_active,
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Thanks!

PR #2020, opened 2026-05-04
xwm: implement support for _NET_WM_PING

kelnos:
## Description

This adds `X11Surface::ping()`, which, if supported by the client, sends a `_NET_WM_PING` message to the window.  When the client responds, `XwmHandler::ping_acked()` is called.

## Checklist

### DIFF
--- a/src/desktop/wayland/window.rs
+++ b/src/desktop/wayland/window.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "wayland_frontend")]
+use crate::utils::Serial;
 #[cfg(feature = "xwayland")]
 use crate::{desktop::space::SpaceElement, xwayland::X11Surface};
 use crate::{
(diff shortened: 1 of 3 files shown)


Drakulix (commented):
Looks good to me and seems mostly analogous to how the ping/pong mechanism of `XdgShell` is implemented.

However I wonder, if we now could also add to our desktop-helpers. E.g. provide a `ping`-method on `desktop::Window`, that dispatches into the right shell's method?

kelnos:
> However I wonder, if we now could also add to our desktop-helpers. E.g. provide a ping-method on `desktop::Window`, that dispatches into the right shell's method?

Ah, good idea.  How should we handle the callback for the ping reply?  Still route it to `XdgShellHandler` or `XwmHandler` depending on which type of window it is?

Looks like there is one semantic difference: the xdg ping will return an error if there's already a ping in flight, but for X11 it sends a new ping and drops the old one.  For X11, the spec doesn't say anything about this, so I think it would be fine to make X11 behave the same way.

Drakulix:
> Looks like there is one semantic difference: the xdg ping will return an error if there's already a ping in flight, but for X11 it sends a new ping and drops the old one. For X11, the spec doesn't say anything about this, so I think it would be fine to make X11 behave the same way.

Yeah, we probably want to adjust this little inconsistency.

> Still route it to XdgShellHandler or XwmHandler depending on which type of window it is?

I think that is the best option. Alternatively we could introduce a new PingHandler, that both these traits depend on, but I don't see a neat way to abstract over the different window types without introducing a dependency on the `desktop`-feature. So I'd rather document the possible `pong`-callbacks.

kelnos:
> Yeah, we probably want to adjust this little inconsistency.

Done!

Implemented `Window::send_ping()`, and documented that the caller needs to implement the surface-specific handler to get replies.

Drakulix (approved):
Looks great! Thanks :)

PR #2021, opened 2026-05-04
drm/surface: Make sure we also reset `pending`

Drakulix:
It is possible to end up in the very unfortunate situation of just setting the `pending` state of a `DrmSurface` before a call to `DrmDevice::pause`.

This *seems* like a good thing, because after `activate`, we presumably will transition to the new pending state without loosing any user operation. However a suspend/resume cycle might render the `Mode`-blob created and set in the pending state invalid... Theoretically a connector could also disappear (with MST).

Since most compositor will re-evaluate the current state and then apply their output configuration anew after a suspend/resume cycle anyway, I argue we should simply reset `pending` to the current state as well.

## Checklist


cmeissl (approved):
LGTM

PR #2022, opened 2026-05-07
xwm: Set `WM_STATE` to match `_NET_WM_STATE_HIDDEN`, instead of mapped

ids1024:
## Description
This fixes an issue on cosmic-comp with Dishonored 2 (the GOG version started in Heroic Game Launcher, though that probably doesn't matter). Prior to this change, when the game loses focus it would auto-minmize, but when unminimized the window would remain black, and not restore itself to fullscreen. With this change, it becomes fullscreen again and works.

It's running through wine, so I'm not sure exactly how that's using `WM_STATE` vs other properties.

I'm not 100% sure this is the correct way to do this, though I guess it's a valid implementation anyway?

## Checklist

### DIFF
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -263,29 +263,14 @@ impl X11Surface {
         if let Some(conn) = self.conn.upgrade() {
             if let Some(frame) = self.state.lock().unwrap().mapped_onto {
                 if mapped {
-                    let property = [1u32 /*NormalState*/, 0 /*WINDOW_NONE*/];
-                    conn.change_property32(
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
I'd see this mostly as cleanup of the new code, that doesn't use unmapping/mapping anymore for minimize/restore. Imo the WM_STATE should correctly reflect that and if it fixes some wine games, that is a good indicator.

hojjatabdollahi:
This introduced an issue:
If you have a wayland native game (add "PROTON_ENABLE_WAYLAND=1") when you open the game, the game controller input goes to steam in the background, and not the game.

PR #2025, opened 2026-05-07
wayland/drm_syncobj: Allow temporarily close the device

Drakulix:
`libseat` [recommends to re-open file descriptors](https://git.sr.ht/~kennylevinsen/seatd/tree/master/item/include/libseat.h#L17) after suspend/resume, since they might have been permanently revoked.

Unfortunately it doesn't wire-up logind's resume event, which provides new file descriptors, to pass those on. As a result compositors need to issue a new open-call, which will not succeed unless the old file descriptor has been closed.

Since we put the `DrmDeviceFd` into internal `Arc`s (yes multiple), this can be a bit annoying and one part where we completely fail to provide a way to re-open the FD is our `drm_syncobj`-protocol implementation.

You can update the internal device in case it disappears and you want to select a new one. But if you simply want to re-open the file descriptor, there is no way to do that without invalidating all existing timelines in the process (which is obviously bad and will hang some clients due to unsignaled release points).

There isn't really an elegant way to do this, so this PR provides a way to temporarily close the internal file descriptor of the global's state.

## Checklist

### DIFF
--- a/src/wayland/drm_syncobj/mod.rs
+++ b/src/wayland/drm_syncobj/mod.rs
@@ -123,7 +123,7 @@ impl Cacheable for DrmSyncobjCachedState {
 #[derive(Debug)]
 pub struct DrmSyncobjState {
     global: GlobalId,
-    import_device: DrmDeviceFd,
+    import_device: Option<DrmDeviceFd>,
     known_timelines: Vec<Weak<DrmTimelineInner>>,
 }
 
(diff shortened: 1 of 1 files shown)


src/wayland/drm_syncobj/mod.rs
```diff
@@ -157,11 +155,19 @@ impl DrmSyncobjState {
...
-            import_device,
+            import_device: Some(import_device),
             known_timelines: Vec::new(),
         }
     }
 
+    /// Closes the current `import_device`, allowing compositors to acquire a new fd.
+    ///
+    /// Note: Any existing timeline objects will error out and hang clients,
+    /// until `update_device` is called with a new device fd.
```
ids1024: Well, if we use `post_error`, that won't just *hang* the client, it will cause the client to crash when its Wayland connection is broken by the protocol error, right?

Drakulix: For new timelines this is correct, but for old timelines `signal` and similar will be noops.

Drakulix:
@ids1024 Ended up adding a `Guard` to make it less easy to break the api.

ids1024 (approved):
Having a `#[muse_use]` `CloseGuard` seems good. Since we have a reference to the `DrmSyncobjState`, we know another thread can't be dispatching Wayland events, so as long as `update_device()` is called, I guess this is good, and clients will continue to import syncobjs afterwards. (While the compositor fails to open a device at all, I guess protocol errors may be reasonable. But hopefully that doesn't happen.)

PR #2027, opened 2026-05-08
Add PointerConstraintsHandler trait for remove_constraint

skygrango:
## Description

According to the protocol [description](https://wayland.app/protocols/pointer-constraints-unstable-v1#zwp_locked_pointer_v1:request:set_cursor_position_hint): 

>If the client is drawing its own cursor, it should update the position hint to the position of its own cursor. A compositor may use this information to warp the pointer upon unlock in order to avoid pointer jumps.

Therefore, we need to ensure that there is a callback to notify the compositor when the constraint is removed.

## Checklist

### DIFF
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -172,6 +172,7 @@ pub struct AnvilState<BackendData: Backend + 'static> {
     pub seat: Seat<AnvilState<BackendData>>,
     pub clock: Clock<Monotonic>,
     pub pointer: PointerHandle<AnvilState<BackendData>>,
+    pub cursor_position_hint: Option<(WlSurface, Point<f64, Logical>)>,
 
     #[cfg(feature = "xwayland")]
     pub xwm: Option<X11Wm>,
@@ -376,6 +377,25 @@ impl<BackendData: Backend> PointerConstraintsHandler for AnvilState<BackendData>
         }
     }
 
+    fn remove_constraint(&mut self, surface: &WlSurface, pointer: &PointerHandle<Self>) {
+        if with_pointer_constraint(surface, pointer, |constraint| constraint.is_none()) {
+            if let Some((hint_surface, hint_location)) = &self.cursor_position_hint {
+                let origin = self
+                    .space
+                    .elements()
+                    .find_map(|window| {
+                        (window.wl_surface().as_deref() == Some(hint_surface)).then(|| window.geometry())
+                    })
+                    .unwrap_or_default()
+                    .loc
+                    .to_f64();
+
+                pointer.set_location(origin + *hint_location);
+            }
+            self.cursor_position_hint = None;
+        }
+    }
+
     fn cursor_position_hint(
         &mut self,
         surface: &WlSurface,
@@ -385,17 +405,7 @@ impl<BackendData: Backend> PointerConstraintsHandler for AnvilState<BackendData>
         if with_pointer_constraint(surface, pointer, |constraint| {
             constraint.is_some_and(|c| c.is_active())
         }) {
-            let origin = self
-                .space
-                .elements()
-                .find_map(|window| {
-                    (window.wl_surface().as_deref() == Some(surface)).then(|| window.geometry())
-                })
-                .unwrap_or_default()
-                .loc
-                .to_f64();
-
-            pointer.set_location(origin + location);
+            self.cursor_position_hint = Some((surface.clone(), location));
         }
     }
 }
@@ -749,6 +759,7 @@ impl<BackendData: Backend + 'static> AnvilState<BackendData> {
             seat_name,
             seat,
             pointer,
+            cursor_position_hint: None,
             clock,
 
             #[cfg(feature = "xwayland")]
--- a/src/wayland/pointer_constraints.rs
+++ b/src/wayland/pointer_constraints.rs
@@ -37,6 +37,11 @@ pub trait PointerConstraintsHandler: SeatHandler {
     /// Use [`with_pointer_constraint`] to access the constraint.
     fn new_constraint(&mut self, surface: &WlSurface, pointer: &PointerHandle<Self>);
 
+    /// Constraint removed for `pointer` on `surface`
+    ///
+    /// Use [`with_pointer_constraint`] to access the constraint.
+    fn remove_constraint(&mut self, surface: &WlSurface, pointer: &PointerHandle<Self>);
+
     /// The client holding a LockedPointer has committed a cursor position hint.
     ///
     /// This is emitted upon a surface commit if the cursor position hint has been updated.
@@ -344,12 +349,23 @@ fn add_constraint<D: SeatHandler + PointerConstraintsHandler + 'static>(
     }
 }
 
-fn remove_constraint<D: SeatHandler + 'static>(surface: &WlSurface, pointer: &PointerHandle<D>) {
-    with_constraint_data::<D, _, _>(surface, |data| {
+fn remove_constraint<D: SeatHandler + PointerConstraintsHandler + 'static>(
+    state: &mut D,
+    surface: &WlSurface,
+    pointer: &PointerHandle<D>,
+) {
+    let is_removed = with_constraint_data::<D, _, _>(surface, |data| {
         if let Some(data) = data {
-            data.constraints.remove(pointer);
+            if let Some(_constraint) = data.constraints.remove(pointer) {
+                return true;
+            }
         }
+        false
     });
+
+    if is_removed {
+        state.remove_constraint(surface, pointer);
+    }
 }
 
 impl<D> Dispatch2<ZwpPointerConstraintsV1, D> for GlobalData
@@ -460,7 +476,7 @@ where
 
 impl<D> Dispatch2<ZwpConfinedPointerV1, D> for PointerConstraintUserData<D>
 where
-    D: SeatHandler,
+    D: SeatHandler + PointerConstraintsHandler,
     D: 'static,
 {
     fn request(
@@ -493,7 +509,7 @@ where
 
     fn destroyed(
         &self,
-        _state: &mut D,
+        state: &mut D,
         _client: wayland_server::backend::ClientId,
         _resource: &ZwpConfinedPointerV1,
     ) {
@@ -501,13 +517,13 @@ where
             return;
         };
 
-        remove_constraint(&self.surface, pointer);
+        remove_constraint(state, &self.surface, pointer);
     }
 }
 
 impl<D> Dispatch2<ZwpLockedPointerV1, D> for PointerConstraintUserData<D>
 where
-    D: SeatHandler,
+    D: SeatHandler + PointerConstraintsHandler,
     D: 'static,
 {
     fn request(
@@ -545,7 +561,7 @@ where
 
     fn destroyed(
         &self,
-        _state: &mut D,
+        state: &mut D,
         _client: wayland_server::backend::ClientId,
         _resource: &ZwpLockedPointerV1,
     ) {
@@ -553,6 +569,6 @@ where
             return;
         };
 
-        remove_constraint(&self.surface, pointer);
+        remove_constraint(state, &self.surface, pointer);
     }
 }


src/input/mod.rs
```diff
@@ -163,6 +164,9 @@ pub trait SeatHandler: Sized {
 
     /// Callback that will be notified whenever the keyboard led state changes.
     fn led_state_changed(&mut self, _seat: &Seat<Self>, _led_state: LedState) {}
+
+    /// Callback that will be notified whenever constraint was removed
+    fn remove_constraint(&mut self, _surface: &WlSurface, _pointer: &PointerHandle<Self>) {}
```
ids1024: This should probably be part of `PointerConstraintsHandler`, which already has a corresponding `new_constraint` method.

skygrango: This is a brilliant idea! Update.

skygrango: update cosmic-comp: https://github.com/pop-os/cosmic-comp/compare/ac3c14907d9fbfa23ebd2924531be3555dd94d4b..6ff99251af34fbbefd00b01e27d0453bd6f7fd77

src/wayland/pointer_constraints.rs
```diff
@@ -344,12 +344,23 @@ fn add_constraint<D: SeatHandler + PointerConstraintsHandler + 'static>(
...
-            data.constraints.remove(pointer);
+            if let Some(constraint) = data.constraints.remove(pointer) {
+                return constraint.is_active();
+            }
         }
+        false
     });
+
+    if was_active {
+        state.new_constraint(surface, pointer);
```
Drakulix: Do we want to call `new_constraint` here or rather add a new `remove_constraint`-method to the `PointerConstraintsHandler`? Because this seems pretty unintuitive.

skygrango: That was indeed my initial idea, but perhaps my original approach wasn't the best, so let's do it this way instead

skygrango: Done :+1:

src/wayland/pointer_constraints.rs
```diff
@@ -344,12 +349,23 @@ fn add_constraint<D: SeatHandler + PointerConstraintsHandler + 'static>(
...
+            if let Some(constraint) = data.constraints.remove(pointer) {
+                return constraint.is_active();
+            }
         }
+        false
     });
+
+    if was_active {
+        state.remove_constraint(surface, pointer);
+    }
```
ids1024: Should this be done unconditionally, whether of not the constraint `is_active()`?

For `persistent` constraints, `activate`/`deactivate` can be called multiple times, and the constraint presumably may or may not be active when it is removed. But we call `add_constraint` and `remove_constraint` when the constraint is added or removed, not when it's activated or deactivated.

(There could be separate methods for handling constraint activation/deactivation, but those aren't needed as long as `activate`/`deactivate` are something that's called by the compositor, not internally be smithay.)

skygrango: good catch. the purpose in using this callback was simply to allow compositor to check whether a cursor_position_hint needed to be applied, because applying the hint at the right time is the best way to ensure compatibility with mouse behavior across different game engines. 

Although this is the only purpose at present, the responsibility and conditions of use should indeed be determined by compositor.

Fix.

ids1024 (approved):

anvil/src/state.rs
```diff
@@ -376,6 +377,25 @@ impl<BackendData: Backend> PointerConstraintsHandler for AnvilState<BackendData>
         }
     }
 
+    fn remove_constraint(&mut self, surface: &WlSurface, pointer: &PointerHandle<Self>) {
+        if with_pointer_constraint(surface, pointer, |constraint| constraint.is_none()) {
```
YaLTeR, line 381: I'm updating niri to these changes and trying to understand them better. What's the purpose of this check, isn't it guaranteed to be true since this constraint was just removed before calling `remove_constaint()`?

anvil/src/state.rs
```diff
@@ -376,6 +377,25 @@ impl<BackendData: Backend> PointerConstraintsHandler for AnvilState<BackendData>
...
+            if let Some((hint_surface, hint_location)) = &self.cursor_position_hint {
```
YaLTeR, line 382: Is it correct to use stored surface and location here? Like, is it possible that `remove_constraint()` is called with a different surface than the stored one?

anvil/src/state.rs
skygrango: Yes, for the `Lifetime::Oneshot` managed by Smithay, this is guaranteed to happen.

I adopted a defensive coding approach for `Lifetime::Persistent`, If `deactivate()` is explicitly called, the current implementation of `remove_constraint()` does not directly handle the cursor, leaving that decision to the compositor.

I don't fully understand the current usage of `Lifetime::Persistent`, so if you think that pattern should be applied here as well, you can switch to using `constraint.is_some_and(|c| !c.is_active())` to handle it.

anvil/src/state.rs
skygrango: In my implementation for cosmic-comp, the call is currently always initiated by the original surface, no exceptions have been found in tests so far

anvil/src/state.rs
YaLTeR: hmm I see thanks

YaLTeR:
So I'm not sure that this PR allows to correctly handle the cursor hint for `Lifetime::Persistent` constraints. Say you have a persistent locked pointer on a surface which set a cursor location hint. You alt-tab out in your compositor or something, causing the pointer to leave the surface, and the constraint to deactivate. However, it's not destroyed since it's persistent, and therefore `remove_constraint()` is not called, so the compositor doesn't get a chance to apply the cursor position hint (which should be applied since the cursor was unlocked). Then, sometime later, if the constraint is finally destroyed, `remove_constraint()` will run and move the pointer even if the user was already doing something unrelated to the surface.

Even with `Lifetime::Oneshot`I believe there's an issue if the client decides to destroy the constraint some delay after the compositor unlocks it? Actually, is `remove_constraint()` even called in deactivate in this case?

I think the right way would be to call `remove_constraint()` right when the constraint is deactivated *or* destroyed?

skygrango:
> So I'm not sure that this PR allows to correctly handle the cursor hint for `Lifetime::Persistent` constraints. Say you have a persistent locked pointer on a surface which set a cursor location hint. You alt-tab out in your compositor or something, causing the pointer to leave the surface, and the constraint to deactivate. However, it's not destroyed since it's persistent, and therefore `remove_constraint()` is not called, so the compositor doesn't get a chance to apply the cursor position hint (which should be applied since the cursor was unlocked). Then, sometime later, if the constraint is finally destroyed, `remove_constraint()` will run and move the pointer even if the user was already doing something unrelated to the surface.

The motivation behind this PR is to address a limitation in the current design. When a client continuously updates its cursor hint, the compositor has no hook to defer applying the hint until the appropriate point in the event lifecycle. Instead, it can only apply each update immediately. Continuously updating the cursor position based on the client's cursor hints can result in incorrect behavior, as it interferes with the client's own cursor state. This is the reason for storing the cursor position hint.

> Even with `Lifetime::Oneshot`I believe there's an issue if the client decides to destroy the constraint some delay after the compositor unlocks it? Actually, is `remove_constraint()` even called in deactivate in this case?

According to the [protocol](https://wayland.app/protocols/pointer-constraints-unstable-v1#zwp_locked_pointer_v1:event:unlocked): If this is a oneshot pointer lock (see wp_pointer_constraints.lifetime) this object is now defunct and should be destroyed. 

> I think the right way would be to call `remove_constraint()` right when the constraint is deactivated _or_ destroyed?

Due to the lack of sufficient use cases, I intentionally did not make `deactivated()` immediately trigger `remove_constraint()` for `Lifetime::Persistent`. I wanted to avoid introducing behavior that might turn out to be incorrect without enough practical validation.

Since the compositor owns the invocation of `deactivate()`, it has enough context to determine whether the constraint is using `Lifetime::Persistent` and, if appropriate, call `remove_constraint()` to apply the stored cursor position hint.

YaLTeR:
> According to the protocol: If this is a oneshot pointer lock (see wp_pointer_constraints.lifetime) this object is now defunct and should be destroyed.

But nothing prevents the client from taking a second or two to process it and only then call destroy.

> Since the compositor owns the invocation of deactivate(), it has enough context to determine whether the constraint is using Lifetime::Persistent and, if appropriate, call remove_constraint() to apply the stored cursor position hint.

I don't think the premise here is true. niri never calls deactivate for example, instead it relies on smithay calling deactivate automatically when the surface loses pointer focus, which the compositor doesn't really have control over.

skygrango:
> But nothing prevents the client from taking a second or two to process it and only then call destroy.

This behavior has not been observed during current usage.

> I don't think the premise here is true. niri never calls deactivate for example, instead it relies on smithay calling deactivate automatically when the surface loses pointer focus, which the compositor doesn't really have control over.

Hmmm. In the constrained state, the cursor cannot escape the surface. In general, the cursor can only be forcibly released from a surface when the window focus changes(for example, when Alt+Tab is used, the newly focused window may end up directly underneath the cursor), which should be under the compositor's control. In `cosmic-comp`, I detect changes to the `KeyboardTarget` and call `deactivate()` whenever the focus changes. This ensures that the cursor is released, and if necessary, it also provides an appropriate place to handle `Lifetime::Persistent`.

YaLTeR:
> This behavior has not been observed during current usage.

I mean, I could write a test client I guess

> In general, the cursor can only be forcibly released from a surface when the window focus changes(for example, when Alt+Tab is used, the newly focused window may end up directly underneath the cursor), which should be under the compositor's control. In cosmic-comp, I handle this by monitoring changes to the KeyboardTarget and forcing a deactivate() when the focus changes. This ensures that the cursor is released, and if necessary, it also provides an appropriate place to handle Lifetime::Persistent.

I think this isn't ideal for two reasons: 
- Need extra tracking in the compositor when smithay could just call this callback automatically at the right spot (as other smithay abstractions tend to do)
- I don't think you can track focus changes like this in the compositor *in all cases* because grabs can modify focus, for example the popup grab and such.

skygrango:
>> This behavior has not been observed during current usage.
>  I mean, I could write a test client I guess

I understand your concerns.

>   I think this isn't ideal for two reasons:
>    * Need extra tracking in the compositor when smithay could just call this callback automatically at the right spot (as other smithay abstractions tend to do)
>    * I don't think you can track focus changes like this in the compositor _in all cases_ because grabs can modify focus, for example the popup grab and such.

Although I consider Smithay's automatic deactivate() call to be more of a safety net than the ideal behavior, connecting deactivate() directly to remove_constraint() does indeed eliminate these potential issues.

If I'm not misunderstanding the proposal, then for Lifetime::Persistent to work correctly, PointerTarget::enter would also need to check for an existing constraint and reactivate it. As far as I can see, Smithay doesn't currently do this.

By the way, do you know of any use cases for Lifetime::Persistent?

skygrango:
> If I'm not misunderstanding the proposal, then for Lifetime::Persistent to work correctly, PointerTarget::enter would also need to check for an existing constraint and reactivate it. As far as I can see, Smithay doesn't currently do this.

I was mistaken. That isn't the correct behavior. It should only happen once both the KeyboardTarget and PointerTarget have been assigned.

YaLTeR:
> If I'm not misunderstanding the proposal, then for Lifetime::Persistent to work correctly, PointerTarget::enter would also need to check for an existing constraint and reactivate it. As far as I can see, Smithay doesn't currently do this.

Currently the compositor is responsible for activating the pointer constraint according to its policy (usually on pointer focus enter). I think that's fine to keep as is, and repeated activations for `Lifetime::Persistent` should keep working as-is with the existing code.

My proposal is for Smithay to notify the compositor immediately upon deactivate or destroy by calling `remove_constraint()`. Smithay can keep automatically deactivating upon pointer focus leave, I think that's fine and expected (and call `remove_constraint()`).

> By the way, do you know of any use cases for Lifetime::Persistent?

I haven't checked what the clients do but that sounds like more or less normal behavior for clients? Say you're a 3D game, you want to lock the pointer. If the user alt-tabs out, the pointer will be unlocked, but the game doesn't want to bother re-locking it again when the user is back, so it just sets persistent lifetime, and the compositor can re-lock automatically afterwards?

skygrango:
> I haven't checked what the clients do but that sounds like more or less normal behavior for clients? Say you're a 3D game, you want to lock the pointer. If the user alt-tabs out, the pointer will be unlocked, but the game doesn't want to bother re-locking it again when the user is back, so it just sets persistent lifetime, and the compositor can re-lock automatically afterwards?

As far as I know, 3D games almost always use `Lifetime::Oneshot`. This isn't just because they frequently switch between the locked and confined modes, which requires destroying the existing constraint object before creating a new one. With `Lifetime::Persistent`, the client also has no control over when the constraint is activated or deactivated, so there doesn't seem to be any obvious benefit for clients to use it. In fact, the common practice today is to recreate a `Lifetime::Oneshot` constraint when the pointer returns.

This also explains why the current implementation doesn't fully cover `Lifetime::Persistent`, yet no issues have been reported so far. It appears to be a fairly uncommon use case.

PR #2029, opened 2026-05-09
fix(xwm): Defer X11 focus release between same-client transitions

mayakwd:
## Description
`X11Wm` set `_NET_ACTIVE_WINDOW = NONE` for any FocusOut event on a tracked top-level window, ignoring `detail` field. It breaks X11 clients like Unity Editor. 

Mechanism of the flaw: 
- the client gets focus at the top level and...
- almost instantly switches focus to its sub-window. 
- X-server generates FocusOut(detail=inferior) for the top-level window. 
- code without fix just set `_NET_ACTIVE_WINDOW` to none, and...
- there is no one to receive `FocusIn`.

But that's not the whole issue. `X11Surface` also synchronously calls `set_input_focus(NONE, NONE)`, producing the intermediate state "focus is nowhere".

Fixes: 
- https://github.com/pop-os/cosmic-comp/issues/2006, 
- https://github.com/pop-os/cosmic-epoch/issues/2311
- and potentially many more issues overall for X11 apps related to focus out, like: https://github.com/pop-os/cosmic-comp/issues/1211

Works not only for cosmic-comp but also for other compositors based on Smithay, all of which suffer the same issue.

### Symptoms
Symptoms are exactly as described in https://github.com/pop-os/cosmic-comp/issues/2006:
- Open a project in Unity
- It's smooth and nice
- Press Play, and experience a framerate dropped to 10 frames per second (Unity turns on throttling when losing focus, to not waste resources).
- More than that, the keyboard and mouse are unresponsive inside Game view
- Alt+tab and back - helps to return focus, but framerate is still down.

And many other related to input-unresponsiveness and sluggish experience, which cannot always be resolved by clicking other panels/alt-tabbing. 

### Diagnostics
Instrumentation wired into Smithay showed that:
```
  02:06:02.408  X11Surface::enter                          WINDOW=29360433 MODEL=LocallyActive
  02:06:02.412  xwm Event::FocusIn   WINDOW=29360433 detail=NONLINEAR tracked=true
  02:06:02.412  xwm: _NET_ACTIVE_WINDOW = window           WINDOW=29360433 
  02:06:02.462  xwm Event::FocusOut  WINDOW=29360433 detail=INFERIOR tracked=true
  02:06:02.462  xwm: _NET_ACTIVE_WINDOW = NONE             WINDOW=29360433   <-- issue
```
Meaning that Unity switched focus from the top level to a sub-window, but _NET_ACTIVE_WINDOW became **none**.

### How the fix works
- `FocusOut` now doesn't touch `_NET_ACTIVE_WINDOW`
- Losing focus now deferred to the next event loop, leaving a space to cancel focus leaving for `X11Wm`.

## What I've checked
- Checked locally with cosmic-comp (no code was changed in cosmic-comp, only path to local version of smithay)
- Unity Editor is no longer throttling
- Switching between X11<->Wayland apps works as expected
- No regression issues found

## Checklist

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -136,7 +136,7 @@ use crate::{
     },
 };
 use atomic_float::AtomicF64;
-use calloop::{Interest, LoopHandle, Mode, PostAction, generic::Generic};
+use calloop::{Interest, LoopHandle, Mode, PostAction, generic::Generic, ping};
 use rustix::fs::OFlags;
 use std::{
     cell::RefCell,
@@ -147,7 +147,10 @@ use std::{
         io::{AsFd, OwnedFd},
         net::UnixStream,
     },
-    sync::{Arc, Weak, atomic::Ordering},
+    sync::{
+        Arc, Weak,
+        atomic::{AtomicBool, Ordering},
+    },
 };
 use tracing::{debug, debug_span, info, trace, warn};
 use wayland_server::{DisplayHandle, Resource};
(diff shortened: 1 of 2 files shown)


Drakulix:
Okay this seems pretty okay as a fix, but I am a bit hesitant on merging the deferred `__NET_ACTIVE_WINDOW` change. Can we not start reading the `detail` instead and only unset it on `NotifyDetailNone` or something?

mayakwd:
> Okay this seems pretty okay as a fix, but I am a bit hesitant on merging the deferred `__NET_ACTIVE_WINDOW` change. Can we not start reading the `detail` instead and only unset it on `NotifyDetailNone` or something?

Well, you are probably right, maybe I jumped the gun a bit. 
I'll try to bring unset of `__NET_ACTIVE_WINDOW` back, and don't smear the logic via deferring, but take a decision in place.

mayakwd:
@Drakulix,
I moved back the unset logic and just added a simple guard. It actually works the same way as deferred (in terms of behavior), but now it's not deferred :-)

Yet we still need to preserve the deferred path for `set_input_focus` to avoid the intermediate `(NONE, NONE)` state in case a transition occurs within the same client X11->X11.

Drakulix (approved):
Still feels a little hacky. At some point we probably should make focus-switches somewhat atomic, so that we can simply infer, if any new target is also an X11Surface or not.

But in the meantime this is a good solution. Thanks!

SotoFidel:
Just tested on Unity 6.4. Looking good 👍🏽 
Closing issue 2006 on cosmic-comp.
You all have my thanks!

PR #2030, opened 2026-05-10
x11: X11Surface bbox/geometry should match committed buffer

kelnos:
## Description

`X11Surface::bbox()` returns the geometry stored internally by `X11Surface`, which is based on the last committed configure attempted by the compositor (via `configure()`).  The problem here is that the client is free to reject that size (size increments, min size, max size, etc.).

In these cases, we can mis-render the window (placement or size) if the last configured rect is not the same as the actual committed buffer.

Instead, use the surface/buffer size, and only fall back to `X11Surface`'s geometry if that's not available.

(This used to be a part of #2018 but I decided to split it off since it's not strictly related.)

## Checklist

### DIFF
--- a/src/desktop/space/wayland/x11.rs
+++ b/src/desktop/space/wayland/x11.rs
@@ -5,9 +5,11 @@ use crate::{
             Kind,
             surface::{WaylandSurfaceRenderElement, render_elements_from_surface_tree},
         },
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Thanks!

PR #2032, opened 2026-05-14
fix: duplicated "the" in desktop space/grab/tablet input comments

vip892766gma:
Three one-line typo fixes for duplicated "the" in source comments:
- `src/desktop/space/wayland/mod.rs` — "// Maybe the the surface got unmapped, send leave on output" → "// Maybe the surface got unmapped, send leave on output"
- `src/desktop/wayland/popup/grab.rs` — "// Check if the the client of the focused surface is still equal to the grabbed surface client" → "// Check if the client of the focused surface..."
- `src/backend/input/tablet.rs` — "/// return the total number of buttons pressed on all devices on the associated seat after the the event was triggered." → "...after the event was triggered."

No code/behavior change.


Drakulix (approved):
Thanks!

PR #2036, opened 2026-05-20
Implement wlr virtual pointer unstable protocol

itaylor:
## Description
Implements the wlr virtual pointer protocol:
https://wayland.app/protocols/wlr-virtual-pointer-unstable-v1

I was able to test it by running anvil:
```
RUST_LOG=smithay::wayland::virtual_pointer=debug WAYLAND_DISPLAY=wayland-1 cargo run --release --manifest-path anvil/Cargo.toml -- --winit
```
in one shell while in another running:
```
WAYLAND_DISPLAY=wayland-2 RUST_LOG=info cargo run --manifest-path test_clients/Cargo.toml --bin test_virtual_pointer
```
And I can see the events logged by anvil:

```
2026-05-20T04:12:54.472274Z DEBUG smithay::wayland::virtual_pointer: virtual_pointer: motion dx=4 dy=3 time=0
2026-05-20T04:12:55.472606Z DEBUG smithay::wayland::virtual_pointer: virtual_pointer: button button=0x00000110 state=Pressed time=0
2026-05-20T04:12:55.472654Z DEBUG smithay::wayland::virtual_pointer: virtual_pointer: button button=0x00000110 state=Released time=0
2026-05-20T04:12:57.473192Z DEBUG smithay::wayland::virtual_pointer: virtual_pointer: dispatching axis frame AxisFrame { source: Some(Wheel), relative_direction: (Identical, Identical), time: 0, axis: (0.0, 30.0), v120: Some((0, 360)), stop: (false, false) }
```
Anvil doesn't appear to draw a mouse pointer when used in this way, but the events do get logged as if they work.  I also have a branch where I wired this into cosmic-comp and was able to use it successfully with `wayland-clickdot` and `wlrctl` to move the mouse around and show clicks.  If there's some better way to have tested this within the scope of this repo I'm happy to add it.
 
AI Disclosure:
I used Claude Sonnet 4.6 to write this, but I read it and tested it myself.  It is similar to the existing virtual_keyboard protocol implementation.

## Checklist

### DIFF
--- a/src/wayland/virtual_pointer/mod.rs
+++ b/src/wayland/virtual_pointer/mod.rs
@@ -316,7 +316,7 @@ where
                         );
                     }
                 } else {
-                    warn!("virtual_pointer: button event dropped — no seat associated with virtual pointer and no default seat provided");
+                    warn!("virtual_pointer: button event dropped - no seat associated with virtual pointer and no default seat provided");
                 }
             }
 
@@ -347,7 +347,7 @@ where
                         ptr.frame(state);
                     }
                 } else if frame.is_some() {
-                    warn!("virtual_pointer: axis frame dropped — no seat associated with virtual pointer and no default seat provided");
+                    warn!("virtual_pointer: axis frame dropped - no seat associated with virtual pointer and no default seat provided");
                 }
             }
 
--- a/src/wayland/virtual_pointer/mod.rs
+++ b/src/wayland/virtual_pointer/mod.rs
@@ -1,0 +1,430 @@
+//! Utilities for wlr-virtual-pointer support
+//!
+//! This module provides utilities to handle virtual pointer instances, allowing
+//! clients to emulate a physical pointer device.
+//!
+//! ```
+//! use smithay::input::{Seat, SeatState, SeatHandler, pointer::CursorImageStatus};
+//! # use smithay::wayland::compositor::{CompositorHandler, CompositorState, CompositorClientState};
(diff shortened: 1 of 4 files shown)


src/wayland/virtual_pointer/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+                            &ButtonEvent {
+                                serial: SERIAL_COUNTER.next_serial(),
+                                time,
+                                button,
+                                state: btn_state,
+                            },
+                        );
+                    }
+                } else {
+                    warn!("virtual_pointer: button event dropped — no seat associated with virtual pointer and no default seat provided");
```
HigherOrderLogic:
```suggestion
                    warn!("virtual_pointer: button event dropped - no seat associated with virtual pointer and no default seat provided");
```

This may look like a normal dash under mono fonts, but it's actually emdash. Imo replacing all of them with normal dash would be better.

itaylor: Replaced them all

itaylor:
Would love to land this, the use-case I want it for is automated testing in Cosmic DE apps.  Now that pointer_warp [landed over there](https://github.com/pop-os/cosmic-comp/pull/2432), this virtual pointer support is really all that's missing to be able to easily drive a test harness to click around in apps in that environment.  Let me know if there's anything else you'd like to see added here.

PR #2038, opened 2026-05-22
Change WM_STATE only if the new value is different

hojjatabdollahi:
This reverts commit 060e9cd9c8ae2633947bdb72cffa1748ac53922d.

## Description
This reverts the change from https://github.com/Smithay/smithay/pull/2022.

It was causing Wayland native games to not receive game controller input. The input was going to steam behind the game.

I tested Dishonored 2 (but from steam, and not GOG) and it did not minimize. My assumption is https://github.com/pop-os/cosmic-comp/commit/6e4551a20019e7790db6bd845fdbd087e41fbb9a fixed the minimizing issue.

## Checklist

### DIFF
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -401,11 +401,20 @@ impl X11Surface {
 
         if let Some(conn) = self.conn.upgrade() {
             if let Some(frame) = self.state.lock().unwrap().mapped_onto {
-                if mapped {
+                let wm_state = if mapped {
                     conn.map_window(frame)?;
+                    1u32 /*NormalState*/
                 } else {
                     conn.unmap_window(frame)?;
-                }
+                    3u32 /*IconicState*/
+                };
+                conn.change_property32(
+                    PropMode::REPLACE,
+                    self.window,
+                    self.atoms.WM_STATE,
+                    self.atoms.WM_STATE,
+                    &[wm_state, 0 /*WINDOW_NONE*/],
+                )?;
                 conn.flush()?;
             }
         }
@@ -1276,6 +1285,21 @@ impl X11Surface {
         } else {
             self.change_net_state(&[], &[self.atoms._NET_WM_STATE_HIDDEN])?;
         }
+
+        if let Some(conn) = self.conn.upgrade() {
+            let wm_state = if suspended {
+                3u32 /*IconicState*/
+            } else {
+                1u32 /*NormalState*/
+            };
+            conn.change_property32(
+                PropMode::REPLACE,
+                self.window,
+                self.atoms.WM_STATE,
+                self.atoms.WM_STATE,
+                &[wm_state, 0 /*WINDOW_NONE*/],
+            )?;
+        }
         Ok(())
     }
 
@@ -1405,19 +1429,6 @@ impl X11Surface {
                 self.atoms._NET_WM_STATE,
                 AtomEnum::ATOM,
                 &new_props,
-            )?;
-
-            let wm_state = if state.net_state.contains(&self.atoms._NET_WM_STATE_HIDDEN) {
-                [3u32 /*IconicState*/, 0 /*WINDOW_NONE*/]
-            } else {
-                [1u32 /*NormalState*/, 0 /*WINDOW_NONE*/]
-            };
-            conn.change_property32(
-                PropMode::REPLACE,
-                self.window,
-                self.atoms.WM_STATE,
-                self.atoms.WM_STATE,
-                &wm_state,
             )?;
         }
 
--- a/src/xwayland/xwm/surface.rs
+++ b/src/xwayland/xwm/surface.rs
@@ -400,13 +400,18 @@ impl X11Surface {
         }
 
         if let Some(conn) = self.conn.upgrade() {
-            if let Some(frame) = self.state.lock().unwrap().mapped_onto {
+            let state = self.state.lock().unwrap();
+            if let Some(frame) = state.mapped_onto {
                 let wm_state = if mapped {
                     conn.map_window(frame)?;
-                    1u32 /*NormalState*/
+                    if state.net_state.contains(&self.atoms._NET_WM_STATE_HIDDEN) {
+                        3u32 /*IconicState*/
+                    } else {
+                        1u32 /*NormalState*/
+                    }
                 } else {
                     conn.unmap_window(frame)?;
-                    3u32 /*IconicState*/
+                    0u32 /*WithdrawnState*/
                 };
                 conn.change_property32(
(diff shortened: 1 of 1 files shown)


ids1024:
> It was causing Wayland native games to not receive game controller input. The input was going to steam behind the game.

I'm not sure how exactly this happens. Game controller input doesn't even go through X11 or Wayland. Would it be Steam Input deciding which application gets controller events here?

hojjatabdollahi:
My best guess is that without this, steam only sees focus loss and gracefully handles it. But with your change, it sees an extra event for its own window and interprets it as being in the UI mode, and not game mode.

(Because in your if block you unconditionally always rewrite WM_STATE, even though the value hasn't changed)

ids1024:
If the multiple writes of the same value is the issue, then it could be best to store the value of `WM_STATE` in a variable and update only if it is different.

hojjatabdollahi:
I think if you write it, it produces a PropertyNotify (even though it's unchanged) and Steam doesn't expect that.

In general it looks like that Map/Unmap should touch WM_STATE and leave _NET_WM_STATE alone. and Focus Change and Fullscreen Toggle should touch _NET_WM_STATE and leave WM_STATE alone.
Your code was touching WM_STATE. 

But of course you know a lot better. I'm learning the minimum I need to polish the experience in https://github.com/pop-os/cosmic-comp/pull/2399. The right solution maybe somewhere else. I'm not too familiar with Smithay's codebase (hence just a Revert).

ids1024:
The [ICCCM spec](https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html) which defines `WM_STATE` is pretty vague (intentionally) about what the state should mean on a given window manager. But testing on Gnome, `xprop` does show minimized windows having `IconicState`. Which we presumably want to match.

hojjatabdollahi:
@ids1024 yeah, but your change was doing it in `change_net_state()` which means changing focus or fullscreen also touches wm_state. 

Here's what I understand after looking into it for half a day (keep that in mind):
```
Event             WM_STATE         _NET_WM_STATE
Map window        NormalState      unchanged
Unmap             WithdrawnState   unchanged
Minimize          IconicState      add HIDDEN
Unminimize        NormalState      remove HIDDEN
Focus change      unchanged        toggle FOCUSED
Fullscreen toggle unchanged        toggle FULLSCREEN
```
So I moved your change to `set_hidden()` in https://github.com/Smithay/smithay/pull/2038/commits/ebefdbb8443f2c1c8f07450b8014a0e29e9a5f51 to update wm_state on minimize/unminimize.


(I tested it, and the game controller input is fine there, but I don't know if that's a good spot for it, and if that achieves what you had in mind)

ids1024:
I think keeping it in `change_net_state()` would be good, but storing the value it was previously set to and only calling `change_property32` if it's different (similarly to how the code currently does for `_NET_WM_STATE`). Then it won't be triggered on changes to focus/fullscreen/etc. (unless it hasn't been set at all yet).

kelnos:
Yes, `WM_STATE` and `_NET_WM_STATE` should be kept in sync when either changes (was just digging through xfwm4 to verify this).

But yeah, best to try to only update properties if they've changed.  Clients should actually fetch the new value and only act if *they* know it's changed or the new value is new information to them, but who knows if there are buggy clients that don't.  xfwm4, for example, only updates `WM_STATE` if it's changed, not for every other state transition.

And regardless, spurious `PropertyNotify` events are a waste of CPU cycles for any app watching for them.

(As an aside, we should also set `WM_STATE` to `WithdrawnState` on unmap, but that's another issue and is arguably not super critical since most modern apps will only unmap right before they destroy a window...)

Drakulix:
I am pretty sure what smithay is doing right now is correct and the steam issue is an effect of:

- Steam input exclusively locks gamepads (this is normal and intended behavior)
- Steam creates a virtual controller to forward (modified) steam input to
- Using the proton wayland backend breaks injecting the steam overlay, which steam uses as a hint to enable forwarding input

The only reason it worked before is, that unmapping probably causes steam to stop steam input temporarily and releases the gamepads.

Reports of steam input not working with the wayland backend are very common on the internet. The fix is disabling steam input for the game in question completely.

As such I don't think we should consider reverting this commit at all.

hojjatabdollahi:
@ids1024's suggestion should also work if the series of events that I described were true. Meaning if we don't set wm_state in `change_net_state` if it hasn't changed, then the issue with steam input that I was seeing should be gone. I'll test it in a minute.

hojjatabdollahi:
My latest commit also fixes the issue.
It checks in `change_net_state` and if wm_state is different it changes it. I had to also add it to `set_mapped`.

I didn't see anybody calling `set_mapped` with `false`. Currently I set it to `IconicState` if you call `set_mapped(false)` but it should be `WithdrawnState`.

Feel free to close this PR and do it properly. But just as a reference this fixes the steam input for wayland games I was experiencing.

kelnos:
> I think keeping it in `change_net_state()` would be good

Logically, updating `WM_STATE` should be a completely separate operation that's called only from `set_mapped()` and `set_hidden()`.  It's tempting to keep it in `change_net_state()`, but honestly that feels a little confused to me.

src/xwayland/xwm/surface.rs
```diff
@@ -401,11 +401,20 @@ impl X11Surface {
...
         if let Some(conn) = self.conn.upgrade() {
             if let Some(frame) = self.state.lock().unwrap().mapped_onto {
-                if mapped {
+                let wm_state = if mapped {
                     conn.map_window(frame)?;
+                    1u32 /*NormalState*/
                 } else {
                     conn.unmap_window(frame)?;
-                }
+                    3u32 /*IconicState*/
```
kelnos: This should be `WithdrawnState` (`0u32`), no?

kelnos:
The latest commit looks more correct to me, but I think we need to handle some other cases.  For example, if the window is not mapped yet, and the compositor calls `set_minimized(true)`, we should not set `IconicState`; it should remain as `WithdrawnState`.  When it eventually does get mapped, it should transition to `IconicState`.

`set_minimized(false)` should also check if the window is mapped or not, and set either `NormalState` or `WithdrawnState` appropriately.

src/xwayland/xwm/surface.rs
hojjatabdollahi: I added it.

hojjatabdollahi:
Now it changes `wm_state` in `st_hidden`, only if the window is already mapped. And when mapping the window, it checks to see if `set_hidden` was called before `set_mapped`

Drakulix (approved):
LGTM now. Thanks!

PR #2039, opened 2026-05-23
drm: propagate acquire_master_lock failure in activate()

rchennau:
## Problem

`DrmDevice::activate()` silently ignores `acquire_master_lock()` failures. When `drmSetMaster()` fails (e.g. because the seat manager hasn't granted DRM master yet during a VT resume race), execution falls through unconditionally and `set_active(true)` is called. The compositor is now marked ready to render but holds no DRM master on the device.

Every subsequent atomic page flip commit returns `EPERM` from the kernel DRM core (`DRM_MASTER` flag on `DRM_MODE_ATOMIC` ioctl). The render loop retries at frame rate (60fps), flooding the journal with permission errors until the compositor is manually killed.

**Affected systems:** Hybrid GPU setups where the render device (e.g. Intel iGPU) and display output device (e.g. NVIDIA dGPU) have separate seat activation timing. Confirmed on Intel+NVIDIA RTX 5060 Ti with `libseat`/`seatd`.

**Upstream reports:** pop-os/cosmic-comp#2331, pop-os/cosmic-comp#2302

## Root cause

```rust
// BEFORE — falls through on failure:
if let Err(err) = self.acquire_master_lock() {
    error!("Failed to acquire drm master again. Error: {}", err);
    // BUG: no return, active=true set unconditionally below
}
if !self.set_active(true) && disable_connectors { ... }
```

## Fix

```rust
// AFTER — propagates error, device stays inactive:
if let Err(err) = self.acquire_master_lock() {
    error!("Failed to acquire drm master on activate. Error: {}", err);
    return Err(Error::DrmMasterFailed);
}
if !self.set_active(true) && disable_connectors { ... }
```

When `activate()` returns `Err`, the caller (compositor's `resume_session()`) skips rendering on that device until the next `ActivateSession` event, at which point `activate()` is retried with a valid master grant.

## SwapBuffersError mapping

This PR also adds `DrmMasterFailed` to the `TemporaryFailure` arm of `From<Error> for SwapBuffersError`. Previously it fell through to `ContextLost`, which signals a fatal, non-recoverable error. DRM master loss during a VT switch is transient — the seat manager will re-grant master on the next VT resume — so `TemporaryFailure` is the correct mapping.

## Testing

Verified on Pop!_OS 24.04 with `cosmic-comp` on Intel+NVIDIA hybrid GPU:
- Before fix: `journalctl --user` shows EPERM flood at 60fps after every VT switch
- After fix: zero EPERM entries across 20 rapid VT switch iterations (`for i in $(seq 1 20); do chvt 2; sleep 0.5; chvt 1; sleep 1; done`)
- Single-GPU regression: `activate()` success path is unchanged — `set_active(true)` is only skipped on `acquire_master_lock()` failure, which cannot occur on non-privileged devices (`is_privileged()` returns false)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/backend/drm/device/mod.rs
+++ b/src/backend/drm/device/mod.rs
@@ -430,7 +430,8 @@ impl DrmDevice {
     pub fn activate(&mut self, disable_connectors: bool) -> Result<(), Error> {
         if self.device_fd().is_privileged() {
             if let Err(err) = self.acquire_master_lock() {
-                error!("Failed to acquire drm master again. Error: {}", err);
+                error!("Failed to acquire drm master on activate. Error: {}", err);
+                return Err(Error::DrmMasterFailed);
             }
         }
         if !self.set_active(true) && disable_connectors {
--- a/src/backend/drm/error.rs
+++ b/src/backend/drm/error.rs
@@ -94,7 +94,9 @@ impl From<Error> for SwapBuffersError {
         // FIXME: replace the special handling for EBUSY with ErrorKind::ResourceBusy once
         // we reach MSRV >= 1.83
         match err {
-            x @ Error::DeviceInactive => SwapBuffersError::TemporaryFailure(Box::new(x)),
+            x @ Error::DeviceInactive | x @ Error::DrmMasterFailed => {
+                SwapBuffersError::TemporaryFailure(Box::new(x))
+            }
             Error::Access(AccessError {
                 errmsg, dev, source, ..
             }) if matches!(


Drakulix:
Why are you running seatd on Pop!_OS? This isn't a supported configuration, you should be running logind instead, which will never use a privileged file descriptor to begin with.

Either way propagating this error is correct, so thanks.

Drakulix (approved):

PR #2041, opened 2026-05-25
 Drive DisplayID-tiled (5K) monitors as one output

algesten:
Some high-res monitors can't fit their full resolution down a single cable, so they present as several DRM connectors, one per tile, each carrying a slice of the image. The connectors share a DisplayID Tile Topology block describing how the tiles fit together (the LG UltraFine 5K and Dell UP2715K are the common ones).

This drives such a monitor as a single logical output. The tiles sit on separate CRTCs but page-flip together in one atomic commit, so they stay frame-locked. The primary tile's vblank stands in for the whole group.

The pieces:

- `DrmDevice::tile_info` reads a connector's `TILE` property into a `TileInfo`. Connectors sharing a `group_id` belong to the same monitor.
- `DrmCompositor::new_tiled` drives N CRTCs as one logical output. It's a general primitive (a single-GPU video wall qualifies too), so it takes geometry as a `TilePlacement` per CRTC rather than `TileInfo` directly. All surfaces must live on one device so they can share the atomic commit.
- `DrmOutputManager::initialize_tiled_output` wraps that as one `DrmOutput`, keyed by the primary tile's CRTC.

Each tile renders its slice of the logical scene into its own framebuffer. The relocation offset is derived through the output transform, so it stays correct under rotation and flip. That part is pure geometry and unit-tested across all eight transforms.

Exercised on a real LG UltraFine 5K over its two DisplayPort streams (via cosmic-comp): display, live rotation, and screencopy all work.

Making use of this requires changes to `cosmic-comp` to read the `TileInfo` and using `new_tiled`. This is deliberate, since `cosmic-comp` needs to present this correctly to the user in settings.

I have prepared the `cosmic-comp` changes, but will keep that PR away until we settled the smithay side. It needs tidying up before PR, but for reference, here it is: https://github.com/pop-os/cosmic-comp/compare/master...algesten:cosmic-comp:tiled-output

Close #2034

* [x] I have used AI in working out this solution. I have carefully reviewed the code presented.

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -1055,6 +1055,12 @@ where
     <F as ExportFramebuffer<A::Buffer>>::Framebuffer: std::fmt::Debug + Send + Sync + 'static,
     G: AsFd + 'static,
 {
+    /// Sub-rect of the logical output this tile scans out, in the output's
+    /// untransformed (mode) coordinate space. The whole output for a
+    /// single-tile compositor; one slice of the grid for a tiled one.
+    // Read by the multi-tile render path (added in a following commit).
+    #[allow(dead_code)]
+    region: Rectangle<i32, Physical>,
     surface: Arc<DrmSurface>,
     planes: Planes,
     overlay_plane_element_ids: OverlayPlaneElementIds,
@@ -1160,14 +1166,16 @@ pub enum TileConfigError {
     /// do not cover a single rectangular logical output.
     #[error("tile regions do not cover the logical output without gaps")]
     NotContiguous,
+    /// The tile surfaces span multiple DRM devices. A tiled compositor requires
+    /// a single device so all its CRTCs can share one atomic commit.
+    #[error("tile surfaces span multiple drm devices")]
+    MultipleDevices,
 }
 
 /// Validate that `regions` exactly tile a rectangle anchored at `(0, 0)` with no
 /// gaps or overlaps, returning that rectangle's size.
 ///
 /// Pure geometry, independent of any DRM state, so it is unit-testable.
-// First non-test caller is `new_tiled`, added in a following commit.
-#[allow(dead_code)]
 fn validate_tile_regions(
     regions: impl IntoIterator<Item = Rectangle<i32, Physical>>,
 ) -> Result<Size<i32, Physical>, TileConfigError> {
@@ -1245,6 +1253,16 @@ fn tile_render_offset(
 ) -> Point<i32, Physical> {
     let loc = tile_logical_region(region, size, transform).loc;
     Point::from((-loc.x, -loc.y))
+}
+
+/// The [`region`](TileState::region) for a single-tile compositor: the whole
+/// logical output. Resolved from `output_mode_source`'s current mode, falling
+/// back to a zero rect when no mode is set yet (the single-tile render path
+/// does not slice, so the value is unused until a mode exists).
+fn single_tile_region(output_mode_source: &OutputModeSource) -> Rectangle<i32, Physical> {
+    <&OutputModeSource as TryInto<(Size<i32, Physical>, Scale<f64>, Transform)>>::try_into(output_mode_source)
+        .map(|(size, _, _)| Rectangle::from_size(size))
+        .unwrap_or_default()
 }
 
 #[cfg(test)]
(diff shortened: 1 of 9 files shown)


algesten:
@Drakulix I hope this is in line with what you were imagining. Happy to take pointers and rework stuff if it's not right. The PR should be possible to both compile and review commit-by-commit.

src/backend/drm/compositor/mod.rs
```diff
@@ -1085,6 +1085,177 @@ where
...
+    for region in &regions {
+        if region.size.w <= 0 || region.size.h <= 0 {
+            return Err(TileConfigError::NonPositiveRegion(*region));
+        }
+        bbox = bbox.merge(*region);
+        covered += i64::from(region.size.w) * i64::from(region.size.h);
+    }
+
+    // The logical output starts at the origin.
+    if bbox.loc != Point::from((0, 0)) {
```
Drakulix: `from` is deprecated, this should use `Point::new`

src/backend/drm/surface/atomic.rs
```diff
@@ -924,6 +924,129 @@ impl AtomicDrmSurface {
...
+    /// Append this surface's scan-out state for the next frame into a shared
+    /// atomic request, without committing.
+    ///
+    /// Several surfaces driving CRTCs on the *same* device can append into one
+    /// request and then be committed together by [`commit_request`](Self::commit_request),
+    /// so they flip in a single atomic commit and stay frame-locked (used for
+    /// tiled outputs). `allow_modeset` selects a full modeset (mode blob +
+    /// connectors, like [`commit`](Self::commit)) over a page-flip (like
+    /// [`page_flip`](Self::page_flip)).
+    pub(in crate::backend::drm) fn append_to_request<'b>(
```
Drakulix, line 936: I would like to make this api available to compositors as well instead of being exclusive to the `DrmCompositor`. The general philosophy of smithay is, that you should be able to build every additional abstraction provided by smithay (like the DrmCompositor) yourself out-of-tree with the next simpler building blocks. (Though this should apply to `DrmSurface` not `AtomicDrmSurface`.)

src/backend/drm/output.rs
```diff
@@ -488,12 +500,123 @@ where
...
+                surface,
+                planes: tile.planes.clone(),
+                region: tile.region,
+            });
+        }
+
+        // NOTE: unlike `initialize_output`, this does not yet retry under
+        // bandwidth pressure (disabling overlay planes / implicit modifiers). A
+        // tiled output is a single dedicated monitor, so the simple path is used
+        // for now.
```
Drakulix: This needs to be addressed before we can merge this MR. The whole purpose of the `DrmOutputManager` is to enable handling of these bandwidth constrainted scenarios. And they can still happen with tiled monitor as many powerful cards can drive 3 or more outputs, which can contribute to the issue.

Drakulix (commented):
So on a first pass, this looks pretty good. I commented on a few bits, that stood out to me.

I am very sure we can proceed with this approach.

The whole PR is quite comment-heavy. This isn't inherently a bad thing, but a good amount of bits and pieces are quite repetitive or needlessly verbose. I assume this is due to the AI-assisted nature of the PR. (Code-wise this is pretty good though, good job cleaning it up.) I'd suggest to focus mostly on documenting reasoning and design decisions, but try to cut down on repetition and obvious explanations.

One thing I am wondering is, whether the code in `DrmCompositor` would already be able to direct-scanout a client buffer spanning multiple outputs (think a fullscreened window). I think it should be able to figure out the necessary parameters for `CRTC_W/H` and `SRC_W/H` properties, which is exciting. For compositing though I think going with multiple swapchains is a good approach (instead of going for a shared framebuffer covering the whole space), since it is likely more compatible.

Last but not least: It isn't 100% obvious to me, if a compositor could use this to make a "tiled output", that doesn't actually use the `TILE`-property, but simply wants to combine multiple outputs. (Assuming the outputs can each use atomic-commits and are connected to the same `DrmDevice`.) I don't see a logical reason, why that shouldn't work, so I just want to make sure the code allows it.

Ping @cmeissl. Since you wrote most of the code that is refactored here, I would really like a review from you, if you have the time.

src/backend/drm/compositor/mod.rs
algesten: Sorted

src/backend/drm/surface/atomic.rs
algesten: Ok! I added a small API for this.

src/backend/drm/output.rs
algesten: Sorted

algesten:
> I am very sure we can proceed with this approach.

:tada: 

> The whole PR is quite comment-heavy. This isn't inherently a bad thing, but a good amount of bits and pieces are quite repetitive or needlessly verbose. I assume this is due to the AI-assisted nature of the PR. (Code-wise this is pretty good though, good job cleaning it up.) I'd suggest to focus mostly on documenting reasoning and design decisions, but try to cut down on repetition and obvious explanations.

I have made a pass, making it terser and removed some redundant comments.

> One thing I am wondering is, whether the code in `DrmCompositor` would already be able to direct-scanout a client buffer spanning multiple outputs (think a fullscreened window). I think it should be able to figure out the necessary parameters for `CRTC_W/H` and `SRC_W/H` properties, which is exciting. For compositing though I think going with multiple swapchains is a good approach (instead of going for a shared framebuffer covering the whole space), since it is likely more compatible.

It may work on drivers that accept clipped/off-screen plane destinations, but the code does not explicitly compute per-tile SRC_X/Y/W/H crops. We could potentially add crops to make it more compatible.

> Last but not least: It isn't 100% obvious to me, if a compositor could use this to make a "tiled output", that doesn't actually use the `TILE`-property, but simply wants to combine multiple outputs. (Assuming the outputs can each use atomic-commits and are connected to the same `DrmDevice`.) I don't see a logical reason, why that shouldn't work, so I just want to make sure the code allows it.

The idea is that it should allow that. The only restriction on `new_tile` should be that it must be the same card, since it's an atomic commit. The values you read out of the TILE property of EDID needs to be converted into the `TilePlacement`, but there is no hard connection between these two.

src/backend/drm/surface/mod.rs
```diff
@@ -187,6 +189,62 @@ impl BasicDevice for DrmSurface {}
...
+    /// # fn rollback_atomic_commit<'a>(
+    /// #     surface_a: &'a DrmSurface,
+    /// #     planes_a: Vec<PlaneState<'a>>,
+    /// # ) -> Result<(), smithay::backend::drm::DrmError> {
+    /// let commit = surface_a.begin_atomic_commit(planes_a, false, true)?;
+    /// commit.rollback();
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn begin_atomic_commit<'a, 'b>(
```
algesten, line 231: It's worth noting that this API could be made more strict. We can't use "the most elegant Rust", of locking the lifetime of a `&'a mut` because inside `DrmSurface` we have `Arc<Mutex<XXXInner>>`, meaning a mut doesn't lock anything.

That means this API currently is not guaranteeing atomic commits across threads (or even re-entrant/interleaved calls on the same thread). This could be fixed by adding some `Arc<Mutex<CommitLock>>` that can be exclusively taken for the lifetime of the `DrmAtomicCommit`.

However I was unsure about introducing more internal `Arc<Mutex` - so didn't do it straight away.

cmeissl:
> Ping @cmeissl. Since you wrote most of the code that is refactored here, I would really like a review from you, if you have the time.

Sure, I am going to review it today.

algesten:
> Sure, I am going to review it today.

Maybe we should close https://github.com/Smithay/smithay/pull/2034 and move the comments here, since it takes the TILE property PR and builds on it?

cmeissl:
> > Sure, I am going to review it today.
> 
> Maybe we should close #2034 and move the comments here, since it takes the TILE property PR and builds on it?

Yeah, sorry, I had both PRs open and picked the wrong one when answering.

> I kept it on top of `FrameState`/`ElementState` rather than tile-aware on purpose. I didn't want the CRTC level to know anything about the TILE property. A tile is just a `DrmSurface` with its own planes and swapchain, and the composition sits above that.
> 
> I don't know if there is a use case here, but I imagined it would be possible to use the tiled output with several normal screens on the same card too.

I did not mean to add it to `DrmSurface` directly. With the current approach it looks like we would end up with a framebuffer per tile which sounds unnecessary. It should work fine to have the same exporter and framebuffer on multiple crtc. What I was thinking about is to make the instances in the `ElementState`  and the planes in `FrameState` CRTC aware. This would also allow to keep the user data in `PendingFrame`, moving it out of the `Frame` opens the door for unnecessary desync of the frame state and the user data.

> You're right. I'll make a new test using the same DrmAtomicCommit builder so it spans all tiles, rather than make `FrameState` tile-aware.

What I though about is adding an abstraction for holding multiple `DrmSurface` and their relationship, just like you did with the tiles, but as a separate abstraction. It could provide the whole test/commit/page_flip logic and allow us to keep a single interface for the frame.

We also have to initialize all planes states once at the start so that each test can incrementally extend the planes. Doing it per Tile and ignoring the atomic properties of all other tiles will lead to uncatched edge cases and might prevent using planes in some cases.
So I would suggest to not special case for tiled/single crtc, but to always run the same logic and only branch in the abstraction responsible for calling test_state and so on.
So for example each `render_frame` first initializes the frame with all primary planes and no overlay planes and then incrementally run the per TILE/CRTC logic.

I think we also have to provide atomic requests for all other functionality like enabling VRR, not sure doing it one-by-one will work.
The previously mentioned abstraction could resemble the `DrmSurface` interface and transparently build the atomic requests for all surfaces. The legacy path could just return an error for now in case there are multiple surfaces defined.

algesten:
Thank you for that feedback! I will process it and fix the PR. A bit busy with work right now so it may take me until the weekend.

Issue #2031, opened 2026-05-10
Support for _GTK_FRAME_EXTENTS in X11Surface (pre-PR discussion)


kelnos:
Wanted to discuss this a bit first before implementing, as it amounts to a behavior change and possibly some API breaks.

`_GTK_FRAME_EXTENTS` (and the much lesser used `_KDE_NET_WM_SHADOW`) are set by the client to describe the pixel widths of drop shadows added around a CSD window.  `_GTK_FRAME_EXTENTS` is perhaps confusingly named, as it is _not_ the same as the WM-set `_NET_FRAME_EXTENTS`, which includes the size of the SSD frame.

Currently, `X11Surface` implements `SpaceElement`'s `bbox()` and `geometry()` using the size of the last-sent configure notify event[0].  For xdg_toplevel surfaces, `bbox()` is the size of the committed buffer, and `geometry()` is the size the client sets using `xdg_surface.set_window_geometry`, which is generally used to exclude drop shadows.

`xdg_surface.set_window_geometry` is sort of the inverse of `_GTK_FRAME_EXTENTS`, but they both have the same purpose: help the WM/compositor understand what the "user-visible bounds" of  a surface is.

What I'd like to do is add support for `_GTK_FRAME_EXTENTS` (and `_KDE_NET_WM_SHADOW`) to `X11Surface`.  `SpaceElement::bbox()` would continue returning what it does today (or, well: [0]).  `::geometry()`, however, would start returning `::bbox()` minus the pixels set in `_GTK_FRAME_EXTENTS`.  In my compositor I have a bunch of special-case code for X11 windows that has to pull out the shadow extents and subtract them before making decisions about initial placement, tiling, maximized size, etc., and it would be nice to get rid of that and just use `SpaceElement::geometry()` for everything.

However, if any compositor is using `SpaceElement::geometry()` for e.g. calculating configure events to send to the client, this will break them, as the configure event needs to include the full size, including the shadows.  Is this possible breakage something y'all would be comfortable with?

I thought about also modifying `X11Surface::configure()` to instead assume user-visible bounds are being passed, but while that would unify the semantics of `xdg_toplevel` and `X11Surface`'s configures, I think making the latter not conform to X11 expectations is confusing and not good API design.  But an alternative could be to add `configure_geometry()`, which internally adds the shadow extents before calling `configure()`.  (I guess we'd then also need `configure_geometry_with_sync()` as a part of #2018.)

[...]

PR #2042, opened 2026-05-26
xwm: Support _GTK_FRAME_EXTENTS and _KDE_NET_WM_SHADOW

kelnos:
## Description

This implements the first part of the discussion in #2031.  Changes:

* `X11Surface` tracks `_GTK_FRAME_EXTENTS`/`_KDE_NET_WM_SHADOW` internally.
* `X11Surface`'s `SpaceElement::geometry` impl returns the surface bbox minus the values in the frame extents (if any).
* Added `X11Surface::frame_extents` to return the extents.
* Added a new `FrameExtents` type along with the other geometry types (`Point`, `Rectangle`, etc.).  The name of the struct is perhaps too specific for something that just has left/right/top/bottom members, but I'm not sure of a better name.
* `XwmHandler::property_notify` notifies on frame extents changes.
* `Window`'s `SpaceElement::geometry` impl delegates to `X11Surface`'s impl for xwayland windows.

I've tested this in xfwl4, and it seems to work well.  I was able to delete a lot of code...

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,9 @@ The `delegate_*!` macros have been replaced with a single `delegate_dispatch2!` 
 of new `Dipsatch2` and `GlobalDispatch2` traits. These will replace `Dispatch` and `GlobalDispatch` in a future version of
 `wayland_server`.
 
+The `SpaceElement::geometry` impl for `X11Surface` is no longer equivalent to its bbox.  For clients that set
+`_GTK_FRAME_EXTENTS`, the shadow extents are subtracted from the bbox to give the geometry.
+
 ### Additions
 
 - ExtBackgroundEffect protocol is now available in `smithay::wayland::background_effect` module.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,9 @@ of new `Dipsatch2` and `GlobalDispatch2` traits. These will replace `Dispatch` a
 The `SpaceElement::geometry` impl for `X11Surface` is no longer equivalent to its bbox.  For clients that set
 `_GTK_FRAME_EXTENTS`, the shadow extents are subtracted from the bbox to give the geometry.
 
+`X11Surface::geometry` has been renamed to `X11Surface::last_configure`. `X11Surface::geometry` now returns the bounding
+box minus the surface's frame extents.
+
 ### Additions
 
 - ExtBackgroundEffect protocol is now available in `smithay::wayland::background_effect` module.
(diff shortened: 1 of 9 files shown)


Drakulix (approved):
LGTM. I like the `FrameExtends` type.

I am just wondering, since `SpaceElement::geometry` requires the `desktop`-feature, any compositor not using that has to go through `X11Surface::geometry` - `X11Surface::frame_extends`, right?

Didn't we want to rename `X11Surface::geometry` (which now also doesn't correspond to `SpaceElement::geometry` anymore, which is even more confusing. Especially since `surface.geometry()` can now mean two things.) and then maybe also provide a way directly query the bbox with removed extends?

kelnos:
Ah, I was going for a smaller change for the first PR, but how about this:

```rust
/// Returns last configure sent to/ACKed by the client
pub fn last_configure(&self) -> Rectangle<> {}

/// Returns current committed surface/buffer size, falls back to last configure
pub fn bbox(&self) -> Rectangle<> {}

/// Returns `bbox() - frame_extents()`
pub fn geometry(&self) -> Rectangle<> {}
```

Then the `SpaceElement` `bbox` and `geometry` items can just delegate to the methods on `X11Surface`.

I'd also rename `pending_geometry()` to `pending_configure()` and `buffered_geometry()` to `buffered_configure()`.

Drakulix:
I like it because it clearly acknowledges, that the size send to the surface may be entirely separate from the size we are actually getting. (Users of the api should be aware that `surface.configure(surface.geometry())` isn't a no-op and this naming does suffice.)

kelnos:
Ok, renaming done.   What do you think?

kelnos:
(I also just removed the `_KDE_NET_WM_SHADOW` code; I misunderstood how it works.  I thought the client draws the shadows and tells the WM about it, but the client is actually asking the WM to draw the shadows for it, based on X pixmap IDs it puts in the property.  That might be an interesting feature to implement, but I think it's out of scope for this PR.)

Drakulix (approved):
Looks great! Thanks :)

Drakulix:
> (I also just removed the `_KDE_NET_WM_SHADOW` code; I misunderstood how it works. I thought the client draws the shadows and tells the WM about it, but the client is actually asking the WM to draw the shadows for it, based on X pixmap IDs it puts in the property. That might be an interesting feature to implement, but I think it's out of scope for this PR.)

I think the best way to support this in smithay would btw be to simply add a getter for the shadow-region on the window. I don't think we want to add a drop shadow shader to smithay directly.

kelnos:
> I don't think we want to add a drop shadow shader to smithay directly.

Yeah, not sure we even *can* do that.  XWayland appears to completely ignore the WM frame window that we reparent the client window into, so we couldn't draw them there.  Compositor has to do it...

PR #2045, opened 2026-05-30
fix: xdg-toplevel-tag setting description

MithicSpirit:
## Description

Fixes typo that led to incorrect behavior.

## Checklist


cmeissl (approved):
Thanks!

PR #2048, opened 2026-06-01
drm: transform plane damage to framebuffer-space for FB_DAMAGE_CLIPS

### LINKED ISSUE #1651: Output transform not applied to damage by `DrmCompositor`
This can be tested in Anvil using `Super+Shift+R` to rotate the screen, then dragging around a terminal. Maybe `FB_DAMAGE_CLIPS` handling varies by GPU, but I see clearly incorrect results. The same issue occurs in cosmic-comp.

It seems `renderer_damage` in `render_frame` doesn't have the output transform applied. Neither `render_frame`, nor `OutputDamageTracker::damage_output_internal` applies the transform to the damage.

Ph4ntomas:
## Description
Right after rendering, DrmCompositor build a PlaneDamageClips object, which is used to populate the FB_DAMAGE_CLIPS plane property. This property may be used by drivers to repaint only the damaged part of the plane, and is expected to be a bunch of rectangles in framebuffer coordinate-space. Current implementation uses the render_damage directly, which are transformed following the output_transform. If the device driver respect the FB_DAMAGE_CLIPS and only repaint those areas, this breaks rendering for any transform but normal since the actual damage and the reported one are misaligned.

Fix: apply the output inverted transform on the damage rectangle before passing them to PlaneDamageClips::from_damage.

Fixes: Smithay/smithay#1651

### Note for review
There are two things I'm not 100% sure about this PR:
1. I went to the change with the least possible side-effect, so damages are transformed to framebuffer-space at the last possible moment. I'm unsure if the transform should be applied to the Vec returned to calling code. My gut feeling is that it shouldn't however.
Transforming the damage in `damage_bag.render_output` would fix winit backend in anvil for free, but also introduce a lot of side effects (I don't think I've seen any code re-transforming the damages returned by that function), but I'd like to have someone else input on this.
2. Size for the transform area is `output_geometry.size` because that's what has been used in damage computation. 
I'm unsure if it's should be changed to `config.properties.dst` which is the un-transformed output size (and would need an additional transform). As far as I can tell, the two are equal anyway, so output_geometry.size felt like a good fit.

EDIT: There's a second call to `PlaneDamageClips` for direct scanout, which might require some changes too. I haven't managed to hit that code while testing yet. I'll add a commit if I do and it's buggy.
EDIT 2: Well, still haven't hit that code, but I *think* `element_config()` does the right transformation before building the `PlaneDamageClips`.

## Checklist

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -2299,10 +2299,16 @@ where
                                     &output_geometry.size.to_logical(1),
                                 )
                             }));
+
+                            // Here we need to apply the output_transform to the damage since we
+                            // haven't rotated our framebuffer. dst is already in the same
+                            // coordinate space src is, so no transform needed.
                             config.damage_clips = PlaneDamageClips::from_damage(
                                 self.surface.device_fd(),
                                 config.properties.src,
                                 config.properties.dst,
+                                Transform::Normal,
+                                output_transform,
                                 render_damage.iter().copied(),
                             )
                             .ok()
@@ -3967,11 +3973,17 @@ where
         let element_damage = element.damage_since(scale, previous_commit);
         let has_element_damage = !element_damage.is_empty();
 
+        // Damage were applied buffer transform to be in physical-space. We need to invert it to go
+        // back to buffer-coordinate. We'll apply the same transform to the element geometry for
+        // scale computation as it's already in physical space.
+        let transform = element.transform().invert();
         let damage_clips = if has_element_damage {
             PlaneDamageClips::from_damage(
                 self.surface.device_fd(),
                 element_config.properties.src,
                 element_config.geometry,
+                transform,
+                transform,
                 element_damage,
             )
             .ok()
--- a/src/backend/drm/surface/mod.rs
+++ b/src/backend/drm/surface/mod.rs
@@ -69,9 +69,11 @@ impl PlaneDamageClips {
         device: &DrmDeviceFd,
         src: Rectangle<f64, Buffer>,
         dst: Rectangle<i32, Physical>,
+        dst_transform: Transform,
+        damage_transform: Transform,
         damage: impl IntoIterator<Item = Rectangle<i32, Physical>>,
     ) -> io::Result<Option<Self>> {
-        let scale = src.size / dst.size.to_logical(1).to_buffer(1, Transform::Normal).to_f64();
+        let scale = src.size / dst.size.to_logical(1).to_buffer(1, dst_transform).to_f64();
 
         let mut rects = damage
             .into_iter()
@@ -81,8 +83,8 @@ impl PlaneDamageClips {
                     .to_logical(1f64)
                     .to_buffer(
                         1f64,
-                        Transform::Normal,
-                        &src.size.to_logical(1f64, Transform::Normal),
+                        damage_transform,
+                        &src.size.to_logical(scale, damage_transform),
                     )
                     .upscale(scale);
                 rect.loc += src.loc;
--- a/src/backend/drm/surface/gbm.rs
+++ b/src/backend/drm/surface/gbm.rs
@@ -349,9 +349,16 @@ where
         let dst = Rectangle::from_size((mode.size().0 as i32, mode.size().1 as i32).into());
 
         let damage_clips = damage.and_then(|damage| {
-            PlaneDamageClips::from_damage(self.drm.device_fd(), src, dst, damage)
-                .ok()
-                .flatten()
+            PlaneDamageClips::from_damage(
+                self.drm.device_fd(),
+                src,
+                dst,
+                Transform::Normal,
+                Transform::Normal,
+                damage,
+            )
+            .ok()
+            .flatten()
         });
 
         // Try to extract a native fence out of the supplied sync point if any


tux-linux:
This fixed the issue we had on [Quill OS](https://github.com/PorQ-Pine/quillstrap) with portrait mode not working. It was previously described here: https://github.com/niri-wm/niri/issues/2741
Since niri does not yet seem to support the latest smithay commits, I found it was easier just making a new commit with this commit's contents on our forked branch to make things work for now.

Kind request: if everything passes your tests and requirements, please merge the PR when it's possible for you. Thanks!

Ph4ntomas:
Ping @cmeissl, IIUC you're the one dealing with most of the DRM stuff. 

Do you mind having a look whenever you have some time?

src/backend/drm/compositor/mod.rs
```diff
@@ -2303,7 +2303,10 @@ where
                                 self.surface.device_fd(),
                                 config.properties.src,
                                 config.properties.dst,
-                                render_damage.iter().copied(),
+                                render_damage
```
cmeissl: We represent different coordinate systems in smithay by specifying a `Kind` for coordinates.

`PlaneDamageClips::from_damage` already uses the src, which is in buffer coordinates and dst which is in physical coordinates to bring the damage into buffer space.
Afaict we should apply the transform for render damage and direct scan-out.
So I wonder if it would be better to extend `PlaneDamageClips::from_damage` to take the transform and apply it there.

I guess it would also be a good opportunity to extract the calculation in `PlaneDamageClips::from_damage` so that it can be tested and add some tests for common cases.

cmeissl (commented):
Thanks for working on this!

src/backend/drm/compositor/mod.rs
Ph4ntomas: > Afaict we should apply the transform for render damage and direct scan-out.

Unfortunately, I haven't found a way to trigger direct scan-out with a device whose driver triggered the bug. OTOH I think the one thing I haven't tested yet is on a black background (which IIUC should allow direct-scanout for the primary plane). I'll give it a shot. Do you have some idea on how to ensure a direct scanout happen? (Or some way to tests things) 

One of the reason I went with the least amount of change is I don't want to break things I can't test :)

> So I wonder if it would be better to extend PlaneDamageClips::from_damage to take the transform and apply it there.

Agreed, if we need to fix it for all plane damages (direct scanout or otherwise) it's best to apply it there. I think there was some transform applied for other planes tho (based on the output and element transform), but I'm not sure in which coordinate system the damages ends up.

cmeissl: It should be enough to change this line to Kind::ScanoutCandidate instead of Unspecified https://github.com/Smithay/smithay/blob/e1fb2496c9d7ec4d994cc50fe61b7439686ba8b9/src/desktop/space/wayland/window.rs#L129

Ph4ntomas: I'm pretty sure I tried at some point, but I had a background color set on clear which IIRC disable direct scanout on the primary plane.

I'll try that again in the next few days, thanks for the hint

cmeissl: Direct scan-out on the primary plane can be a bit tricky. By default we limit the format to the same as of the swapchain, this has something to do with bandwidth limitations and multiple outputs. But you can re-enable all formats by changing anvil to pass FrameFlags that allow any format on the primary plane.

Beside that it depends on driver behavior, some drivers won't allow direct scan-out on the primary plane if it does not cover the whole crtc for example. iirc intel works fine, but the drm driver for the i.mx6 will not allow it for example.

A clear color other than black will block it if the element does not cover the whole crtc.

Ph4ntomas: > Direct scan-out on the primary plane can be a bit tricky.

Yeah, that's what make troubleshooting this issue tricky for me. Out of all my device, I've got only one whose driver use the FB_DAMAGE_CLIPS and AFAICT it only has the primary plane available, so let's hope it also supports direct scan-out for the primary plane.

If it doesn't, I might just extract the logic and see how the various rectangles ends up. As long as the math is correct, I may be able to confirm things that way

Ph4ntomas: Okay, so I wasn't able to test the actual code, but that

> If it doesn't, I might just extract the logic and see how the various rectangles ends up. As long as the math is correct, I may be able to confirm things that way

seems to confirm some of my assumptions (with actual data, not just by reading the code), but also raised some questions.

In `try_assign_plane`, we use `PlaneProperty::src` as the `src` rectangle when we call `PlaneDamageClips`. This value is returned by the element `Element::src()` function.
That same value is used in `AtomicRequest::set_plane` to set the `SRC_*` plane props, which confirms it's our framebuffer coordinates.
We also have `PlaneProperty::transform` used to set the plane's `rotation` property, and `PlaneProperty::dst` used for the `CRTC_*` properties.

At least in my tests, element's damage was already in the same coordinate system the src rectangle is, so we don't want to apply any rotation. OTOH, element.transform() returned `Transform::Normal`, so that was to be expected.

Digging into the code, to find where elements get there damage, I found this for `WaylandSurfaceRenderElement` :
https://github.com/Smithay/smithay/blob/e1fb2496c9d7ec4d994cc50fe61b7439686ba8b9/src/backend/renderer/utils/wayland.rs#L199-L211

Both case seem to have the computed damage with the buffer transform already applied, so I don't think we want to transform the element's damage in any way. 
I'll gladly test this even, but I'm not sure how to create an element with a rotated underlying buffer

cmeissl: The damage in the linked code is in buffer space, yes. But the damage used during rendering and also in the DrmCompositor is in physical space. It is transformed in https://github.com/Smithay/smithay/blob/e1fb2496c9d7ec4d994cc50fe61b7439686ba8b9/src/backend/renderer/element/surface.rs#L379

You can use weston-transformed to get a transformed buffer, unfortunately it uses shm and won't be directly scanned out. But it helps to see that the damage might need to be tranaformed to buffer space again.

Ph4ntomas: Thanks!

Okay so I was able to confirm we need to apply the buffer transform to damages during scanout.

We need two transform however.
https://github.com/Smithay/smithay/blob/e1fb2496c9d7ec4d994cc50fe61b7439686ba8b9/src/backend/drm/surface/mod.rs#L74

If the dest had a transform applied (compared to the src, we need to invert it or the scale computation will fail).

And the second one is applied to the damage rect itself, either to invert the buffer location, to go from logical  space to physical space

Ph4ntomas: I've amended my commit. The only part I'm not sure about is the `GbmBufferedSurface`. I couldn't find a usage example (anvil define a type alias, but then doesn't use it and I couldn't find any compositor making use of the type).

I wonder if the damage transform should be added to the `QueuedFb` object so it can be applied in `GbmBufferedSurface::submit`. For the time being, I used `Transform::Normal` so it doesn't change the current behavior.

Ph4ntomas: Okay that last change broke something somehow.

EDIT: Yeah forgot to remove the transformed applied in render_frame. That's fixed.

src/backend/drm/surface/mod.rs
```diff
@@ -81,8 +83,8 @@ impl PlaneDamageClips {
                     .to_logical(1f64)
                     .to_buffer(
                         1f64,
-                        Transform::Normal,
-                        &src.size.to_logical(1f64, Transform::Normal),
+                        damage_transform,
+                        &src.size.to_logical(1f64, damage_transform),
```
cmeissl: On a second thought I believe we also have to specify the scale here or otherwise a cropped region will be wrong.

Ph4ntomas: Yeah, make sense, the damage gets the buffer scale applied in https://github.com/Smithay/smithay/blob/e1fb2496c9d7ec4d994cc50fe61b7439686ba8b9/src/backend/renderer/element/surface.rs#L379, so we need to undo that as well

cmeissl (approved):
Thanks!

PR #2051, opened 2026-06-05
Ei protocol support using reis

hojjatabdollahi:
## Description
based on https://github.com/Smithay/smithay/pull/1388

It also pulls in the changes in https://github.com/ids1024/reis/pull/12

## Checklist

### DIFF
--- a/src/backend/libei/input.rs
+++ b/src/backend/libei/input.rs
@@ -1,0 +1,301 @@
+// Implementation of `InputBackend` for reis requests
+
+use reis::{
+    eis,
+    request::{self, DeviceCapability},
+};
+use std::path::PathBuf;
+
+use crate::backend::input::{self, InputBackend};
+
+use super::EiInput;
+
+/// Generic ei scroll event
+#[derive(Debug)]
+pub enum ScrollEvent {
+    /// Continuous scroll event
+    Delta(request::ScrollDelta),
+    /// Scroll cancel event
+    Cancel(request::ScrollCancel),
+    /// Discrete scroll event
+    Discrete(request::ScrollDiscrete),
+    /// Scroll stop event
+    Stop(request::ScrollStop),
+}
+
+impl InputBackend for EiInput {
+    type Device = request::Device;
+    type KeyboardKeyEvent = request::KeyboardKey;
+    type PointerAxisEvent = ScrollEvent;
+    type PointerButtonEvent = request::Button;
+    type PointerMotionEvent = request::PointerMotion;
+    type PointerMotionAbsoluteEvent = request::PointerMotionAbsolute;
+
+    type GestureSwipeBeginEvent = input::UnusedEvent;
+    type GestureSwipeUpdateEvent = input::UnusedEvent;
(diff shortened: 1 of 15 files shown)


src/backend/libei/input.rs
```diff
@@ -0,0 +1,300 @@
...
+impl input::Device for request::Device {
+    fn id(&self) -> String {
+        use reis::Interface;
+        use rustix::fd::{AsFd, AsRawFd};
+        let object = self.device().as_object();
+        let id = object.id();
+        match object.backend() {
+            Some(backend) => format!("ei-{}-{id}", backend.as_fd().as_raw_fd()),
+            None => format!("ei-{id}"),
+        }
```
ids1024: Having an ID that changes value when the backend weak pointer fails to upgrade is probably not helpful.

If we decide the ID can be backed-specific and specific to the EI connection, the FD isn't needed here.

hojjatabdollahi: Yeah. It should probably be just `format!("ei-{id}}")`.

hojjatabdollahi: I did update the doc saying it's backend-specific so we can just do "ei-{id}". Thinking more about it the changing id() after the backend is destroyed is pretty uncool as a key!

ids1024:
`ei_text` support is merged in `reis` now, and should work, though the code here still needs support for it (or there's not that much need to be pointing at that branch instead of the current release). There's also a `DeviceClosed` now that should work with just:

```rust
EisRequest::DeviceClosed(event) => Some(InputEvent::DeviceRemoved { device: event.device })
```

`reis` should be released with this soon.

ids1024:
I've released reis 0.7.0.

It would be good to get this merged. If anything needs to be changed in the future, smithay doesn't have a particularly stable API. The main blocker here was coming up with a decent API, which I think the API I added here in January is pretty solid. The cosmic-comp support needn't block this since we have support in `anvil`.

The `WIP`/`TODO`/`XXX` commits can be squashed into my first commit here. For `ei_text` support, we'll need `add_text()` etc. like we have for other types of input. Though we don't necessarily need that to merge this.

ids1024:
https://github.com/Smithay/smithay/pull/2051/changes/1d235ca040acf21756ab275b021d9fef00cca099 seems out of place here? `XWayland::spawn` doesn't really have anything to do with this.

hojjatabdollahi:
It's needed for [enabling ei_portal in xwayland](https://github.com/pop-os/cosmic-comp/pull/2442/changes#diff-a839b80920bf6042f047c1a140b6be4ca1367fd4de162e012be065264d6cde97R118). Do you suggest I open a separate PR for that?

ids1024:
Ah, right. I guess that's reasonable then.

ids1024 (approved):
This will need another review from someone other than me, but I think it should all be good to merge. We can add anything like `ei_text` later. It's nice to eventually merge things instead of repeatedly re-basing and updating the same commits.

Drakulix (approved):
LGTM as well. Thanks!

PR #2054, opened 2026-06-09
gles: Fix blitting from dmabufs

Drakulix:
## Description

This PR does two things which seems to be necessary to use blitting for blur successfully in certain setups:

1. Return `SyncPoint`s from `BlitFrame` as well. These are necessary to wait upon for the results of the blit to be visible, when sampling from the same framebuffer in e.g. a sub-renderer/frame. (Observed with the multi-gpu renderer and an intel-device being the render gpu)
2. Use the bind-texture path to import dmabufs for binding instead of `EGLImageTargetRenderbufferStorageOES`. This fixes blits to dmabufs being visible at all on nvidia gpus. (Which definitely sounds like a driver bug, but oh well.) In limited profiling I have not observed any performance differences, binding dmabufs imported as `external` will still fail, but in `GlesRenderer::bind_texture` instead.

## Checklist

### DIFF
--- a/src/backend/drm/dumb.rs
+++ b/src/backend/drm/dumb.rs
@@ -177,7 +177,7 @@ pub fn framebuffer_from_dumb_buffer(
 
             let fourcc = format.code;
             let (depth, bpp) = get_depth(fourcc)
-                .and_then(|d| get_bpp(fourcc).map(|b| (d, b)))
+                .zip(get_bpp(fourcc))
                 .ok_or_else(|| AccessError {
                     errmsg: "Unknown format for legacy framebuffer",
                     dev: drm.dev_path(),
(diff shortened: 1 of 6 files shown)


cmeissl (approved):
> Use the bind-texture path to import dmabufs for binding instead of EGLImageTargetRenderbufferStorageOES. This fixes blits to dmabufs being visible at all on nvidia gpus. (Which definitely sounds like a driver bug, but oh well.) In limited profiling I have not observed any performance differences, binding dmabufs imported as external will still fail, but in GlesRenderer::bind_texture instead.

afaict it shouldn't make a difference, the advantage of the render buffer being able to choose an optimal format won't work for dmabuf anyway.

YaLTeR:
> Use the bind-texture path to import dmabufs for binding instead of EGLImageTargetRenderbufferStorageOES. This fixes blits to dmabufs being visible at all on nvidia gpus. (Which definitely sounds like a driver bug, but oh well.) In limited profiling I have not observed any performance differences, binding dmabufs imported as external will still fail, but in GlesRenderer::bind_texture instead.

Seems like this commit breaks niri on the TTY on my laptop with AMD iGPU (and unused dGPU)

PR #2056, opened 2026-06-09
feat(drm_syncobj): add sync point accessors and sync-file import

endigma:
## Description
This PR adds focused DRM syncobj timeline-point accessors and an explicit sync-file import path for `DrmSyncPoint`.

`DrmSyncPoint::timeline()` and `DrmSyncPoint::point()` let callers inspect the timeline and numeric point without reaching through private state. The direct `AsFd` exposure on `DrmTimeline` is replaced with `DrmSyncPoint::import_sync_file(...)`, which better represents the operation callers need: materializing an exported sync fd as the fence for a specific timeline point.

The import implementation now follows the kernel/libdrm contract directly. It creates a temporary binary syncobj, imports the sync fd into that existing handle with `DRM_IOCTL_SYNCOBJ_FD_TO_HANDLE` and `IMPORT_SYNC_FILE`, then transfers point `0` into the target timeline point before destroying the temporary syncobj. This avoids the `drm` crate helper path that passes `handle = 0`, which the kernel rejects with `ENOENT` for sync-file import.

This is intended for compositors that drive Vulkan explicit sync with `VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SYNC_FD_BIT` and need to import the resulting sync fd into a client's release point.

## Checklist

### DIFF
--- a/src/wayland/drm_syncobj/sync_point.rs
+++ b/src/wayland/drm_syncobj/sync_point.rs
@@ -81,6 +81,12 @@ impl PartialEq for DrmTimeline {
     }
 }
 
+impl AsFd for DrmTimeline {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        self.0.timeline_fd.as_fd()
+    }
+}
+
 impl DrmTimeline {
     /// Import DRM timeline from file descriptor
     pub fn new(device: &DrmDeviceFd, fd: OwnedFd) -> io::Result<Self> {
@@ -113,6 +119,16 @@ pub struct DrmSyncPoint {
 }
 
 impl DrmSyncPoint {
+    /// Borrow the [`DrmTimeline`] this point lives on.
+    pub fn timeline(&self) -> &DrmTimeline {
+        &self.timeline
+    }
+
+    /// Numeric timeline value for this point.
+    pub fn point(&self) -> u64 {
+        self.point
+    }
+
     /// Create an eventfd that will be signaled by the syncpoint
     pub fn eventfd(&self) -> io::Result<Arc<OwnedFd>> {
         let fd = rustix::event::eventfd(
--- a/src/wayland/drm_syncobj/sync_point.rs
+++ b/src/wayland/drm_syncobj/sync_point.rs
@@ -81,12 +81,6 @@ impl PartialEq for DrmTimeline {
     }
 }
 
-impl AsFd for DrmTimeline {
-    fn as_fd(&self) -> BorrowedFd<'_> {
-        self.0.timeline_fd.as_fd()
-    }
-}
-
 impl DrmTimeline {
     /// Import DRM timeline from file descriptor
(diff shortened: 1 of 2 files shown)


ids1024:
We really should get https://github.com/Smithay/drm-rs/pull/226 merged so that can be used here, though defining the ioctl directly in smithay is probably okay in the meantime...

src/wayland/drm_syncobj/sync_point.rs
```diff
@@ -168,6 +178,82 @@ impl DrmSyncPoint {
...
+        const DRM_SYNCOBJ_FD_TO_HANDLE_FLAGS_IMPORT_SYNC_FILE: u32 = 1;
+
+        #[repr(C)]
+        struct DrmSyncobjHandle {
+            handle: u32,
+            flags: u32,
+            fd: i32,
+            pad: u32,
+            point: u64,
+        }
```
ids1024: https://docs.rs/drm-ffi/0.9.1/drm_ffi/struct.drm_syncobj_handle.html is public, so it doesn't need to be duplicated here.

ids1024:
Defining the ioctl number and calling the ioctl might be a little cleaner using `rustix::ioctl::{opcode::read_write, ioctl, Updater}` and `drm_ffi::DRM_IOCTL_BASE`.

But otherwise this seems like a good thing to have, and appears to match `wlr_drm_syncobj_timeline_import_sync_file()` and be correctly implemented. When https://github.com/Smithay/drm-rs/pull/226 or similar is released, we won't need the duplicated ioctl logic here.

Drakulix:
> But otherwise this seems like a good thing to have, and appears to match `wlr_drm_syncobj_timeline_import_sync_file()` and be correctly implemented. When [Smithay/drm-rs#226](https://github.com/Smithay/drm-rs/pull/226) or similar is released, we won't need the duplicated ioctl logic here.

Looks like the drm-rs discussion is pretty stalled right now. Imo it would be fine to go ahead implementing this in smithay for now. @ids1024 wanna do a full review of this?

ids1024:
I haven't actually tested the implementation here (not sure how this would be used outside a full Vulkan renderer? I don't know what @endigma has used it for), but it looks perfectly correct, other than the above-mentioned ways it could be updated to use `rustix::ioctl` and `drm_ffi`.

So if we don't care about those things for a "temporary" implementation, until the drm-rs release, this is fine to merge as is.

endigma:
Fixed the ioctl thing, this is used in https://github.com/caedenhq/wmpl, which is something approximating a full Vulkan renderer I suppose?

src/wayland/drm_syncobj/sync_point.rs
endigma: Should be fixed

Drakulix (approved):
Thanks!

Drakulix:
Looks like this needs to bump dri-sys to make sure it contains the struct for the check-minimal check.

endigma:
@Drakulix should be fixed

PR #2057, opened 2026-06-09
drm: check for EINVAL when dropping master

amshafer:
## Description

When using DRM leasing on the NVIDIA driver smithay may run into an error while trying to release master during `DrmLeaseState::new()->get_non_master_fd()`. The NVIDIA driver marks all clients as authenticated, so when smithay does the `get_client()` call it erroneously assumes the fd has master and tries to drop it. Dropping master fails since the fd never had it to begin with.

This change checks for EINVAL while releasing master. If we see EINVAL returned then we know that the fd, while potentially marked as authenticated, did not have master and therefore was unable to drop it. 

Open to suggestions if you do not think this is the right fix.

## Checklist


Drakulix (approved):
LGTM. Thanks!

PR #2058, opened 2026-06-09
drm/legacy: recover from EINVAL from a failed page flip

kelnos:
## Description

In the legacy path, we can't be sure that the hardware is set up as it was the last time we did a commit, as a VT switch or some other DRM master operation might change things from underneath us.

There's also a case where a compositor might call reset_buffers() right before a page flip is queued, which will drop the swapchain and destroy all framebuffers, which will cause the page flip to fail, as there's nothing left to flip.

So, if we get EINVAL, try to reset state, and if that causes there to be a pending commit, do the commit and then retry the page flip.

For reference, here's a (lightly edited and annotated for clarity and to remove repetitive prefixes) excerpt from the user's kernel log with DRM debugging enabled:

```
# LVDS-1 (CRTC 44) is modeset with FB:56 as its scanout buffer:
[ 8668.920949] [drm:drm_mode_setcrtc] [CRTC:44:crtc-0]
[ 8668.920977] [drm:drm_mode_setcrtc] [CONNECTOR:48:LVDS-1]
[ 8668.921003] [drm:drm_crtc_helper_set_config] [CRTC:44:crtc-0] [FB:56] #connectors=1 (x y) (0 0)

# FB:56 is removed while still bound, so the kernel tears the whole pipe down (this corresponds to
# xfwl4 calling reset_buffers() on a scale change, which it perhaps doesn't need to do, but the
# expectation is that this would be safe):
[ 8670.156213] DRM_IOCTL_MODE_RMFB
[ 8670.156632] [drm:drm_mode_rmfb_work_fn]  Removing [FB:56] from all active usage due to RMFB ioctl
[ 8670.156654] [drm:drm_framebuffer_remove] Disabling [CRTC:44:crtc-0] because [FB:56] is removed
[ 8670.156679] [drm:drm_crtc_helper_set_config] [CRTC:44:crtc-0] [NOFB]

# Then we get to the first render: a new fb is created:
[ 8671.449317] [drm:drm_mode_addfb2] [FB:56]
# ... then a page-flip onto the now-modeless CRTC, which leads to EINVAL:
[ 8671.455982] DRM_IOCTL_MODE_PAGE_FLIP
[ 8671.456002] comm="xfwl4.orig", pid=31653, ret=-22
```

Smithay translates this into a seemingly-unrecoverable error:

```
ContextLost(
    Access(
        AccessError {
            errmsg: "Failed to page flip",
            dev: Some(
                "/dev/dri/card0",
            ),
            source: Os {
                code: 22,
                kind: InvalidInput,
                message: "Invalid argument",
            },
        },
    ),
)
```

User reports that the issue goes away with this patch.

## Checklist

### DIFF
--- a/src/backend/drm/surface/legacy.rs
+++ b/src/backend/drm/surface/legacy.rs
@@ -1,6 +1,7 @@
 use drm::control::{Device as ControlDevice, Mode, PageFlipFlags, connector, crtc, encoder, framebuffer};
 
 use std::collections::HashSet;
+use std::io::ErrorKind;
 use std::sync::{
     Arc, Mutex, RwLock,
     atomic::{AtomicBool, Ordering},
@@ -9,7 +10,7 @@ use std::sync::{
 use crate::backend::drm::error::AccessError;
 use crate::{
     backend::drm::{
-        DrmDeviceFd, device::DrmDeviceInternal, device::legacy::set_connector_state, error::Error,
+        DrmDeviceFd, DrmSurface, device::DrmDeviceInternal, device::legacy::set_connector_state, error::Error,
     },
     utils::DevPath,
 };
@@ -333,7 +334,7 @@ impl LegacyDrmSurface {
 
     #[instrument(level = "trace", parent = &self.span, skip(self))]
     #[profiling::function]
-    pub fn page_flip(&self, framebuffer: framebuffer::Handle, event: bool) -> Result<(), Error> {
+    fn do_page_flip(&self, framebuffer: framebuffer::Handle, event: bool) -> Result<(), Error> {
         trace!("Queueing Page flip");
 
         if !self.active.load(Ordering::SeqCst) {
@@ -365,6 +366,29 @@ impl LegacyDrmSurface {
                 source,
             })
         })
+    }
+
+    #[instrument(level = "trace", parent = &self.span, skip(self))]
+    #[profiling::function]
+    pub fn page_flip(&self, framebuffer: framebuffer::Handle, event: bool) -> Result<(), Error> {
+        match self.do_page_flip(framebuffer, event) {
+            Ok(_) => Ok(()),
+            Err(err) => {
+                let lost_mode = matches!(err, Error::Access(AccessError { ref source, .. }) if source.kind() == ErrorKind::InvalidInput);
+
+                if lost_mode {
+                    self.reset_state::<DrmSurface>(None)?;
+                    if self.commit_pending() {
+                        self.commit(framebuffer, false)?;
+                        self.do_page_flip(framebuffer, event)
+                    } else {
+                        Err(err)
+                    }
+                } else {
+                    Err(err)
+                }
+            }
+        }
     }
 
     #[instrument(level = "trace", parent = &self.span, skip(self))]


cmeissl:
I am not against this change in general, but have a few questions about the cause.

> In the legacy path, we can't be sure that the hardware is set up as it was the last time we did a commit, as a VT switch or some other DRM master operation might change things from underneath us.

We already provide `DrmDevice::activate` for session resume after VT switch. When calling with `disable_connectors=true` it does already reset the device and surfaces. In case of optimistic (for flicker free operation) with `disable_connectors=false` anvil tries to detect the error during `queue_frame` and will reset the device. Seems the error case for the legacy path is currently not detected, but I guess this could be solved by either a change in the anvil logic or by returning a different error?

> There's also a case where a compositor might call reset_buffers() right before a page flip is queued, which will drop the swapchain and destroy all framebuffers, which will cause the page flip to fail, as there's nothing left to flip.

This sounds like a bug to me, a queued/pending frame should still contain a reference to framebuffer and prevent its destruction. The swapchain slot is also stored.

kelnos:
> We already provide `DrmDevice::activate` for session resume after VT switch. When calling with `disable_connectors=true` it does already reset the device and surfaces.

Right, perhaps I was a bit overzealous in thinking of possible causes.

> Seems the error case for the legacy path is currently not detected, but I guess this could be solved by either a change in the anvil logic or by returning a different error?

Possibly, but I'm concerned that just a blanket "`EINVAL` is recoverable" type of error could be incorrect: could there be cases where that does indicate an actual misconfiguration of the connector/CRTC/etc?

> This sounds like a bug to me, a queued/pending frame should still contain a reference to framebuffer and prevent its destruction.

This is probably what I'm hitting, as I can pretty reliably see that it doesn't happen if I don't call `reset_buffers()` after certain kinds of output changes that occur right on startup after enabling the connector.  (I'd rather not have to deal with this; though: it would be preferable if `reset_buffers()` was safe to call whenever, without things getting into a bad state internally.)

If the real cause is as you suggest above (and my change isn't needed to avoid the non-recoverable error), I'm open to trying to fix it at the source.  Will dig around and take a look.

cmeissl:
> > Seems the error case for the legacy path is currently not detected, but I guess this could be solved by either a change in the anvil logic or by returning a different error?
> 
> Possibly, but I'm concerned that just a blanket "`EINVAL` is recoverable" type of error could be incorrect: could there be cases where that does indicate an actual misconfiguration of the connector/CRTC/etc?

I haven't fully thought that through, but maybe it works by using the check you added to return a "recoverable" error. Handling the error like with atomic will result in `reset_state`, so the next call should commit which might fail if the error is not recoverable? But your approach might also be valid for the legacy case. @Drakulix Any preference from your side?
 
> > This sounds like a bug to me, a queued/pending frame should still contain a reference to framebuffer and prevent its destruction.
> 
> This is probably what I'm hitting, as I can pretty reliably see that it doesn't happen if I don't call `reset_buffers()` after certain kinds of output changes that occur right on startup after enabling the connector. (I'd rather not have to deal with this; though: it would be preferable if `reset_buffers()` was safe to call whenever, without things getting into a bad state internally.)
> 
> If the real cause is as you suggest above (and my change isn't needed to avoid the non-recoverable error), I'm open to trying to fix it at the source. Will dig around and take a look.

Can you describe how I could try to reproduce the issue? When is `reset_buffers` called, also on which abstraction?

kelnos:
> Can you describe how I could try to reproduce the issue? When is reset_buffers called, also on which abstraction?

I call `reset_buffers()` on the `DrmOutput`, when reconfiguring `Output`s.  Looking at the code, it should probably only be called when I actually change the mode, but I also seem to call it in other situations where it might not be necessary, like when only the output scale or transform has changed.  I *think* the only trigger for this issue is when I call it when it's not strictly needed, but I'm not sure if it can also happen when calling it after setting a new mode.  A user is reporting this; I don't have hardware that uses legacy modesetting, so my ability to test is limited.  I don't see this problem on my own hardware, which uses atomic modesetting.

But back to:

>This sounds like a bug to me, a queued/pending frame should still contain a reference to framebuffer and prevent its destruction.

I found something that maybe fits the bill on this?  [This bit](https://github.com/Smithay/smithay/blob/master/src/backend/drm/compositor/mod.rs#L2564-L2567) looks a little suspicious: if `user_data` is `None`, which is always the case when `handle_flip()` is called from `commit_frame()`, then `prepared_frame` will get dropped (or, maybe, the swapchain will have the only reference to it, and then lose it in `reset_buffers()`).  Thinking about a change such as this:

```diff
diff --git a/src/backend/drm/compositor/mod.rs b/src/backend/drm/compositor/mod.rs
index 289ce5bfe..c80163123 100644
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -2561,10 +2561,14 @@ where
                     self.reset_pending = false;
                 }
 
-                self.pending_frame = user_data.map(|user_data| PendingFrame {
-                    frame: prepared_frame.frame,
-                    user_data,
-                });
+                if let Some(user_data) = user_data {
+                    self.pending_frame = Some(PendingFrame {
+                        frame: prepared_frame.frame,
+                        user_data
+                    });
+                } else {
+                    self.current_frame = prepared_frame.frame;
+                }
             }
             Err(crate::backend::drm::error::Error::Access(ref access))
                 if access.source.kind() == ErrorKind::InvalidInput =>
```

That is, if `user_data` is none, instead store the new frame in `self.current_frame`.  Does that seem right?  Also not sure if `self.pending_frame` should be cleared out to `None` in that `else` branch as well, since the old code would effectively do that.  I can have the user with the older hardware test that if you think it makes sense and would fix the issue.

Edit: it looks like `commit_frame()` sets `self.pending_frame` to `None` on success before calling `handle_flip()`, so setting it here may not be strictly needed, but it might be a good idea to do so in order to be explicit?

cmeissl:
> I don't have hardware that uses legacy modesetting, so my ability to test is limited. I don't see this problem on my own hardware, which uses atomic modesetting.

You can use `SMITHAY_USE_LEGACY=1` to force using the legacy path. (technically also `yes`/`true`/`y` work)

> I found something that maybe fits the bill on this? [This bit](https://github.com/Smithay/smithay/blob/master/src/backend/drm/compositor/mod.rs#L2564-L2567) looks a little suspicious: if user_data is None, which is always the case when handle_flip() is called from commit_frame(), then prepared_frame will get dropped (or, maybe, the swapchain will have the only reference to it, and then lose it in reset_buffers()).

Wow, yeah, that indeed looks like it might be able to cause such an issue and is clearly wrong as it drops the frame.
You are also most likely right that without `reset_buffers` we are just lucky that the swapchain keeps the cached fb alive.

> That is, if user_data is none, instead store the new frame in self.current_frame. Does that seem right? Also not sure if self.pending_frame should be cleared out to None in that else branch as well, since the old code would effectively do that. I can have the user with the older hardware test that if you think it makes sense and would fix the issue.
>
> Edit: it looks like commit_frame() sets self.pending_frame to None on success before calling handle_flip(), so setting it here may not be strictly needed, but it might be a good idea to do so in order to be explicit?

Looking at the code I personally do not like to base it on the user_data option to be present/absent. Seems we actually only have two callers, but which represent different things. So maybe we should not try to handle both in the same `handle_flip`  function.

What do you think about splitting the function up a bit like this:

```
diff --git a/src/backend/drm/compositor/mod.rs b/src/backend/drm/compositor/mod.rs
index 289ce5bfe6..a5db046904 100644
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -2505,12 +2505,15 @@ where
             .frame
             .commit(&self.surface, self.supports_fencing, false, false);
 
-        if flip.is_ok() {
+        let res = self.handle_flip(&prepared_frame, flip);
+
+        if res.is_ok() {
             self.queued_frame = None;
             self.pending_frame = None;
+            self.current_frame = prepared_frame.frame;
         }
 
-        self.handle_flip(prepared_frame, None, flip)
+        res
     }
 
     /// Re-evaluates the current state of the crtc and forces calls to [`render_frame`](DrmCompositor::render_frame)
@@ -2546,13 +2549,21 @@ where
                 .page_flip(&self.surface, self.supports_fencing, allow_partial_update, true)
         };
 
-        self.handle_flip(prepared_frame, Some(user_data), flip)
+        let result = self.handle_flip(&prepared_frame, flip);
+
+        if result.is_ok() {
+            self.pending_frame = Some(PendingFrame {
+                frame: prepared_frame.frame,
+                user_data,
+            });
+        }
+
+        result
     }
 
     fn handle_flip(
         &mut self,
-        prepared_frame: PreparedFrame<A, F>,
-        user_data: Option<U>,
+        prepared_frame: &PreparedFrame<A, F>,
         flip: Result<(), crate::backend::drm::error::Error>,
     ) -> FrameResult<(), A, F> {
         match flip {
@@ -2560,11 +2571,6 @@ where
                 if prepared_frame.kind == PreparedFrameKind::Full {
                     self.reset_pending = false;
                 }
-
-                self.pending_frame = user_data.map(|user_data| PendingFrame {
-                    frame: prepared_frame.frame,
-                    user_data,
-                });
             }
             Err(crate::backend::drm::error::Error::Access(ref access))
                 if access.source.kind() == ErrorKind::InvalidInput =>

```

kelnos:
> You can use `SMITHAY_USE_LEGACY=1` to force using the legacy path.

Hm, I get an `EBUSY` on the first render, and then a blank screen thereafter (keyboard input still works to kill it at least).  Odd, it's wrapped in `TemporaryFailure`, so you'd think it should retry/recover.  Oh well.

> What do you think about splitting the function up a bit

Definitely, your suggestion makes it simpler and clearer what's going on.  Will work that up and have my legacy-path user test it (not sure I have the energy to debug why legacy isn't working on my laptop).  Should I close this and open a new PR, or ok to re-purpose this one?

kelnos:
Decided to open a new one (#2060).  Incidentally, this new patch also fixes the legacy path on my hardware with `SMITHAY_USE_LEGACY=1` set.

PR #2059, opened 2026-06-10
renderer/multigpu: Allow manually creating `MultiTexture`s

Drakulix:
## Description

Adds a method to the `MultiTexture` type to wrap an arbitrary texture from an underlying renderer. This is useful when doing e.g. highly renderer-specific offscreen rendering (e.g. blur), but want to use a `MultiRenderer` for compositing. You can't necessarily just get a reference to the underlying frame, as the `MultiFrame` internally tracks damaged regions.

## Checklist


cmeissl (approved):
LGTM

PR #2060, opened 2026-06-10
drm: store the current frame after committing it

kelnos:
## Description

`DrmCompositor::commit_frame()` would pass `None` for the `user_data` arg to `DrmCompositor::handle_flip()`.  When `handle_flip()` sees no user-data, it dropped the prepared_frame entirely.  The swapchain would still hold a reference to the frame, but if `reset_buffers()` is called after `commit_frame()` but before the page flip actually occurs, the swapchain would drop anything in all of its slots, and frame would be fully dropped.

On the legacy DRM path, this would cause `EINVAL` to be returned on page flip, because dropping the frames would also cause all framebuffers to be deleted, and the driver would have nothing to flip to.

This modifies `handle_flip()` to no longer take the `user_data` parameter; instead, callers (there are only two of them) are expected to store the pending frame in the appropriate place.  `commit_frame()` now stores the pending frame in `self.current_frame`, which will keep it alive in the case where all data in the swapchain is dropped.

## Checklist

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -2505,12 +2505,15 @@ where
             .frame
             .commit(&self.surface, self.supports_fencing, false, false);
 
-        if flip.is_ok() {
+        let res = self.handle_flip(&prepared_frame, flip);
+
+        if res.is_ok() {
             self.queued_frame = None;
             self.pending_frame = None;
+            self.current_frame = prepared_frame.frame;
         }
 
-        self.handle_flip(prepared_frame, None, flip)
+        res
     }
 
     /// Re-evaluates the current state of the crtc and forces calls to [`render_frame`](DrmCompositor::render_frame)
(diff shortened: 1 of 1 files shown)


kelnos:
Likely supersedes #2058.  Waiting on user testing.

cmeissl (approved):
Thanks!

Just to document this here too, we might have a regression on master: https://github.com/Smithay/smithay/pull/2061

kelnos:
Ah gotcha. I see from matrix that it looks like #2061 is bad too. Will hold off on repinning xfwl4 until that's fixed. 

On the plus side, my user with the old hardware reports this PR fixes his problem, so... progress!

cmeissl:
> Ah gotcha. I see from matrix that it looks like #2061 is bad too. Will hold off on repinning xfwl4 until that's fixed.

#2061 is merged and should fix the regression introduced by #2021 

> On the plus side, my user with the old hardware reports this PR fixes his problem, so... progress!

Great, thanks for reporting back.

kelnos:
Ah I mixed up 2021 and 2061, good to know.

PR #2061, opened 2026-06-11
Fix broken drm pending state

cmeissl:
## Description

Reverts 13f07dd1f1f20ef0b3a0daa1f860d87f27a816d7 and tries to re-initialize the mode blob.

@Drakulix Not sure about the connectors, but just using the current state might also not match the user expectation.
An alternative to this might be to not revert the commit, but document the expectations and require all callers to
re-issue the mode and connectors (not sure this is actually stored anywhere else as in the surface itself).

## Checklist


Drakulix (approved):
I think this is the better fix, so lets go with that. thx!

PR #2062, opened 2026-06-11
xwm: handle _NET_WM_STATE_MODAL client messages

mayakwd:
## Description
Currently, the XWM tracks `_NET_WM_STATE_MODAL` only as a part of inital/updated `_NET_WM_STATE` property exposed via `X11Surface::is_popup`, but `_NET_WM_STATE` client messages toggling the modal state at runtime are silently swallowed.  A client that maps a window first and then requests modality for it never gets the state applied, so the compositor will never be notified.

This PR wires the modal state through the same path as the existing fullscreen/above/below handling:
- New `XwmHandler::modal_request` / `XwmHandler::unmodal_request` callbacks, invoked on `_NET_WM_STATE_MODAL` client messages.
- New `X11Surface::set_modal` so the compositor can acknowledge the request by updating `_NET_WM_STATE` on the window, as it does for `set_fullscreen`/`set_above`.

Both callbacks have empty default implementations, so this is not a breaking change for existing `XwmHandler` implementations.

## Small note
`is_popup` in `X11Surface` can be a misleading term for a modal window, and `is_modal` sounds more reasonable IMO.
So we might want to consider renaming it (yet it's cosmetics and a breaking change as well).

## Motivation
This PR is also part of the larger work on bringing modality to comsic-comp ([xdg-dialog-v1](https://wayland.app/protocols/xdg-dialog-v1) protocol support).

## Checklist

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -425,6 +425,14 @@ pub trait XwmHandler {
     fn unfullscreen_request(&mut self, xwm: XwmId, window: X11Surface) {
         let _ = (xwm, window);
     }
+    /// Window requests to be set as a modal dialog (see [`X11Surface::is_popup`]).
+    fn modal_request(&mut self, xwm: XwmId, window: X11Surface) {
+        let _ = (xwm, window);
+    }
+    /// Window requests to no longer be a modal dialog.
+    fn unmodal_request(&mut self, xwm: XwmId, window: X11Surface) {
+        let _ = (xwm, window);
+    }
     /// Window requests to be minimized.
     fn minimize_request(&mut self, xwm: XwmId, window: X11Surface) {
         let _ = (xwm, window);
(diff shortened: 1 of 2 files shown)


Drakulix:
> ## Small note
> 
> `is_popup` in `X11Surface` can be a misleading term for a modal window, and `is_modal` sounds more reasonable IMO. So we might want to consider renaming it (yet it's cosmetics and a breaking change as well).

Probably a good idea. The original Xwm code was modelled to map onto wayland concepts where possible, but that turned out to be a bad match. At this point we try to use the X11 terminology to make it more obvious what which method refers to.

You could add `is_model`, make `is_popup` call `is_modal` internally and deprecate it.

mayakwd:
> Probably a good idea. The original Xwm code was modelled to map onto wayland concepts where possible, but that turned out to be a bad match. At this point we try to use the X11 terminology to make it more obvious what which method refers to.
> 
> You could add `is_model`, make `is_popup` call `is_modal` internally and deprecate it.

Good addition, I've updated the code with your suggestion.

Drakulix (approved):
thanks!

PR #2067, opened 2026-06-13
2066 touch input rework

### LINKED ISSUE #2066: RFC: Rework `input::touch` module
# Context
My original goal was to add a function similar to `PointerHandle::client_pointers` so a compositor has a simple way to identify if a given client bound `wl_touch` (e.g. to implement some pointer emulation). This led me to dig a bit into the touch handling code.

Following a discussion on Matrix with @cmeissl, I have some improvement ideas, but I'd like to have some feedback ~~before jumping to some implementations~~.
EDIT: I lied, I now have an implementation to try things out here: https://github.com/Ph4ntomas/smithay/tree/2066-touch-input-rework

## Event Serial
Currently, `TouchGrab` and `TouchTarget` have an additional `seq` parameter for each events. 
This parameter is used to de-duplicate events when propagating them to every `wl_touch`.

The serial is used in two place:
- Inside `TouchInternal`: the serial is stored for each touch slot as a way to know whether the slot was affected. 
This in turn is used during `frame` & `cancel` events, to pick `TouchTarget` to propagate the event to.
- Inside `wayland::seat::touch`, the serial is used to avoid sending events more than once to a given `wl_touch` instance.

## `frame` and `cancel` de-duplication

As mentioned above, `TouchInternal` doesn't perform event de-duplication itself. This is only an issue for `frame` and `cancel` events, since they are currently sent once per touch-slot participating in the frame.
However, this push the responsibility to handle de-duplication to `TouchTarget`.

# Proposed Changes

I see several issues with the current implementation:
- the implementation push event de-duplication toward the TouchTarget implementation. This isn't necessarily a bad thing per-se, but it's currently undocumented, and could lead to code duplication. It also means `TouchHandle` does not behave as one would expect by reading libinput or wayland documentation.
- The `seq` is only useful for `frame` and `cancel` but is passed to every function
- I believe the current implementation in `wayland::seat::touch` can eat up events under certain circumstances. The de-duplication code is:
`if ptr.id().same_client_as(&surface.id()) && last_seq.map(|last| last < seq).unwrap_or(true) { [...] }`
Since `seq` is shared among client, it's technically possible for a given client to have received `u32::MAX/2 - 1` as it's `last_seq`, then for `seq` to overflow before one of the client's surface gets focus again and start receiving new events.

### Keep the serial as an implementation detail:
With the exception of `frame` and `cancel`, the pending-frame serial doesn't need to be passed around, and could be kept inside `TouchInternal`. If a `TouchGrab` need it, an accessor could be added to `TouchInnerHandle`.
This isn't strictly required, but clean the API and avoid user error (it's currently possible for a user to pass an arbitrary serial).

The serial would be stored as:
```
struct TouchInternal {
[...]
pending_frame: Option<Serial>,
}
```
A call to any event would create the serial if it doesn't exist. `frame` and `cancel` would `Option::take()` it before proceeding.

### Move de-duplication logic to TouchInternal
This is harder to achieve. The main issue, as pointed out by @cmeissl, is that `TouchTarget` can't be compared. `PartialEq` is required as a super trait, but it's possible for a given `TouchTarget` to wrap another one, which would make the comparison return a false negative.

I'd like to propose the following changes:
Add the following method to TouchTarget:
```rust
/// Return last serial
fn last_serial(&self) -> Option<Serial>;
```

With this method, `TouchInternal` could compare the serial itself instead of the instance, and decide whether to call `frame` or `cancel`. The `TouchTarget` would only be responsible for storing the event serial on a call to `frame` or `cancel`.

EDIT: For that last part, we may need to make `frame` and `cancel` take `&mut self` instead of `&self` (or let mutability as an exercise for the user)

Ph4ntomas:
## Description

This series rework the touch handling API with the following changes:
- frame serial are now internal to `TouchInternal`. They are only sent
  to the `TouchTarget` as part of a frame() or cancel() call.
  EDIT: Instead of a Serial, a new FrameMarker is used.
  - The frame serial are generated once per frame, and discarded on a call
  to frame or cancel.
  EDIT: That's still true, but since the FrameMarker type is an id, it's not as simple. TouchInternal::frame no longer discard the marker, as it may not be called. The id for the pending frame is now removed by TouchHandle on a call to `frame` or `cancel` instead. 
- `TouchInternal` is now responsible for de-duplicating frame and cancel
  events. It does so by retrieving the `last_serial` from the
  `TouchTarget` before sending the event.
  It's possible for a target to opt-out of this behavior by always returning None in `last_serial`
  EDIT: last serial was changed to `last_frame`

Closes #2066 

## Checklist

### DIFF
--- a/src/input/touch/mod.rs
+++ b/src/input/touch/mod.rs
@@ -9,13 +9,27 @@ use tracing::{info_span, instrument};
 use wayland_server::Weak;
 
 use crate::backend::input::TouchSlot;
-use crate::utils::{IsAlive, Logical, Point, Serial, SerialCounter};
+use crate::utils::{IsAlive, Logical, Point, Serial};
 
 pub use grab::{DefaultGrab, GrabStartData, TouchDownGrab, TouchGrab};
 
 use super::{GrabStatus, Seat, SeatHandler};
 
 mod grab;
+
+crate::utils::ids::id_gen!(frame_marker);
+
+/// A marker to identify a given touch frame.
+///
+/// This marker is sent to [`TouchTarget`] during `frame` and `cancel`, and can be returned via the
+/// [`TouchTarget::last_frame`] function. They are used internally by [`TouchHandle`] to avoid
+/// sending `frame` and `cancel` event more than once.
+///
+/// FrameMarker are used instead of comparing [`TouchTarget`] as different [`TouchTarget`] could
+/// refer to the same underlying object (e.g. multiple surfaces belonging to the same wayland
+/// `Client`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct FrameMarker(usize);
 
 /// An handle to a touch handler
 ///
@@ -29,8 +43,7 @@ mod grab;
 pub struct TouchHandle<D: SeatHandler> {
     pub(crate) inner: Arc<Mutex<TouchInternal<D>>>,
     #[cfg(feature = "wayland_frontend")]
-    pub(crate) known_instances:
-        Arc<Mutex<Vec<(Weak<wayland_server::protocol::wl_touch::WlTouch>, Option<Serial>)>>>,
+    pub(crate) known_instances: Arc<Mutex<Vec<Weak<wayland_server::protocol::wl_touch::WlTouch>>>>,
     pub(crate) span: tracing::Span,
 }
 
@@ -81,16 +94,24 @@ impl<D: SeatHandler> std::cmp::Eq for TouchHandle<D> {}
 
 pub(crate) struct TouchInternal<D: SeatHandler> {
     focus: HashMap<TouchSlot, TouchSlotState<D>>,
-    seq_counter: SerialCounter,
+    pending_frame: Option<FrameMarker>,
     default_grab: Box<dyn Fn() -> Box<dyn TouchGrab<D>> + Send + 'static>,
     grab: GrabStatus<dyn TouchGrab<D>>,
+}
+
+impl<D: SeatHandler> Drop for TouchInternal<D> {
+    fn drop(&mut self) {
+        if let Some(marker) = self.pending_frame.take() {
+            frame_marker::remove(marker.0);
+        }
+    }
 }
 
 struct TouchSlotState<D: SeatHandler> {
     focus: Option<(<D as SeatHandler>::TouchFocus, Point<f64, Logical>)>,
     frame_pending: Option<<D as SeatHandler>::TouchFocus>,
-    pending: Serial,
-    current: Option<Serial>,
+    pending: FrameMarker,
+    current: Option<FrameMarker>,
 }
 
 impl<D: SeatHandler> fmt::Debug for TouchSlotState<D> {
@@ -178,26 +199,34 @@ where
     ///
     /// This touch point is assigned a unique ID. Future events from this touch point reference this ID.
     /// The ID ceases to be valid after a touch up event and may be reused in the future.
-    fn down(&self, seat: &Seat<D>, data: &mut D, event: &DownEvent, seq: Serial);
+    fn down(&self, seat: &Seat<D>, data: &mut D, event: &DownEvent);
 
     /// The touch point has disappeared.
     ///
     /// No further events will be sent for this touch point and the touch point's ID
     /// is released and may be reused in a future touch down event.
-    fn up(&self, seat: &Seat<D>, data: &mut D, event: &UpEvent, seq: Serial);
+    fn up(&self, seat: &Seat<D>, data: &mut D, event: &UpEvent);
 
     /// A touch point has changed coordinates.
-    fn motion(&self, seat: &Seat<D>, data: &mut D, event: &MotionEvent, seq: Serial);
+    fn motion(&self, seat: &Seat<D>, data: &mut D, event: &MotionEvent);
 
     /// Indicates the end of a set of events that logically belong together.
-    fn frame(&self, seat: &Seat<D>, data: &mut D, seq: Serial);
+    ///
+    /// The `marker` parameter is used to avoid re-sending the cancel event if the [`TouchTarget`]
+    /// has multiple active touch slot, or if multiple [`TouchTarget`] belong to the same underlying
+    /// target (e.g. a single client). See [`TouchTarget::last_frame`].
+    fn frame(&self, seat: &Seat<D>, data: &mut D, marker: FrameMarker);
 
     /// Touch session cancelled.
     ///
     /// Touch cancellation applies to all touch points currently active on this target.
     /// The client is responsible for finalizing the touch points, future touch points on
     /// this target may reuse the touch point ID.
-    fn cancel(&self, seat: &Seat<D>, data: &mut D, seq: Serial);
+    ///
+    /// The `marker` parameter is used to avoid re-sending the cancel event if the [`TouchTarget`]
+    /// has multiple active touch slot, or if multiple [`TouchTarget`] belong to the same underlying
+    /// target (e.g. a single client). See [`TouchTarget::last_frame`].
+    fn cancel(&self, seat: &Seat<D>, data: &mut D, marker: FrameMarker);
 
     /// Sent when a touch point has changed its shape.
     ///
@@ -206,14 +235,20 @@ where
     /// length describes the shorter diameter. Major and minor are orthogonal and both are specified
     /// in surface-local coordinates. The center of the ellipse is always at the touch point location
     /// as reported by [`TouchTarget::down`] or [`TouchTarget::motion`].
-    fn shape(&self, seat: &Seat<D>, data: &mut D, event: &ShapeEvent, seq: Serial);
+    fn shape(&self, seat: &Seat<D>, data: &mut D, event: &ShapeEvent);
 
     /// Sent when a touch point has changed its orientation.
     ///
     /// The orientation describes the clockwise angle of a touch point's major axis to the positive surface
     /// y-axis and is normalized to the -180 to +180 degree range. The granularity of orientation depends
     /// on the touch device, some devices only support binary rotation values between 0 and 90 degrees.
-    fn orientation(&self, seat: &Seat<D>, data: &mut D, event: &OrientationEvent, seq: Serial);
+    fn orientation(&self, seat: &Seat<D>, data: &mut D, event: &OrientationEvent);
+
+    /// Returns last know [`FrameMarker`].
+    ///
+    /// If this function returns `Some(marker)`, [`TouchHandle`] will use that value to avoid
+    /// sending the `frame` or `cancel` event more than once.
+    fn last_frame(&self, seat: &Seat<D>, data: &mut D) -> Option<FrameMarker>;
 }
 
 impl<D: SeatHandler + 'static> TouchHandle<D> {
@@ -295,9 +330,8 @@ impl<D: SeatHandler + 'static> TouchHandle<D> {
     ) {
         let mut inner = self.inner.lock().unwrap();
         let seat = self.get_seat(data);
-        let seq = inner.seq_counter.next_serial();
         inner.with_grab(data, &seat, |data, handle, grab| {
-            grab.down(data, handle, focus, event, seq);
+            grab.down(data, handle, focus, event);
         });
     }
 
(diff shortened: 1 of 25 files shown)


src/wayland/seat/touch.rs
```diff
@@ -32,38 +35,58 @@ impl<D: SeatHandler + 'static> TouchHandle<D> {
...
+fn with_touch_target_data<T, F: FnOnce(&mut TouchTargetData) -> T>(surface: &WlSurface, f: F) -> T {
+    compositor::with_states(surface, |states| {
+        let data = states
+            .data_map
+            .get_or_insert_threadsafe(|| Mutex::new(TouchTargetData::default()));
+
+        f(&mut data.lock().unwrap())
+    })
+}
+
```
Ph4ntomas: Okay, this is not what I had planned.
This is meant to be stored per client, but I forgot that when implementing the poc ...

src/input/touch/mod.rs
```diff
@@ -80,17 +94,16 @@ impl<D: SeatHandler> std::cmp::Eq for TouchHandle<D> {}
 
 pub(crate) struct TouchInternal<D: SeatHandler> {
     focus: HashMap<TouchSlot, TouchSlotState<D>>,
-    seq_counter: SerialCounter,
-    pending_frame: Option<Serial>,
+    pending_frame: Option<FrameMarker>,
```
Ph4ntomas, line 97: A drop implementation is missing here. We won't remove the pending frame if the TouchHandle is dropped. 

I'll look into that.

src/input/touch/mod.rs
```diff
@@ -545,36 +571,38 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
...
         focus: Option<(<D as SeatHandler>::TouchFocus, Point<f64, Logical>)>,
         event: &DownEvent,
-        seq: Serial,
     ) {
+        let marker = self.frame_marker();
         self.focus
             .entry(event.slot)
             .and_modify(|state| {
-                state.pending = seq;
+                state.pending = marker;
```
cmeissl, line 579: I guess here we have the same issue that a pending frame might get overridden leaving the marker dangling.

I wonder if it is actually a good idea to use gen_id or if a simple u64 would also work fine?

src/input/touch/mod.rs
```diff
@@ -588,66 +616,83 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
...
             let mut new_event = event.clone();
             new_event.location -= *loc;
-            focus.motion(seat, data, &new_event, seq);
+            focus.motion(seat, data, &new_event);
         }
     }
 
-    fn frame(&mut self, data: &mut D, seat: &Seat<D>, seq: Serial) {
+    fn frame(&mut self, data: &mut D, seat: &Seat<D>) {
+        let seq = self.pending_frame.expect("frame called without prior events");
```
cmeissl: Not sure we really want to panic here, maybe we should just do something like:

```
let Some(fame) = self.pending_frame.take() else {
    tracing::warn!("frame called without prior events");
    return;
}
```

If necessary we could add a debug assert before that.

src/input/touch/mod.rs
```diff
@@ -588,66 +616,83 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
...
+                if focus.last_frame(seat, data) != Some(seq) {
+                    focus.frame(seat, data, seq);
+                }
             }
         }
     }
 
-    fn cancel(&mut self, data: &mut D, seat: &Seat<D>, seq: Serial) {
+    fn cancel(&mut self, data: &mut D, seat: &Seat<D>) {
+        let seq = self.pending_frame.expect("cancel called without prior events");
```
cmeissl: same here

src/wayland/seat/touch.rs
```diff
@@ -32,38 +37,69 @@ impl<D: SeatHandler + 'static> TouchHandle<D> {
...
+                    .unwrap()
+                    .last_frame
+                    .lock()
+                    .unwrap());
+                break;
+            }
+        }
+    }
+}
+
```
cmeissl: Isn't the last frame now stored per client? It seems unnecessary in that case to iterate through all instances and read/overwrite the value multiple times?

src/wayland/compositor/mod.rs
```diff
@@ -620,13 +621,15 @@ pub struct CompositorState {
...
     scale_override: Arc<AtomicF64>,
+    last_touch_frame: Arc<Mutex<Option<FrameMarker>>>,
 }
 
 impl Default for CompositorClientState {
     fn default() -> Self {
         CompositorClientState {
             queue: Mutex::new(None),
             scale_override: Arc::new(AtomicF64::new(1.)),
+            last_touch_frame: Arc::new(Mutex::new(None)),
```
cmeissl: same here with storing the marker

cmeissl (changes requested):
Thanks for working through this!

In general this seems like a nice simplification of the API.

src/input/touch/mod.rs
Ph4ntomas: Ah right, I wasn't sure whether to panic (it's a compositor bug) or to log and return.
IMO both make sense, but there's no need to crash the compositor.

> we could add a debug assert before that.

That could make sense, yeah

src/wayland/compositor/mod.rs
Ph4ntomas: > same here with storing the marker

I'm not sure what you mean

EDIT: Or I guess you mean we don't need the Arc<Mutex<>> here if we set it only  once

src/wayland/seat/touch.rs
Ph4ntomas: Yeah this is a bit of a dirty trick.

The issue is there's no way to retrieve the client state without further constraining the Data generics in the impl block for `WlSurface`.

Adding `CompositorHandler` would solve the issue, but I wasn't sure about that and wanted to do some testing before.

I'll add the constraint and drop the trick, that will be cleaner.

src/input/touch/mod.rs
Ph4ntomas: > pending frame might get overridden leaving the marker dangling.

As long as the marker is removed after a `frame`, a `cancel` or on `drop` (this one is pending), there's no way to make it dangling, since it's reused for every intermediate events until `frames` or `cancel`. 
We could make the remove() automatic however:

```rust
enum FrameMarkerInner {
    Primary(usize),
    Secondary(usize),
}

pub struct FrameMarker(FrameMarkerInner);

impl Drop for FrameMarkerInner {
   fn drop(&mut self) {
       if let Self::Primary(id) {
           frame_marker::remove(id);
       };
   }
}
```

So long as the Primary stay inside the TouchInternal, there's no way for the id to become dangling.

(The full wrapping with drop impl is safer, but I find it a bit over-engineered for an opaque type meant to disambiguate two frames)

> I wonder if it is actually a good idea to use gen_id or if a simple u64 would also work fine?

Both work. The marker has the nice property that if a client has multiple seats with multiple wl_touch, each frames are guaranteed to be unique.

src/input/touch/mod.rs
Ph4ntomas: Added a Drop impl on TouchHandle

src/input/touch/mod.rs
Ph4ntomas: I've added a Drop impl on `TouchInternal`, but kept the `id` impl.

The marker can only be created on an event, and only if there are no active marker.

Active markers are cleared & removed:
- when leaving TouchHandle::frame
- when leaving TouchHandle::cancel
- on TouchIntenal::drop.

We're doing it in `TouchHandle` instead of `TouchInternal` because `TouchInternal::frame()` might be called more than once if e.g. nested/composite grabs are implemented. 
Ideally you'd avoid calling it more than once, but I'd rather have a somewhat sane behavior if it happen.

Ph4ntomas:
I think that's everything. 

I've resolved the comment. Let me know if there's a policy as to who should mark discussion as resolved (e.g. the PR author or the reviewer)

src/input/touch/mod.rs
cmeissl: > We're doing it in `TouchHandle` instead of `TouchInternal` because `TouchInternal::frame()` might be called more than once if e.g. nested/composite grabs are implemented.
> Ideally you'd avoid calling it more than once, but I'd rather have a somewhat sane behavior if it happen.

If a grab decides to delay `frame`/`cancel` for whatever reason (not that I have something specific in mind) that might be problematic.

Ph4ntomas: Yeah, missed that case.

We're protected against sending frame/cancel more than once too if we just `Option::take()` the marker in the `TouchInternal`. If the calling code calls `TouchInnerHandle::frame` or `TouchInnerHandle::cancel` more than once, the first time will send the frame/cancel event, subsequent call will warn that there's a bug in the implementation.

We could add a debug assert to crash the user code, too.

This is still a sane behavior IMO, so I've move that in TouchInternal.

src/input/touch/mod.rs
```diff
@@ -545,36 +571,38 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
...
             let mut new_event = event.clone();
             new_event.location -= *loc;
-            focus.down(seat, data, &new_event, seq);
+            focus.down(seat, data, &new_event);
         }
     }
 
-    fn up(&mut self, data: &mut D, seat: &Seat<D>, event: &UpEvent, seq: Serial) {
+    fn up(&mut self, data: &mut D, seat: &Seat<D>, event: &UpEvent) {
+        let seq = self.frame_marker();
```
cmeissl: nit: should be updated to `marker`

src/input/touch/mod.rs
```diff
@@ -588,66 +616,93 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
         seat: &Seat<D>,
         _focus: Option<(<D as SeatHandler>::TouchFocus, Point<f64, Logical>)>,
         event: &MotionEvent,
-        seq: Serial,
     ) {
+        let seq = self.frame_marker();
```
cmeissl: nit: should be updated to `marker`

src/input/touch/mod.rs
```diff
@@ -588,66 +616,93 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
...
             let mut new_event = event.clone();
             new_event.location -= *loc;
-            focus.motion(seat, data, &new_event, seq);
+            focus.motion(seat, data, &new_event);
         }
     }
 
-    fn frame(&mut self, data: &mut D, seat: &Seat<D>, seq: Serial) {
+    fn frame(&mut self, data: &mut D, seat: &Seat<D>) {
+        let Some(seq) = self.pending_frame.take() else {
```
cmeissl: nit: should be updated to `marker`

src/input/touch/mod.rs
```diff
@@ -588,66 +616,93 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
...
+                }
             }
         }
+
+        frame_marker::remove(seq.0);
     }
 
-    fn cancel(&mut self, data: &mut D, seat: &Seat<D>, seq: Serial) {
+    fn cancel(&mut self, data: &mut D, seat: &Seat<D>) {
+        let Some(seq) = self.pending_frame.take() else {
```
cmeissl: nit: should be updated to `marker`

src/input/touch/mod.rs
```diff
@@ -588,66 +616,93 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
...
+                }
             }
         }
+
+        frame_marker::remove(seq.0);
     }
 
-    fn shape(&mut self, data: &mut D, seat: &Seat<D>, event: &ShapeEvent, seq: Serial) {
+    fn shape(&mut self, data: &mut D, seat: &Seat<D>, event: &ShapeEvent) {
+        let seq = self.frame_marker();
```
cmeissl: nit: should be updated to `marker`

src/input/touch/mod.rs
```diff
@@ -588,66 +616,93 @@ impl<D: SeatHandler + 'static> TouchInternal<D> {
...
         state.pending = seq;
         if let Some((focus, _)) = state.focus.as_ref() {
-            focus.shape(seat, data, event, seq);
+            focus.shape(seat, data, event);
         }
     }
 
-    fn orientation(&mut self, data: &mut D, seat: &Seat<D>, event: &OrientationEvent, seq: Serial) {
+    fn orientation(&mut self, data: &mut D, seat: &Seat<D>, event: &OrientationEvent) {
+        let seq = self.frame_marker();
```
cmeissl: nit: should be updated to `marker`

cmeissl (approved):
Thanks! Just some small nits about `seq`

@Ph4ntomas Do you prefer to clean up the history and a rebase or shall we squash the changes?

Ph4ntomas:
> Thanks! Just some small nits about `seq`
> 
> @Ph4ntomas Do you prefer to clean up the history and a rebase or shall we squash the changes?

I've rebased on latest master, and cleaned-up the history with the fix for the remaining nit (along with writing a better commit for the main api change).

You can still squash if you feel it should be done or if it's the project policy.

PR #2071, opened 2026-06-15
wlr-layer-shell: skip pre-commit validation for destroyed surfaces

BVMiko:
## Description
I hit an issue while using gtk4-layer-shell: the client gets disconnected the moment it hides a layer-shell window.

To hide a window, gtk4-layer-shell destroys the `zwlr_layer_surface_v1` and then commits the `wl_surface` (with a nil buffer).
On destroy the cached layer state is reset to defaults (size 0, no anchors), but the pre-commit hook is still attached to the surface, so that commit runs it. The state now matches the `pending.size.w == 0 && !pending.anchor.anchored_horizontally()` check, which posts InvalidSize ("width 0 requested without setting left and right anchors") and takes down the whole client.

This PR just skips the pre-commit validation once the `zwlr_layer_surface_v1` has been destroyed. It only short-circuits the hook; it doesn't affect how the surface is torn down.

## Checklist

### DIFF
--- a/src/wayland/shell/wlr_layer/mod.rs
+++ b/src/wayland/shell/wlr_layer/mod.rs
@@ -484,6 +484,10 @@ impl LayerSurface {
         compositor::with_states(surface, |states| {
             let mut role = states.data_map.get::<LayerSurfaceData>().unwrap().lock().unwrap();
 
+            if !role.surface.alive() {
+                return;
+            }
+
             let mut guard_layer = states.cached_state.get::<LayerSurfaceCachedState>();
             let pending = guard_layer.pending();
 


Drakulix:
> This PR just skips the pre-commit validation once the `zwlr_layer_surface_v1` has been destroyed. It only short-circuits the hook; it doesn't affect how the surface is torn down.

Hmm, I am not sure this is fully correct. We obviously don't want to validate any state that would be specific to the layer-surface (like geometry or anchors), but attaching a null buffer actually has an effect further down the commit hook, which is resetting the unmapped surface for another initial configure.

Arguably gtk4-layer-shell should really send the unmap (null-buffer commit) before destroying the surface, but we can work-around it.

PR #2073, opened 2026-06-17
feat: add `ei_text` support

hojjatabdollahi:
## Description
Adds `Text` capability to the seat. Also adds `TextKeySym` and `TextUtf8` as EiInputEvents. 

## Checklist

### DIFF
--- a/anvil/src/libei.rs
+++ b/anvil/src/libei.rs
@@ -39,7 +39,9 @@ pub fn listen_eis(handle: &calloop::LoopHandle<'static, AnvilState<UdevData>>) {
                         let dh = data.display_handle.clone();
                         data.process_input_event(&dh, event);
                     }
-                    EiInputEvent::TextKeysym { .. } | EiInputEvent::TextUtf8 { .. } => {}
+                    EiInputEvent::TextKeysym { .. } | EiInputEvent::TextUtf8 { .. } => {
+                        // Anvil doesn't add a text device to the libei seat
+                    }
                 })
                 .unwrap();
             Ok(calloop::PostAction::Continue)
--- a/anvil/src/libei.rs
+++ b/anvil/src/libei.rs
@@ -39,6 +39,7 @@ pub fn listen_eis(handle: &calloop::LoopHandle<'static, AnvilState<UdevData>>) {
                         let dh = data.display_handle.clone();
                         data.process_input_event(&dh, event);
                     }
+                    EiInputEvent::TextKeysym { .. } | EiInputEvent::TextUtf8 { .. } => {}
                 })
                 .unwrap();
             Ok(calloop::PostAction::Continue)
--- a/anvil/src/libei.rs
+++ b/anvil/src/libei.rs
@@ -31,7 +31,7 @@ pub fn listen_eis(handle: &calloop::LoopHandle<'static, AnvilState<UdevData>>) {
                         let seat = connection.add_seat("default");
                         let _ = seat.add_keyboard("virtual keyboard", XkbConfig::default());
                         seat.add_pointer("virtual pointer");
-                        seat.add_pointer_absolute("virtual absolute pointer");
+                        seat.add_pointer_absolute("virtual absolute pointer", &[]);
                         seat.add_touch("virtual touch");
                     }
                     EiInputEvent::Disconnected => {}
--- a/src/backend/libei/seat.rs
+++ b/src/backend/libei/seat.rs
@@ -57,10 +57,12 @@ impl EiInputSeat {
             pointer_absolute: None,
             pointer_absolute_regions: Vec::new(),
             touch: None,
+            text: None,
             device_keyboard: None,
             device_pointer: None,
             device_pointer_absolute: None,
             device_touch: None,
+            device_text: None,
             bound_capabilities: BitFlags::empty(),
         })))
     }
@@ -98,6 +100,26 @@ impl EiInputSeat {
         let mut inner = self.0.lock().unwrap();
         inner.device_keyboard = None;
         inner.keyboard = None;
+    }
+
+    /// Send the compositor's current modifier state to this seat's keyboard device via
+    /// `ei_keyboard.modifiers`, so the client keeps its keyboard state in sync with the
+    /// compositor (mirroring `wl_keyboard.modifiers`). No-op if there is no bound keyboard.
+    pub fn keyboard_modifiers(&self, depressed: u32, locked: u32, latched: u32, group: u32) {
+        let inner = self.0.lock().unwrap();
+        let Some(device) = inner.device_keyboard.as_ref() else {
+            return;
+        };
+        let Some(keyboard) = device.device.interface::<reis::eis::Keyboard>() else {
+            return;
+        };
+        let Some(connection) = inner.connection.upgrade() else {
+            return;
+        };
+        connection.connection.with_next_serial(|serial| {
+            keyboard.modifiers(serial, depressed, locked, latched, group);
+        });
+        let _ = connection.connection.flush();
     }
 
     /// Add a pointer device to the EI seat
@@ -168,6 +190,25 @@ impl EiInputSeat {
         inner.touch = None;
     }
 
+    /// Add a text device to the EI seat.
+    ///
+    /// A text device lets clients inject input by keysym or UTF-8 string (`ei_text`), independent
+    /// of any keymap. Calling on a seat that already has a text device will remove that device and
+    /// add a new one.
+    pub fn add_text(&self, name: &str) {
+        let mut inner = self.0.lock().unwrap();
+        inner.device_text = None;
+        inner.text = Some(name.to_string());
+        inner.refresh_devices();
+    }
+
+    /// Remove text device from the EI seat
+    pub fn remove_text(&self) {
+        let mut inner = self.0.lock().unwrap();
+        inner.device_text = None;
+        inner.text = None;
+    }
+
     /// Remove seat from EI connection
     pub fn remove(&self) {
         let inner = self.0.lock().unwrap();
@@ -195,11 +236,13 @@ struct EiInputSeatInner {
     // coordinate space(s) the client may address (see `EiRegion`).
     pointer_absolute_regions: Vec<EiRegion>,
     touch: Option<String>,
+    text: Option<String>,
     // Devices created in response to client bind
     device_keyboard: Option<DeviceDropWrapper>,
     device_pointer: Option<DeviceDropWrapper>,
     device_pointer_absolute: Option<DeviceDropWrapper>,
     device_touch: Option<DeviceDropWrapper>,
+    device_text: Option<DeviceDropWrapper>,
 }
 
 impl EiInputSeatInner {
@@ -295,6 +338,22 @@ impl EiInputSeatInner {
                     device: device.clone(),
                 });
                 self.device_touch = Some(DeviceDropWrapper::new(device, &self.event_sender));
+            }
+        }
+
+        if self.device_text.is_none() && self.bound_capabilities.contains(DeviceCapability::Text) {
+            if let Some(name) = self.text.as_ref() {
+                let device = self.seat.add_device(
+                    Some(name),
+                    DeviceType::Virtual,
+                    DeviceCapability::Text.into(),
+                    |_| {},
+                );
+                device.resumed();
+                let _ = self.event_sender.send(InputEvent::DeviceAdded {
+                    device: device.clone(),
+                });
+                self.device_text = Some(DeviceDropWrapper::new(device, &self.event_sender));
             }
         }
     }
--- a/src/backend/libei/seat.rs
+++ b/src/backend/libei/seat.rs
@@ -5,9 +5,32 @@ use xkbcommon::xkb;
 use crate::{
     backend::input::InputEvent,
     input::keyboard::{KeymapFile, XkbConfig},
+    utils::{Logical, Rectangle},
 };
 
 use super::{EiInput, EiInputConnection, EiInputConnectionInner};
+
+/// A region advertised on an EI absolute pointer device.
+///
+/// Describes a rectangular coordinate space the client may point within, in the
+/// compositor's logical coordinate system. Clients that capture at a different
+/// (e.g. physical) resolution use `scale` to convert their coordinates into the
+/// logical space the compositor expects.
+///
+/// `rect` is the logical offset and size of the region; `scale` the
+/// physical-to-logical scale of the underlying output (e.g. `2.0` for a 200%
+/// display). `mapping_id`, when set, identifies which screen/stream the region
+/// maps to (it must match the `mapping_id` the client sees for the corresponding
+/// screencast stream), letting a client correlate a region with the video it shows.
+#[derive(Clone, Debug, PartialEq)]
+pub struct EiRegion {
+    /// Logical offset and size of the region.
+    pub rect: Rectangle<i32, Logical>,
+    /// Physical-to-logical scale of the region (e.g. `2.0` at 200%).
+    pub scale: f32,
+    /// Identifier tying this region to a screencast stream (e.g. the output name).
+    pub mapping_id: Option<String>,
+}
 
 /// A seat advertised on an EI sender context
 #[derive(Clone, Debug)]
@@ -32,6 +55,7 @@ impl EiInputSeat {
             keyboard: None,
             pointer: None,
             pointer_absolute: None,
+            pointer_absolute_regions: Vec::new(),
             touch: None,
             device_keyboard: None,
             device_pointer: None,
@@ -102,12 +126,19 @@ impl EiInputSeat {
     /// The EI device will have button and scroll capabilities in addition
     /// to the pointer absolute capability.
     ///
+    /// `regions` describes the logical coordinate space(s) the client may point
+    /// within (see [`EiRegion`]). Pass an empty slice to advertise no region, in
+    /// which case the client must guess the coordinate space. Advertising a
+    /// region lets clients that capture at a different resolution (e.g. physical
+    /// pixels on a HiDPI output) convert their coordinates correctly.
+    ///
     /// Calling on a seat that already has an absolute pointer device will remove
     /// that device and add a new one.
-    pub fn add_pointer_absolute(&self, name: &str) {
+    pub fn add_pointer_absolute(&self, name: &str, regions: &[EiRegion]) {
         let mut inner = self.0.lock().unwrap();
         inner.device_pointer_absolute = None;
         inner.pointer_absolute = Some(name.to_string());
+        inner.pointer_absolute_regions = regions.to_vec();
         inner.refresh_devices();
     }
 
@@ -116,6 +147,7 @@ impl EiInputSeat {
         let mut inner = self.0.lock().unwrap();
         inner.device_pointer_absolute = None;
         inner.pointer_absolute = None;
+        inner.pointer_absolute_regions.clear();
     }
 
     /// Add a touch device to the EI seat
@@ -159,6 +191,9 @@ struct EiInputSeatInner {
     keyboard: Option<(String, KeymapFile)>,
     pointer: Option<String>,
     pointer_absolute: Option<String>,
+    // Regions advertised on the absolute pointer device, describing the logical
+    // coordinate space(s) the client may address (see `EiRegion`).
+    pointer_absolute_regions: Vec<EiRegion>,
     touch: Option<String>,
     // Devices created in response to client bind
     device_keyboard: Option<DeviceDropWrapper>,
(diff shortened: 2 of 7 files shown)


src/input/keyboard/mod.rs
```diff
@@ -1201,6 +1345,68 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
...
+                focus.modifiers(&seat, data, mods, SERIAL_COUNTER.next_serial());
+            }
+            if let Some(surface) = focus.wl_surface() {
+                let wl_state = match key_state {
+                    KeyState::Pressed => wl_keyboard::KeyState::Pressed,
+                    KeyState::Released => wl_keyboard::KeyState::Released,
+                };
+                crate::wayland::seat::keyboard::for_each_focused_kbds(&seat, &surface, |kbd| {
+                    kbd.key(serial.0, time, keycode.raw() - 8, wl_state);
+                });
```
ids1024: I guess this is an issue with the previous virtual keyboard code, but this is entirely bypassing the `KeyboardGrab` and `KeyboardTarget` APIs to send Wayland events directly here?

I assume that will have various issues (we don't send keyboard events to any server side elements that can take keyboard focus and input but don't have a wl_surface, bypasses `PopupKeyboardGrab` and `XWaylandKeyboardGrabHandler`, etc.)

src/input/keyboard/mod.rs
```diff
@@ -480,6 +480,150 @@ impl<'a> KeysymHandle<'a> {
...
+    /// The serializable keymap for this source, used to send the layout to focused clients.
+    #[cfg(feature = "wayland_frontend")]
+    pub(crate) fn keymap_file(&self) -> &KeymapFile {
+        &self.keymap
+    }
+
+    /// Resolve a keysym to a keycode that produces it in the currently active layout.
+    ///
+    /// Returns `None` if no key in the active layout produces the keysym. The required
+    /// shift level / modifiers are not synthesized
```
ids1024: Is it actually useful to get the `Keycode` without also returning what shift level it has to be used with?

hojjatabdollahi: well that's a mistake, probably a combination of me trying to avoid turning a string into a set of keyboard shortcuts, and my tests being keycode+keysyms from Rustdesk.

I'm experimenting with doing it [similar to KWin](https://github.com/KDE/kwin/blob/master/src/backends/fakeinput/fakeinputbackend.cpp#L339C5-L342C1). Which means a hybrid of `keysym_to_keycode` plus in case of missing keysym in current layout, change layout using `keymap_for_keysym`. But I don't know if it's worth it, and how our stack, and clients will react to repeated changing of keymap even though it happens in the isolated input...

src/backend/libei/seat.rs
```diff
@@ -9,6 +9,30 @@ use crate::{
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct EiRegion {
+    /// Logical x offset of the region.
+    pub x: u32,
+    /// Logical y offset of the region.
+    pub y: u32,
+    /// Logical width of the region.
+    pub width: u32,
+    /// Logical height of the region.
+    pub height: u32,
```
ids1024: Generally in smithay we'd combine these into a `Rectangle`. Though I guess we don't use `u32` coordinates for any points in smithay, and consequently don't have methods for converting to it, etc...

src/input/keyboard/mod.rs
```diff
@@ -480,6 +497,321 @@ impl<'a> KeysymHandle<'a> {
...
+                keymap,
+                state,
+            })),
+            mods_state: ModifiersState::default(),
+            pressed_keys: HashSet::new(),
+            forwarded_pressed_keys: HashSet::new(),
+            keymap: keymap_file,
+            active_remap: None,
+        }
+    }
```
ids1024: If `::new` doesn't require `wayland_frontend`, this probably shouldn't either?

` #[cfg(feature = "wayland_frontend")]` can be moved to just the definitions of `keymap_file` and `active_remap`.

src/input/keyboard/mod.rs
```diff
@@ -480,6 +497,321 @@ impl<'a> KeysymHandle<'a> {
...
+            })),
+            mods_state: ModifiersState::default(),
+            pressed_keys: HashSet::new(),
+            forwarded_pressed_keys: HashSet::new(),
+            #[cfg(feature = "wayland_frontend")]
+            keymap: keymap_file,
+            #[cfg(feature = "wayland_frontend")]
+            active_remap: None,
+        })
+    }
```
ids1024: This can be replaced with a call to `Self::new_from_keymap`.

src/input/keyboard/mod.rs
```diff
@@ -480,6 +497,321 @@ impl<'a> KeysymHandle<'a> {
...
+                }
+            }
+        }
+        None
+    }
+
+    /// The currently serialized `(depressed, latched, locked, effective_layout)` modifier
+    /// state, for saving/restoring around a synthetic modifier change.
+    #[cfg(feature = "wayland_frontend")]
+    pub fn serialized_mods(&self) -> (u32, u32, u32, u32) {
```
ids1024: Presumably this should return `SerializedMods` rather than a tuple.

src/input/keyboard/mod.rs
```diff
@@ -480,6 +497,321 @@ impl<'a> KeysymHandle<'a> {
...
+                    return Some((keycode, mask));
+                }
+            }
+        }
+        None
+    }
+
+    /// The currently serialized `(depressed, latched, locked, effective_layout)` modifier
+    /// state, for saving/restoring around a synthetic modifier change.
+    #[cfg(feature = "wayland_frontend")]
```
ids1024: This `#[cfg]` seems to be unnecessary.

src/input/keyboard/mod.rs
```diff
@@ -480,6 +497,321 @@ impl<'a> KeysymHandle<'a> {
...
+pub struct IsolatedKeyboardState {
+    xkb: Arc<Mutex<Xkb>>,
+    mods_state: ModifiersState,
+    pressed_keys: HashSet<Keycode>,
+    forwarded_pressed_keys: HashSet<Keycode>,
+    #[cfg(feature = "wayland_frontend")]
+    keymap: KeymapFile,
+    // holds the original keymap text so it can be restored afterwards
+    #[cfg(feature = "wayland_frontend")]
+    active_remap: Option<String>,
```
ids1024: Hm, I guess this is stored as a `String` rather than an `xkb::Keymap` since the latter isn't `Send`/`Sync`? Not sure if that's ideal, though it's a bit awkward to deal with...

src/input/keyboard/mod.rs
```diff
@@ -480,6 +497,321 @@ impl<'a> KeysymHandle<'a> {
...
+    keymap: xkb::Keymap,
+) {
+    let mut state = xkb::State::new(&keymap);
+    for keycode in pressed_keys {
+        state.update_key(*keycode, xkb::KeyDirection::Down);
+    }
+    mods_state.update_with(&state);
+    *keymap_file = KeymapFile::new(&keymap);
+    xkb.keymap = keymap;
+    xkb.state = state;
```
ids1024: So we throw away the old version of `xkb::State`, and don't restore that part? I wonder if that could have any negative consequences.

One alternative could be to make `active_remap` the whole `Xkb`.

src/backend/libei/seat.rs
hojjatabdollahi: I switched to Rectangle, and kept the i32. That could become useful in a multi-monitor scenario (currently we don't restrict the cursor to the shared screen, which allows the guest to move the cursor to another screen, but then fail to return it back since we don't support negative values).

src/input/keyboard/mod.rs
hojjatabdollahi: Done

src/input/keyboard/mod.rs
hojjatabdollahi: Thanks

src/input/keyboard/mod.rs
hojjatabdollahi: Switched to xkb, should be fine.

src/input/keyboard/mod.rs
hojjatabdollahi: done.

src/input/keyboard/mod.rs
hojjatabdollahi: Updated the API to return that.

ids1024:
This may not be changed by this PR, but testing with https://github.com/pop-os/cosmic-osk, virtual-keyboard input doesn't really seem "isolated"? Shift or caps lock on a physical keyboard seems to capitalize input when pressing a key on the OSK. And holding shift on the OSK likewise affects input on the physical keyboard.

That's not supposed to happen here, right?

Also, since https://wayland.app/protocols/virtual-keyboard-unstable-v1#zwp_virtual_keyboard_v1:request:modifiers is sent from the client to the server, is the virtual-keyboard protocol not supposed to update the xkb state automatically on handling `key`, and only on when the client sends `modifiers`?

May need to compare other implementations.

src/input/keyboard/mod.rs
```diff
@@ -1201,6 +1538,79 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
...
+                state.forwarded_pressed_keys.insert(keycode);
+            }
+            KeyState::Released => {
+                state.forwarded_pressed_keys.remove(&keycode);
+            }
+        }
+
+        let mut guard = self.arc.internal.lock().unwrap();
+        let seat = self.get_seat(data);
+        let modifiers = mods_changed.then_some(mods);
```
ids1024: I guess we also need to pass modifiers to the keyboard grab if the last key press was on a different isolated keyboard state, or non-isolated? (Likewise `input_forward` for normal input.)

We probably need to track somewhere what the last send modifiers were.

hojjatabdollahi: I added something to track it. It gets reset on `set_focus`.

hojjatabdollahi:
> This may not be changed by this PR, but testing with https://github.com/pop-os/cosmic-osk, virtual-keyboard input doesn't really seem "isolated"? Shift or caps lock on a physical keyboard seems to capitalize input when pressing a key on the OSK. And holding shift on the OSK likewise affects input on the physical keyboard.
> 
> That's not supposed to happen here, right?

I think the idea was that shortcuts that the compositor should intercept (such as super+t) work even if you're holding shift on the physical keyboard. Modifiers otherwise go to the client.

But with the latest commit I implemented `last_sent_modifier`, and now the modifiers on physical keyboard doesn't affect the osk, but I don't know if that's desirable.


> Also, since https://wayland.app/protocols/virtual-keyboard-unstable-v1#zwp_virtual_keyboard_v1:request:modifiers is sent from the client to the server, is the virtual-keyboard protocol not supposed to update the xkb state automatically on handling `key`, and only on when the client sends `modifiers`?
> 
> May need to compare other implementations.

Gnome and KDE don't use zwp_virtual_keyboard. I think the only implementation would be wl_roots and it just uses the shared keyboard state for the seat (so do Gnome and KDE with their virtual keyboards).

So, maybe for the virtual keyboard we want to do the same as everybody else and share it with the physical keyboard.


Also had to add https://github.com/pop-os/cosmic-comp/pull/2442/commits/990525aa33b28bf3defb8b39593a03c7e60eca07 to make sure esc works in cosmic-osk.

What do you think?

ids1024:
Yeah, I guess if `zwp_virtual_keyboard_v1` is basically a wlroots protocol, we want to be consistent with their implementation since that is the behavior clients expect. Even if the protocol looks like it should behave differently (the spec isn't as clear as official Wayland protocols usually are).

Though the client can also use `zwp_virtual_keyboard_v1::keymap` to set a different keymap from what the compositor is otherwise using, so shift level and group don't necessarily even mean the same thing outside the virtual keyboard? Mixing up the group index with isolated keyboard using different keymaps presumably wouldn't make much sense.

For libei, Gnome is probably the implementation to reference, but hopefully the intended behavior of libei is better specified.

ids1024:
Oh, it's separate from this, but I guess sending [`ei_keyboard.modifiers`](https://libinput.pages.freedesktop.org/libei/interfaces/ei_keyboard/index.html#ei_keyboardmodifiers) is also something that we need to add. Looks like we should be send that to all libei sender clients with an `ei_keyboard` any time the compositor's modifier state changes.

hojjatabdollahi:
Alright, I'm rewriting everything to be more in line with the protocol and other compositors.

- I'm keeping a map of held modifiers keyed by the source device, that enables us to have a consistent final compiled modifiers list.
- When a modifier changes, we send a `wl_keyboard.modifiers` and `ei_keyboard.modifiers` to all the clients, everybody needs to update their internal state based on compositors state. So pressing shift on a physical keyboard should show up on cosmic-osk too.
- I'm removing all the isolated keyboard stuff, if you're typing on a remote machine and a person there hold down a key, then depending on your remote desktop software (if it's sending keysym or keycode) they're gonna be affected by the modifier (that's the intended behavior).
- The clients (vk, ei) must release everything when they go away. So probably a cosmic-osk PR is needed here.

Concerns:

The reason we had isolated keyboard stuff, was because if a remote user holds down "super" you can't do anything anymore on your local machine! You can't even click on "quit" (on Rustdesk for example) to kick the remote user off because that just drags the window. You can't type, that would just do all the super+* shortcuts. But apparently that's by design.
Also buggy clients may mess up your modifiers list and you have no way of resetting it. For libei clients, when they get disconnected the server can clear all their modifiers (when removing the device). For virtual keyboard, it's up to the osk app, for example if I press super (latch) on the osk, if for some reason the keyboard interface is going to hide or go away the app is required to release that, otherwise we're gonna have the super key held down while trying to bring osk back up and clicking on it to release.

ids1024:
I've made some changes to `cosmic-osk` to test a keymap that's different from what the compositor has. I need to do a bit of testing, but if it shares modifiers, including `group`, with different keymaps, that seems like a bug in wlroots, or the design of the virtual-keyboard protocol...

I guess it's not really explicit with libei/xdg-desktop-portal if the modifier state should be independent. Libei itself provides a way to expose devices, but isn't explict about what exactly they represent, and https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.RemoteDesktop.html#org-freedesktop-impl-portal-remotedesktop-connecttoeis doesn't really provide additional details.

But we probably want `ei_keyboard` to not be isolated from the main keyboard (probably should verify that's what Gnome does).

ids1024:
If we decide we don't need `IsolatedKeyboardState` for anything related to libei, we can just leave any changes to virtual-keyboard out of this PR. It may also need various improvements, but those can be a separate PR. Then we can hopefully merge the libei part sooner.

hojjatabdollahi:
I removed virtual keyboard and isolated keyboard stuff. I can squash the commits if you want.

ids1024:
I can just do a squash merge if you don't want this divided into multiple commits in some cleaner way.

It's easier to review this without worrying about if virtual-keyboard is regressed. Still worth looking into if that behaves correctly, though the more I look at virtual-keyboard, it seems like not a great protocol; maybe cosmic-osk should somehow use libei instead.

hojjatabdollahi:
> maybe cosmic-osk should somehow use libei instead.

I think that's why gnome doesn't support it, they suggest using libei.

ids1024:
For an on-screen keyboard, Gnome presumably also implements theirs inside the compositor process and isn't really concerned about supporting third party versions of things like that outside the compositor (it wouldn't really work on Gnome without layer-shell support anyway).

But yeah, Sway also has a preference to be able to do things through Wayland protocols and not rely on DBus, which Gnome isn't concerned with.

hojjatabdollahi:
> I can just do a squash merge if you don't want this divided into multiple commits in some cleaner way.

I think a single commit is fine.
Kinda getting tired of rebasing two PRs, especially because you can't simply rebase on the latest simthay commit, since the relevant cosmic-comp PRs are not merged yet! So, I'd appreciate if we do a push to get at least this PR merged.

> For an on-screen keyboard, Gnome presumably also implements theirs inside the compositor process and isn't really concerned about supporting third party versions of things like that outside the compositor (it wouldn't really work on Gnome without layer-shell support anyway).

yeah, and KDE has their own thing. But maintaining just libei (instead of libei AND virtual keyboard) is easier. However they handle keymaps differently, kv allows you to upload your own, but ei is supposed to work with the compositors one (however we've implemented ei_text which kinda allows you to enter whatever you want!)

ids1024:
Yeah, I'd really like to get these things merged. It would also make testing cosmic-comp changes easier if the released version of the portal has `RemoteDesktop`. We can improve things later as long as it doesn't have major issues.

That said, testing https://github.com/ids1024/reis/blob/master/examples/type-text.rs modified to not release the keys, it looks like the `release_source()` logic isn't working properly? When I close the terminal, the keys still are stuck. I had to ssh in from another system to terminate cosmic-comp since even tty switching key bindings were affected (because shift was held).

Edit: Looks like `release_source` doesn't get called in this case.

ids1024:
Oh, that one I guess is my fault. It should have `cb(EiInputEvent::Disconnected, connection)` for `Err` on the connection.

ids1024:
Rebased and added a commit for that, squashing don't the commits that aren't already cleanly separate.

ids1024:
I guess the three `EiRegion` commits may as well be squashed too.

ids1024 (approved):

ids1024:
I think the way this tracks `KeyboardSource`s looks good. And everything else here. I'd like to merge this, though I'd like to see what @Drakulix thinks about the keyboard changes before clicking the merge button.

hojjatabdollahi:
Thank you. Looks good to me!

anvil/src/libei.rs
```diff
@@ -39,6 +39,7 @@ pub fn listen_eis(handle: &calloop::LoopHandle<'static, AnvilState<UdevData>>) {
                         let dh = data.display_handle.clone();
                         data.process_input_event(&dh, event);
                     }
+                    EiInputEvent::TextKeysym { .. } | EiInputEvent::TextUtf8 { .. } => {}
```
Drakulix: Can we maybe not completely ignore this in anvil? After all anvil should be some reference at least.

I would be fine with a simply comment saying:
> Anvil doesn't add a text device to the libei seat.

src/input/keyboard/mod.rs
```diff
@@ -203,10 +203,32 @@ impl fmt::Debug for Xkb {
...
+}
+
+impl KeyboardSource {
+    /// The default source used by [`KeyboardHandle::input`] (the physical keyboard).
+    pub const MAIN: KeyboardSource = KeyboardSource::Physical;
+
+    /// Mint a fresh, process-unique auxiliary source.
+    pub fn new_auxiliary() -> Self {
+        static NEXT_AUX_SOURCE: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(1);
+        KeyboardSource::Auxiliary(NEXT_AUX_SOURCE.fetch_add(1, std::sync::atomic::Ordering::Relaxed))
```
Drakulix, line 223: I know this is basically impossible to hit, but is `fetch_add` wrapping?

src/input/keyboard/mod.rs
```diff
@@ -979,6 +1084,24 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
...
     }
 
+    /// Release every key currently held by `source` (e.g. when a virtual keyboard is destroyed
+    /// or a libei connection drops), forwarding the resulting releases to the focused client.
+    pub fn release_source(&self, data: &mut D, source: KeyboardSource) {
+        let (transitioned, mods) = {
+            let mut guard = self.arc.internal.lock().unwrap();
+            let transitioned = guard.release_source_keys(source);
+            (transitioned, guard.mods_state)
+        };
```
Drakulix, line 1097: So can these releases not be intercepted? This would mess up e.g. cosmic's single-modifier bindings, no?

src/input/keyboard/mod.rs
```diff
@@ -1201,6 +1350,99 @@ impl<D: SeatHandler + 'static> KeyboardHandle<D> {
...
+{
+    /// Inject a batch of keysyms as text into the currently focused client (KWin-style).
+    ///
+    /// Builds a throwaway keymap that binds each keysym to its own spare keycode at base level
+    /// (so no modifiers are ever needed), hands that keymap to clients, taps each keycode on the
+    /// focused client, then restores the seat keymap. The compositor's own seat xkb state is
+    /// **never** touched, so shortcut handling and physical-keyboard state are unaffected.
+    /// `modifiers(0)` accompanies the injection keymap so any modifier the seat currently holds
+    /// (e.g. a physically-held Shift) doesn't alter the injected characters.
+    pub fn inject_text_keysyms(&self, data: &mut D, keysyms: &[Keysym]) {
```
Drakulix, line 1374: So this is meant as a helper function, right? A compositor would attempt to use `text_input` first and otherwise use this?

Drakulix (approved):
The changes look mostly good to me.

I still find it somewhat mind-boggling, that compositors aren't isolating the virtual keyboards from real ones, but it is better to match behavior imo.

varbhat:
> that compositors aren't isolating the virtual keyboards from real ones

Can we provide a configurable option to do that atleast, if we are not going to isolate the virtual keyboards from real ones?

ids1024:
> Can we provide a configurable option to do that atleast, if we are not going to isolate the virtual keyboards from real ones?

The libei API here at least is compatible with compositors wanting to isolate the keyboard or not. The compositor can add a keyboard with whatever keymap it wants, and handle the input events how it wants. So if you can do isolation just using the public API of smithay, that's possible already.

I do think `virtual-keyboard-unstable-v1` logically needs to be isolated since the client can send its own `keymap` and `modifiers`. If we do need to not isolate it for compatibility with other implementations, the protocol is just kind of broken.

(Anyway, if we have any changes to that, that would be another PR.)

src/input/keyboard/mod.rs
hojjatabdollahi: It is wrapping.

src/input/keyboard/mod.rs
hojjatabdollahi: This only happens on teardown, normal key releases don't go through this. I update the comment.

src/input/keyboard/mod.rs
hojjatabdollahi: Yes. I'll add a comment mentioning the order.

src/input/keyboard/mod.rs
hojjatabdollahi: The only thing that I thought would break was super+drag. I just tested it. It does stays dragging a window when the client gets disconnected because we don't release mouse buttons, but since it's not isolated, you can just click on your mouse and it resets it. We can later also release mouse buttons

hojjatabdollahi: And releasing mouse buttons is just a cosmic-comp issue, doesn't need a smithay change. Unless we want to add per-source pointer-button tracking in smithay (mirroring the keyboard), not sure if it's worth it!

ids1024:
Based on @Drakulix's review and the added comments that seem good, I think we can merge.

PR #2076, opened 2026-06-20
drm_scanner: release CRTC reservations for vanished connectors

### LINKED ISSUE #2075: `SimpleCrtcMapper` leaks CRTC reservations for connectors that vanish from the resource list
### Summary

`SimpleCrtcMapper` never releases the CRTC reservation of a connector whose handle disappears from the DRM resource list entirely (as opposed to being reported as `Disconnected`). These stale reservations accumulate until every CRTC is "taken" by a ghost handle, after which a genuinely connected connector can no longer be assigned a CRTC and `crtc_for_connector` returns `None` for it — so the compositor silently never brings up that output.

### Impact

I have been observing this in Niri. I have a home set up where I plug my laptop into a Thunderbolt docking station with 2 external monitors. Over time, as I connect/disconnect from the dock and/or suspend and wake, I often had one or both of my external monitors not come up. I could try unplugging/plugging the dock: it might switch which one of the external monitors was dead, but wouldn't bring them both back. The only way to fix was to end my Niri session and log in again.

### Reproduction

1. Drive ≥1 display through a DP-MST hub (e.g. a USB-C/Thunderbolt dock with multiple external monitors).
2. Suspend/resume (or unplug/replug the dock) repeatedly while docked, so the MST sink connectors are destroyed and recreated under new handles.
3. After enough cycles, a connected monitor stops being assigned a CRTC: `crtc_for_connector` returns `None` for it even though `/sys/class/drm/*/status` reports it connected.

### Possible fix

At the start of `map()`, retain only reservations whose connector is in the set of currently connected handles, which releases both present-but-disconnected and vanished connectors. I have submitted #2076 which contains this fix.

### Affected crate

`smithay-drm-extras` — `drm_scanner::crtc_mapper::SimpleCrtcMapper`

### Details

On each call, `SimpleCrtcMapper::map()` releases reservations like this:

```rust
for connector in connectors
    .clone()
    .filter(|conn| conn.state() != connector::State::Connected)
{
    self.crtcs.remove(&connector.handle());
}
```

This only releases connectors that are **still present** in `connectors` and reported as non-connected. A connector whose handle is gone from the resource list is never passed to `map()` again, so its entry in `self.crtcs` is never removed.

This is the normal lifecycle for DP-MST sink connectors: the connectors behind an MST hub are created and destroyed dynamically. Unplugging a USB-C/Thunderbolt dock, or suspending and resuming while docked (the connectors are torn down and recreated under fresh handles while the compositor isn't scanning), makes the old handles vanish rather than transition to `Disconnected`. Each such event leaks one reservation per vanished connector.

Because `is_taken()` still counts a leaked reservation, the leaks accumulate across cycles. Once all CRTCs are reserved by ghost handles, `pick_next_available_for_connector` can't find a free CRTC for a newly connected connector, and it never gets mapped. The only recovery is recreating the `DrmScanner`.

spacedentist:
Fixes #2075

## Description

`SimpleCrtcMapper` keeps a `connector -> crtc` reservation map and, on each `map()` call, releases the reservation of any connector it is told about that is no longer connected:

```rust
for connector in connectors
    .clone()
    .filter(|conn| conn.state() != connector::State::Connected)
{
    self.crtcs.remove(&connector.handle());
}
```

This only releases connectors that are still present in the resource list and reported as disconnected. A connector whose handle disappears from the resource list entirely is never passed to `map()` again, so its reservation is never released and leaks for the lifetime of the mapper.

This happens routinely with DP-MST: the sink connectors behind an MST hub are dynamically created and destroyed, so unplugging a USB-C/Thunderbolt dock — or suspending and resuming while docked, where the connectors are torn down and recreated under fresh handles while the compositor isn't scanning — makes the old handles vanish rather than transition to `Disconnected`. Each such event leaks one reservation per vanished connector. Because a leaked reservation still counts as "taken", the reservations accumulate across cycles until every CRTC is reserved by a ghost handle. At that point `pick_next_available_for_connector` can no longer find a free CRTC for a genuinely connected connector, so `crtc_for_connector` returns `None` for it and the compositor silently never brings up that output. The only recovery is recreating the `DrmScanner`.

This was observed in niri as external monitors on a dock going missing after a number of suspend/resume cycles, getting progressively more likely over time, and only recoverable by restarting the compositor.

Fix it by releasing the reservation of every connector that is not currently connected, including connectors that are no longer present at all. At the start of `map()`, collect the set of currently connected connector handles and retain only reservations whose handle is in that set. The previously handled case (a connector present but reported as disconnected) is still covered, since such a connector is excluded from the connected set.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -202,6 +202,15 @@ default implementations, which result in skipping the new functionality. As such
 - `lower_element()`: lowers an element to the bottom of the stack, respecting its z-index group.
 - `relocate_element()`: moves an element to a new location in the space without changing the stacking order.
 
+### Bugfixes
+
+`SimpleCrtcMapper` (in `smithay-drm-extras`) now releases the CRTC reservation of any connector that
+is no longer connected, including connectors that have disappeared from the resource list entirely
+rather than being reported as disconnected. Previously such connectors (for example DP-MST sink
+connectors when a dock is unplugged, or across suspend/resume) leaked their CRTC reservation
+indefinitely, which could accumulate until a newly connected output could no longer be assigned a
+CRTC.
+
 ## 0.7.0
 
 ### Breaking changes
--- a/smithay-drm-extras/src/drm_scanner/crtc_mapper.rs
+++ b/smithay-drm-extras/src/drm_scanner/crtc_mapper.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use drm::control::{Device as ControlDevice, connector, crtc};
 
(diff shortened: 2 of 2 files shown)


cmeissl (approved):
LGTM

@PolyMeilex if you have some time would appreciate a second opinion :) thanks!

Drakulix (approved):
LGTM as well, I'll go ahead and merge this.

PR #2078, opened 2026-06-22
tablet: add infrastructure for proper tablet support

### LINKED ISSUE #1753: Tablet grabs
There are pointer and touch grabs, but no tablet grabs. They are needed for much the same reasons, e.g. putting the pen down inside a window, then moving it outside while holding it down, then back inside—similar to the pointer ClickGrab. Or for compositor-specific grabs like interactive window movement with a tablet pen.

The [protocol](https://wayland.app/protocols/tablet-v2) does mention grabs:

> If the tool moves out of the surface but stays in proximity (i.e. between windows), compositor-specific grab policies apply. This usually means that the proximity-out is delayed until all buttons are released.

This, too, is in line with how pointer and touch grabs work.

Ph4ntomas:
## Description
This PR brings generic tablet support to the `input` module with similar semantics we already have for pointer, touch and keyboard.

The TabletSeat no longer depends on the `wayland_frontend` feature, and was moved to `input::tablet`. The main interfaces, `TabletToolHandle` and `Tablet` can be created via `TabletSeat::add_tool` and `TabletSeat::add_tablet` or their `add_wp_*` counterpart which additionally setup wayland integration.

`TabletToolHandle` is the main type user will interact with. It's API resemble the other `*Handle` and supports grab and abstract targets for focus handling.

`TabletToolTarget` trait can be used to implements arbitrary targets, and has been implemented for `WlSurface` and `X11Surface`.

`TabletSeatHandler` now has a `ToolFocus`, similar to `SeatHandler::PointerFocus` and `SeatHandler::TouchFocus`.

I'm leaving this in draft for now as some doc test must still be updated, comments need proofreading, and I want to have some more real-world tests done.

Fixes #1753 

## Checklist

### DIFF
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -1098,7 +1098,6 @@ impl AnvilState<UdevData> {
 
         if let Some(pointer_location) = self.touch_location_transformed(&evt) {
             let tool = evt.tool();
-            tablet_seat.add_wp_tool(self, dh, &tool);
 
             let pointer = self.pointer.clone();
             let under = self.surface_under(pointer_location);
--- a/anvil/src/input_handler.rs
+++ b/anvil/src/input_handler.rs
@@ -5,7 +5,7 @@ use crate::{AnvilState, focus::PointerFocusTarget, shell::FullscreenSurface};
 #[cfg(feature = "udev")]
 use crate::udev::UdevData;
 #[cfg(feature = "udev")]
-use smithay::backend::renderer::DebugFlags;
+use smithay::{backend::renderer::DebugFlags, input::tablet};
 
 use smithay::{
     backend::input::{
@@ -16,6 +16,7 @@ use smithay::{
     input::{
         keyboard::{FilterResult, Keysym, ModifiersState, keysyms as xkb},
         pointer::{AxisFrame, ButtonEvent, MotionEvent},
+        tablet::{TabletDescriptor, TabletSeatTrait},
         touch::{DownEvent, UpEvent},
     },
     output::Scale,
@@ -28,7 +29,6 @@ use smithay::{
         input_method::InputMethodSeat,
         keyboard_shortcuts_inhibit::KeyboardShortcutsInhibitorSeat,
         shell::wlr_layer::{KeyboardInteractivity, Layer as WlrLayer},
-        tablet_manager::{TabletDescriptor, TabletSeatTrait},
     },
 };
 
@@ -534,7 +534,7 @@ impl<BackendData: Backend> AnvilState<BackendData> {
         if device.has_capability(DeviceCapability::TabletTool) {
             self.seat
                 .tablet_seat()
-                .add_tablet::<Self>(dh, &TabletDescriptor::from(&device));
+                .add_wp_tablet(dh, &TabletDescriptor::from(&device));
         }
         if device.has_capability(DeviceCapability::Touch) && self.seat.get_touch().is_none() {
             self.seat.add_touch();
@@ -1045,8 +1045,8 @@ impl AnvilState<UdevData> {
         if let Some(pointer_location) = self.touch_location_transformed(&evt) {
             let pointer = self.pointer.clone();
             let under = self.surface_under(pointer_location);
-            let tablet = tablet_seat.get_tablet(&TabletDescriptor::from(&evt.device()));
             let tool = tablet_seat.get_tool(&evt.tool());
+            let time = self.clock.now().as_millis();
 
             pointer.motion(
                 self,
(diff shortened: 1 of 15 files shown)


Ph4ntomas:
Mmmh the error seems unrelated (curl failed while fetching a package), so I think this is good for review

Drakulix:
Thanks for taking the time, this should enable more advanced use-cases like pointer emulation with a tablet much easier.

I see the draft status was removed, do you consider this ready for review now? Which clients did you test to verify that nothing was regressed?

Ph4ntomas:
Thanks for having a look!

> I see the draft status was removed, do you consider this ready for review now? 

When I originally removed the draft yes, but I'm kinda second guessing some choices (OTOH that what reviews are for), mainly removing grabs on (physical) proximity out. 

It make sense IMO to have this default behavior, but I discovered that at least with my testing HW (a pinenote), pressing a button on a tool sends a proximity out immediately followed by a proximity in, so I wonder if the ungrab on proximity out should be left the responsibility of the compositor, since removing the grab in this context might be surprising.
OTOH the grab being kept would be weird too in a multi-tablet context, so I'd love some input on this.

> Which clients did you test to verify that nothing was regressed?

I did some tests using `xournalpp` and `wev` on a PineNote, without noticeable issues, but in a currently private compositor I uses to test my changes (anvil doesn't run on my HW and I haven't found the time to troubleshoot it). I could tweak cosmic/niri/pinnacle to add support for these changes for a more 'formal' testing environment.

Also did a bunch of tests using non tablet-enabled apps  (In those tests, I was using the grab to emulate regular pointer)

src/input/tablet/tool/mod.rs
```diff
@@ -0,0 +1,1235 @@
...
+        self.unset_grab(data, seat, descriptor, event.serial, event.time, false);
+
+        // Now, reset everything so we don't get stale data later.
+        self.tip_down = false;
+        self.pressed_buttons.clear();
+        self.current_buttons.clear();
+        self.focus = None;
+        self.pending_focus = None;
+        self.tablet = None;
+        self.location = None;
```
Ph4ntomas: As mentioned in a previous comment, I've found some weird behavior on the HW I'm testing with, where some buttons release event are immediately followed by a proximity_out and a proximity_in. These are coming from libinput, so there's not much that can be done, but it kinda break my initial assumption that a proximity_out should release grabs since the tool is gone.

The cleanup itself is fine, but I'm leaning toward leaving the decision to unset the grab up to the compositor here.

Drakulix (approved):
A lot of code, but this looks good to me.

@ids1024 @YaLTeR @cmeissl Anybody any other input? Otherwise I am going to merge this by the end of the week.

anvil/src/input_handler.rs
```diff
@@ -1118,20 +1118,49 @@ impl AnvilState<UdevData> {
...
+                        tool.proximity_out(
+                            self,
+                            &tablet::tool::ProximityOutEvent {
+                                serial: SCOUNTER.next_serial(),
+                                time: evt.time_msec(),
+                            },
+                        );
+
+                        // The tool left, we don't want to send a frame event here as it's sent as
+                        // part of the proximity out logic.
```
Ph4ntomas: Whoops, sorry about that, this no longer holds true. I'll fix it in an hour of two.

Ph4ntomas: Fixed it in a new commit to ease on the re-review, but I'm happy to rebase/squash once it's not needed.

src/input/tablet/tool/grab.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    fn motion(
+        &mut self,
+        data: &mut D,
+        handle: &mut TabletToolInnerHandle<'_, D>,
+        _focus: Option<(<D as TabletSeatHandler>::ToolFocus, Point<f64, Logical>)>,
+        event: &MotionEvent,
+    ) {
+        handle.motion(data, self.start_data.focus.clone(), event);
```
YaLTeR: Do we want to update the start location here like in the recently merged https://github.com/Smithay/smithay/pull/2091/changes#diff-194103ede3152640cbf728b93c01c390a02862d971670bbfea91b3a6072a48f2R367?

anvil/src/input_handler.rs
```diff
@@ -1100,12 +1098,14 @@ impl AnvilState<UdevData> {
 
         if let Some(pointer_location) = self.touch_location_transformed(&evt) {
             let tool = evt.tool();
-            tablet_seat.add_tool::<Self>(self, dh, &tool);
+            tablet_seat.add_wp_tool(self, dh, &tool);
```
YaLTeR: I don't think this line is correct since it will readd the tool every time. I think `unwrap_or_else` right below already handles adding the tool?

src/input/tablet/tool/grab.rs
```diff
@@ -0,0 +1,308 @@
...
+        handle.proximity_in(data, focus, event);
+    }
+
+    fn proximity_out(
+        &mut self,
+        data: &mut D,
+        handle: &mut TabletToolInnerHandle<'_, D>,
+        event: &ProximityOutEvent,
+    ) {
+        handle.proximity_out(data, event);
```
YaLTeR, line 282: Not sure, should this unset the grab? The doc comment elsewhere says "until an up or physical proximity out event"; `up` unsets the grab but this doesn't

YaLTeR (commented):
Unfortunately I don't think I'll have time for a proper review or test for the next 1-2 weeks as I'm leaving to travel, but here are a few minor comments

anvil/src/input_handler.rs
Ph4ntomas: Oh yeah, you're right, I forgot to remove this line

src/input/tablet/tool/grab.rs
Ph4ntomas: I don't think there's a valid way to get a `proximity_out` without an `up` in this specific case (the grab was initiated following a `down` event), but I don't mind adding it in since it's indeed more correct.

src/input/tablet/tool/grab.rs
Ph4ntomas: I guess we should for consistency, ~~although I find it weird to update any of the start data in the default/automatic grab. If that behavior is needed, it should be up to the compositor to compensate by updating the various locations at the time the grab is overridden as the opposite can't be done.~~

EDIT: Right, I read the changes too fast and missed that the updated focus isn't the one from StartData but an additional fiels private to the grab. That solve the issues I was seeing, I'll align on ClickGrab.

Ph4ntomas:
Updated to take @YaLTeR comments into account.

For anyone interested in quickly testing, here's the compositor I'm using (it only has the bare minimum to test things out, so here be dragons.)
https://git.sr.ht/~phantomas/slate/tree/dev/tablet-support

As for usage example, my pointer emulation code is here:
https://git.sr.ht/~phantomas/slate/tree/dev/tablet-support/item/slate-comp/src/shell/grab/pointer_emu/tablet.rs
https://git.sr.ht/~phantomas/slate/tree/dev/tablet-support/item/slate-comp/src/input/tablet.rs#L140

Drakulix (approved):
Thanks again!

PR #2079, opened 2026-06-23
anvil: perform cursor x and y movements separately (fix #1890)

### LINKED ISSUE #1890: Confined cursors "stick" to region borders
Copied from https://github.com/YaLTeR/niri/issues/3128

In niri, when a client attempts to confine the cursor to a given region, and the mouse approaches the borders of the region, it never actually reaches the borders of the region and instead gets stuck right before it. This happens in floating and tiled windows, but not in fullscreen windows as long as there is not a another monitor sharing region edge. This issue occurs with Wine, SDL3, and libwayland applications (I think Wine and SDL3 both use libwayland). The program in the video is in [this](https://gist.github.com/emekoi/68fc5a70356baf09c33e618b82d3bd7d) gist. I don't have a good way of recording two monitors at once, so that isn't included in the video but the behavior on that shared edge is the same as in the floating case shown.

This behavior is reproducible in Anvil as well.

https://github.com/user-attachments/assets/79ac1fc5-050d-4bb4-b32f-0744045cdaa3

emekoi:
## Description
This fixes #1890 (c.f. https://github.com/YaLTeR/niri/issues/3128). Previously, the pointer would "stick" to the edges of confinement regions since X and Y movement were linked. If one or the other would move the cursor outside of the region, we would perform no movement at all. Instead we perform X and Y movements separately and send each component to 0 individually if they would move the pointer outside the region. 

NOTE: The arguably correct fix would be to take the displacement vector and if it moves the pointer outside of the region, project the portion outside the region against the closest edge and then apply that. I had a solution that did this by using the confinement rectangles to build a signed distance function and raymarching against that to find contact edges and projecting along them as necessary, but I had bunch of issues with reliability. This is a lot simpler and people probably wouldn't notice the difference anyways.

The pointer constraint test client was also updated to test a more complex confinement region (the whole window with the upper 1/16 and lower 1/4 missing).

No AI was used in this PR.

## Checklist


Drakulix (approved):
LGTM. thanks!

PR #2081, opened 2026-06-25
multigpu: Re-import shadow dmabufs before rendering

Drakulix:
## Description

I haven't seen any actual issues without this code, but I wrote this up trying to debug something else and I believe it to be technically correct/required to make changes visible.

Especially because our shadow buffer could be external to the target gpu (likely for nvidia).

## Checklist

### DIFF
--- a/src/backend/renderer/multigpu/mod.rs
+++ b/src/backend/renderer/multigpu/mod.rs
@@ -1800,6 +1800,22 @@ impl MultiTexture {
             })
     }
 
+    fn reimport<A: GraphicsApi + 'static>(
+        &self,
+        renderer: &mut <A::Device as ApiDevice>::Renderer,
+    ) -> Result<(), <<A::Device as ApiDevice>::Renderer as RendererSuper>::Error>
+    where
+        <A::Device as ApiDevice>::Renderer: ImportDma,
+    {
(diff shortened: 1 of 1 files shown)


src/backend/renderer/multigpu/mod.rs
```diff
@@ -1983,17 +1999,21 @@ where
     ) -> Result<(), Error<R, T>> {
         let render_id = self.frame.as_mut().unwrap().context_id();
         let sync = texture.needs_synchronization::<R>(&render_id);
+        if let Some(sync) = sync {
+            if let Err(err) = self.frame.as_mut().unwrap().wait(&sync) {
+                trace!(?err, "Failed to import sync point, blocking");
+                while let Err(_) = sync.wait() {}
```
cmeissl: Are we 100% sure this won't be able to busy loop forever? Like for example when the gpu is detached.

Drakulix: Well looking at `eglClientWaitSync` there is a difference between "error" and "timeout reached". The way we bubble this up however, we don't handle this difference correctly. So indeed, this should be a single wait.

cmeissl (approved):

PR #2083, opened 2026-06-29
renderer/multigpu: Allow disabling cross-device dmabuf exports

Drakulix:
This is useful for downstream to implement quirks in the compositor similar to https://github.com/pop-os/cosmic-workspaces-epoch/pull/218.

## Checklist


ids1024 (approved):
Seems fairly uncontroversial to have this, analogously to `can_do_cross_device_imports`, can the implementation fairly obviously should be fine.

Issue #2044, opened 2026-05-28
Add a helper for importing shm buffers via udmabuf


valpackett:
As it turns out, the import of shm buffers into the GPU can be massively optimized (skipping copies) by using udmabuf:

https://zamundaaa.github.io/wayland/2026/05/06/making-wl-shm-fast.html

Smithay should provide a helper function that attempts udmabuf import of a shm backed WlBuffer…

YaLTeR:
Will this also help with #1652?

valpackett:
Yeah, I think so!

ids1024:
I wonder if `VK_EXT_external_memory_host` could also be useful here, once we have a Vulkan renderer.

valpackett:
@ids1024 the linked article mentions it, saying

> Unfortunately, the implementation at least on AMD comes with some limitations. Because of potential security issues, pointers to anything associated with a file descriptor (which shm buffers always are) can’t be imported this way by amdgpu

PR #2084, opened 2026-06-29
allocator: Add udmabuf

Drakulix:
## Description

Implementation of our `Allocator`-trait using udmabuf.

Tested in https://github.com/pop-os/cosmic-comp/pull/2501.

Should be usable for solving https://github.com/Smithay/smithay/issues/2044 as well, but I didn't test that (yet).

## Checklist

### DIFF
--- a/src/backend/allocator/udmabuf.rs
+++ b/src/backend/allocator/udmabuf.rs
@@ -1,0 +1,169 @@
+//! Module containing an implementation of the [`Allocator`]-trait using linux `udmabuf` api.
+
+#![cfg(target_os = "linux")]
+
+use std::{
+    io,
+    os::fd::{AsFd, AsRawFd, BorrowedFd, FromRawFd, OwnedFd, RawFd},
+    sync::LazyLock,
+};
+
+use super::{
+    Allocator, Fourcc, Modifier,
+    dmabuf::{Dmabuf, DmabufFlags},
+    format::get_bpp,
+};
+use crate::utils::Size;
+
+use rustix::{
+    fs::{MemfdFlags, Mode, OFlags, SealFlags, fcntl_add_seals, ftruncate, memfd_create},
+    ioctl::{Ioctl, Opcode, ioctl, opcode},
+    param::page_size,
+};
+
+static PAGE_SIZE: LazyLock<usize> = LazyLock::new(page_size);
+
+/// [`Allocator`] creating [`Dmabuf`]s for system memory via the `udmabuf` uapi.
+///
+/// Note: [`Allocator::create_buffer`] will always return linear buffers for a
+/// [`UdmabufAllocator`] and fail if [`Modifier::Linear`] is not in the list
+/// of passed modifiers.
+#[derive(Debug)]
+pub struct UdmabufAllocator {
+    dev: OwnedFd,
+}
+
+/// Minimum alignment requirement for the stride for `radv` (which seems to be the highest among drivers).
+pub const STRIDE_ALIGN: usize = 256;
+
+impl UdmabufAllocator {
+    /// Tries to open `/dev/udmabuf` and create a [`UdmabufAllocator`].
+    pub fn new() -> io::Result<UdmabufAllocator> {
+        let fd = rustix::fs::open("/dev/udmabuf", OFlags::RDONLY | OFlags::CLOEXEC, Mode::empty())?;
+        Ok(UdmabufAllocator { dev: fd })
+    }
+
+    /// Creates a [`Dmabuf`] from an existing allocation represented by a memfd.
+    ///
+    /// For importing to succeed a few additional requirements need to be satisfied:
+    /// - the width and height cannot exceed 65535.
+    /// - the stride needs to be aligned to [`STRIDE_ALIGN`].
+    /// - size and offset need to be aligned to the platforms page size.
+    ///
+    /// Note: dmabufs created through this interface will always be interpreted as linear.
+    #[allow(clippy::too_many_arguments)]
+    pub fn create_buffer_from_memfd(
+        &self,
+        mem_fd: impl AsFd,
+        offset: usize,
+        size: usize,
+        format: Fourcc,
+        width: u32,
+        height: u32,
+        stride: u32,
+    ) -> Result<Dmabuf, io::Error> {
+        if width >= u16::MAX as u32 || height >= u16::MAX as u32 {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+        if !(stride as usize).is_multiple_of(STRIDE_ALIGN) {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+
+        if !size.is_multiple_of(*PAGE_SIZE) || size < (height * stride) as usize {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+
+        let dma_fd = udmabuf_from_memfd(self.dev.as_fd(), mem_fd.as_fd(), offset as u64, size as u64)?;
+        let mut dmabuf = Dmabuf::builder(
+            Size::new(width as i32, height as i32),
+            format,
+            Modifier::Linear,
+            DmabufFlags::empty(),
+        );
+        dmabuf.add_plane(dma_fd, 0, 0, stride);
+        dmabuf.build().ok_or_else(|| io::ErrorKind::Unsupported.into())
+    }
+}
+
+impl Allocator for UdmabufAllocator {
+    type Buffer = Dmabuf;
+    type Error = io::Error;
+
+    fn create_buffer(
+        &mut self,
+        width: u32,
+        height: u32,
+        fourcc: Fourcc,
+        modifiers: &[Modifier],
+    ) -> Result<Self::Buffer, Self::Error> {
+        if width >= u16::MAX as u32 || height >= u16::MAX as u32 {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+        if !modifiers.contains(&Modifier::Linear) {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+        let stride = (width as usize * get_bpp(fourcc).ok_or(io::ErrorKind::InvalidData)?)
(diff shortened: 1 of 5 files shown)


src/backend/allocator/udmabuf.rs
```diff
@@ -0,0 +1,166 @@
...
+/// Note: [`Allocator::create_buffer`] will always return linear buffers for a
+/// [`UdmabufAllocator`] and fail if [`Modifier::Linear`] is not in the list
+/// of passed modifiers.
+#[derive(Debug)]
+pub struct UdmabufAllocator {
+    dev: OwnedFd,
+}
+
+/// Minimum alignment requirement for the stride of mem-fds to import them as [`Dmabuf`]s.
+pub const STRIDE_ALIGN: usize = 256;
```
ids1024: Where does this number come from?

`udmabuf` itself doesn't have a stride requirement since it has no concept of strides (it's a generic buffer that doesn't necessarily contain pixel data). I seem to recall that Intel didn't require any stride alignment (at least with older iGPUs), while AMD and Nvidia were stricter, but I don't think required as much as 256 byte alignment (grepping through my notes, it seems I saw 32 byte alignment needed?).

(Ideally dmabuf feedback would advertise what stride of dmabufs are supported, but EGL and Vulkan don't seem to expose that.)

Drakulix: amd requires a 256 alignment, if I am not mistaken?

src/backend/allocator/udmabuf.rs
```diff
@@ -0,0 +1,166 @@
...
+        let fd = rustix::fs::open("/dev/udmabuf", OFlags::RDONLY | OFlags::CLOEXEC, Mode::empty())?;
+        Ok(UdmabufAllocator { dev: fd })
+    }
+
+    /// Creates a [`Dmabuf`] from an existing allocation represented by a memfd.
+    ///
+    /// For importing to succeed a few additional requirements need to be satisfied:
+    /// - the width and height cannot exceed 65535.
+    /// - the stride needs to be aligned to [`STRIDE_ALIGN`].
+    /// - size and offset need to be aligned to the platforms page size.
```
ids1024, line 51: I suppose if we instead pass an `offset` to `Dmabuf::add_plane` we only need an offset aligned to `STRIDE_ALIGN`? Not sure if it otherwise makes much difference whether we use the stride for creating the udmabuf or pass it when importing the dmabuf.

src/backend/allocator/udmabuf.rs
```diff
@@ -0,0 +1,166 @@
...
+    ///
+    /// Note: dmabufs created through this interface will always be interpreted as linear.
+    #[allow(clippy::too_many_arguments)]
+    pub fn create_buffer_from_memfd(
+        &self,
+        mem_fd: impl AsRawFd,
```
ids1024: Generally, Rust APIs should use `AsFd` now rather than `AsRawFd`. Technically it would be unsafe to accept `AsRawFd`.

src/backend/allocator/udmabuf.rs
```diff
@@ -0,0 +1,166 @@
...
+        let size = (height as usize * stride + (*PAGE_SIZE - 1)) & page_mask;
+
+        let mem_fd = memfd_create("udmabuf", MemfdFlags::ALLOW_SEALING | MemfdFlags::CLOEXEC)?;
+        ftruncate(&mem_fd, size as u64)?;
+        fcntl_add_seals(&mem_fd, SealFlags::SHRINK)?;
+        self.create_buffer_from_memfd(mem_fd, 0, size, fourcc, width, height, stride as u32)
+    }
+}
+
+fn udmabuf_from_memfd(fd: impl AsFd, mem_fd: impl AsRawFd, offset: u64, size: u64) -> io::Result<OwnedFd> {
```
ids1024: This likewise probably shouldn't use `AsRawFd`.

Since it's a private function used in only one place, it should probably just use `BorrowedFd` for both arguments, then call `AsRawFd` in the implementation.

src/backend/allocator/udmabuf.rs
ids1024: I thought at some point I tested it and saw less that a 256 byte alignment required on an AMD integrated GPU, but I may be wrong; in any case, I don't have any discrete AMD hardware if that's different.

src/backend/allocator/udmabuf.rs
```diff
@@ -0,0 +1,166 @@
...
+    ) -> Result<Dmabuf, io::Error> {
+        if width >= u16::MAX as u32 || height >= u16::MAX as u32 {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+        if stride as usize % STRIDE_ALIGN != 0 {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+
+        let page_mask = !(*PAGE_SIZE - 1);
+        if (size & page_mask) != size || size < (height * stride) as usize {
```
ids1024: Isn't `size & page_mask) != size` equivalent to just `size.is_multiple_of(PAGE_SIZE)`?

src/backend/allocator/udmabuf.rs
```diff
@@ -0,0 +1,166 @@
...
+        if width >= u16::MAX as u32 || height >= u16::MAX as u32 {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+        if !modifiers.contains(&Modifier::Linear) {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+        let stride = (width as usize * get_bpp(fourcc).ok_or(io::ErrorKind::InvalidData)?)
+            .next_multiple_of(STRIDE_ALIGN);
+        let page_mask = !(*PAGE_SIZE - 1);
+        let size = (height as usize * stride + (*PAGE_SIZE - 1)) & page_mask;
```
ids1024: Unless I'm missing something, the way `page_mask` is used here looks wrong. `(*PAGE_SIZE - 1)) & page_mask` is just `(*PAGE_SIZE - 1)) & !(*PAGE_SIZE - 1)`?

What we want is `.next_multiple_of(*PAGE_SIZE)`, right?

src/backend/allocator/udmabuf.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+    const UDMABUF_FLAGS_CLOEXEC: u32 = 0x01;
+
+    unsafe impl Ioctl for udmabuf_create {
+        type Output = RawFd;
+
+        const IS_MUTATING: bool = false;
+
+        fn opcode(&self) -> Opcode {
+            opcode::read_write::<udmabuf_create>(b'u', 0x42)
```
ids1024: It looks like this should be `opcode::write` rather than `opcode::read_write`? https://github.com/torvalds/linux/blob/master/include/uapi/linux/udmabuf.h defines it using `_IOW`, not `_IOWR`, and I got an invalid ioctl error trying to call this function. It seems to work after changing it.

src/backend/allocator/udmabuf.rs
ids1024: Testing it, it does indeed look like AMD needs a 256 byte alignment (while Intel requires none). So this should be good. :+1: 

Probably it should have a quick comment explaining that this is a value that generally works across GPUs.

It would definitely be nice if EGL/Vulkan and the Wayland dmabuf protocol could expose stride requirements; but anyway, the fact we can set whatever stride we want is another advantage of udmabuf. With gbm buffers, `GbmBufferFlags::SCANOUT` is the only option we have to increase the stride alignment beyond what the driver would default to.

Drakulix: https://gitlab.freedesktop.org/mesa/mesa/-/work_items/14027

src/backend/allocator/udmabuf.rs
Drakulix: I am not sure, if it works this way or if the start of the plane needs to be page-aligned for the dmabuf to be created successfully. Documentation on the ioctl is a bit sparse, so I generally coded this to be very conservative.

src/backend/allocator/udmabuf.rs
Drakulix: updated comment

src/backend/allocator/udmabuf.rs
Drakulix: fixed

src/backend/allocator/udmabuf.rs
Drakulix: yeah, I was obviously too inspired by C code to clean this up...

src/backend/allocator/udmabuf.rs
Drakulix: fixed

src/backend/allocator/udmabuf.rs
Drakulix: fixed

src/backend/allocator/udmabuf.rs
Drakulix: fixed

src/backend/allocator/udmabuf.rs
ids1024: We could always experiment with that later if we want to try https://github.com/Smithay/smithay/issues/2044 / https://zamundaaa.github.io/wayland/2026/05/06/making-wl-shm-fast.html. We don't really need a stable API here or anything.

But if we want to use `udmabuf` for `wl_shm`, I suspect we'd want to create a single `udmabuf` for the whole pool, that we can slice into multiple `Dmabuf`s. We may want to make `Dmabuf` use an `Arc` for the file descriptor (we already have to dup an fd in `AsDmabuf for VulkanImage` for multiple planes).

Drakulix: yeah the idea wasn't to solve that here, but just provide the api already, that can then be used/adjusted as needed.

ids1024 (approved):
As noted in the review comment, we may want to consider some changes to the implementation or API if we want to use this for `wl_shm`. But the `Allocator` here seems good; we can experiment with what works best for `wl_shm` later.

PR #2088, opened 2026-07-07
xwayland/xwm: Don't error when setting primary output if already set

RipleyTom:
## Description
Currently if setting output to already primary output set_randr_primary_output returns `Err(PrimaryOutputError::OutputUnknown)`.
Changed it to return `Ok(())` without any extra work.

## Checklist


Drakulix (approved):
Thanks!

PR #2089, opened 2026-07-08
xwayland/xwm: Fix missing WM_STATE update on Unmap

RipleyTom:
## Description
WM_STATE was never set to WithdrawnState after root window was unmaped.
This was found as wine's x11drv expect this to be set and games wait on that signal to map the window again(and thus would leave the game rendering in the background, invisible).
ref: https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.3.1

## Checklist

### DIFF
--- a/src/xwayland/xwm/mod.rs
+++ b/src/xwayland/xwm/mod.rs
@@ -1855,6 +1855,13 @@ where
                         if let Some(frame) = state.mapped_onto.take() {
                             conn.destroy_window(frame)?;
                         }
+                        conn.change_property32(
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Oh I see, because we destroy the frame window on `UnmapNotify`, the `set_mapped(false)` call on `State::unmapped_window` is a no-op.

Since `UnmapNotify` means we already have an unmapped window anyway (while `set_mapped` is meant as a way for the compositor to issue an unmap at will) this looks like a correct fix.

LGTM. Thanks!

PR #2090, opened 2026-07-10
define Yuv420 and Nv12 formats

karthago1:
use case:
application can advertise to the wayland client that it supports these formats. and can create shaders to convert dmabufs or shared memory buffers into RGBA formats

## Checklist


Drakulix (approved):
thanks

PR #2091, opened 2026-07-11
Make ClickGrab's motion focus better when the surface has moved

kelnos:
## Description

If the focus target has moved since the grab has started, the focus
parameter in the start data will be out of date, and motion events will
be sent relative to the old/original position of the surface.

We ran into this with xfce4-panel's move code.  It's a layer-shell
surface, and when the user grabs the move handle and drags, it moves
itself by watching pointer motion events and calculating the diff from
the previous position, and then updating its margins.

With the motion events delivered with respect to the start data's focus
position, the panel move goes wild, because it is (unknowingly)
recomputing its margins based on the cumulative difference in the
pointer's location, rather than the diff.

I've broken this into three commits, because the first may be less objectionable than the second, which may be less objectionable than the third, even though the third is the only one that truly fixes the problem.

The first commit prefers the coordinates in the `focus` param to the function over that in `start_data.focus`.  This helps, but if the user moves the pointer fast enough that it's outside the focus target for any length of time, it has to go back to the start data, which again causes problems.

The second commit stores the last-known focus (that matches the target of the start focus) in `ClickGrab`, and when the `focus` param to the function is `None`, it falls back to the stored location.  While this is better, honestly it still doesn't behave very well in my testing; it's still very easy to get the panel to start moving wildly around the screen, completely divorced from the pointer's position (though it *is* a bit easier to carefully get the pointer back on top of the panel to fix the situation).

The third commit is the real fix: I've added an optional hook to `SeatHandler` to allow `ClickGrab` to ask the compositor the current location of the `PointerFocus`.  I'm not quite sure about the API here, if that's a reasonable place to put it, but this does completely fix the issue for me, at least.

Anyhow, happy to make any changes, and/or squash things down to whatever solution y'all feel is the correct one.  I looked at labwc and wayfire, and this is how they both work in the face of moving surfaces during an implicit pointer grab.

## Checklist

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -137,7 +137,10 @@ use self::{
     pointer::{CursorImageStatus, PointerHandle, PointerTarget},
     touch::TouchGrab,
 };
-use crate::utils::{Serial, user_data::UserDataMap};
+use crate::{
+    input::pointer::{ClickGrab, GrabStartData as PointerGrabStartData, PointerGrab},
+    utils::{Serial, user_data::UserDataMap},
+};
 
 pub mod dnd;
 pub mod keyboard;
@@ -145,7 +148,7 @@ pub mod pointer;
 pub mod touch;
 
 /// Handler trait for Seats
-pub trait SeatHandler: Sized {
+pub trait SeatHandler: Sized + 'static {
     /// Type used to represent the target currently holding the keyboard focus
     type KeyboardFocus: KeyboardTarget<Self> + PartialEq + Clone + 'static;
     /// Type used to represent the target currently holding the pointer focus
@@ -164,6 +167,15 @@ pub trait SeatHandler: Sized {
 
     /// Callback that will be notified whenever the keyboard led state changes.
     fn led_state_changed(&mut self, _seat: &Seat<Self>, _led_state: LedState) {}
+
+    /// Provides the implicit pointer grab for clicks
+    ///
+    /// When the user presses a pointer button, an implicit pointer grab is installed. If your
+    /// compositor needs custom behavior for this grab, you can implement this trait item and
+    /// return your own [`PointerGrab`] implementation.
+    fn click_grab(&mut self, start_data: PointerGrabStartData<Self>) -> impl PointerGrab<Self> {
+        ClickGrab::new(start_data)
+    }
 }
 /// Delegate type for all [Seat] globals.
 ///
--- a/src/input/pointer/grab.rs
+++ b/src/input/pointer/grab.rs
@@ -244,6 +244,7 @@ impl<D: SeatHandler + 'static> PointerGrab<D> for DefaultGrab {
                         button: event.button,
                         location: handle.current_location(),
                     },
+                    focus: handle.current_focus(),
                 },
             );
         }
@@ -343,6 +344,7 @@ impl<D: SeatHandler + 'static> PointerGrab<D> for DefaultGrab {
 /// the grab once all are released.
 pub struct ClickGrab<D: SeatHandler> {
     start_data: GrabStartData<D>,
+    focus: Option<(D::PointerFocus, Point<f64, Logical>)>,
 }
 
 impl<D: SeatHandler + 'static> fmt::Debug for ClickGrab<D> {
@@ -361,11 +363,14 @@ impl<D: SeatHandler + 'static> PointerGrab<D> for ClickGrab<D> {
         focus: Option<(<D as SeatHandler>::PointerFocus, Point<f64, Logical>)>,
         event: &MotionEvent,
     ) {
-        let focus = match (&focus, &self.start_data.focus) {
-            (Some((current, loc)), Some((start, _))) if current == start => Some((current.clone(), *loc)),
-            _ => self.start_data.focus.clone(),
-        };
-        handle.motion(data, focus, event);
+        if let (Some((new_target, new_location)), Some((grab_target, grab_location))) =
+            (&focus, self.focus.as_mut())
+        {
+            if *new_target == *grab_target {
+                *grab_location = *new_location;
+            }
+        }
+        handle.motion(data, self.focus.clone(), event);
     }
 
     fn relative_motion(
--- a/src/input/pointer/grab.rs
+++ b/src/input/pointer/grab.rs
@@ -358,10 +358,14 @@ impl<D: SeatHandler + 'static> PointerGrab<D> for ClickGrab<D> {
         &mut self,
         data: &mut D,
         handle: &mut PointerInnerHandle<'_, D>,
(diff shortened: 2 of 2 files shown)


Drakulix:
> We ran into this with xfce4-panel's move code. It's a layer-shell surface, and when the user grabs the move handle and drags, it moves itself by watching pointer motion events and calculating the diff from the previous position, and then updating its margins.

Ouch. This sounds like quite the hack for something that really should have a move request (something layer surface really was never designed to provide).

Does this behave correctly in other compositors? I would really like to avoid adding api, that assumes anything in global coordinate space. (At the very least we need to update the wording to mean the cursor coordinate space, but ideally I would love to figure out a way to avoid commit 3 in it's current form all together.)

kelnos:
> Ouch. This sounds like quite the hack for something that really should have a move request (something layer surface really was never designed to provide).

Yeah, I don't love it.  We've always offered the ability (on X11) for users to move panels around by grabbing them, and there wasn't really another way to implement this.  I thought about an extension protocol on top of layer-shell, but as you say, how the surface works isn't really designed for that in the first place.  And it would be difficult to implement for us client-side as we're using gtk3-layer-shell, and I suspect an extension protocol that messes with margins and stuff from the compositor side would break the library's internal state.

> Does this behave correctly in other compositors?

Yes, it works fine in labwc and wayfire at least.

> I would really like to avoid adding api, that assumes anything in global coordinate space. (At the very least we need to update the wording to mean the cursor coordinate space, but ideally I would love to figure out a way to avoid commit 3 in it's current form all together.)

Understood, I'm not really happy with the third commit either.  My fallback plan was going to be to just copy/paste ClickGrab into my compositor, edit it, and manually install my custom grab on click, since of course I can fix this myself in my own grab code.

But I do believe this represents a real issue, regardless of our hacky move code/process.  If ClickGrab is active and then the surface (not just layer-surface, any surface) initially under it moves (for any reason, even if not user-initiated), then motion events during the grab will be incorrect.  This is maybe a very rare corner case, so maybe it isn't worth fixing.  labwc and wayfire don't have this problem because AFAICT wlroots doesn't have an implicit grab there, and the compositors themselves implement them, and have the information needed to not have this problem.

Anyway, I'm open to whatever you'd like to do here, even if that turns out to be nothing :)

Drakulix:
> Yeah, I don't love it. We've always offered the ability (on X11) for users to move panels around by grabbing them, and there wasn't really another way to implement this. I thought about an extension protocol on top of layer-shell, but as you say, how the surface works isn't really designed for that in the first place. And it would be difficult to implement for us client-side as we're using gtk3-layer-shell, and I suspect an extension protocol that messes with margins and stuff from the compositor side would break the library's internal state.

Fair enough.

> > I would really like to avoid adding api, that assumes anything in global coordinate space. (At the very least we need to update the wording to mean the cursor coordinate space, but ideally I would love to figure out a way to avoid commit 3 in it's current form all together.)
> 
> Understood, I'm not really happy with the third commit either. My fallback plan was going to be to just copy/paste ClickGrab into my compositor, edit it, and manually install my custom grab on click, since of course I can fix this myself in my own grab code.
> 
> But I do believe this represents a real issue, regardless of our hacky move code/process. If ClickGrab is active and then the surface (not just layer-surface, any surface) initially under it moves (for any reason, even if not user-initiated), then motion events during the grab will be incorrect. This is maybe a very rare corner case, so maybe it isn't worth fixing. labwc and wayfire don't have this problem because AFAICT wlroots doesn't have an implicit grab there, and the compositors themselves implement them, and have the information needed to not have this problem.
> 
> Anyway, I'm open to whatever you'd like to do here, even if that turns out to be nothing :)

Yeah thinking of solutions to this mess is the difficult part I think.

I think the sanest option to fix this in smithay would be to expose more of the DefaultGrab and ClickGrab and allow downstream to set it's own. This could be as simple as having a `ClickGrabBehavior`-enum on creation (with variants `Auto` or `Manual`) so that you don't have to fight the built-in grab code.

But maybe something like a generic + trait and have smithay provide a `DefaultClickGrab` would be cleaner? So that the `DefaultGrab` could still engage the provided ClickGrab. Something that needs to be explored.

kelnos:
I think I like the trait idea, and providing a default implementation.  How would the impl get provided to smithay though?  A new method on `SeatHandler`? That way there'd be no changes needed for current compositors that already do the default thing, and compositors like mine could return a custom impl.

Do we need a new trait for this, even?  I guess for a simple impl of `PointerGrab` there's a bunch of boilerplate, but maybe that's ok...

Drakulix:
> I think I like the trait idea, and providing a default implementation. How would the impl get provided to smithay though? A new method on SeatHandler?

That would make sense I think.

> Do we need a new trait for this, even? I guess for a simple impl of `PointerGrab` there's a bunch of boilerplate, but maybe that's ok...

Probably not, thinking about it.

kelnos:
Ok, I dropped the 3rd commit and added a new one that adds a new `SeatHandler::click_grab()` item.  The `SeatHandler` impl now needs to be `'static`, as `click_grab()` returns `PointerGrab<Self>`.  Hopefully that's ok?

I decided to leave the first two commits, as they are helpful in situations where there's some mild/simple (and perhaps one-time?) motion of the surface after a click starts.  But I can remove them if you think this is a case where the compositor should just implement it if it needs it.

What do you think?

Drakulix:
Looks good!

> The SeatHandler impl now needs to be 'static, as click_grab() returns PointerGrab<Self>. Hopefully that's ok?

I think I need to see if implementing this in cosmic-comp causes any issues. @YaLTeR Can you check niri?

Otherwise we could make the click-grab an associated type?

> I decided to leave the first two commits, as they are helpful in situations where there's some mild/simple (and perhaps one-time?) motion of the surface after a click starts. But I can remove them if you think this is a case where the compositor should just implement it if it needs it.

Yeah I think that is fine and a good improvement.

Can you link xfce implementation of this? I like to see this in action.

YaLTeR:
niri built fine with this PR (I haven't tried making a custom grab)

kelnos:
> Can you link xfce implementation of this? I like to see this in action.

https://gitlab.xfce.org/xfce/xfwl4/-/blob/custom-clickgrab/src/core/shell/grabs/click.rs?ref_type=heads

With this setup, the xfce4-panel grab handle for movement works perfectly, no matter how quickly I move the pointer around.

(It won't build for you, as it's pointing to my local checkout of smithay, which is not the same as this branch.  xfwl4 is a bit behind smithay master right now, and there are incompatible changes between xfwl4's pinned rev and this branch.  After this PR is merged I'll update and fix the incompatibilities, but not ready to do that yet...)

Drakulix (approved):
Also seems to compile fine with cosmic. Merging :)

PR #2092, opened 2026-07-11
fix compilation errors

sophietheopossum:
## Description
fixes some minor compilation errors, changes are fairly self explanatory, mostly related to imports. don't think CI normally tries compiling these, which is why it might have not been noticed until now.

## Checklist
* [ X ] I agree to smithay's [Developer Certificate of Origin](https://github.com/Smithay/smithay/blob/master/DCO.md).


Drakulix:
Hmm, wlcs_anvil should definitely be compiled by CI, however smithay-drm-extras indeed isn't, we should fix that. (Doesn't have to be addressed by this PR.) Thanks!

PR #2093, opened 2026-07-11
fix doc tests

sophietheopossum:
## Description
makes sure tests come back fully successful. inspired by the example code in [simple.rs](https://github.com/Smithay/smithay/blob/master/smithay-drm-extras/examples/simple.rs)

## Checklist
* [ x ] I agree to smithay's [Developer Certificate of Origin](https://github.com/Smithay/smithay/blob/master/DCO.md).


Drakulix (approved):
Thanks!

PR #2095, opened 2026-07-14
renderer: `cgmath` -> `glam`

### LINKED ISSUE #2087: cgmath dependency is unmaintained
`cargo deny` started warning me that `cgmath` is unmaintained: https://rustsec.org/advisories/RUSTSEC-2026-0196

Unfortunately there doesn't seem to be an alternative at present...

alterization:
## Description

Resolves #2087.

## Checklist

### DIFF
--- a/src/backend/renderer/gles/mod.rs
+++ b/src/backend/renderer/gles/mod.rs
@@ -3,8 +3,8 @@
 // GL calls are all unsafe, so not very helpful in this module.
 #![allow(unsafe_op_in_unsafe_fn)]
 
-use cgmath::{Matrix3, Vector2, prelude::*};
 use core::slice;
+use glam::{Mat3, Vec2};
 use std::{
     collections::HashMap,
     ffi::{CStr, CString},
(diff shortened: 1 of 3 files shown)


src/backend/renderer/gles/mod.rs
```diff
@@ -2519,7 +2522,7 @@ impl GlesFrame<'_, '_> {
                 self.renderer.solid_program.uniform_matrix,
                 1,
                 ffi::FALSE,
-                mat.as_ptr(),
+                &mat.to_cols_array() as *const GLfloat,
```
Drakulix: I would like to avoid new allocations in the hot-path. It looks like `glam` has guarantees about the memory layout and supports `bytemuck` or `zerocopy` for casting. So we should be able to just use a `Mat3` (instead of `Mat3A`) and skip the array allocation.

Drakulix (commented):
Looks mostly good, thanks for working on this.

Drakulix (approved):
Thanks. Looks correct to me.

@ids1024 @cmeissl Anyone wanna lend a second pair of eyes to make sure we don't introduce any regressions here?

git-f0x:
Seems to work fine from some light testing in my [`cosmic-comp` fork](https://github.com/git-f0x/cosmic-comp/tree/glam). The only thing that changed on that end is that the `Rectangle` size and loc fields need to be cast to `f32` for `Vec2` (and `DVec2` doesn't mix with everything else). Not sure if that's an issue.

PolyMeilex (approved):
Looks correct, nothing catches my attention

PR #2096, opened 2026-07-14
dmabuf: expose and preserve buffer flags

thelipe7:
## Description
Expose `DmabufFlags` in the backend and preserve unknown flags received via the linux-dmabuf protocol.

## Checklist

### DIFF
--- a/src/backend/allocator/dmabuf.rs
+++ b/src/backend/allocator/dmabuf.rs
@@ -231,6 +231,8 @@ impl Dmabuf {
     }
 
     /// Returns the flags stored on this buffer without interpreting them.
+    ///
+    /// Bits unknown to this Smithay version are retained.
     pub fn flags(&self) -> DmabufFlags {
         self.0.flags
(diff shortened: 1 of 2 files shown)


Drakulix:
What is the use-case here? The allowed bits are clearly defined by the protocol.

thelipe7:
Compositors implementing only a progressive composition path must reject `INTERLACED` / `BOTTOM_FIRST` flags at import time, as recommended by the linux-dmabuf protocol.

Previously Smithay only exposed `y_inverted()`, so the other flags were stored internally but not accessible to the compositor.

Switched to `from_bits_retain` when parsing flags from the wire. This preserves unknown bits instead of silently dropping them, allowing compositors to explicitly reject anything they do not support (instead of treating unknown bits as zero).

This change gives compositors full control over buffer policy decisions based on the complete set of flags sent by the client.

Drakulix (approved):

PR #2098, opened 2026-07-16
ci: Deply docs using `deploy-pages` action

ids1024:
## Description
This new mechanism of deploying Github Pages does not require pushing to a branch.

I've already updated wayland-rs like this (https://github.com/Smithay/wayland-rs/pull/945), so it should work here too. An admin will need to go to the settings of the repo, and change it to deploy Pages from github actions instead of a branch. Presumably that should be done before merging.

The `gh-pages` branch should then be safe to delete.

## Checklist


Drakulix (approved):
Seems nice. I'll update the settings.

PR #2100, opened 2026-07-17
[Renderer] Use `Affine2` where possible

git-f0x:
## Description

This changes things that don't directly interact with OpenGL to use `Affine2` instead of `Mat3`, since it's more efficient for 2D affine transformations, and is 16 byte aligned for SIMD. Also removes the `float-types` feature from glam, since it's only for f64 types (e.g. DMat3).

Also addresses deprecated method warnings in benches for rand.

## Checklist


Drakulix (approved):
Thanks!

PR #2104, opened 2026-07-19
renderer: allow early release of cached implicit buffers

thelipe7:
## Description

Add `RendererSurfaceState::release_implicit_buffer()` for compositors that have
fully imported an implicitly synchronized client buffer into renderer-owned
storage.

Normally, Smithay releases a `wl_buffer` when the last `Buffer` wrapper drops.
For SHM renderers that copy client pixels into an internal texture, that can
keep a client buffer unavailable while the surface remains attached even though
future redraws no longer read the client backing.

This API releases the implicit `wl_buffer` while preserving the current
renderer surface state and cached texture. It is intended to be called only
after a successful import into compositor-owned storage.

The release is idempotent. A new `BufferAssignment::NewBuffer` for the same
`wl_buffer` rearms release so each attachment generation can be released once.

Explicitly synchronized buffers are intentionally excluded: their release
points must remain tied to GPU completion.

## Motivation

This lets SHM clients reuse a buffer after the compositor has copied it,
without requiring detach, replacement, or surface destruction first. It is
particularly useful for single-buffer or memory-constrained clients.

## Checklist

### DIFF
--- a/src/backend/renderer/utils/wayland.rs
+++ b/src/backend/renderer/utils/wayland.rs
@@ -19,7 +19,10 @@ use crate::{
 use std::{
     any::Any,
     collections::{HashMap, hash_map::Entry},
-    sync::{Arc, Mutex},
+    sync::{
+        Arc, Mutex,
+        atomic::{AtomicBool, Ordering},
+    },
 };
 
 use super::{CommitCounter, DamageBag, DamageSet, DamageSnapshot, SurfaceView};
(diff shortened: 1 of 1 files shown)


Drakulix:
The reason we don't have this feature is because smithay doesn't bind you to use just one renderer. Quite the contrary, the `MultiRenderer` incentives you to have multiple on multi-gpu systems. In those cases we cannot know, if a compositor might switch the renderer for a given window (e.g. because the window is dragged to a different output) in which case we need to re-import from shared memory (or worse we would have to transfer from the other GPU, which isn't always possible). Another case that is impacted by this even on systems with a single GPU are GPU resets, which can invalidate any resources on the GPU.

Since there is no way to ask the client to give us a fresh buffer in those cases (looking at the still work-in-progress surface-invalidation protocol), we need to keep the buffer around.

PR #2105, opened 2026-07-19
add patch for build crate in e2k cpu architecture

sanua356:
Good afternoon!

When generating code for the "linux-raw-sys" crate using "bindgen" on the "E2K (Elbrus 2000)" processor architecture, one of the types for the "st_rdev" field does not match the type implemented for the "x86_64" architecture (u32 instead of u64). A small patch is required to fix the error.

- [x]  I agree to smithay's [Developer Certificate of Origin](https://github.com/Smithay/smithay/blob/master/DCO.md).

### DIFF
--- a/src/backend/drm/device/fd.rs
+++ b/src/backend/drm/device/fd.rs
@@ -1,3 +1,5 @@
+#![allow(unexpected_cfgs)]
+
 use drm::{Device as BasicDevice, control::Device as ControlDevice};
 use std::{
     os::unix::io::{AsFd, AsRawFd, BorrowedFd, RawFd},
@@ -94,7 +96,18 @@ impl DrmDeviceFd {
 
     /// Returns the `dev_t` of the underlying device
     pub fn dev_id(&self) -> rustix::io::Result<libc::dev_t> {
-        Ok(rustix::fs::fstat(&self.0.fd)?.st_rdev)
+        let dev: libc::dev_t = {
+            #[cfg(not(target_arch = "e2k"))]
+            {
+                rustix::fs::fstat(&self.0.fd)?.st_rdev
+            }
+
+            #[cfg(target_arch = "e2k")]
+            {
+                rustix::fs::fstat(&self.0.fd)?.st_rdev as u64
+            }
+        };
+        Ok(dev)
     }
 
     /// Returns a weak reference to the underlying device
--- a/src/backend/drm/device/fd.rs
+++ b/src/backend/drm/device/fd.rs
@@ -1,5 +1,3 @@
-#![allow(unexpected_cfgs)]
-
 use drm::{Device as BasicDevice, control::Device as ControlDevice};
 use std::{
     os::unix::io::{AsFd, AsRawFd, BorrowedFd, RawFd},
@@ -96,17 +94,9 @@ impl DrmDeviceFd {
 
     /// Returns the `dev_t` of the underlying device
     pub fn dev_id(&self) -> rustix::io::Result<libc::dev_t> {
-        let dev: libc::dev_t = {
-            #[cfg(not(target_arch = "e2k"))]
-            {
-                rustix::fs::fstat(&self.0.fd)?.st_rdev
-            }
-
-            #[cfg(target_arch = "e2k")]
-            {
-                rustix::fs::fstat(&self.0.fd)?.st_rdev as u64
-            }
-        };
+        // Hack (u64 as libc::dev_t) for E2K CPU architecture
+        #[allow(clippy::unnecessary_cast)]
+        let dev = rustix::fs::fstat(&self.0.fd)?.st_rdev as libc::dev_t;
         Ok(dev)
     }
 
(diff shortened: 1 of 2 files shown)


src/backend/drm/device/fd.rs
```diff
@@ -94,7 +94,10 @@ impl DrmDeviceFd {
 
     /// Returns the `dev_t` of the underlying device
     pub fn dev_id(&self) -> rustix::io::Result<libc::dev_t> {
-        Ok(rustix::fs::fstat(&self.0.fd)?.st_rdev)
+        // Hack (u64 as libc::dev_t) for E2K CPU architecture
+        #[allow(clippy::unnecessary_cast)]
+        let dev = rustix::fs::fstat(&self.0.fd)?.st_rdev as libc::dev_t;
```
Drakulix, line 99: Hmm, isn't think rather a rustix or libc error? `man 2 stat` defines `st_rdev` as `dev_t`, so either one of these crates isn't returning the right type-alias for your architecture. (and you certainly don't want to accidentally truncate the result, that is just asking for trouble.)

src/backend/udev.rs
```diff
@@ -98,7 +98,11 @@ impl UdevBackend {
             .into_iter()
             // Create devices
             .flat_map(|path| match stat(&path) {
-                Ok(stat) => Some((stat.st_rdev, path)),
+                Ok(stat) => Some({
+                    // Hack (u64 as u64) for E2K CPU architecture
+                    #[allow(clippy::unnecessary_cast)]
+                    (stat.st_rdev as u64, path)
```
Drakulix, line 104: Same thing here. What we are storing is a `libc::dev_t` inside the hashmap, so this should by all means match and I don't want to silence any errors, that could potentially show wrong casts.

src/backend/drm/device/fd.rs
sanua356: I also thought that the error was in libc or rustix. However, after looking at the stat structure headers, I found that the "st_rdev" field is indeed of type "u32".

```
cat /usr/include/asm/stat.h


/*
 * SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note
 * Copyright (c) 2023 MCST
 */

#ifndef _E2K_STAT_H_
#define _E2K_STAT_H_

/*
 * Tuned up to match GNU libc defaults.
 */

#include <linux/types.h>

struct stat {
  unsigned long st_dev;
  unsigned long st_ino;
  unsigned int  st_mode;
  __u32 st_nlink;
  unsigned int st_uid;
  unsigned int st_gid;
  __u32 st_rdev;
  long st_size;
  long st_blksize;
  long st_blocks;
  long st_atime;
  unsigned long st_atime_nsec;
  long st_mtime;
  unsigned long st_mtime_nsec;
  long st_ctime;
  unsigned long st_ctime_nsec;
};

#endif /* _E2K_STAT_H_ */

```

src/backend/udev.rs
sanua356: I understand. Maybe I should return conditional compilation only for "e2k" with type casting to dev_t? (Check reply upper)

src/backend/drm/device/fd.rs
nekopsykose: `/usr/include/asm/stat.h` isn't what the libc api exposes the types as though. For x86_64-linux for instance, the type of `st_rdev` is `__kernel_ulong_t` in that file. But that's obviously not the name that the libc uses, and in both rust `libc` and the actual system libc headers it is marked as `dev_t` (which is a u64 under the hood).

I think the point being made is that the e2k rust-libc fork probably shouldn't put a bare `u32` there. The kernel headers aren't related to that - it depends on the actual libc api. As an example, the posix api [defines it as dev_t](https://pubs.opengroup.org/onlinepubs/9799919799/), and libc's that follow that also call it dev_t.

Now, if what you're saying is that the kernel headers have a `u32` but the libc `dev_t` definition on e2k is actually not a `u32` but rather a `u64` or something, then that sounds very peculiar.

sanua356: @nekopsykose Everything is right. In the kernel headers, u32 is used, and dev_t in the libc fork is of type u64. The "rustix" crate describes the stat structure as it exists in the kernel headers, so the types do not actually match.

Rustix (line 766): https://github.com/helce/rustix/blob/devel_e2k/src/backend/linux_raw/fs/types.rs
Libc (line 11): https://github.com/helce/libc/blob/devel_e2k/src/unix/linux_like/linux/mod.rs

I'll probably just ask to change the type in crate "rustix" and the problem will go away. Even though the types overlap, it's a bit concerning.

nekopsykose: Huh, that's quite interesting, I think it's like that because in the kernel header the types of st_dev and st_rdev don't match either, so there can't be just a universal `dev_t`. Very strange, I wonder if there was a good reason for that or if the e2k kernel ABI just made an error defining it that way ages ago.

But yeah, I guess you can't really fix it in rust-libc then and need this cast.

PR #2107, opened 2026-07-22
wayland/pointer_constraints: call remove_constraint when pointer leave surface

skygrango:
## Description

related: https://github.com/Smithay/smithay/pull/2027#issuecomment-5032154662

The implementation ensures that remove_constraint() is invoked when the pointer leaves the surface, allowing the last saved cursor position hint to be applied.

This both improves the expected behavior of `Lifetime::Persistent` and ensures that pointer focus changes are handled immediately.

## Checklist

### DIFF
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -64,7 +64,9 @@ use smithay::{
             KeyboardShortcutsInhibitHandler, KeyboardShortcutsInhibitState, KeyboardShortcutsInhibitor,
         },
         output::{OutputHandler, OutputManagerState},
-        pointer_constraints::{PointerConstraintsHandler, PointerConstraintsState, with_pointer_constraint},
+        pointer_constraints::{
+            PointerConstraint, PointerConstraintsHandler, PointerConstraintsState, with_pointer_constraint,
+        },
         pointer_gestures::PointerGesturesState,
         presentation::PresentationState,
         relative_pointer::RelativePointerManagerState,
@@ -380,23 +382,26 @@ impl<BackendData: Backend> PointerConstraintsHandler for AnvilState<BackendData>
         }
     }
 
-    fn remove_constraint(&mut self, surface: &WlSurface, pointer: &PointerHandle<Self>) {
-        if with_pointer_constraint(surface, pointer, |constraint| constraint.is_none()) {
-            if let Some((hint_surface, hint_location)) = &self.cursor_position_hint {
-                let origin = self
-                    .space
-                    .elements()
-                    .find_map(|window| {
-                        (window.wl_surface().as_deref() == Some(hint_surface)).then(|| window.geometry())
-                    })
-                    .unwrap_or_default()
-                    .loc
-                    .to_f64();
+    fn remove_constraint(
+        &mut self,
+        _surface: &WlSurface,
+        pointer: &PointerHandle<Self>,
+        _constraint: Option<&PointerConstraint>,
+    ) {
+        if let Some((hint_surface, hint_location)) = &self.cursor_position_hint {
+            let origin = self
+                .space
+                .elements()
+                .find_map(|window| {
+                    (window.wl_surface().as_deref() == Some(hint_surface)).then(|| window.geometry())
+                })
+                .unwrap_or_default()
+                .loc
+                .to_f64();
 
-                pointer.set_location(origin + *hint_location);
-            }
-            self.cursor_position_hint = None;
+            pointer.set_location(origin + *hint_location);
         }
+        self.cursor_position_hint = None;
     }
 
     fn cursor_position_hint(
--- a/src/wayland/seat/pointer.rs
+++ b/src/wayland/seat/pointer.rs
@@ -25,7 +25,10 @@ use crate::{
         },
     },
     utils::{Client as ClientCoords, Point, Serial, iter::new_locked_obj_iter_from_vec},
-    wayland::{Dispatch2, compositor, pointer_constraints::with_pointer_constraint},
+    wayland::{
+        Dispatch2, compositor,
+        pointer_constraints::{PointerConstraintsHandler, with_pointer_constraint},
+    },
 };
 
 use super::{SeatHandler, WaylandFocus};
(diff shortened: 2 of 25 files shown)


skygrango:
@YaLTeR Does this meet your needs ?

skygrango:
Introducing the `PointerConstraintsHandler` trait significantly increases the scope of the changes. Perhaps we should move `remove_constraint()` out of PointerConstraintsHandler

src/wayland/seat/pointer.rs
```diff
@@ -235,24 +238,26 @@ impl WlPointerHandle {
...
-                if let Some(constraint) = constraint {
-                    constraint.deactivate();
-                }
+            let deactivated = with_pointer_constraint(self, &pointer, |constraint| {
+                constraint.map(|c| c.deactivate()).unwrap_or(false)
             });
+
+            if deactivated {
+                data.remove_constraint(self, &pointer);
+            }
```
YaLTeR: I would imagine that this code should go inside `deactivate()` itself to avoid putting it into every caller?

Or I guess `data` is inaccessible from there? Hmm, not sure how to structure this

skygrango: I've updated the signature of `deactivate()`, which I believe addresses your suggestion.

However, the downside is that `PointerConstraintsHandler` has now become a trait dependency of `PointerTarget`.

YaLTeR: That looks like the change that I had in mind, yeah.

It also doesn't seem like the repeated `remove_constraint()` call from `destroyed()` is a problem, because the constraint is required to be unique per surface, so the surface *must* destroy the previous constraint before making a new one, so it shouldn't be possible for a repeated `remove_constraint()` call to operate on a cursor position hint from a different constraint. Is my reasoning correct here?

skygrango: For `Lifetime::Oneshot`, explicitly calling `deactivate()` does result in one extra invocation of `remove_constraint()`. However, I don't think this has any practical impact. The compositor should ensure that only a single surface's cursor position hint is stored at any given time(per seat), and that the hint is consumed after it has been been applied. Consequently, the second invocation becomes a no-op because there is no cursor position hint remaining to apply.

YaLTeR: Yes, and I am thinking through whether it's possible to end up with the incorrect stored hint through overlapping pointer constraint objects, and so far I don't see any way.

skygrango:
I moved `remove_constraint()` to `SeatHandler`, which significantly reduces the amount of trait boilerplate and maintenance. 

Since `remove_constraint()` may be called while `with_pointer_constraint()` is active, the `PointerConstraint` is passed to the compositor so it can inspect the `RegionAttributes` without re-entering `with_pointer_constraint()`.

anvil/src/state.rs
```diff
@@ -321,6 +323,28 @@ impl<BackendData: Backend> SeatHandler for AnvilState<BackendData> {
     fn led_state_changed(&mut self, _seat: &Seat<Self>, led_state: LedState) {
         self.backend_data.update_led_state(led_state)
     }
+
+    fn remove_constraint(
```
Drakulix: Urgh, I am not sure I like the move here. So if you don't expose the constrains-protocol, you still have to implement this function? That makes no sense.

src/wayland/seat/pointer.rs
```diff
@@ -243,14 +243,14 @@ where
...
-    fn leave(&self, seat: &Seat<D>, _data: &mut D, serial: Serial, time: u32) {
+    fn leave(&self, seat: &Seat<D>, data: &mut D, serial: Serial, time: u32) {
         if let Some(pointer) = seat.get_pointer() {
             pointer.wp_pointer_gestures.leave::<D>(self, serial, time);
             pointer.wl_pointer.leave(self, serial, time);
 
             with_pointer_constraint(self, &pointer, |constraint| {
                 if let Some(constraint) = constraint {
-                    constraint.deactivate();
+                    constraint.deactivate(data, &self, &pointer);
```
Drakulix: So the issue is this call, right? (This would be so easy, if we had specialization.)

I would honestly rather have the `PointerTarget` impl for `WlSurface` require the `PointerContraintsHandler`, than `remove_constraint` to be part of the `SeatHandler`, if nobody has a better idea.

skygrango: how do you think about 
```rust
/// Callback that returns a reference to the pointer constraints handler, if implemented.
fn pointer_constraints_handler(&mut self) -> Option<&mut dyn PointerConstraintsHandler<Self>> {
    None
}
``` 
for `SeatHandler`

Drakulix: No please. The `SeatHandler` should not expose anything related to pointer constraints. If our `WlSurface` implementation of `PointerTarget` needs it, just add a requirement for the `PointerConstraintsHandler`.

skygrango:
@Drakulix Following your suggestion, I added `PointerConstraintsHandler` to `PointerTarget`. Could you please let me know if this is good enough? I'll clean up the Git history later.

Drakulix (approved):
LGTM

>  I'll clean up the Git history later.

I can simply squash the commits on merge, if you want.

YaLTeR:
I think this just needs to be merged now?

skygrango:
Thank you for your review. This can be merged now.

Drakulix (approved):

YaLTeR:
Uh oh, so there's actually a deadlock problem. `PointerHandle::motion()` locks the pointer inner, then calls `<WlSurface as PointerTarget>::leave()` which then calls compositor's `remove_constraint()`. The compositor is expected to call `pointer.set_location()` inside, which then again tries locking the pointer handle, resulting in a deadlock.

YaLTeR:
Hm, I guess it's debatable whether you even want to warp the pointer if the constraint was broken by pointer leaving the surface. Is there a way to find out in `remove_constraint()` whether this is the case? Then just skip set_location.

YaLTeR:
Perhaps need to pass a new parameter like `enum RemoveReason { LeftSurface, Deactivated, Destroyed }`?

YaLTeR:
Luckily with the current code it happens to be possible to tell if I'm not mistaken:
```
        if pointer.last_enter().is_some() {
            pointer.set_location(target);
        }
```
Since `pointer.wl_pointer.leave(self, serial, time);` is called right before `constraint.deactivate(data, self, &pointer);`, then by the time `remove_constraint()` is called, `last_enter()` is already cleared out. And it doesn't lock the inner handle.

And when the constraint is deactivated by the client, `last_enter()` will still point at that client, so the pointer will warp fine.

PR #2108, opened 2026-07-24
Add an `InputTime` type to use instead of `u32`/`u64`, with a `::now()` method

ids1024:
## Description
I've been wondering if it would be good if `time`  was a newtype instead of just an integer for a while.

Making the event structs in `src/input` using microsecond precision instead of just millisecond would presumably be a prerequisite for https://wayland.app/protocols/input-timestamps-unstable-v1.

But I ran into that trying to add `wl_keyboard.warp`, as an `is_warp: bool` field of `MotionEvent`. (I'm still not entirely sure on doing that vs having a separate `WarpEvent`; if all the logic for handling it is the same as motion events, it may make sense and save a lot of duplication). But with a fallback to send `motion` and a 0 `relative_motion` on clients not binding the latest version. `relative_motion` requires a microsecond time.

Going through this also reminds me that in some cases smithay synthesizes input events with a time of 0, which is probably not correct. In at  least one place it synthesizes an event using `CLOCK_MONOTONIC`. That does match libinput (https://wayland.freedesktop.org/libinput/doc/latest/timestamps.html) but technically isn't guranteed if we had a Wayland backend. I think X also was typically `CLOCK_MONOTONIC` but not technically guaranteed?

I still need to update anvil/etc.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,7 +120,8 @@ box minus the surface's frame extents.
 time. `LockSurface::ext_session_lock` has been added to allow the compositor to disambiguate, and associate lock surfaces
 with the correct lock instance.
 
-`backend::input` and `input` now use `backend::input::InputTime` for timestamps instead of `u32` or `u64`.
+`backend::input` and `input` now use `backend::input::InputTime` for timestamps instead of `u32` or `u64`. `InputTime::now()`
+is used for timestamps for synthesized events.
 
 ### Additions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,8 @@ box minus the surface's frame extents.
 time. `LockSurface::ext_session_lock` has been added to allow the compositor to disambiguate, and associate lock surfaces
 with the correct lock instance.
 
+`backend::input` and `input` now use `backend::input::InputTime` for timestamps instead of `u32` or `u64`.
+
 ### Additions
 
 - ExtBackgroundEffect protocol is now available in `smithay::wayland::background_effect` module.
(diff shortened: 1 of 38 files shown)


ids1024:
Compositors having to update `u32` to `InputTime` is a bit annoying, but mostly a simple search and replace. I think we probably want to make this change.

ids1024:
I've added an `InputTime::now()`  method. This assumes `CLOCK_MONOTONIC` (as `libinput` and `libei` specify; though Wayland doesn't; if I recall X11 also generally is `CLOCK_MONOTONIC` but it isn't guaranteed by the spec).

This makes it easier to generate a timestamp for synthesized events that don't get a timestamp from a backend event. Previously code using `Clock<Monotonic>` was duplicated in a few places for this. And some places just used a time of `0`, which seems to be invalid.

I considered instead defining a `SeatHandler::input_time_now` that compositors could implement to produce a timestamp appropriate for the backend being used. But it seemed unnecessarily complicated since I can't think of any way that would be implemented other than just using `CLOCK_MONOTONIC`. Except maybe for a custom input backend that for some reason uses different timestamps. (Also `libei` generated events will still use `CLOCK_MONOTONIC`).

If for some reason we do decide that's needed, a dedicated `InputTime::now()` function makes it easy; we can simply remove it then it will be obvious that we need to change to some other method in all the places using it. I don't expect any compositors using smithay will need that though.

YaLTeR:
Haven't looked in *too* muchdetail, but looks reasonable to me

Drakulix (approved):
LGTM, but this definitely needs a CHANGELOG entry.

ids1024:
Updated to add changelog entry.

Drakulix (approved):

PR #2109, opened 2026-07-24
fix: flush ei connection after adding/removing a device

hojjatabdollahi:
## Description
If we add/remove devices on an ei_connection we should flush the connection so the client is notified.

Noticed this when my client didn't see the output scale changing.

## Checklist


PolyMeilex (approved):
LGTM

Issue #2110, opened 2026-07-27
Feature request: A new feature for removing sha2 dependencies


jaewan-github:
`sha2` brings a huge number of dependencies while `smithay` only uses it for generating an unique key.

Here's the required dependencies.

```
$ cargo tree
smithay v0.7.0
├── sha2 v0.10.9
│   ├── cfg-if v1.0.4
│   ├── cpufeatures v0.2.17
│   └── digest v0.10.7
│       ├── block-buffer v0.10.4
│       │   └── generic-array v0.14.7
│       │       └── typenum v1.20.1
│       │       [build-dependencies]
│       │       └── version_check v0.9.5
│       └── crypto-common v0.1.7
│           ├── generic-array v0.14.7 (*)
│           └── typenum v1.20.1
...
```

It's a burden for building smithay in restricted environment where all dependencies are strictly managed.
For example, Android framework wants to limit the crypto library but sha2 adds a new one.

It would be better if we can conditionally remove the dependency, especially because it now has `#[allow(deprecated)]`.

One idea is to use built-in hash to generate the unique key. Let me propose a PR with it.

PR #2111, opened 2026-07-27
Add new feature use_sha for opt-in sha2 dependency

jaewan-github:
Introduce a new opt-in feature 'use_sha' to enable the 'sha2' dependency and use `Sha256` for keymap hashing.

When 'use_sha' is disabled (default), a custom hash function using built-in `DefaultHasher` is used instead, without 'sha2' dependency.

BREAKING CHANGE: If you want to keep the old behavior (using Sha256), you must explicitly enable the 'use_sha' feature. This breaking change is necessary because Cargo does not support excluding dependencies when a feature is enabled (features are strictly additive), so the dependency must be opt-in.

Tested:
- cargo test --lib --no-default-features --features wayland_frontend (default, using builtin hash)
- cargo test --lib --no-default-features --features wayland_frontend,use_sha (using Sha256)
- Added 'test_keymap_file_creation' test to verify KeymapFile works.

## Description

This adds a new `use_sha` feature and change default keymap id generation with chaining `DefaultHasher`.
To keep the old behavior explicit feature `use_sha` is required.

### Note

This introduces breaking change. Tell me if you prefer not doing this, or completely remove the sha2 dependency

### Motivation
sha2 crates brings huge dependency while smithay only uses it for generating good hash value.

It would be helpful to provide an option to opt-in the sha2 crate usage.

### Existing Issue

https://github.com/Smithay/smithay/issues/2110

I created the issue with bit more details.

### Disclaimer about the AI uses

I haven't used AI for writing CL, but only used to proofread my commit message.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,7 +72,6 @@ smallvec = "1.11"
 pixman = { version = "0.2.1", features = ["drm-fourcc", "sync"], optional = true }
 aliasable = { version = "0.1.3", optional = true }
 portable-atomic = { version = "1", features = ["float"] }
-sha2 = "0.10.9"
 tracy-client = { version = "0.18.4", default-features = false, optional = true }
 reis = { version = "0.7.0", features = ["calloop"] }
 glam = { version = "0.33.2", default-features = false, features = ["std"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,7 +72,7 @@ smallvec = "1.11"
 pixman = { version = "0.2.1", features = ["drm-fourcc", "sync"], optional = true }
 aliasable = { version = "0.1.3", optional = true }
 portable-atomic = { version = "1", features = ["float"] }
-sha2 = "0.10.9"
+sha2 = { version = "0.10.9", optional = true }
 tracy-client = { version = "0.18.4", default-features = false, optional = true }
 reis = { version = "0.7.0", features = ["calloop"] }
 glam = { version = "0.33.2", default-features = false, features = ["std"] }
@@ -116,6 +116,7 @@ wayland_frontend = ["wayland-server", "wayland-protocols", "wayland-protocols-wl
 x11rb_event_source = ["x11rb"]
 xwayland = ["encoding_rs", "wayland_frontend", "x11rb/composite", "x11rb/xfixes", "x11rb/randr", "x11rb_event_source", "scopeguard"]
 test_all_features = ["default", "use_system_lib", "renderer_glow", "renderer_test"]
+use_sha = ["dep:sha2"]
 
 [[example]]
 name = "minimal"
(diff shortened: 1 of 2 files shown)


kelnos:
Not sure what you mean about the feature situation.  `use_sha` could be in default features, and then if you don't want it, pass `--no-default-features` (and then explicitly pass the features you do want).  Or are you saying that even that won't remove the dependency?  If so, that feels like a cargo bug...

jaewan-github:
> Not sure what you mean about the feature situation. `use_sha` could be in default features, and then if you don't want it, pass `--no-default-features` (and then explicitly pass the features you do want). Or are you saying that even that won't remove the dependency? If so, that feels like a cargo bug...

Your understanding is correct. I'll follow your suggest to make `use_sha` as default.

jaewan-github:
Done as you suggested. It would be a great help if you take another look on this.

ids1024:
We might not need a `use_sha2` feature here, if this method of hashing is effective for preventing unintentional collisions. I used `sha2` since SipHash technically shouldn't be sufficient for that, and `sha2` just seemed to be the most common hash crate (and was already in `cosmic-comp`'s dependency tree).

Technically maybe a maliciously constructed keymap could be sent with `virtual-keyboard-unstable-v1`, but I'm not sure what that would achieve, and a client using that is already privileged.

jaewan-github:
Thank you for the suggestions. Uploaded another revision as you suggested.

More comments are inlined.

> We might not need a use_sha2 feature here, if this method of hashing is effective for preventing unintentional collisions.

I believe that this is enough to prevent unintentional collisions, and provides the same level collision resistance with SHA-256.

If you're fine with removing sha2 usages, I'll completely remove it. Please take another look.

> I used sha2 since SipHash technically shouldn't be sufficient for that, and sha2 just seemed to be the most common hash crate (and was already in cosmic-comp's dependency tree).

Well, not sure what whether `smithay` is depending on `cosmic-comp` (i.e. `cargo tree` doesn't have it).
Probably it's removed?

> Technically maybe a maliciously constructed keymap could be sent with virtual-keyboard-unstable-v1, but I'm not sure what that would achieve, and a client using that is already privileged.

Agreed. It doesn't need to be cryptographically secure.

jaewan-github:
Dear @ids1024, is there anything missing from my side?

It would be a great help  if you take a look at this again.

jaewan-github:
@ids1024 Gentle reminder.

jaewan-github:
Another reminder here:

If this approach doesn't look right or simple, I can switch to other approaches such as:
  - Use more simpler hash crates with less dependencies.
  - Just use built-in hash function, and use u64 as a key.

jaewan-github:
Dear @ids1024 , would you take another look here? I updated the PR as we discussed at the chatroom.

ids1024 (approved):
This looks good now. Thanks!

jaewan-github:
Thank you so much for spending your time on this.

PR #2115, opened 2026-07-28
fix(ei): add regions for touch support

hojjatabdollahi:
## Description
Adds ei region for ei touch. And unifies the ei regions for pointer and touch.

Depends on:
- [ ] https://github.com/ids1024/reis/pull/33

## Checklist

### DIFF
--- a/src/backend/libei/mod.rs
+++ b/src/backend/libei/mod.rs
@@ -21,7 +21,7 @@
 //!                 let _ = seat.add_keyboard("virtual keyboard", XkbConfig::default());
 //!                 seat.add_pointer("virtual pointer");
 //!                 seat.add_pointer_absolute("virtual absolute pointer", &[]);
-//!                 seat.add_touch("virtual touch");
+//!                 seat.add_touch("virtual touch", &[]);
 //!             }
 //!             EiInputEvent::Disconnected => {}
 //!             EiInputEvent::Event(event) => {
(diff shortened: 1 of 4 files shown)


src/backend/libei/mod.rs
```diff
@@ -299,7 +299,10 @@ fn convert_request(request: EisRequest) -> Option<InputEvent<EiInput>> {
         EisRequest::TouchMotion(event) => Some(InputEvent::TouchMotion { event }),
         EisRequest::TouchCancel(event) => Some(InputEvent::TouchCancel { event }),
         EisRequest::DeviceClosed(event) => Some(InputEvent::DeviceRemoved { device: event.device }),
-        EisRequest::Frame(_) => None,
+        EisRequest::Frame(event) => event
+            .device
+            .has_capability(reis::request::DeviceCapability::Touch)
+            .then_some(InputEvent::TouchFrame { event }),
```
ids1024, line 311: Hm, `ei_device.frame` isn't specific to touch input. Should we somehow restrict this to only frames that include at least one touch event?

hojjatabdollahi: AFAIK devices are created per-capability, so touch gets its own device and reis already drops empty frames.

But I think pending_requests could cause issues since it's one queue per connection, not per device, so a frame on the touch device can commit another devices's events. I have been looking into that today, so I understand it better.

hojjatabdollahi: Alright, I'm not 100% sure about my solution.

ids1024: Ah, I guess this is actually a `reis` bug then. It looks like `eis_queue_event` in libeis uses a `device->pending_event_queue`.

ids1024: Since smithay's abstraction doesn't add more than one capability to the touch device, I guess nothing needs to be done for this here if `reis` used a per-device queue.

hojjatabdollahi: Yeah, there is also a worse symptom, one frame drains the whole queue, so if two devices are framed, the second one is silently dropped.

Should I remove my second commit, and wait for an update to reis?

hojjatabdollahi: I went ahead and opened a PR, in my tests, it sends the frames when it should and drops unframed events. Leave the rest up to you.

ids1024:
I've released reis 0.7.1.

ids1024 (approved):

PR #2117, opened 2026-07-31
Add Output::owns_xdg_output()

kelnos:
## Description

Returns true if the passed `ZxdgOutputV1` protocol instance is owned by the `Output`.

## Checklist


kelnos:
In the end I changed something in my design, so I don't actually need this anymore, but feel free to merge if you think it might be generally useful.  (Or close it, that's fine too :) )

Drakulix (approved):
This seems good and useful. Thanks!

PR #2118, opened 2026-08-02
Add remove_xsettings() and clear_xsettings() to X11Wm

kelnos:
## Description

Noticed you can add xsettings values, but not remove or clear them.  I need to be able to remove them when the corresponding setting is deleted from our configuration store (something a user can do), so that applications revert to the default for that setting.

## Checklist


Drakulix (approved):
thanks!

PR #2120, opened 2026-08-03
fix: clear stale outputs in Space::refresh first

RipleyTom:
## Description
Smithay could emit surface.enter events with removed wl_output. This was encountered in mpv. turning the monitor off and back on would crash mpv with a nullpointer dereference as the enter events was forwarded by libwayland with a nullptr for the wl_output(as it was sent a zombie value, libwayland replaces those with nullptr).

e.outputs was cleared only after the enter were emitted and some(specifically https://github.com/Smithay/smithay/blob/master/src/desktop/space/wayland/window.rs#L39 ) calls self.refresh() => output_update => output.enter on a stale output.

This just reorders things, making sure the stale outputs in the elements are removed first before checking if enter events need to be sent.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -216,6 +216,9 @@ CRTC.
 XWayland now honors WM_HINTS request to map the window in IconicState(minimized). Compositors should
 check X11Surface::is_hidden() at mapping time and minimize the window appropriately.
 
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
thanks for tracking this down!

PR #2121, opened 2026-08-05
small blur fixes

hojjatabdollahi:
## Description

Before: update to one window, causes update to the panel, which then causes update to the windows touching that area. (damage moves up one layer to the panel and then down to all the windows below it, maybe a video in floating mode would've made more sense)

https://github.com/user-attachments/assets/ec0b5a05-c5a3-4ebd-8278-65a673439cd2

After:
update to one window only causes update to the panel

https://github.com/user-attachments/assets/eaee7fd2-ece4-411d-8164-c6c5f30764f1

## Checklist

### DIFF
--- a/src/backend/renderer/damage/mod.rs
+++ b/src/backend/renderer/damage/mod.rs
@@ -690,6 +690,7 @@ impl OutputDamageTracker {
         for (z_index, element) in render_elements
             .iter()
             .enumerate()
+            .rev()
             .filter(|(_, e)| e.is_framebuffer_effect())
         {
             let damage_index = if force_effect_redraw {
--- a/src/backend/renderer/damage/mod.rs
+++ b/src/backend/renderer/damage/mod.rs
@@ -659,7 +659,7 @@ impl OutputDamageTracker {
         element_damage.extend_from_slice(&self.last_state.opaque_regions);
         element_damage =
             Rectangle::subtract_rects_many_in_place(element_damage, self.opaque_regions.iter().copied());
-        if !element_damage.is_empty() {
+        if element_damage.iter().any(|rect| !rect.is_empty()) {
             force_effect_redraw = true;
         }
         self.damage.extend_from_slice(&element_damage);


src/backend/renderer/damage/mod.rs
```diff
@@ -659,7 +659,7 @@ impl OutputDamageTracker {
         element_damage.extend_from_slice(&self.last_state.opaque_regions);
         element_damage =
             Rectangle::subtract_rects_many_in_place(element_damage, self.opaque_regions.iter().copied());
-        if !element_damage.is_empty() {
+        if element_damage.iter().any(|rect| !rect.is_empty()) {
```
Drakulix, line 662: This looks sensible.

src/backend/renderer/damage/mod.rs
```diff
@@ -690,6 +690,7 @@ impl OutputDamageTracker {
         for (z_index, element) in render_elements
             .iter()
             .enumerate()
+            .rev()
```
Drakulix, line 693: This doesn't, as this would change all the indices to go in reverse. Can you explain why you think this is necessary?

hojjatabdollahi: It's supposed to check if anything below an element changed, and adds full damage to anything above. But it's going in order (front to back) so if a window in the front reports damage windows in the back will also get full damage.

This video shows you what I mean. I have a big window in the back, cmatrix in the middle creating change, and a window above it. Only the window above it is supposed to report full damage, but you can see the window behind it will also report full damage, because we're going front to back.


https://github.com/user-attachments/assets/569c37e7-53ca-4fe6-a4cd-1fa6136de0d3

hojjatabdollahi: And with `.rev()`:

https://github.com/user-attachments/assets/a6b8602a-22fb-4d98-83b3-d17f4b34662a

YaLTeR:
Haven't looked into this in detail, but during initial framebuffer effect development we found a bunch of edge cases where moving an opaque window on top of the blurred window required redrawing the contents *behind* the blurred window afterwards, e.g. https://github.com/Smithay/smithay/pull/1910#issuecomment-3845835487

So this damaging may be required for correct operation.

src/backend/renderer/damage/mod.rs
Drakulix: > It's supposed to check if anything below an element changed, and adds full damage to anything above. But it's going in order (front to back) so if a window in the front reports damage windows in the back will also get full damage.

Oh, I think I see it now. We use the z_index to filter window damage above the effect, but since we add to `self.damage`, the next iteration will pick up the blur-damage of the previous effect even though it is above the element.

But the rev happens after enumerate, meaning we don't break the z_index. Looks fine to me then.

Drakulix:
> Haven't looked into this in detail, but during initial framebuffer effect development we found a bunch of edge cases where moving an opaque window on top of the blurred window required redrawing the contents _behind_ the blurred window afterwards, e.g. [#1910 (comment)](https://github.com/Smithay/smithay/pull/1910#issuecomment-3845835487)

This only applied to opaque windows, this case is about multiple framebuffer effects on top of each other and I think @hojjatabdollahi is right and we do overdraw a bit in this case.

Drakulix (approved):
LGTM. But I would like a second review.

Ping @YaLTeR, @cmeissl or @ids1024 maybe?

Drakulix:
I have been running this locally for a while without noticing any render corruption in cosmic-comp at least. I'll merge it to avoid bit-rot. We can always revert it, should anybody find issues with it my review didn't.

PR #2129, opened 2026-08-11
anvil: Fix deadlock in remove_constraint()

YaLTeR:
## Description

Fixes a deadlock, see https://github.com/Smithay/smithay/pull/2107#issuecomment-5249816540

## Checklist


Drakulix (approved):
Hmm, it would be nice, if we could break the deadlock in smithays abstractions, but this looks good either way.

PR #2132, opened 2026-08-11
Send enter/down/proximity_in to late-bound pointer/touch/tablet_tool

### LINKED ISSUE #2131: Client that does seat.get_pointer() or get_touch() late doesn't get wl_pointer.enter()
If a client shows the surface on screen, the compositor does pointer enter/motion on top of it, and *then* the client does wl_seat.get_pointer(), then the client won't get enter() on that pointer, and will get motion() right away. This is incorrect. sctk for example will filter out those motion events with a warning.

The same issue affects wl_touch (spawn window, touch down, then client binds wl_touch, doesn't get wl_touch.down() but gets motion()).

The same issue affects tablet tools (spawn window, proximity in, then client binds tablet tool, doesn't get proximity in but gets motion).

Keyboard is not affected.

Reproduces on Anvil. Test client for all of these: https://github.com/YaLTeR/wl-test-clients/blob/main/src/delayed_get_input_devices.rs

YaLTeR:
## Description

Resolves https://github.com/Smithay/smithay/issues/2131.

Sending the rest of the state is left as an exercise for the future readers. As well as pointer button I guess.

The alternative is to not bother sending anything, but then we need to gate all motion() and other events behind "did this object enter() before", which sounds more error prone?

Both `same_client_as()` and `wl_surface()` are checked because apparently in Anvil's impl of `WaylandFocus`, when aiming at an SSD, it's possible to get `true` from the former and `None` from the latter.

## Checklist

### DIFF
--- a/src/input/pointer/mod.rs
+++ b/src/input/pointer/mod.rs
@@ -702,7 +702,7 @@ impl<D: SeatHandler + 'static> PointerInnerHandle<'_, D> {
 pub(crate) struct PointerInternal<D: SeatHandler> {
     pub(crate) focus: Option<(<D as SeatHandler>::PointerFocus, Point<f64, Logical>)>,
     pending_focus: Option<(<D as SeatHandler>::PointerFocus, Point<f64, Logical>)>,
-    location: Point<f64, Logical>,
+    pub(crate) location: Point<f64, Logical>,
     grab: GrabStatus<dyn PointerGrab<D>>,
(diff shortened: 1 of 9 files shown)


Drakulix (approved):
Seems good to me. I am not sure if we really need to bother sending an action like an already pressed pointer button, but enter/proximity_in/down is definitely useful.

PR #2135, opened 2026-08-12
Update rand dep to 0.10

kelnos:
## Description

There appear to be no changes needed to smithay API-wise, and I don't see anything in https://rust-random.github.io/book/update-0.10.html that needs to be updated behavior-wise.  0.10's MSRV is 1.85.0, which is still below smithay's.

## Checklist


Drakulix (approved):
thx!

PR #2136, opened 2026-08-12
Update frame throttling time unconditionally

YaLTeR:
## Description

Before, normal (unthrottled) frame callbacks didn't update the stored time, meaning that the throttled frame callbacks were firing unconditionally on their timer, even if the surface was otherwise continuously receiving frame callbacks.

## Checklist


Drakulix (approved):
Thanks!

PR #2137, opened 2026-08-12
input/pointer: Support `wl_pointer.warp` in `MotionEvent`

ids1024:
## Description
Based on https://github.com/Smithay/smithay/pull/2108.

`cosmic-comp` needs to be updated with this to test properly with pointer constraints, then it needs testing in `XWayland`/etc. This is meant to replace the `motion`/`relative_motion` code in https://github.com/pop-os/cosmic-comp/commit/dda2d14832.

As far as I understand, it should be correct to use `.motion` on older versions of the protocol, and `.warp` on newer versions. If the `.relative_pointer` event with a delta of 0 is just a workaround for the lack of a "warp" in earlier versions of the protocol, it presumably isn't necessary to send with newer versions? In which case, handling here seems appropriate.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ wayland-egl = { version = "0.32.11", optional = true }
 wayland-protocols = { version = "0.32.13", features = ["unstable", "staging", "server"], optional = true }
 wayland-protocols-wlr = { version = "0.3.12", features = ["server"], optional = true }
 wayland-protocols-misc = { version = "0.3.12", features = ["server"], optional = true }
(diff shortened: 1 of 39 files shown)


YaLTeR:
Hm. What about cases like the new logic in ClickGrab that updates the focus start data? I guess it doesn't need any explicit handling because if the compositor changes the location of the surface without a pointer event, it will send `is_warp: true`, which will pass through ClickGrab correctly, and if it's coincided with an input event, then the compositor will send `is_warp: false`, which will also be correct?

PR #2139, opened 2026-08-13
renderer: add `Renderer::invalidate_caches` and clean up multi-GPU texture caches on every device

minyek:
## Summary

Companion PR to pop-os/cosmic-comp#2500, which makes cache cleanup event-based.
`Renderer::invalidate_caches` lets a compositor drop cached imports it knows
won't be needed again soon; alongside it are leak fixes in `MultiRenderer` and
`GpuManager`.

## What's Changed

1. **New `Renderer::invalidate_caches`** — `cleanup_texture_cache` only frees
   resources whose source buffer is already gone, so a renderer that draws
   infrequently keeps every cached import (and its GPU memory) pinned between
   renders. `invalidate_caches` discards every cached import and bind target
   unconditionally; the trait method defaults to `cleanup_texture_cache`,
   keeping external implementors source-compatible. Caches held outside the
   renderer (e.g. per-surface textures in surface user-data) are unaffected.

2. **`MultiRenderer::cleanup_texture_cache` skipped the "other" devices** — it
   visited only the render and target device. Buffers that can't be imported on
   the render node are imported on their source device instead, so those devices
   accumulated `dmabuf_cache` entries and queued GL destructions that nothing
   ever released. Now every device is visited, best-effort, so one failing
   device can't leave the rest uncleaned.

3. **New `GpuManager::cleanup_texture_cache` / `invalidate_caches`** — outside a
   draw a compositor holds a `GpuManager`, not a `MultiRenderer`, so clearing
   caches across devices meant hand-rolling a loop over `devices_mut`. Such a
   loop can't reach the render→target copy buffers, which the manager keys by
   node pair — one full output-sized dmabuf per pair, left pinned.
   `GpuManager::invalidate_caches` clears those along with every device's caches.

   cosmic-comp currently works around this by iterating `devices_mut` itself at
   end-of-draw (f86cd933); this commit supersedes that workaround.

### Error handling across devices

All four multi-device methods are best-effort: every device is attempted even
if one fails, each failure is logged with its node, and the first error is
returned. The `MultiRenderer` methods visit the render node first, so its
failure — the one that breaks the next frame — takes precedence.

Smithay has no established convention here: `GpuManager::early_import` and
`import_surface_tree` keep the last error and log nothing, while
`DrmCompositor` discards the result outright. Happy to align with whichever
convention the maintainers prefer.

## Testing

Three pixman unit tests pin the retention-vs-invalidation contract that live entries survive `cleanup_texture_cache`,
dead ones don't, and live ones are dropped by `invalidate_caches`.

The `GpuManager` methods were also exercised end-to-end on an instrumented
cosmic-comp build that calls `GpuManager::invalidate_caches` after each
infrequent main-thread render (dual-output NVIDIA system). Across three
output-reconfiguration cycles the cycled output's swapchain advanced from
generation 4 to 112 with live slots pinned at 4 — every superseded
generation's imports were released — and the renderer's cleanup queues drained
at every census with no cache holding a dead entry.

Not covered by that run, verified by inspection and compilation only:

- **The multi-GPU path** — the machine has a single GPU, so `other_renderers`
  is always empty.
- **The error paths** — no device failed during the run.

`cargo test --features test_all_features` passes (74 unit, 96 doctests),
clippy is clean on the touched files, and each commit builds independently.

## AI assistance

Per Smithay's [AI policy](https://github.com/Smithay/smithay/blob/master/AI.md),
AI assistance (Claude Code) is disclosed in each commit message; the author
understands the changes and can respond to review.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -219,6 +219,11 @@ Added `Renderer::invalidate_caches`, which unconditionally drops all renderer-in
 including entries `Renderer::cleanup_texture_cache` retains while their source buffers are still
 alive.
 
+Added `GpuManager::cleanup_texture_cache` and `GpuManager::invalidate_caches`, which apply the
+corresponding `Renderer` method to every enumerated device. `GpuManager::invalidate_caches`
+additionally drops the buffers cached for copying between a render and a target node, which no
+`Renderer` owns.
+
 ### Bugfixes
 
 `SimpleCrtcMapper` (in `smithay-drm-extras`) now releases the CRTC reservation of any connector that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -237,6 +237,12 @@ are emitted with a stale wl_output anymore.
 The XWayland WM implementation previously incorrectly prefixed large transfers from X11 windows with
 the four INCR bytes.
 
+`MultiRenderer::cleanup_texture_cache` now also cleans up the devices other than the render and
+target device. Buffers that cannot be imported on the render node directly are imported on their
+source device, so those devices accumulate cached imports that previously were never released.
+Cleanup is now also attempted on every device even if it fails on one of them, with every failure
+logged and the first error returned.
+
 ## 0.7.0
 
 ### Breaking changes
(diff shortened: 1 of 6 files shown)


Drakulix (approved):
LGTM, thanks

PR #2143, opened 2026-08-17
foreign-toplevel-list: allow compositor to pick the toplevel identifier.

Ph4ntomas:
## Description
Smithay generates the identifier using a random 32 character alphanumeric string.

This is generally fine, although it doesn't check for collisions. There are however two issue with having Smithay generate the id on toplevel_handle creation:
- A compositor wanting a special generation scheme must re-implement the whole protocol instead of the generation only.
- Since `ext_foreign_toplevel_handle_v1` are meant to represent a mapped toplevel, it's not possible for a compositor to make use of the identifier before the toplevel is mapped (e.g. for configuration-as-code purpose) without risking weird edge cases. As it is, the compositor either need to create the toplevel handle before the toplevel is mapped (and risk weird ordering of operation) , or re-implement the protocol.

This PR address both by providing an API for compositors to specify the identifier on handle creation.

## Checklist

### DIFF
--- a/src/wayland/foreign_toplevel_list/mod.rs
+++ b/src/wayland/foreign_toplevel_list/mod.rs
@@ -128,11 +128,16 @@ pub struct ForeignToplevelHandle {
 }
 
 impl ForeignToplevelHandle {
-    fn new(title: String, app_id: String, instances: Vec<Weak<ExtForeignToplevelHandleV1>>) -> Self {
+    fn new(
+        title: String,
+        app_id: String,
+        identifier: String,
+        instances: Vec<Weak<ExtForeignToplevelHandleV1>>,
+    ) -> Self {
         Self {
             inner: Arc::new((
                 Mutex::new(ForeignToplevelHandleInner {
-                    identifier: Alphanumeric.sample_string(&mut rand::rng(), 32),
+                    identifier,
                     title,
                     app_id,
                     instances,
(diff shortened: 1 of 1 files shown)


src/wayland/foreign_toplevel_list/mod.rs
```diff
@@ -337,9 +342,38 @@ impl ForeignToplevelListState {
...
+        &mut self,
+        title: impl Into<String>,
+        app_id: impl Into<String>,
+        identifier: impl Into<String>,
+    ) -> ForeignToplevelHandle
+    where
+        D: ForeignToplevelListHandler + Dispatch<ExtForeignToplevelHandleV1, ForeignToplevelHandle>,
+    {
+        let identifier = identifier.into();
+        debug_assert!(
```
YaLTeR: Maybe move len check second, and make this a non-debug assert? Should be cheap I think

Ph4ntomas: Yeah I agree, the bound check is cheaper, and then once there's only 32 ascii char to check it should be cheap as well. I wonder if `is_ascii()` should be replaced by a proper validation function, too.

From the protocol:
> An identifier is a string that contains up to 32 printable ASCII bytes. 

I could swap `is_ascii` by a loop to check all char are `is_ascii_graphic()`. OTOH, this is just a sanity check, as compositor that want to use this should already ensure they only uses valid characters.

Drakulix (approved):
LGTM. Thanks!

PR #2144, opened 2026-08-25
wayland/pointer_constraints: reorganize call sequence to resolve deadlock issue

skygrango:
## Description

This avoids deadlocks while restoring the interface requirements for `PointerConstraintsHandler::remove_constraint`.

## Checklist

### DIFF
--- a/src/wayland/seat/pointer.rs
+++ b/src/wayland/seat/pointer.rs
@@ -27,7 +27,7 @@ use crate::{
     utils::{Client as ClientCoords, Point, Serial, iter::new_locked_obj_iter_from_vec},
     wayland::{
         Dispatch2, compositor,
-        pointer_constraints::{PointerConstraintsHandler, with_pointer_constraint},
+        pointer_constraints::{ConstraintRemove, PointerConstraintsHandler, with_pointer_constraint},
     },
 };
 
@@ -277,11 +277,21 @@ where
             pointer.wp_pointer_gestures.leave::<D>(self, serial, time);
             pointer.wl_pointer.leave(self, serial, time);
 
-            with_pointer_constraint(self, &pointer, |constraint| {
+            if let Some(region) = with_pointer_constraint(self, &pointer, |constraint| {
                 if let Some(constraint) = constraint {
-                    constraint.deactivate(data, self, &pointer);
+                    if constraint.is_active() {
+                        let region = constraint.region().cloned();
+                        constraint.deactivate();
+                        Some(region)
+                    } else {
+                        None
+                    }
+                } else {
+                    None
                 }
-            });
+            }) {
+                data.remove_constraint(self, &pointer, ConstraintRemove::PointerLeave(region));
+            }
         }
 
         compositor::with_states(self, |states| {
--- a/anvil/src/state.rs
+++ b/anvil/src/state.rs
@@ -65,7 +65,8 @@ use smithay::{
         },
         output::{OutputHandler, OutputManagerState},
         pointer_constraints::{
-            PointerConstraint, PointerConstraintsHandler, PointerConstraintsState, with_pointer_constraint,
+            ConstraintRemove, PointerConstraint, PointerConstraintsHandler, PointerConstraintsState,
+            with_pointer_constraint,
         },
         pointer_gestures::PointerGesturesState,
         presentation::PresentationState,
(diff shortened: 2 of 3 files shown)


Drakulix (approved):
LGTM, but CI isn't happy

skygrango:
@Drakulix  Thank you for the reminder, I fixed anvil.

@YaLTeR  I revert your changes and rolled back `PointerConstraintsHandler::remove_constraint` to its original state, so you might want to double-check that.

skygrango:
Since the deadlock issue is more serious than the API change, this can be merged as soon as possible if necessary

src/wayland/seat/pointer.rs
```diff
@@ -277,11 +277,19 @@ where
...
             });
+
+            if is_constraint_active {
+                // we are fine with calling `remove_constraint` first, and then actually performing `deactivate`.
+                data.remove_constraint(self, &pointer);
+                with_pointer_constraint(self, &pointer, |constraint| {
+                    if let Some(constraint) = constraint {
+                        constraint.deactivate();
+                    };
+                });
```
YaLTeR: This is a bit weird because if the compositor code inside remove_constraint() causes a different, new constraint to activate (which is not entirely unlikely given that the compositor is supposed to warp the pointer there), then this next line will immediately deactivate it. A sort of TOCTOU problem

skygrango: hmm, I think `new_constraint` should be impossible to be called between these two?

YaLTeR: hm maybe the data.remove_constraint() can just be moved to below deactivating?

skygrango: You cannot do this directly, because `PointerConstraint` must be provided to the compositor.

I have another [approach](https://github.com/skygrango/smithay/commit/c3a8e1fc976babf80f1e1284c477fbf48590eef4), but I think the current PR is simpler and clearer.

YaLTeR: I'm not sure I understand. You mean the compositor should be able to get the constraint inside remove_constraint()? But it already cannot in the destroyed case I believe since it's removed earlier

skygrango: If you want to simply ignore `cursor_position_hint` with region of constraint, this could work. but I'm not sure if we should do that.

YaLTeR: I'm still not sure what you mean. It's late here so I might not be thinking clearly. The remove_constraint() handler in anvil doesn't use the current constraint. Why can't the `data.remove_constraint(&pointer)` line be moved below the `with_pointer_constraint()` call?

skygrango: This is where the compositor actively deactivate the constraint; I think client should not have reached the point of issuing the destroy request yet. but we can indeed apply the final `cursor_position_hint` without region of constraint

skygrango: I misunderstood it a bit just now, but your idea does work. I'm just not sure whether it could break some existing behavior, so I'm trying to provide the compositor with as much information as possible to make the behavior more consistent.

YaLTeR: Yeah, that's what I had in mind. I can't test it for now so perhaps someone else could double check that my reasoning is ok

skygrango: I updated commit, which should help address your concerns about the TOCTOU problem.

skygrango: I have a clearer picture of the behavioral issue now, so I’d like to explain the reasoning behind the possible approaches.

First, `cursor_position_hint` is behavior specific to the pointer-locked state. If the pointer is unexpectedly unlocked by the forced deactivation we perform here, the most likely reason is that pointer focus has been taken over for some external reason.

This raises an important question: **where should the cursor be positioned when the pointer lock is broken?**

With our current implementation, the cursor remains at the position where it was when the pointer was locked. However, applications can handle `cursor_position_hint` in different ways:

* Some applications use `cursor_position_hint` to reposition the cursor to the center of the window.
* Some continuously update the cursor position according to their own requirements.
* Some provide their own software cursor and handle the cursor position entirely themselves.
* Xwayland is particularly interesting because it continuously updates the cursor coordinates through `set_cursor_position_hint`.

Given this diversity, there are three reasonable behaviors when an unexpected focus takeover causes the pointer lock to be broken:

1. **Ignore `cursor_position_hint`.**
   After unlocking, leave the cursor at the position it had before the pointer was locked.

2. **Apply the last `cursor_position_hint`.**
   Trust the position most recently provided by the client and move the cursor there when the lock is released.

3. **Move the cursor to the center of the screen.**
   This is a variation of the first approach, specifically intended to handle unexpected focus loss. Re-centering the cursor could make it easier for the user to locate it immediately after the focus transition.

If we look at KWin's `breakPointerConstraints` behavior, it appears to follow the first approach: after breaking the pointer lock, it preserves the cursor position from before the lock was established.

Therefore, I think simply calling `deactivate` before `remove_constraint` at this point is the appropriate approach for now. Although strictly following the specification is one possible approach, KWin's behavior may provide a more consistent user experience in practice.

skygrango:
I added `ConstraintRemove` so that developers can handle constraint removal based on the reason it was removed. This ensures that when the client explicitly requests the constraint to be destroyed, the compositor can correctly check  region.

YaLTeR:
> I added `ConstraintRemove` so that developers can handle constraint removal based on the reason it was removed. This ensures that when the client explicitly requests the constraint to be destroyed, the compositor can correctly check region.

At a glance I think this looks good. Would be nice for someone else to check

skygrango:
> At a glance I think this looks good. Would be nice for someone else to check

I still need to make some adjustments, because according to the approach we originally agreed upon, the `cursor_position_hint` should be applied on `PointerLeave`

skygrango:
> I still need to make some adjustments, because according to the approach we originally agreed upon, the `cursor_position_hint` should be applied on `PointerLeave`

Oh, we just need to make sure `cursor_position_hint` is cleared on `PointerLeave`.

That should be enough for now.

skygrango:
I added `Option<RegionAttributes>` to `PointerLeave` so that if we decide to apply `cursor_position_hint` here in the future, we will not need to make another change to this part.

Drakulix (approved):
Seems to work nicely with cosmic-comp now and the code also looks reasonable. Gonna merge this, thanks :)

PR #2145, opened 2026-08-25
fix(drm/gles): survive GPU resets — detect context loss, keep dmabufs valid, stop the log flood

gustavx404:
Thanks for the work and the detailed description — and sorry for the noise.

After the last update, everything is already working well on my setup with `COSMIC_COMP_GPU_RESET_KILL_CLIENTS=1`, which kills clients on GPU reset instead of trying to keep the session alive. That’s been enough for my use, so I’m not actually validating the full recovery path implemented here.

I’m closing this PR to avoid blocking review or implying active validation from my side. The changes still look useful for anyone pursuing true session-surviving GPU reset recovery; I hope they can be picked up in that context.

Thanks again, and apologies for the churn.


ToxicMushroom:
taps the sign https://github.com/Smithay/smithay/blob/master/AI.md

PR #2146, opened 2026-08-28
input/dnd: add Source::accepted so proxying sources can relay the target's choice

dahankzter:
`Source` has hooks for the other two target->source requests — `send` for `receive`, `choose_action` for `set_actions` — but nothing for `accept`. `handle_dnd` sets `data.accepted` and that's the end of it.

That breaks proxying sources. A nested compositor forwarding a drag to its own clients implements `Source`, and the host compositor only sends `drop` once that nested compositor has accepted on the *upstream* offer. It currently has no way to know when to do that, so the drop never arrives and the drag just ends in `leave`.

This adds `Source::accepted(Option<String>)`, called from the `Accept` arm. Default no-op, so nothing existing changes.

Found while fixing drag-and-drop onto cosmic-panel applets, which is broken for exactly this reason: pop-os/cosmic-panel#675, which uses this hook and is tested working against it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DCQQAWhxyYJwT49Zvcko6z

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,12 @@ operations (see below).
 types, which can be targets of DnD operations. `Source` is a new trait for types that represent
 sources of data for DnD operations.
 
+`Source::accepted` is called when a target accepts a mime-type, or rejects the offer with `None`.
+Sources that proxy an upstream offer need it: a nested compositor forwarding a drag to its own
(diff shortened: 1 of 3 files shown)


Drakulix (approved):
Thanks. I don't see any issues with this approach. :)

PR #2148, opened 2026-08-29
renderer/gles: disable fence export after runtime failure

krissedout:
## Description

- disable ExportFence after EGLFence::create fails at runtime
- use the existing synchronized glFinish fallback for the failed frame and subsequent frames
- avoid retrying a broken eglCreateSync call on every rendered frame

Some context:
GlesRenderer enables ExportFence when the required GL extension is advertised. Some drivers advertise support but still fail when Smithay attempts to create the EGL fence, occasionally without setting an EGL error.

export_sync_point currently returns None after that failure but retains the capability, causing the same failing EGL call and warning to repeat every frame.

Once fence creation fails for a renderer context, this change removes ExportFence from that renderer’s capabilities. Smithay then follows its existing glFinish fallback path, preserving synchronization and correctness without repeated EGL calls.

## Checklist

### DIFF
--- a/src/wayland/shell/xdg/mod.rs
+++ b/src/wayland/shell/xdg/mod.rs
@@ -436,6 +436,8 @@ xdg_role!(
     /// for the xdg_popup state to take effect.
     #[derive(Debug)]
     XdgPopupSurfaceRoleAttributes {
+        active: bool,
+
         /// Holds the parent for the xdg_popup.
         ///
         /// The parent is allowed to remain unset as long
@@ -1971,6 +1973,9 @@ impl PopupSurface {
                 .unwrap()
                 .lock()
                 .unwrap();
+            if !role.active {
+                return Ok(None);
+            }
             if role.parent.is_none() {
                 return Err((
                     xdg_surface::Error::NotConstructed,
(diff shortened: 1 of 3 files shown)


src/wayland/shell/xdg/mod.rs
```diff
@@ -1971,6 +1973,9 @@ impl PopupSurface {
                 .unwrap()
                 .lock()
                 .unwrap();
+            if !role.active {
+                return Ok(None);
+            }
```
Drakulix, line 1978: What is this second commit doing in this PR. I don't see anything related to it in the description nor a way how role.active could ever be false.

Drakulix (commented):
It seems very odd to me, that the extension would be supported if no fence can be created. Which driver are we talking about here?

krissedout:
I thought the repeated EGL warning in my DE was caused by fence creation because it disappeared while i was testing this path, after tracing it properly, both fence types create successfully and the warning i can reproduce is from a rejected 10-bit EGLConfig probe before the normal 8-bit fallback, so this change seems to have been unnecessary because i misdiagnosed an issue on my DE side. The second commit was an unrelated and incomplete popup teardown experiment (i forgot to do it on a separate branch), sorry for the noise, i’m closing this.

PR #2151, opened 2026-09-02
drm/compositor: option to keep previous frame's primary buffer alive

### LINKED ISSUE #1862: Flickering in bottom right of screen with NVIDIA GPUs and high refresh rates
Both niri and cosmic have this issue.

https://github.com/YaLTeR/niri/issues/2597
https://github.com/pop-os/cosmic-epoch/issues/2352

Probably a smithay issue. Gnome and hyprland work fine. 

screen recording collected from other repos: 

https://github.com/user-attachments/assets/7c4e2275-4c00-4954-b64f-f6ad5a882c5e

aoli-al:
## Description

Introduce an option to enable 3 buffer rotation. Currently, `Swapchain::acquire` returns the first free slot, which is nearlly always the most recent released buffer. The nvidia driver appears to keep reading the buffer for a short while. Holding the buffer for 1 frame longer will avoid flickering. 

This matches the behavior in Hyprland, which does not exhibit flickering.

This fixes: https://github.com/Smithay/smithay/issues/1862

I use claude code + fable to understand the implementation differences between smithay and hyprland and draft the PR

## Checklist

### DIFF
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -1067,6 +1067,8 @@ where
     framebuffer_exporter: F,
 
     current_frame: CompositorFrameState<A, F>,
+    previous_primary_slot: Option<Arc<Slot<A::Buffer>>>,
+    keep_previous_frame_buffer: bool,
     pending_frame: Option<PendingFrame<A, F, U>>,
     queued_frame: Option<QueuedFrame<A, F, U>>,
     next_frame: Option<PreparedFrame<A, F>>,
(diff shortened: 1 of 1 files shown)


Drakulix:
Thanks for debugging this, but that seems like a more problematic synchronization issue, that I wouldn't want to solely solve with a workaround in `DrmCompositor`.

@cmeissl Maybe we should set `OUT_FENCE_PTR` and attach that as a sort of "blocker" to the buffer? I would expect the DRM_EVENT_FLIP_COMPLETE to be enough, though it is actually hard to figure out what guarantees that holds.

aoli-al:
Thanks! Yes,  I agree the current solution is a bit unfortunate. This is definitely a driver bug but it seems that Nvidia is not doing anything 🙁

Happy to test on my end if you have other solutions.

Binary-Eater:
@Drakulix @aoli-al we just recently root caused and have a solution internally for the issue. We hope to have the fix in a future driver release. Also, `OUT_FENCE_PTR` is similarly impacted by the issue.....

PR #2156, opened 2026-09-05
Update winit to 0.31-beta.3

kelnos:
## Description

Upstream API breaks:

* `WindowEvent`, `PointerSource`, `ButtonSource`, and `MouseScrollDelta` have been marked `#[non_exhaustive]`.

## Checklist

* [x] I agree to smithay's [Developer Certificate of Origin](https://github.com/Smithay/smithay/blob/master/DCO.md).

PR #2157, opened 2026-09-05
Fix CI/test issues

kelnos:
## Description

* Fixes clippy warnings against current stable.
* Fixes `benches/geometry.rs` build failure due to changed trait import in rand 0.10.

## Checklist


kelnos:
Argh, this fails because the winit beta pin is floating, not `=`, and master doesn't compile against the new winit beta.  #2156 fixes that, but doesn't include the fixes in *this* PR...

Drakulix (approved):

PR #2158, opened 2026-09-06
renderer: add finish_without_export for GLES frames

abhighansala:
## Description

This PR adds `GlesFrame::finish_without_export()` to Smithay's GLES renderer.

The new method allows a GLES frame to be finalized without exporting an EGL `SyncPoint` or falling back to `glFinish()`, while still performing the normal frame finalization and texture synchronization required by Smithay.

This is useful for rendering pipelines that perform multiple intermediate passes where exporting a native EGL fence after every pass is unnecessary. The final rendering pass can still use the existing `finish()` method when a `SyncPoint` is required for synchronization outside the intermediate pipeline.

The implementation keeps the existing `finish()` behavior unchanged and factors the common finalization logic into `finish_internal_impl()` with an option controlling whether an export fence should be produced.

This change is intended to be used by COSMIC's blur rendering pipeline to avoid unnecessary synchronization overhead between intermediate blur passes. The corresponding COSMIC PR has not yet been submitted.

This PR was developed with assistance from generative AI.

## Checklist

### DIFF
--- a/src/backend/drm/compositor/frame_result.rs
+++ b/src/backend/drm/compositor/frame_result.rs
@@ -343,7 +343,7 @@ where
                 .clear(Color32F::BLACK, &clear_damage)
                 .map_err(BlitFrameResultError::Rendering)?;
 
-            sync = Some(frame.finish().map_err(BlitFrameResultError::Rendering)?);
+            sync = Some(frame.finish(true).map_err(BlitFrameResultError::Rendering)?);
         }
 
         // first do the potential blit
@@ -410,7 +410,7 @@ where
                     .map_err(BlitFrameResultError::Rendering)?;
             }
 
-            Ok(frame.finish().map_err(BlitFrameResultError::Rendering)?)
+            Ok(frame.finish(true).map_err(BlitFrameResultError::Rendering)?)
         } else {
             Ok(sync.unwrap_or_default())
         }
--- a/src/backend/drm/compositor/mod.rs
+++ b/src/backend/drm/compositor/mod.rs
@@ -3358,7 +3358,7 @@ where
                             element.transform(),
                             element.alpha(),
                         )?;
-                        let _ = frame.finish()?.wait(); // what can we do?
+                        let _ = frame.finish(true)?.wait(); // what can we do?
                         Ok(())
                     },
                 )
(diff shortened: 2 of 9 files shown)


Drakulix:
Hmm, I wonder if it would be better to add a flag to `finish` instead, something like `exportable: bool`. I would still like to return a fence, but we could call `eglCreateSync` without the native_fence flags, which would still produce a syncpoint, that just cannot be imported into other GPUs. That *should* be just as cheap.

abhighansala:
> Hmm, I wonder if it would be better to add a flag to `finish` instead, something like `exportable: bool`. I would still like to return a fence, but we could call `eglCreateSync` without the native_fence flags, which would still produce a syncpoint, that just cannot be imported into other GPUs. That _should_ be just as cheap.

That makes sense. I initially introduced finish_without_export() because the intermediate Dual-Kawase passes only need same-context synchronization and don't need an exportable/native fence. Returning a regular EGL syncpoint from finish(false) would be cleaner since the caller can still explicitly synchronize each intermediate pass.

I'm happy to rework the PR around an exportable flag. Would you prefer finish(exportable: bool) directly on the Frame trait, or a GLES-specific variant/API so that the existing Frame::finish() interface remains unchanged?

Drakulix:
> I'm happy to rework the PR around an exportable flag. Would you prefer finish(exportable: bool) directly on the Frame trait, or a GLES-specific variant/API so that the existing Frame::finish() interface remains unchanged?

Directly on the `Frame` trait please. This will also be relevant for the upcoming Vulkan renderer and the other renderer implementations can likely all simply ignore the exportable-flag as they should always return signaled fences anyway. This also allows us to easily handle edge-cases in the `MultiRenderer` with this, should they come up.

Thanks for taking the time!

abhighansala:
> > I'm happy to rework the PR around an exportable flag. Would you prefer finish(exportable: bool) directly on the Frame trait, or a GLES-specific variant/API so that the existing Frame::finish() interface remains unchanged?
> 
> Directly on the `Frame` trait please. This will also be relevant for the upcoming Vulkan renderer and the other renderer implementations can likely all simply ignore the exportable-flag as they should always return signaled fences anyway. This also allows us to easily handle edge-cases in the `MultiRenderer` with this, should they come up.
> 
> Thanks for taking the time!

Thanks! That makes sense. I'll rework the PR around Frame::finish(exportable: bool), keeping the synchronization behavior for the non-exportable GLES path while avoiding the native-fence export overhead. I'll also update the other renderer implementations to accept the flag and preserve their existing signaled-fence behavior.

abhighansala (commented):
Reworked the PR based on your feedback.

exportable is now part of the generic Frame::finish(exportable: bool) API rather than a GLES-specific finish_without_export() API.

The flag is propagated through the relevant renderer implementations and MultiRenderer. GLES retains the internal synchronization behavior while only using the exportable fence path when requested.

The latest implementation is in f529531.

Thanks for the guidance!

PR #2159, opened 2026-09-08
Rework virtual_keyboard, implement zwp_virtual_keyboard_v1 and zwlr_virtual_pointer_v1

### LINKED ISSUE #1937: New keyboard impl
As in #1903 said and other fact shows, current keyboard impl have something strange.

~~1. Handling multiple keyboard in more than one wl_keyboard object(yeah currently it's not allowed).~~(Maybe my misunderstanding)
1. Properly merge keymap together.
2. impl virtual_keyboard properly.

pgaskin:
#### Keyboard

Previously, virtual keyboards would bypass the compositor's event handling, directly writing events to wl_keyboards. This results in compositor bindings, keyboard grabs, shortcut inhibition, and focus not being handled correctly.

This commit makes virtual keyboards work the same way as real input devices, similar to how the libei backend works.

The device is only created once a keymap is uploaded from the client since all keycodes are interpreted using it.

The compositor is now responsible for switching the keymap when handling events from the virtual device. Previously, it was changed silently by smithay behind the compositor's back.

Since KeymapChanged and Modifiers don't have a counterpart in InputEvent, they are sent as InputEvent::Special.

This also fixes a bug where the client could cause the server to panic by specifying an invalid seat in create_virtual_keyboard.

The new XkbContext::set_modifier_mask function allows the modifier state to be set from a raw xkb mask, and is necessary since KeyboardHandle::set_modifier_state is lossy due to serialize_back. It handles leds_state the same way as set_layout.

The keyboard behaviour differs slightly from wlroots in that modifiers are shared. On wlroots, the per-keyboard modifiers are more a result of how input works than an explicit design choice for virtual keyboards. The `zwp_virtual_keyboard_v1` protocol only states that it should "emulate the behaviour of a physical keyboard", which is what this does.

Also see #1903 and niri-wm/niri#403.

#### Pointer

The new virtual pointer implementation works similarly.

There are some protocol deviations for wlroots compatibility:
- Button state is treated as a C-style boolean rather than an enum.
- Missing axis_source defaults to Wheel.
- Accumulated axis values are cleared on axis_stop.

The absolute positions from the client are normalized to 0.0..=1.0 based on the requested extents.

#### AI usage

I used AI to help figure out a design consistent with the rest of the codebase, and to compare the implementation to wlroots for compatibility. I also used it to generate the Dispatch boilerplate.

The actual logic and all prose (comments, commit messages, this PR) was hand-written by me.

#### Testing

I previously implemented [`wl-uinput-proxy`](https://github.com/pgaskin/wl-uinput-proxy) as a workaround, and learned a LOT about wayland and xkb while doing so (this was my first ever wayland-related project, one of my first rust projects, but not my first virtual input or remote-desktop related project).

I have been using this (along with RealVNC via [`xwlrvnc`](https://github.com/pgaskin/xwlrvnc), and wayvnc) daily since mid-June, and am pretty confident in the correctness and completeness.

I have also tested this with wtype and a few one-off scripts.

See niri-wm/niri#4548 for the niri side.

## Checklist

---

closes #2036 (vibe-coded incomplete virtual-pointer impl)
closes #1903 (modifies virtual_keyboard to do a similar keymap-swapping trick, but less cleanly)
closes #1597 (same idea as the set_modifier_mask thing I added)
fixes #1937
see niri-wm/niri#403
see niri-wm/niri#3595


pgaskin:
cc @Bot-wxt1221 @itaylor @YaLTeR

anvil/src/input_handler.rs
```diff
@@ -1399,3 +1400,48 @@ fn process_keyboard_shortcut(modifiers: ModifiersState, keysym: Keysym) -> Optio
...
+
+impl<BackendData: Backend> VirtualKeyboardHandler for AnvilState<BackendData> {
+    fn process_virtual_keyboard_event(&mut self, event: InputEvent<VirtualKeyboardBackend>) {
+        let keyboard = self.seat.get_keyboard().unwrap();
+
+        match event {
+            // The device's keycodes only mean anything under its own keymap, so
+            // activate it before processing events. Anvil never switches back;
+            // a real compositor should track which device the seat keymap came
+            // from and restore its configured one when a physical key arrives.
```
YaLTeR, line 1412: I think it would be good to have an example that does this properly because that's the behavior that actual compositors will want in 100% of the cases.

anvil/src/input_handler.rs
```diff
@@ -1399,3 +1400,48 @@ fn process_keyboard_shortcut(modifiers: ModifiersState, keysym: Keysym) -> Optio
...
+                mods_depressed,
+                mods_latched,
+                mods_locked,
+                group,
+                ..
+            }) => {
+                keyboard.with_xkb_state(self, |mut context| {
+                    context.set_modifier_mask(mods_depressed, mods_latched, mods_locked, group)
+                });
+            }
```
YaLTeR, line 1443: Is this correct to handle like this? What if, let's say:
1. virtual kb sends modifiers held Super
2. real kb sends Enter
3. virtual kb sends Enter

Enter from virtual kb should probably be Super+Enter, but should Enter from real kb be Super+Enter?

I guess if you have two real kbs plugged in, then modifiers from either of them are shared (unless the compositor really goes out of its way to split them), so two Super+Enter in this virtual case is fine too? Would be good to document.

src/wayland/virtual_pointer/input.rs
YaLTeR, line 1: I suppose niri should be able to just switch to this and drop its vendored impl of the same?

YaLTeR: Ah yeah, I noticed the niri PR now.

anvil/src/input_handler.rs
```diff
@@ -1399,3 +1400,48 @@ fn process_keyboard_shortcut(modifiers: ModifiersState, keysym: Keysym) -> Optio
...
+                    action @ (KeyAction::None
+                    | KeyAction::Quit
+                    | KeyAction::Run(_)
+                    | KeyAction::TogglePreview
+                    | KeyAction::ToggleDecorations) => self.process_common_key_action(action),
+                    // VT switching and output changes are out-of-scope for virtual keyboards.
+                    action => debug!(?action, "ignoring key action from a virtual keyboard"),
+                }
+            }
+            InputEvent::Special(VirtualKeyboardSpecialEvent::Modifiers {
```
YaLTeR, line 1433: Also I guess it would be good to have an example here for resetting held modifiers and keys when a virtual keyboard is destroyed?

anvil/src/input_handler.rs
pgaskin: I chose this handling since it's simplest and aligns with how X11 behaves, how some other compositors handle it, and how physical keyboards behave.

The alternative is that key modifiers change from underneath things which may have already received key events for the physical keys for those modifiers, causing hard-to-explain weirdness in applications.

The protocol doesn't dictate the behaviour one way or another, and only states that is should "emulate the behaviour of a physical keyboard".

I think shared modifiers is the correct way to go.

Edit: I tested this again, and wlroots maintains the xkb state for all keyboards separately, so it isn't shared on wlroots, but not necessarily by design, rather just how input works on wlroots.

anvil/src/input_handler.rs
pgaskin: I think niri is the best example, as it's simple enough (this part is only a fraction of it), and it's realistic.

anvil/src/input_handler.rs
pgaskin: I now do this automatically (similarly to how libinput does it for unplugged keyboards).

src/wayland/virtual_pointer/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+            }
+            zwlr_virtual_pointer_v1::Request::Destroy => {
+                // no-op; the device is removed in `destroyed`
+            }
+            _ => unreachable!(),
+        }
+    }
+
+    fn destroyed(&self, state: &mut D, _client: ClientId, virtual_pointer: &ZwlrVirtualPointerV1) {
+        state.process_virtual_pointer_event(InputEvent::DeviceRemoved {
```
YaLTeR: Does this need to release held buttons and whatnot too?

src/wayland/virtual_keyboard/mod.rs
```diff
@@ -151,17 +154,228 @@ where
...
+                        device: self.device(virtual_keyboard),
+                        mods_depressed,
+                        mods_latched,
+                        mods_locked,
+                        group,
+                    },
+                ));
+            }
+            zwp_virtual_keyboard_v1::Request::Destroy => {
+                // no-op; the device is removed in `destroyed`
```
YaLTeR, line 325: I forget, do we want to destroy stuff here right away to prevent some kind of delay between the Destroy request and `destroyed()`? Or that is not a thing in Smithay?

pgaskin: Nope, `destroyed` gets all of cases, and there's no delay.

src/wayland/virtual_pointer/mod.rs
pgaskin: Done.

src/wayland/virtual_keyboard/mod.rs
```diff
@@ -1,69 +1,81 @@
...
+pub use input::{
+    VirtualKeyboardBackend, VirtualKeyboardDevice, VirtualKeyboardKeyEvent, VirtualKeyboardSpecialEvent,
+};
 
-pub use virtual_keyboard_handle::VirtualKeyboardUserData;
+/// Handler for virtual keyboard input events.
+pub trait VirtualKeyboardHandler {
+    /// Handle an event from a virtual keyboard.
+    ///
+    /// c it the same way as any other
```
YaLTeR: something happened here

src/wayland/virtual_keyboard/mod.rs
```diff
@@ -151,17 +154,228 @@ where
...
+        _client: &Client,
+        virtual_keyboard: &ZwpVirtualKeyboardV1,
+        request: zwp_virtual_keyboard_v1::Request,
+        _dh: &DisplayHandle,
+        _data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            zwp_virtual_keyboard_v1::Request::Keymap { format, fd, size } => {
+                let Some(keymap) = compile_keymap(format, fd, size as usize) else {
+                    return;
```
YaLTeR: Maybe a warn!() or something here

src/wayland/virtual_keyboard/mod.rs
```diff
@@ -151,17 +154,228 @@ where
...
+                    virtual_keyboard.post_error(
+                        zwp_virtual_keyboard_v1::Error::NoKeymap,
+                        "`key` sent before keymap.",
+                    );
+                    return;
+                }
+
+                // This should be wl_keyboard::KeyState, but the protocol does
+                // not define it as an enum.
+                let key_state = if key_state == 1 {
```
YaLTeR: This `== 1` matches wlroots? It shouldn't be `!= 0`?

src/wayland/virtual_keyboard/mod.rs
```diff
@@ -151,17 +154,228 @@ where
...
+                    KeyState::Released
+                };
+
+                // Tracked so `destroyed` can release held keys.
+                {
+                    let mut pressed = self.data.pressed_keys.lock().unwrap();
+                    match key_state {
+                        KeyState::Pressed => {
+                            if !pressed.contains(&key) {
+                                pressed.push(key);
```
YaLTeR: Maybe it should be a `HashSet`? Idk

src/wayland/virtual_keyboard/mod.rs
pgaskin: ugh, must have middle-clicked from my phone by accident lol

src/wayland/virtual_keyboard/mod.rs
pgaskin: Ah, I remember why I did that... I was thinking about what would happen if an unknown value was passed, and decided that it'd be better to treat it as a released key to avoid stuck keys from buggy clients or future changes.

OTOH, I don't recall ever seeing a client do that, so I'll probably change it to match the wlroots pointer (since the smithay enum only gives two options).

anvil/src/input_handler.rs
pgaskin: Correction (I updated the previous comment): wlroots is an odd one compared to other compositors, it has per-keyboard xkb state, but not specifically for virtual keyboards.

I still think shared modifiers is best.

pgaskin:
Made a few changes:

- Added support for the `no_keymap` format (which uses the existing seat keymap).
- Added logging for keymap compilation errors.
- Fixed some copy-paste errors in doc comments between pointer and keyboard impls.
- ~~Made the pointer always send wheel steps, even if 0 (to match the trait).~~ Edit: reverted this; I realized why I didn't do this to begin when I originally implemented it (it breaks stuff which based on wheel events if the client doesn't send discrete values).
- Squashed the fixups.

pgaskin:
I did some more testing and review of niri-wm/niri#4548 against wlroots. The differences are:

- Keyboard
  - XKB state is per-seat rather than per-keyboard. This is wlroots-specific. The protocol states that it should "emulate the behaviour of a physical keyboard", so arguable, both are correct. Regardless, I think it makes the most sense for modifiers to be shared across all keyboards on a seat.
  - Virtual keyboards added with the keymap format `no_keymap` still have modifiers (this is basically the same thing as above).
  - Invalid keymap formats (or keymap mapping errors) are ignored (and logged), resulting in a `no_keymap` when the client attempts to send a `key`. In wlroots, it emits a `invalid_keymap_format` protocol error when parsing the keymap (but this can't be done with smithay's bindings).
- Pointer
  - An `axis` event is emitted per-frame rather than per-axis. This isn't actually different though, since wlroots will merge it itself, but smithay doesn't.
  - No `frame` event is emitted if there aren't any `axis` ones pending (basically the same as above).
  - The absolute position is clamped.

And the wlroots quirks I implemented on purpose here for compatibility:

- Keyboard
  - The `no_keymap` format results in the seat keymap being used.
- Pointer
  - Button state is interpreted as a C boolean rather than an enum.
  - Default axis source is wheel if none is sent with an event.
  - Absolute motion with a zero extent (i.e., would result in a divide by zero) is ignored silently.
  - High-resolution scroll wheel axis events emitting one logical wheel step for every delta of 120 (i.e. WLR_POINTER_AXIS_DISCRETE_STEP).

pgaskin:
I need to test this a bit more to ensure the recent tweaks and fixes I did didn't introduce any regressions compared to the build I've been using for the last few months.

I'll post another comment here once I do.

pgaskin:
I think I'm happy with this now, and it should be ready to fully review/merge (as long as no issues are found with niri-wm/niri#4548).

I tested this on niri with various edge cases including:

- Scrolling with different combinations of axis values.
- Sharing modifiers with real input device.
- Pressing and releasing keys at the same time as a real keyboard.
- Pressing and releasing keys at the same time as another keyboard.
- Releasing keys/buttons when a virtual keyboard/pointer is removed.
- Rotated and scaled outputs, including as part of mixed layouts.
  - I fixed a bug with this (see `absolute_position_is_logical`).
- Multiple virtual keyboards with the same and different layouts
  - I fixed a few bugs around keymap switching here (mostly for the `no_keymap` case).
  - I did some optimizations to reduce unnecessary layout switches (see `configured_keymap`).
- Modifiers on both the physical and real keyboards.
  - I fixed a bug where clients would forget modifiers in some cases due to keymap changes (see `restore_keyboard_state`).
- Virtual keymaps with different shifted layouts.

I've also been using this (along with niri-wm/niri#4554) for multiple hours today with RealVNC on [xwlrvnc](https://github.com/pgaskin/xwlrvnc) (also with x11vnc on xwlrvnc), and I've also tested it with wayvnc.

YaLTeR (approved):
This and the niri change look ok to me; someone else please take a look

src/wayland/virtual_pointer/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+            } => {
+                let Some(axis) = parse_axis(virtual_pointer, axis) else {
+                    return;
+                };
+
+                let mut frame = self.data.frame.lock().unwrap();
+                frame.time = frame.time.max(time);
+                let pending = frame.axes[axis_index(axis)].get_or_insert_with(Default::default);
+                pending.value = value;
+                pending.v120 = Some(discrete * 120);
```
YaLTeR:
```suggestion
                pending.v120 = Some(discrete.saturating_mul(120));
```

Fix to avoid overflow panic in a client-provided value. In niri I believe that's the only one that was problematic, here I didn't look in detail.

pgaskin: Done, also for the keycode offset.

pgaskin:
Uh, am I supposed to be doing something about @codecov-commenter?

YaLTeR:
uh not really

YaLTeR:
I just found that Smithay's `KeyboardHandle::input_from_source()` has a `KeyboardSource` input that is apparently intended to keep track of multiple keyboards and is used for things like keeping track of pressed keys, so press A one one => press A on second => release A on second => release A on one won't send key up until it's released on both keyboards. Does this PR need to do something about this? Or maybe the niri PR needs to pass KeyboardSource based on input device? (Not sure)

Though it's also not completely hooked up, e.g. `input_intercept()` hardcodes `KeyboardSource::MAIN`. So maybe instead it needs to be removed from Smithay?

pgaskin:
I think that's out of scope for this PR specifically, which implements the same semantics as other physical keyboards (as specified in the virtual keyboard spec).

About per-keyboard tracking, I personally prefer key states being shared across keyboards on the same seat, so...

xlsigned:
Hopefully this gets merged soon. As far as I understand this would fix the issue that `niri` entirely ignores keyboard shortcuts arriving via `wayvnc` (https://github.com/niri-wm/niri/issues/403), right?

pgaskin:
Yes, it fixes that and other remote desktop keyboard bugs.

I think this PR is waiting for a second review from someone experienced with this part.

pgaskin:
Anyone who could do the second review of this and merge it? I've already reviewed it myself (though I lack experience with the Smithay codebase in general) and tested it throughly with niri. @YaLTeR has reviewed it comprehensively.

There doesn't seem to be anyone else here with significant work on virtual input since the initial implementation other than @Bot-wxt1221.

Maybe @ids1024 since you did libei and have done a lot with Smithay as a whole?

PR #2160, opened 2026-09-08
feat: expose X11 window icons

FreddyFunk:
## Description

Parse bounded `_NET_WM_ICON` image records and expose all valid sizes through `X11Surface`. Advertise and report runtime icon property updates and removals. This allows compositors to properly propagate window metadata.

I noticed that icons and application names are not properly shown for video games launched through Steam in COSMIC dock. This is one of the building blocks to make it work.

## Checklist


src/xwayland/xwm/surface.rs
```diff
@@ -157,6 +168,7 @@ pub(crate) struct SharedSurfaceState {
     title: String,
     class: String,
     instance: String,
+    icons: Vec<X11SurfaceIcon>,
```
Drakulix: Thanks! Can we avoid permanently storing the pixel data in each X11Surface?

If a compositor decides to this (e.g. via UserData), that is one thing. But I think smithay should simply notify the compositor, when there is new image data and then fetch the data upon request from Xwayland.

FreddyFunk: Yes, that's a way better design. I implemented it now according to your suggestion.

PR #2165, opened 2026-09-09
vulkan renderer

Drakulix:
## Description

Very early draft of a vulkan renderer.

1. This moves around a lot of logic from backend/allocator/vulkan into
   backend/vulkan to generalize for use in the renderer.
2. It also expands the modules in backend/vulkan significantly to
   provide more useful abstractions around vulkan objects and their
   initialization logic.
3. The vulkan renderer then itself deals mostly with specialized drawing
   related functionality as well as implementating the various
   renderer-trait semantics by creating the right vulkan objects with
   the right set of flags.
4. Most testing so far has been done with examples/vulkan.rs. The render
   itself still lacks a lot of functionality needed to run a full
   compositor.

## Checklist


Pheoxy:
https://github.com/Pheoxy/smithay/tree/add-vulkan-renderer-support-cosmic-e3d461a

This is my fork branch where I'm working on figuring out Vulkan myself.

It's a bit dirty but it was never intended to be a PR branch.

src/backend/vulkan/device.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.0.ext_host_image_copy.as_ref()
+    }
+
+    pub fn memory_properties(&self) -> &PhysicalDeviceMemoryProperties {
+        &self.0.mem_properties
+    }
+
+    pub fn queue(&self) -> &Queue {
+        &self.0.queue
+    }
```
ids1024, line 199: https://docs.vulkan.org/refpages/latest/refpages/source/vkQueueSubmit.html

> Host access to queue must be externally synchronized if it was not created with [VK_DEVICE_QUEUE_CREATE_INTERNALLY_SYNCHRONIZED_BIT_KHR](https://docs.vulkan.org/refpages/latest/refpages/source/VkDeviceQueueCreateFlagBits.html)

Maybe this should return a `&Mutex<vk::Queue>`, which would enforce the necessary synchronization.

src/backend/renderer/vulkan/mod.rs
```diff
@@ -0,0 +1,1158 @@
...
+                        .signal_semaphores(&[self.renderer.timeline.vk])
+                        .push_next(&mut timeline_info)],
+                    Fence::null(),
+                )
+                .map_err(Error::SubmitError)?;
+        }
+
+        self.renderer
+            .cmd_pool
+            .store_pending_buffer(buf, next_seq_no[0], descriptor);
```
ids1024, line 1080: It seems like the best practice is generally to call `vkQueueSubmit` only once per frame. For instance in https://gpuopen.com/learn/rdna-performance-guide:

> Ideally, submissions would only happen when syncing queues or at the end of a frame.

So we probably want to store a `vk::CommandBufer` in `VulkanFrame` and submit in `VulkanFrame::finish`?

src/backend/renderer/vulkan/mod.rs
```diff
@@ -0,0 +1,1158 @@
...
+    pub fn new(phd: &PhysicalDevice, drm: Option<DrmDeviceFd>) -> Result<VulkanRenderer, Error> {
+        // Check matching drm descriptor, if provided
+        let node = if let Some(fd) = drm.as_ref() {
+            let node = DrmNode::from_file(fd).map_err(|_| Error::MismatchedDrmDevice)?;
+            if !(phd.render_node().ok().flatten().is_some_and(|node| node == node)
+                || phd.primary_node().ok().flatten().is_some_and(|node| node == node))
+            {
+                return Err(Error::MismatchedDrmDevice);
+            }
+            Some(node)
```
skygrango, line 133: I noticed a typo here:
current
```rust
let node = DrmNode::from_file(fd).map_err(|_| Error::MismatchedDrmDevice)?;
if !(phd.render_node().ok().flatten().is_some_and(|node| node == node)
                || phd.primary_node().ok().flatten().is_some_and(|node| node == node))
```
it should be
```rust
let node = DrmNode::from_file(fd).map_err(|_| Error::MismatchedDrmDevice)?;
if !(phd.render_node().ok().flatten().is_some_and(|n| n == node)
                || phd.primary_node().ok().flatten().is_some_and(|n| n == node))
```

Issue #2138, opened 2026-08-13
Smithay is sending wl_surface.frame to windows out of sight


Quackdoc:
Cosmic comp and niri are currently sending `wl_surface.frame` events to windows even when the window is not visible, It slows the cadence down to roughly 1 call a second. the protocol states  [A server should avoid signaling the frame callbacks if the surface is not visible in any way, e.g. the surface is off-screen, or completely obscured by other opaque surfaces](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_surface:~:text=A%20server%20should%20avoid%20signaling%20the%20frame%20callbacks%20if%20the%20surface%20is%20not%20visible%20in%20any%20way%2C%20e%2Eg%2E%20the%20surface%20is%20off%2Dscreen%2C%20or%20completely%20obscured%20by%20other%20opaque%20surfaces).

I am not sure if this is a smithay or a cosmic issue so I am reporting it in both, but niri also has the issue so it leads me to believe smithay.

Reported to cosmic here https://github.com/pop-os/cosmic-comp/issues/2732

YaLTeR:
It's intended behavior. All compositors do this because various video games and other clients don't handle it when the compositor stops sending frame callbacks completely.

There's xdg-toplevel Suspended state which lets the compositor stop sending frame callbacks, but niri doesn't currently use it.

Quackdoc:
I only came across this because of a bug in an app I was hitting in cosmic that I wasn't hitting in sway but was here. Sway does not send the events, I'm not sure about gnome or kde. 

It seems like applications that bug out should be fixed and the protocols text should be followed, or the protocol itself should be changed if most of the major compositors aren't going to follow the reccomendation.

kchibisov:
should avoid is not a `must`, so application is broken, and no spec should be updated as a result.

Quackdoc:
Sounds good, I was able to fix it for the client since they simply forgot pre present notify causing queue stalls, I'll go ahead and close this then, just more behavioral fragmentation guess. :/

PR #2168, opened 2026-09-11
desktop: don't send frame callbacks to unmapped surfaces

rayslava:
## Description
`send_frames_surface_tree` sends frame callbacks to surfaces without a buffer, which the renderer never draws. Firefox with `gfx.webrender.compositor.force-enabled=true` takes such a callback as proof its layer-root subsurface is presented, never attaches a buffer to it, and shows a black window.

This skips surfaces whose renderer state has no view, together with their children. Surfaces without renderer state keep the old behaviour, and mapped but hidden surfaces still get the throttled fallback (unlike #2138).

AI assistance is disclosed in the commit message, per the [AI policy](https://github.com/Smithay/smithay/blob/master/AI.md).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -259,6 +259,9 @@ source device, so those devices accumulate cached imports that previously were n
 Cleanup is now also attempted on every device even if it fails on one of them, with every failure
 logged and the first error returned.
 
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
Looks good. Thanks

PR #2170, opened 2026-09-12
image-copy: properly handle capture session destruction for cursor sessions

pgaskin:
Previously, destroying the capture session wouldn't destroy the cursor session and fail the frames like it does when destroying the cursor session itself, or regular sessions.

Tested with niri-wm/niri#4554.

cc @YaLTeR 

## Checklist

### DIFF
--- a/src/wayland/image_copy_capture/mod.rs
+++ b/src/wayland/image_copy_capture/mod.rs
@@ -426,7 +426,7 @@ impl CursorSessionRef {
             return;
         }
 
-        if let Some(session_obj) = inner.session_obj.as_ref() {
+        if let Some(session_obj) = inner.session_obj.as_ref().filter(|obj| obj.is_alive()) {
             session_obj.buffer_size(constraints.size.w as u32, constraints.size.h as u32);
             for fmt in &constraints.shm {
                 session_obj.shm_format(*fmt);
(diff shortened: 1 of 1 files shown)


pgaskin:
Not directly related to this PR, but @ids1024 may be interested in https://github.com/niri-wm/niri/pull/4554#discussion_r3995427661 (cosmic implements frame handling on session destruction differently than the others).

src/wayland/image_copy_capture/mod.rs
```diff
@@ -868,6 +873,8 @@ pub struct SessionData {
 pub struct CursorSessionData {
     inner: Arc<Mutex<CursorSessionInner>>,
     user_data: Arc<UserDataMap>,
+    /// Cursor session to destroy.
+    cursor_obj: Option<ExtImageCopyCaptureCursorSessionV1>,
```
ids1024, line 877: Instead of using `Option`, it would probably make sense to split `CursorSessionData` into two types; one used to implement dispatch for `ExtImageCopyCaptureSessionV1` and another for `ExtImageCopyCaptureSessionV1`?

I would also make this a `Weak<ExtImageCopyCaptureCursorSessionV1>` if we don't need a strong reference.

PR #2002, opened 2026-04-18
fix(gles): use STREAM_DRAW for PIXEL_PACK_BUFFER (fixes #1595)

xdagiz:
## Description

The GLES renderer was using GL_STREAM_READ when allocating PBOs for
glReadPixels operations (i don't know if this was a bug or intentional). However, PIXEL_PACK_BUFFER is used to pack
pixel data FROM the GPU (GPU writes, CPU reads), so using GL_STREAM_DRAW fixed it.

This caused screenshots in [niri](https://github.com/niri-wm/niri/issues/807) to fail with "failed to copy to output" and with debug logs:
```
ERROR smithay::backend::renderer::gles: [GL] GL_INVALID_ENUM in glBufferData(invalid usage: GL_STREAM_READ)
ERROR smithay::backend::renderer::gles: [GL] GL_INVALID_OPERATION in glReadPixels(out of bounds PBO access)
```
 on older CPUs (particularly Intel integrated graphics), where drivers
enforce the usage hint more strictly.

## Checklist

* [x] I agree to smithay's [Developer Certificate of Origin](https://github.com/Smithay/smithay/blob/master/DCO.md).

PR #2173, opened 2026-09-13
fix(gles): fix texture binding on GLES 2.0 GPUs

xdagiz:
## Description

I made a pr (#2002) a while ago to fix this and i just updated niri to the latest version and i started running into the same old issue (screenshots not working on opengl 2.0) 

i also added comments so the gles 2.0 fallback is clealer now

## Checklist


ids1024 (approved):
Looking at the OpenGL ES 2.0 and OpenGL ES 3.0 specs, this seems to be accurate.

Thanks!

PR #2174, opened 2026-09-16
wlr_layer, session_lock: skip the pre-commit hook once the role object is destroyed

### LINKED ISSUE #1979: Commits to unmapped layer surfaces error after role object has been destroyed
Originally from https://github.com/wmww/gtk4-layer-shell/pull/119. GTK4 Layer Shell (a library I maintain) does some weird but technically legal things with the layer shell protocol due to how it interacts with GTK. In this case it calls:
1. `zwlr_layer_surface_v1.destroy()`
2. `wl_surface.attach(NULL, 0, 0)`
3. `wl_surface.commit()`

Which I believe should be allowed, however Smithay errors the client with "width 0 requested without setting left and right anchors".

Side note: this only happens on the native Rust Wayland stack, the libwayland once silently drops the error (I think because the error is being sent on a dead Wayland object).

## Repro
1. Install ghostty (it uses GTK4 Layer Shell)
2. Put `keybind = global:cmd+backquote=toggle_quick_terminal` in `~/.config/ghostty/config`
3. Run `cargo run -p anvil --no-default-features --features winit -- --winit`
4. Run ghostty inside anvil
5. Press Super+\` to bring up the quick terminal (uses layer shell)
6. Press Super+\` again
7. Note the the client is killed and main window appears, instead of just the quick terminal closing

## Root Cause
I had my Claude investigate and she thinks the the problem is `pre_commit_hook` getting run after the layer surface object is destroyed. Indeed https://github.com/wmww/smithay/commit/6a18fc4ee04230b91e81931047d765b96333f1a4 appears to fix the bug, but I'm unconvinced this is the *correct* solution. Just adding a check seems like a hack, and this approach doesn't seem to apply as cleanly to XDG surface types, which also appear to have the same problem (untested).

MSILycanthropy:
## Description
Fixes #1979 

Destroy a layer surface (or lock surface), attach a null buffer and commit. The pre-commit hook still runs, and errors out. Looks like `libwayland` drops errors on dead objects and `wayland-rs` sends them, so the client ends up getting killed. 

Ran into this with Quickshell (dies every time it hides a panel https://github.com/pop-os/cosmic-comp/issues/1590) and with Vicinae (which currently turns layer-shell off on COSMIC [entirely](https://github.com/vicinaehq/vicinae/blob/cd2aa4ae1e4c38c4f2086a647861df37cb5e5b43/src/server/src/utils/environment.hpp#L81-L86)). 

Both hooks now return early if the role object is dead. 

Been running a fork with the fix for a few days without issue.

_AI Disclosure_: AI was used to research the problem and point me to the relevant bug reports. The actual fix and text written are solely my own. Please let me know if that disqualifies the PR, happy to have it taken down if it does. 

## Checklist


src/wayland/session_lock/surface.rs
```diff
@@ -302,6 +302,10 @@ impl LockSurface {
         compositor::with_states(surface, |states| {
             let role = states.data_map.get::<LockSurfaceData>().unwrap().lock().unwrap();
 
+            if !role.surface.is_alice() {
```
kelnos: I assume you mean that to be `is_alive`?

MSILycanthropy: 🤦🏼 yes, yes I do. 

Fixed in [7c5c592](https://github.com/Smithay/smithay/pull/2174/commits/7c5c5924f9fafe0b1a9983d5a464c2314f52835e)

Drakulix (approved):
Thanks!

YaLTeR:
I guess this works in various edge cases because the role cannot be reassigned to a different role after destruction, and because assigning the same role will make it alive once again?

Drakulix:
> I guess this works in various edge cases because the role cannot be reassigned to a different role after destruction, and because assigning the same role will make it alive once again?

This is how `is_alive` should work, otherwise it is conceptually broken. But I have to admit I didn't check in full detail.

PR #2176, opened 2026-09-17
Stop using `dlopen` for `libwayland` by default

HigherOrderLogic:
## Description

Currently, the `dlopen` feature in `wayland-sys` is turned on by default by `winit`, so Smithay (and downstream compositors as a result) all use `dlopen` to open `libwayland` instead of static linking it. Remove this feature make static linking the default.

If you want, I can add a feature to make this behavior toggleable.

## Checklist


YaLTeR:
I hope that by static linking you mean dynamic linking

HigherOrderLogic:
I'm not really sure which kind of linking `wayland-sys` does, but I just want to eliminate `dlopen` usage.

Ph4ntomas:
Wouldn't that break running the compositor on x11?

If so, it should be possible to re-enable it via a flag

HigherOrderLogic:
> Wouldn't that break running the compositor on x11?

Not really I think, the `dlopen` is intended for running Winit app on X11 without the `libwayland` blob, but Smithay already uses `libwayland` unconditionally anyway.

 

> If so, it should be possible to re-enable it via a flag

Sure I can, but I was waiting for maintainer input as noted in the description.

ids1024:
It makes sense to me that `backend_winit` shouldn't force `winit/wayland-dlopen` to be enabled. Even if compositors might often want that, Smithay is generally supposed to be modular and not require things like that unnecessarily.

> but Smithay already uses libwayland unconditionally anyway.

I don't think that should be true?

Smithay doesn't actually require libwayland server-side. Unless the `use_system_lib` feature is enabled, it will use the Pure-Rust implementation of server side wayland in the `wayland-backend` crate.

The `winit` backend also uses the `wayland-egl` crate, but with dlopen that should be used at runtime only when winit creates a Wayland window.

Granted, it's pretty unusual to want to use a Smithay compositor on a system without `libwayland-client.so.0`. But in principle it should be possible, and sounds like a bug if it isn't currently.

HigherOrderLogic:
> I don't think that should be true?

Yeah, I forgot that Wayland-rs also has a Rust backend.



> The `winit` backend also uses the `wayland-egl` crate, but with dlopen that should be used at runtime only when winit creates a Wayland window.

The `dlopen` feature of `wayland-egl` is already off by default, so I dont think there's any issue with that.

mati865:
> I hope that by static linking you mean dynamic linking

Yeah, static linking would be terrible. Much worse than dlopen.

> > The `winit` backend also uses the `wayland-egl` crate, but with dlopen that should be used at runtime only when winit creates a Wayland window.
> 
> The `dlopen` feature of `wayland-egl` is already off by default, so I dont think there's any issue with that.

I didn't inspect the code, but for the latest versions of cosmic-comp and Niri the shared wayland library is not dynamically linked (adding gtk4 for reference):

<details>
<summary>Details</summary>

```
❯ ldd target/debug/anvil | rg wayland || echo "libwayland is not linked"
libwayland is not linked

❯ ldd /bin/cosmic-comp | rg wayland || echo "libwayland is not linked"
libwayland is not linked

❯ ldd /bin/niri | rg wayland || echo "libwayland is not linked"
libwayland is not linked

❯ ldd /bin/gtk4-launch | rg wayland || echo "libwayland is not linked"
	libwayland-client.so.0 => /usr/lib/libwayland-client.so.0 (0x00007f243b003000)
	libwayland-egl.so.1 => /usr/lib/libwayland-egl.so.1 (0x00007f2439b27000)
	libwayland-cursor.so.0 => /usr/lib/libwayland-cursor.so.0 (0x00007f2438cf4000)
```

</details>

You can see which libraries are dlopen by running:
```
❯ gdb --args target/debug/anvil --winit

(gdb) catch load
Catchpoint 1 (load)
(gdb) command
Type commands for breakpoint(s) 1, one per line.
End with a line saying just "end".
>bt
>c
>end
(gdb) r
```
That will be a lot of text but there you can see things like:
```
Catchpoint 1
  Inferior loaded /usr/lib/libwayland-egl.so.1
0x00007ffff7fc8ded in ?? () from /lib64/ld-linux-x86-64.so.2
#0  0x00007ffff7fc8ded in ?? () from /lib64/ld-linux-x86-64.so.2
#1  0x00007ffff7fba579 in _dl_catch_exception () from /lib64/ld-linux-x86-64.so.2
#2  0x00007ffff7fc7d52 in ?? () from /lib64/ld-linux-x86-64.so.2
#3  0x00007ffff7fba579 in _dl_catch_exception () from /lib64/ld-linux-x86-64.so.2
#4  0x00007ffff7fc8306 in ?? () from /lib64/ld-linux-x86-64.so.2
#5  0x00007ffff78a8d9c in ?? () from /usr/lib/libc.so.6
#6  0x00007ffff7fba579 in _dl_catch_exception () from /lib64/ld-linux-x86-64.so.2
#7  0x00007ffff7fba6a3 in ?? () from /lib64/ld-linux-x86-64.so.2
#8  0x00007ffff78a8817 in ?? () from /usr/lib/libc.so.6
#9  0x00007ffff78a8e69 in dlopen () from /usr/lib/libc.so.6
...
#43 0x000055555729958e in wayland_sys::egl::wayland_egl_handle () at src/egl.rs:48
#44 0x00005555571db76a in wayland_egl::WlEglSurface::new_from_raw (surface=0x555557668370, width=1280, height=800) at src/lib.rs:74
#45 0x0000555556509564 in smithay::backend::winit::init_from_attributes_with_gl_attr<smithay::backend::renderer::gles::GlesRenderer> (attributes=..., gl_attributes=...) at src/backend/winit/mod.rs:141
#46 0x00005555565037cd in smithay::backend::winit::init_from_attributes<smithay::backend::renderer::gles::GlesRenderer> (attributes=...) at src/backend/winit/mod.rs:87
#47 0x000055555650af4c in smithay::backend::winit::init<smithay::backend::renderer::gles::GlesRenderer> () at src/backend/winit/mod.rs:69
#48 0x00005555563eaaa0 in anvil::winit::run_winit () at anvil/src/winit.rs:100
#49 0x0000555555c7b1be in anvil::main () at anvil/src/main.rs:42
```
Which means that `wayland_egl` has called `dlopen` on `libwayland-egl.so.1`.

EDIT: Now looking at the code, `winit-wayland/dlopen` feature enables `wayland-sys/dlopen` feature. Remember that Cargo builds each crate only once (unless multiple versions are used at the same time).

HigherOrderLogic:
> I didn't inspect the code, but for the latest versions of cosmic-comp and Niri the shared wayland library is not dynamically linked (adding gtk4 for reference):
> 
> [...]
> 
> Which means that `wayland_egl` has called `dlopen` on `libwayland-egl.so.1`.

I believe this was caused by `winit` enabling the `dlopen` feature on the `wayland-backend` crate, which in turn [makes `wayland-egl` believe that it should use `dlopen`](https://github.com/Smithay/wayland-rs/blob/master/wayland-egl/src/lib.rs#L23-L25). 

Turning the feature off here should eliminate `dlopen` usage for the whole Wayland-rs stack. `cargo tree -p wayland-sys -e features --invert | grep -i "dlopen"` returns nothing for me with this change.

PR #2177, opened 2026-09-17
Replace `downcast-rs` crate with trait upcasting

ids1024:
## Description
Since Rust 1.86, we can upcast dyn traits to `&dyn Any` using `as &dyn Any`. Or we can use `<dyn Any>::` to invoke `Any` methods without an explicit cast.

We could provide convenience methods on `dyn *Grab` for `is` and `downcast_ref`, etc. so compositors don't need to do explict casts, but cosmic-comp and niri seem easy enough to update to this without any extra functions.

## Checklist


ids1024:
Hm, I guess with `<dyn Any>::` we need to be careful it gets invoked on the exact right type (CI errors may be from trying to invoke it on the `Box`?).

PR #2180, opened 2026-09-20
renderer/gles: only make the context current when there is cleanup to do

### LINKED ISSUE #2178: multigpu: cleanup_texture_cache now makes every device's EGL context current per frame, which wakes a runtime-suspended dGPU and stalls the compositor
Since 2800e70e MultiRenderer::cleanup_texture_cache also visits other_renderers.

DrmCompositor::render_frame calls `cleanup_texture_cache` on every no-damage or direct-scanout frame, and `cleanup_texture_cache` starts with `egl.make_current()`. On a hybrid laptop that renders on the iGPU with the dGPU in runtime D3, that means the compositor touches the NVIDIA context on basically every frame.

Setup: niri v26.04-93 (smithay 0ff00983), AMD iGPU as render-drm-device, NVIDIA in the gpu manager as a non-render device, no outputs on it. Kernel 7.2.6, nvidia-open 615.71.09 and 610.57.04 both. Outputs at 144Hz and 240Hz.

What I see:

- ftrace on the nvidia-drm entry points: niri's main thread calls nv_drm_semsurf_fence_create_ioctl 6726 times in 120s. About every 4-8ms or seemingly once per frame about.
- With the screen static the calls stop and the dGPU runtime-suspends about 7s later. On the next frame it wakes. The wake comes from the driver's own work queue:

nv_queue-1009  rpm_resume: 0000:64:00.0
  => __pm_runtime_resume
  => nv_indicate_not_idle
  => RmForceGpuNotIdle
  => rm_execute_work_item

So the leak fix is right for a desktop with two active GPUs, but on a laptop it turns a sleeping device into one that wakes on every idle-to-active transition and blocks the main loop while it does. A device with nothing in its caches probably should not be made current at all.

minyek:
## Description

Fixes #2178

`GlesRenderer::cleanup_texture_cache` and `invalidate_caches` make the EGL context current even
when there is nothing to destroy. Since multi-GPU cleanup visits every device on no-damage and
direct-scanout frames, this wakes an idle GPU every frame and stalls the compositor while it
resumes.

Now the context is only made current when a resource is queued for destruction, and an
already-current context (and its bound surface) is left alone. If activation fails the resource
is requeued.

On a hybrid laptop the secondary GPU sits in runtime suspend whenever the compositor is not
rendering on it. Making its EGL context current wakes it, and the main loop blocks until it has
resumed. Since the multi-GPU cleanup visited that device on every no-damage and direct-scanout
frame, the compositor paid that wake once a frame with nothing to free. With the queue checked
first, an idle device's context is never activated, so the GPU stays suspended and the per-frame
stall goes away. A device that does have something queued is still woken, but only then.

## Testing

Existing tests pass. New tests in `gles/mod.rs` and `multigpu/mod.rs` run on Mesa's software
EGL device, so they need no GPU, and skip where no EGL device yields a renderer. They fail
without the fix by pinning that an idle renderer, and an idle device under
`GpuManager::cleanup_texture_cache`, never has its context made current. They also pin that
deferral does not leak: both entry points still free dropped textures from an unbound context,
and repeated rounds of several queued textures leave nothing behind.

An instrumented cosmic-comp run on direct scanout went from making the context current on
every cleanup entry to none, and a resource census over a desktop workload reconciled
`queued == drained` for every GL object class.

Not covered: the hybrid-laptop scenario itself as I don't have a hybrid setup and haven't seen 
the issue. @bbedward could you check whether this change resolves it?

## AI assistance

Per Smithay's [AI policy](https://github.com/Smithay/smithay/blob/master/AI.md), AI assistance
(Claude Code) is disclosed in the commit message; the author understands the change and can
respond to review.

## Checklist


bbedward:
Can confirm, I built niri with your smithay rev and there's no more `nv_drm_semsurf` calls and no dGPU wakeups.

Drakulix (commented):
Thanks, the changes to the GLES renderer look good, but I find the tests to be quite verbose, while not adding much meaningful functionality.

Please drop the multi-renderer tests. For GLES lets keep `cleanup_texture_cache_leaves_an_idle_context_unbound` and `cleanup_texture_cache_frees_a_dropped_texture_from_an_unbound_context` to make sure we don't regress these changes.

But I don't think we need the rest. `invalidate_caches` is already a function that should be called really rarely.

minyek:
I've addressed your feedback and rebased into a single commit. I've also made an additional change: the tests now fail when no EGL device is available, rather than skipping with a warning. I shouldn't have added tests which pass when they can't be executed, they should fail. I'm happy to revert this change, if you don't agree.

minyek:
There was  conflict in CHANGELOG.md, so I've resolved and repushed

Drakulix (approved):
Looks good to me now, thanks!

